PR #1, opened 2014-10-13
KNOX-440: HttpFS impersonation issue

roengram:
KNOX puts 'doAs' parameter to impersonate certain user. However, HttpFS does not accept it; it accepts 'doas' instead. According to WebHDFS REST API specification (http://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-hdfs/WebHDFS.html#Proxy_Users), 'doas' is the correct one.


risdenk:
@roengram It looks like this issue was resolved by Hadoop proper according to the comments in KNOX-440. Can you close this PR?

PR #2, opened 2015-11-10
pac4j provider evolutions

leleuj:


risdenk:
@leleuj It looks like this functionality was merged. I don't know which JIRA specifically since it wasn't called out here. Is this PR still necessary? If not can the PR be closed?

leleuj:
Yes, I think so.

PR #6, opened 2017-05-09
KNOX-768 - Pass-through service definition and rewrite rules for Knox & Kafka

rmkellogg:
KNOX-768 - Preliminary pass-through service definition and rewrite rules for Knox & Confluent Kafka REST (v3.2.1).

Only supports producer at this time.


rmkellogg:
Pull request resolved via direct JIRA patch.

PR #7, opened 2017-05-09
KNOX-931:asType will not work for body filter when using IdentityAsse…

Wancy:
…rterFilter


risdenk:
Closing since file related changes don't match PR title.

PR #8, opened 2017-05-24
Provide informative error in place on NPE

rvesse:
When Shiro provider is used if the user doesn't enable authentication or
enables anonymous authentication then Knox will hit an NPE because it
assumes the Shiro Subject is populated.  This commit adds a check and
adds a specific error message which indicates what the problem is and
how to fix it


rvesse:
Closing in favour of open review board request and JIRA https://issues.apache.org/jira/browse/KNOX-966

PR #17, opened 2018-04-24
KNOX-1275 Add json-smart explicitly to gateway-spi

saravankrish:


risdenk:
@saravankrish - I think this PR can be closed since the change was committed. Can you close it?

PR #34, opened 2018-10-05
KNOX-1517 - Tests should be able to run from IDE

risdenk:
This is for review only. Patches will be attached to KNOX-1517


risdenk:
Closing this since it causes other problems later in the build.

PR #48, opened 2018-12-21
KNOX-1679 - Add Alias API service

moresandeep:


gateway-service-admin/pom.xml
```diff
@@ -90,6 +90,14 @@
             <groupId>com.sun.xml.bind</groupId>
             <artifactId>jaxb-impl</artifactId>
         </dependency>
+        <dependency>
```
risdenk: Do we need a full commons-lang3 dependency just to check if a string is blank? Seems overkill.

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
```diff
@@ -0,0 +1,220 @@
...
+   */
+  @POST
+  @Produces({ APPLICATION_JSON })
+  @Consumes(APPLICATION_FORM_URLENCODED)
+  @Path(ALIAS_API_PATH)
+  public Response postAlias(final @PathParam("topology") String topology,
+      final @PathParam("alias") String alias,
+      final @FormParam("value") String value) {
+
+    if (StringUtils.isBlank(value)) {
```
risdenk: Seems overkill for a full commons-langs3 dependency.

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
```diff
@@ -0,0 +1,220 @@
...
+  private static final String ALIASES_API_PATH = "aliases";
+  private static final String ALIASES_TOPOLOGY_API_PATH =
+      ALIASES_API_PATH + "/{topology}";
+  private static final String ALIAS_API_PATH =
+      ALIASES_TOPOLOGY_API_PATH + "/{alias}";
+
+  @Context
+  private HttpServletRequest request;
+
+  public AliasResource() {
```
risdenk: Unnecessary super() call.

gateway-test/src/test/java/org/apache/knox/gateway/GatewayAdminFuncTest.java
```diff
@@ -159,4 +160,81 @@ public void testAdminService() throws ClassNotFoundException {
 
     TestUtils.LOG_EXIT();
   }
+
+  @Test(timeout = TestUtils.MEDIUM_TIMEOUT)
+  public void testAliasAPI() {
+    String username = "guest";
+    String password = "guest-password";
+
+    String addAliasUrl = clusterUrl + "/api/v1/aliases/test-cluster/myalias";
```
risdenk: Seems weird to be hardcoded here and in the AliasResource class. Can there be a constant for this?

gateway-service-admin/pom.xml
moresandeep: Well, you are right, I prefer all the StringUtils handling of nulls, but we can do with 
`value != null && ! value.isEmpty()`

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
moresandeep: super() stuck around as a part of best practice and habit, I don't see any issues with it.

gateway-test/src/test/java/org/apache/knox/gateway/GatewayAdminFuncTest.java
moresandeep: good catch, this can be a constant.

PR #56, opened 2019-02-21
KNOX-1756 - Knox Gateway TLS Keystore and Alias Should be Configurable

rlevas:

**Problem**
The location of the keystore housing the Knox Gateway TLS certificate is hardcoded to `<calculated from configs>/keystores/gateway.jks` and the certificate alias is hardcoded to “`gateway-identity`”. This limits the ability for external management facilities to setup a custom TLS key and certificate for the Knox Gateway. For example, a host-wide, CA-signed, TLS certificate.

Knox has configuration hooks for the following (optional) properties
- Home Directory - GatewayConfig.getGatewayHomeDir()
   - Gateway-site property: GATEWAY_HOME
   - System property: GATEWAY_HOME
   - Environment variable: GATEWAY_HOME
- Data Directory - GatewayConfig.getGatewayDataDir()
   - System property: GATEWAY_DATA_HOME
   - Environment variable: GATEWAY_DATA_HOME
   - Gateway-site property: gateway.data.dir
   - Calculated: [Home Directory] + [Path Separator] + “data”
- Security Directory - GatewayConfig.getGatewaySecurityDir()
   - Gateway-site property: gateway.security.dir
   - Calculated: [Data Directory] + [Path Separator] + “security”

*Note*: the calculation for the home directory is inconsistent with the other directory calculations. This inconsistency may be confusing to users and thus should be fixed to be
- System property: GATEWAY_HOME
- Environment variable: GATEWAY_HOME
- Gateway-site property: gateway.home.dir

The path to the Knox Gateway TLS keystore is calculated as
```
[Security Directory] + [Path Separator] + “keystores” + [Path Separator] + “gateway.jks”
```

**Solution**
To make it easier to use an externally provided TLS key and certificate, the Knox Gateway should allow the TLS keystore file and alias name to be configurable. The following properties should be made available:
- TLS Keystore File Path - GatewayConfig.getIdentityKeystorePath()
   - Gateway-site property: gateway.tls.keystore.path
   - Calculated: [Security Directory] + [Path Separator] + "keystores"  +[Path Separator]+  "gateway.jks"
- TLS Keystore Password Alias (value to be stored in the Knox Gateway credential store) - GatewayConfig.getIdentityKeystorePasswordAlias()
   - Gateway-site property: gateway.tls.keystore.password.alias
   - Calculated: "gateway-identity-keystore-password"
- TLS Keystore Type - GatewayConfig.getIdentityKeystoreType()
   - Gateway-site property: gateway.tls.keystore.type
   - Calculated: :”jks”
- TLS Key Alias - GatewayConfig.getIdentityKeyAlias()
   - Gateway-site property: gateway.tls.key.alias
   - Calculated: “gateway-identity”
- TLS Key Passphrase Alias - GatewayConfig.getIdentityKeyPassphraseAlias()
   - Gateway-site property: gateway.tls.key.passphrase.alias
   - Calculated: "gateway-identity-passphrase"

Additional methods for GatewayConfig should be added to improve consistency and prepare for potential changes to signing key related configurations
- Keystore Directory - GatewayConfig.getKeystoreDir()
   - Calculated: [Security Directory] + [Path Separator] + “keystores”
- Signing Keystore File Path - GatewayConfig.getSigningKeystorePath()
   - Calculated: If gateway.signing.keystore.name is set, [Keystore Directory] + [Path Separator] + [Signing Keystore Name]; else [TLS Keystore File Path]
- Signing Keystore File Type - GatewayConfig.getSigningKeystoreType()
   - Calculated: If gateway.signing.keystore.name is set, value of gateway.signing.keystore.type (default: "jks"); else [TLS Keystore File Type]
- Signing Keystore Password Alias - GatewayConfig.getSigningKeystorePasswordAlias()
   - Calculated: If gateway.signing.keystore.name is set, value of gateway.signing.keystore.password.alias (default: "signing.keystore.password"); else [TLS Keystore Password Alias]
- Signing Key Alias - GatewayConfig.getSigningKeyAlias()
   - Calculated: If gateway.signing.keystore.name is set, value of gateway.signing.key.alias (default: "gateway-identity"); else [TLS Key Alias]
- Signing Key Passphrase Alias - GatewayConfig.getSigningKeyPassphraseAlias()
   - Calculated: If gateway.signing.keystore.name is set, value of gateway.signing.key.passphrase.alias (default: "signing.key.passphrase"); else [TLS Key Passphrase Alias]

Note: the calculated values are set so they are backwards compatible with older versions of Knox to ease the upgrading process.

Hardcoded values related to the identity and signing keystore location and relevant alias names should be removed, using the GatewayConfig methods to provide configured or calculated values.  Assumptions that the keystore and key passwords are the master secret should be removed; however, if a password cannot be found via the AliasService, the fallback value should the master secret. 

## How was this patch tested?

Updated unit tests

Manually tested using
- default keystore for identity and signing
- custom keystore for identity, default keystore for signing (falls to custom keystore for identitiy)
- custom keystore for identity, custom keystore for signing 


gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -300,6 +310,27 @@ public Key getKeyForGateway(String alias, char[] passphrase) throws KeystoreServ
...
 
+  @Override
+  public Key getKeyForGateway(char[] passphrase) throws KeystoreServiceException {
+    return getKeyForGateway(config.getIdentityKeyAlias(), passphrase);
+  }
+
+  @Override
+  public Certificate getCertificateForGateway() throws KeystoreServiceException, KeyStoreException {
+    Certificate cert = null;
+    readLock.lock();
```
smolnar82: Just a minor note: `getKeystoreForGateway` already applies read lock; do we need to lock here? I mean the only critical step here is to fetch the `ks` instance which is already guarded by the `readLock`.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -450,6 +481,22 @@ private void removeFromCache(String clusterName, String alias) {
 
   @Override
   public String getKeystorePath() {
-    return keyStoreDir + GATEWAY_KEYSTORE;
+    return config.getIdentityKeystorePath();
+  }
+
+  private char[] getIdentityKeystorePassword() throws KeystoreServiceException {
```
smolnar82: `getIdentityKeystorePassword` and `getSigningKeystorePassword` has the same implementation except for the password alias they are fetching the credential for. What if you create a new private method `getKeystorePassword(String passwordAlias)` and you invoke this one with the proper params (`config.getIdentityKeystorePasswordAlias()` or `config.getSigningKeystorePasswordAlias()`)?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/RemoteAliasService.java
```diff
@@ -379,8 +378,44 @@ public void generateAliasForCluster(final String clusterName,
...
+  public char[] getGatewayIdentityKeystorePassword() throws AliasServiceException {
+    char[] passphrase = getPasswordFromAliasForGateway(config.getIdentityKeystorePasswordAlias());
+    if (passphrase == null) {
+      passphrase = ms.getMasterSecret();
+    }
+    return passphrase;
+  }
+
+  @Override
+  public char[] getSigningKeyPassphrase() throws AliasServiceException {
```
smolnar82: This is almost the same as `getGatewayIdentityPassphrase` except for the passphrase alias; consider creating a private method and invoke it using the proper parameters.

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -652,34 +653,49 @@ public void execute() throws Exception {
...
-         }
-         else {
+         } else {
 //           log.keyStoreForGatewayFoundNotCreating();
          }
-         char[] passphrase = as.getPasswordFromAliasForCluster(GATEWAY_CREDENTIAL_STORE_NAME, GATEWAY_IDENTITY_PASSPHRASE);
-         if (passphrase == null) {
-           MasterService ms = services.getService("MasterService");
-           passphrase = ms.getMasterSecret();
+         boolean isSelfSigned;
```
smolnar82: Moving this declaration where you initiate the variable?

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestConfig.java
```diff
@@ -91,6 +91,11 @@ public String getGatewaySecurityDir() {
     return gatewayHomeDir + "/security";
   }
 
+  @Override
+  public String getGatewayKeystoreDir() {
+    return getGatewaySecurityDir() + "/keystores";
```
risdenk: Use Paths.get or non os specific separator

gateway-test/src/test/java/org/apache/knox/gateway/AmbariServiceDefinitionTest.java
```diff
@@ -352,4 +366,28 @@ public void contextPathInViewsResponse() throws Exception {
     LOG_EXIT();
   }
 
+  private static void createTestKeystore()
```
risdenk: Duplicated in gatewaytestdriver?

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestConfig.java
```diff
@@ -135,6 +140,31 @@ public long getGatewayIdleTimeout() {
     return 0L;
   }
 
+  @Override
+  public String getIdentityKeystorePath() {
+    return getGatewayKeystoreDir() + "/" + DEFAULT_GATEWAY_KEYSTORE_NAME;
```
risdenk: Use Paths.get or non os specific separator

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestDriver.java
```diff
@@ -261,6 +273,39 @@ public String getLdapUrl() {
...
+    keyStore.setCertificateEntry(alias, cert);
+    keyStore.setKeyEntry(alias, keyPair.getPrivate(), password, new java.security.cert.Certificate[]{cert});
+    try (OutputStream out = Files.newOutputStream(keystoreFile)) {
+      keyStore.store(out, password);
+    }
+  }
+
+  public void cleanupTestKeystore(GatewayConfig config) {
+    try {
+      Files.delete(Paths.get(config.getIdentityKeystorePath()).toAbsolutePath());
```
risdenk: Fileutils delete quietly maybe?

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestConfig.java
```diff
@@ -376,9 +406,29 @@ public String getSigningKeystoreName() {
     return null;
   }
 
+  @Override
+  public String getSigningKeystorePath() {
+    return getGatewayKeystoreDir() + "/" + DEFAULT_GATEWAY_KEYSTORE_NAME;
```
risdenk: Use Paths.get or non os specific separator

gateway-server/src/test/java/org/apache/knox/gateway/websockets/WebsocketEchoTest.java
```diff
@@ -369,4 +447,28 @@ public void handleTopologyEvent(List<TopologyEvent> events) {
       }
     }
   }
+
+  private static void createTestKeystore()
```
risdenk: Duplicated in gatewaytestdriver?

gateway-server/src/test/java/org/apache/knox/gateway/websockets/BadUrlTest.java
```diff
@@ -290,4 +369,28 @@ public void handleTopologyEvent(List<TopologyEvent> events) {
       }
     }
   }
+
+  private static void createTestKeystore()
```
risdenk: Duplicated in gatewaytestdriver?

gateway-server/src/test/java/org/apache/knox/gateway/config/impl/GatewayConfigImplTest.java
```diff
@@ -292,5 +295,55 @@ public void testGetReadOnlyOverrideTopologyNames() {
...
+
+    config.set("gateway.signing.key.alias", "custom_key_alias");
+    config.set("gateway.signing.key.passphrase.alias", "custom_key_passphrase_alias");
+    config.set("gateway.signing.keystore.name", "custom_keystore_name");
+    config.set("gateway.signing.keystore.type", "PKCS12");
+    config.set("gateway.signing.keystore.password.alias", "custom_keystore_password_alias");
+
+    assertEquals("custom_key_alias", config.getIdentityKeyAlias());
+    assertEquals("custom_key_passphrase_alias", config.getIdentityKeyPassphraseAlias());
+    assertEquals("/custom/keystore/path/keystore.p12", config.getIdentityKeystorePath());
```
risdenk: Might to be OS non specific separator?

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/Pac4jMessages.java
```diff
@@ -47,4 +47,8 @@
   @Message( level = MessageLevel.INFO, text =
       "No private key passphrase alias found. Defaulting to master. Exception encountered: {0}")
   void noPrivateKeyPasshraseProvisioned(Exception e);
+
+  @Message( level = MessageLevel.ERROR, text =
+      "No keystore password alias found. Defaulting to master. Exception encountered: {0}")
```
risdenk: Defaulting to master secret

gateway-server/src/test/java/org/apache/knox/gateway/websockets/WebsocketMultipleConnectionTest.java
```diff
@@ -352,4 +430,28 @@ public void onOpen(Session session, EndpointConfig config) {
       session.addMessageHandler(this);
     }
   }
+
+  private static void createTestKeystore()
+      throws NoSuchAlgorithmException, CertificateException, KeyStoreException, IOException {
```
risdenk: Duplicated in gatewaytestdriver?

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
```diff
@@ -190,8 +190,18 @@ private void addDefaultConfig(String clientNameParameter, Map<String, String> pr
       properties.put(PropertiesConfigFactory.SAML_KEYSTORE_PATH,
           keystoreService.getKeystorePath());
 
-      properties.put(PropertiesConfigFactory.SAML_KEYSTORE_PASSWORD,
-          new String(masterService.getMasterSecret()));
+      // check for provisioned alias for keystore password
+      char[] giksp = null;
```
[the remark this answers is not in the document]
rlevas: It is possible, but I was trying to keep with the style of the file (and method impl) where `gip` (Gateway Identity Password) was already used.  In an effort to not add more lines to the patch, I left `giksp` and `gip` rather than changing both to `gatewayIdentityKeystorePassword` and `gatewayIdentityPassword` or `identityKeystorePassword` and `identityPassword`, or whatnot.

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
```diff
@@ -201,15 +211,11 @@ private void addDefaultConfig(String clientNameParameter, Map<String, String> pr
       catch(AliasServiceException ase) {
         log.noPrivateKeyPasshraseProvisioned(ase);
       }
-      if (gip != null) {
-        properties.put(PropertiesConfigFactory.SAML_PRIVATE_KEY_PASSWORD,
-            new String(gip));
-      }
-      else {
+      if (gip == null) {
```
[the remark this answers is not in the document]
rlevas: That would be outside the scope of this patch. But I would have loved to do it.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultAliasService.java
```diff
@@ -39,14 +39,20 @@
 public class DefaultAliasService implements AliasService {
   private static final GatewayMessages LOG = MessagesFactory.get( GatewayMessages.class );
 
-  private static final String GATEWAY_IDENTITY_PASSPHRASE = "gateway-identity-passphrase";
+  protected static char[] chars = { 'a', 'b', 'c', 'd', 'e', 'f', 'g',
```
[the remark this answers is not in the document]
rlevas: I will look into that. Must be a merging issue. Thanks.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
rlevas: Agree.. nice catch. Fixing.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
rlevas: Agreed... will change.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/RemoteAliasService.java
rlevas: I thought of this, but the fall back call `getSigningKeystorePassword` vs `getSigningKeystorePassword` would cause the need to provide some implemented interface.  I think that may be more work than the duplicated code.  However `getSigningKeystorePassword` and `getGatewayIdentityKeystorePassword` may all into this category for code reuse.

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestConfig.java
rlevas: I thought about that since I did it other places, but it goes against the general style of the file... so I didn't do it that way.  I will fix though.

rlevas: Should I convert all of the path concatenations in the file?

gateway-test/src/test/java/org/apache/knox/gateway/AmbariServiceDefinitionTest.java
rlevas: It is.. I had issue with that too, but I was not sure if introducing the GatewayDriver was appropriate.  If we think it is ok, then I am all for the code reuse.

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestConfig.java
risdenk: Ew - didn't realize that there were other examples that are hardcoded with `/`. I think either we fix them all in this file or don't fix it with this PR and do it in a followup.

gateway-test/src/test/java/org/apache/knox/gateway/AmbariServiceDefinitionTest.java
risdenk: Hmmmm - Not sure if it is possible to move the create test keystore method from `GatewayTestDriver` to `gateway-test-utils` module `TestUtils`? It looks like it relies on stuff in `gateway-server` like `X509CertificateUtil`. 

Just seems really ugly to have this block of code copied around.

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/Pac4jMessages.java
rlevas: The message in the line above reads
```
No private key passphrase alias found. Defaulting to master. Exception encountered: {0}
```

Should I change both?

risdenk: Yea probably.

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestConfig.java
rlevas: I am happy to fix them all now.

gateway-test/src/test/java/org/apache/knox/gateway/AmbariServiceDefinitionTest.java
risdenk: Actually - `X509CertificateUtil` is in `gateway-spi` so shouldn't be too bad to move to `TestUtils`. This should avoid needing an additional dependency in most cases. Much easier to add a test dependency on `gateway-test-utils` than on `gateway-test-release-utils`.

rlevas: Adding gateway-spi as a dependency of gateway-test-utils creates a cycle...
```
[ERROR] The projects in the reactor contain a cyclic reference: Edge between 'Vertex{label='org.apache.knox:gateway-spi:1.3.0-SNAPSHOT'}' and 'Vertex{label='org.apache.knox:gateway-i18n:1.3.0-SNAPSHOT'}' introduces to cycle in the graph org.apache.knox:gateway-i18n:1.3.0-SNAPSHOT --> org.apache.knox:gateway-test-utils:1.3.0-SNAPSHOT --> org.apache.knox:gateway-spi:1.3.0-SNAPSHOT --> org.apache.knox:gateway-i18n:1.3.0-SNAPSHOT -> [Help 1]
```

rlevas: I will copy the `X509CertificateUtil` into the project to make things simple.

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestConfig.java
rlevas: fixing.

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestDriver.java
rlevas: Fixing

gateway-test-release-utils/src/main/java/org/apache/knox/gateway/GatewayTestConfig.java
rlevas: fixing

gateway-server/src/test/java/org/apache/knox/gateway/websockets/WebsocketEchoTest.java
rlevas: fixing

gateway-server/src/test/java/org/apache/knox/gateway/websockets/BadUrlTest.java
rlevas: fixing

gateway-server/src/test/java/org/apache/knox/gateway/config/impl/GatewayConfigImplTest.java
rlevas: fixing

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/Pac4jMessages.java
rlevas: fixing

gateway-server/src/test/java/org/apache/knox/gateway/websockets/WebsocketMultipleConnectionTest.java
rlevas: fixing

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
rlevas: fixing

risdenk:
Changes look reasonable but getting test failures with `mvn -T.5C verify -Prelease,package`

```
[ERROR] Errors:
[ERROR] org.apache.knox.gateway.GatewayDeployFuncTest.testDeployRedeployUndeploy(org.apache.knox.gateway.GatewayDeployFuncTest)
[ERROR]   Run 1: GatewayDeployFuncTest.setupGateway:110 » NullPointer
[ERROR]   Run 2: GatewayDeployFuncTest.cleanupGateway:121 NullPointer
[INFO]
[ERROR] org.apache.knox.gateway.GatewayMultiFuncTest.org.apache.knox.gateway.GatewayMultiFuncTest
[ERROR]   Run 1: GatewayMultiFuncTest.setUpBeforeClass:87->setupGateway:124->startGatewayServer:139 » NullPointer
[ERROR]   Run 2: GatewayMultiFuncTest.tearDownAfterClass:94 NullPointer
[INFO]
[ERROR] org.apache.knox.gateway.GatewaySslFuncTest.org.apache.knox.gateway.GatewaySslFuncTest
[ERROR]   Run 1: GatewaySslFuncTest.setUpBeforeClass:100->setupGateway:141->startGatewayServer:170 » NullPointer
[ERROR]   Run 2: GatewaySslFuncTest.tearDownAfterClass:107 NullPointer
[INFO]
[INFO]
[ERROR] Tests run: 62, Failures: 0, Errors: 3, Skipped: 3
```

The three tests above I checked in an IDE and they all broke at `DefaultKeystoreService` line 172 checking if the configured keystore exists or not.

risdenk:
All tests passed locally now. Performing some sanity checks to make sure this works as designed.

risdenk:
@rlevas while testing I noticed the following:

- There are strings like the following still present when there are errors loading the certificate "provided (or persisted) master secret"
- `./bin/knoxcli.sh create-cert` seemed to overwrite my existing certificate. Checking into this more to see what happened here.

risdenk:
For the `./bin/knoxcli.sh create-cert`, I was using a self signed certificate for testing and create-cert overwrites the existing certificate if it is self signed. Based on rereading the PR this is expected behavior.

I guess I was expecting to have to force to overwrite the existing certificate since it was a certificate that I had created.

risdenk:
Found an interesting edge case when configuring Knox.

If you configured a custom signing certificate in `gateway-site.xml` and then go to create the aliases with `./bin/knoxcli.sh create-alias ...` you will run into the following error:

```
2019-02-22 14:25:36,775 ERROR knox.gateway (BaseKeystoreService.java:getKeystore(113)) - Failed to load keystore [filename=keystore-signing.jks, type=JKS]: java.io.IOException: Keystore was tampered with, or password was incorrect
```

`DefaultKeystoreService.init(...)` is doing checks to see if the signing certificate is setup correctly. This was not added in this PR but breaks the ability to add aliases if there is a custom signing certificate not fully setup. The check in `DefaultKeystoreService` should be removed because it doesn't belong in the `KeystoreService`. The `KeystoreService` impl should be generic to what is stored in the keystore. Currently the signing certificate check breaks that design.

rlevas:
> For the `./bin/knoxcli.sh create-cert`, I was using a self signed certificate for testing and create-cert overwrites the existing certificate if it is self signed. Based on rereading the PR this is expected behavior.
> 
> I guess I was expecting to have to force to overwrite the existing certificate since it was a certificate that I had created.

Would you like this behavior changed to require `--force` if the entry exists in the keystore?  However I think overwriting was the behavior before the changes in this patch.

risdenk:
`Would you like this behavior changed to require `--force` if the entry exists in the keystore? However I think overwriting was the behavior before the changes in this patch.`

I think the existing behavior made sense when it was in `data/security/keystores/gateway.jks`. This was a path that was fully owned by Knox. Since we are now following the path to the custom keystore, we should be more careful about what we overwrite. Not sure how easy it would be to check if the identity path is the default vs a configured one?

rlevas:
> The check in `DefaultKeystoreService` should be removed because it doesn't belong in the `KeystoreService`. The `KeystoreService` impl should be generic to what is stored in the keystore. Currently the signing certificate check breaks that design.

Would you like this changed in this patch.  It seems like maybe it would fall under a different issue, but I will fix it here is desired.

rlevas:
> I think the existing behavior made sense when it was in `data/security/keystores/gateway.jks`. This was a path that was fully owned by Knox. Since we are now following the path to the custom keystore, we should be more careful about what we overwrite. Not sure how easy it would be to check if the identity path is the default vs a configured one?

I guess we can see if the path is the expected default and maybe also see if the cert is self-signed and has the DN of the Gateway-created one. If so, we blindly overwrite, else we require a `--force`. This seems like it may fall under a different issue, but I can fix it in this patch if desired.

risdenk:
> Would you like this changed in this patch. It seems like maybe it would fall under a different issue, but I will fix it here is desired.

@rlevas can you file a separate JIRA to fix this? They are definitely related but can be fixed independent of this PR.

There is an easy work around, define the aliases upfront before configuring `gateway-site.xml`. This wouldn't work in a Docker scenario potentially but something that would be possible in a lot of cases.

risdenk:
> I guess we can see if the path is the expected default and maybe also see if the cert is self-signed and has the DN of the Gateway-created one. If so, we blindly overwrite, else we require a `--force`. This seems like it may fall under a different issue, but I can fix it in this patch if desired.

I think it should be fixed here since the check for self-signed was introduced in this patch. I worry that someone will have their own self signed certificate overridden by mistake if we introduce this patch.

rlevas:
> * Create a JIRA to address the `DefaultKeystoreService` check issue

Created https://issues.apache.org/jira/browse/KNOX-1793.

rlevas:
@risdenk tasks completed with latest push...
- [x] Fix the logging messages to be consistent - remove remaining references to "provided (or persisted) master secret"
- [x] Address the ./bin/knoxcli.sh create-cert check for selfsigned to check if it is the default path

risdenk:
Manual tests all look good as well. Ran the following tests:

* Checked that with no configuration Knox still works
* Checked that TLS and signed certificate configurations work independently
* Misconfiguring on purpose TLS/signed configuration properties
* Checked that error messages are reasonable
* Checked both self signed and CA signed certificates work correctly
* Ensure that JKS and PKCS12 certificates work

PR #57, opened 2019-02-22
KNOX-1191 - Azure AD support for KnoxSSO

moresandeep:

Currently OIDC support for Azure AD was broken. This is because Azure AD does not allow ? in callback urls. This fix updates the callback flow in Knox so that the callback urls now can have callback parameter as a path element rather than a query parameter.

## How was this patch tested?

This patch was manually tested with Azure AD (OIDC), Google (OIDC) and Okta (SAML) authentication providers.

### DIFF
--- a/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
+++ b/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
@@ -32,11 +32,13 @@ import org.pac4j.core.client.Client;
 import org.pac4j.core.config.Config;
 import org.pac4j.core.context.session.J2ESessionStore;
 import org.pac4j.core.context.session.SessionStore;
+import org.pac4j.core.http.callback.PathParameterCallbackUrlResolver;
 import org.pac4j.core.util.CommonHelper;
 import org.pac4j.http.client.indirect.IndirectBasicAuthClient;
 import org.pac4j.http.credentials.authenticator.test.SimpleTestUsernamePasswordAuthenticator;
 import org.pac4j.j2e.filter.CallbackFilter;
 import org.pac4j.j2e.filter.SecurityFilter;
+import org.pac4j.oidc.client.AzureAdClient;
 
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
@@ -75,11 +77,17 @@ public class Pac4jDispatcherFilter implements Filter {
 
   public static final String PAC4J_CALLBACK_PARAMETER = "pac4jCallback";
 
+  public static final String PAC4J_AZURE_AD_CLIENT = "AzureAdClient";
+
+  public static final String URL_PATH_SEPARATOR = "/";
+
   private static final String PAC4J_COOKIE_DOMAIN_SUFFIX_PARAM = "pac4j.cookie.domain.suffix";
 
   private static final String PAC4J_CONFIG = "pac4j.config";
 
   private static final String PAC4J_SESSION_STORE = "pac4j.session.store";
+
+  private static final String PAC4J_CLIENT_NAME_PARAM = "clientName";
 
   private CallbackFilter callbackFilter;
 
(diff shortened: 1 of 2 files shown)


gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
```diff
@@ -75,12 +77,18 @@
 
   public static final String PAC4J_CALLBACK_PARAMETER = "pac4jCallback";
 
+  public static final String PAC4J_AZURE_AD_CLIENT = "AzureAdClient";
+
+  public static final String URL_PATH_SEPARATOR = "/";
```
risdenk: nit: I would guess that there is already a constant for this from HTTP client or something?

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/session/KnoxSessionStore.java
```diff
@@ -176,6 +211,25 @@ public void set(WebContext context, String key, Object value) {
...
 
+    /**
+     * Keep only the fileds that are needed for Pac4J.
+     * Used to reduce the cookie size.
+     * @param value profile object
+     * @return trimmed profile object
+     * @since 1.3.0
+     */
+    private Object clearUserProfile(final Object value) {
+        if(value instanceof LinkedHashMap<?,?>) {
```
risdenk: Could this be any Map instead of a LinkedHashMap? Just thinking if profiles Map impl could change later but would still have the same underlying interface.

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
moresandeep: Ah, I couldn't find any constants that are exposed.

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
```diff
@@ -160,8 +179,17 @@ public void init( FilterConfig filterConfig ) throws ServletException {
       } else {
         clientName = clientNameParameter;
       }
+
+      /* special handling for Azure AD, use path separators instead of query params */
+      clients.forEach( client -> {
+        if(client.getName().equalsIgnoreCase(PAC4J_AZURE_AD_CLIENT)) {
+          ((AzureAdClient)client).setCallbackUrlResolver(new PathParameterCallbackUrlResolver());
```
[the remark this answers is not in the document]
moresandeep: This is a good suggestion, actually now I think there might be a bug if someone just uses Oidc client and not specify Azure (the code should work for both), I'll fix this, thanks !

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/session/KnoxSessionStore.java
```diff
@@ -143,6 +162,22 @@ public void set(WebContext context, String key, Object value) {
...
+         *  set the correct path for setting pac4j profile cookie.
+         *  This is because, Pac4jDispatcherFilter.PAC4J_CALLBACK_PARAMETER in the path
+         *  indicates callback when ? cannot be used.
+         */
+        if (context.getPath() != null && context.getPath()
+            .contains(Pac4jDispatcherFilter.PAC4J_CALLBACK_PARAMETER)) {
+
+            final String[] parts = ((J2EContext) context).getRequest().getRequestURI()
+                .split(
+                    "websso"+ Pac4jDispatcherFilter.URL_PATH_SEPARATOR + Pac4jDispatcherFilter.PAC4J_CALLBACK_PARAMETER);
```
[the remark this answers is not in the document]
moresandeep: I didn't like this approach either, the problem here is that the string does not end with PAC4J_CALLBACK_PARAMETER, it will end with the name of the client which could be just simple OIDC or subclasses of OIDC clients like AzureADClient. This is due to he path normalization that Pac4J can do for Azure clients.

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/session/KnoxSessionStore.java
moresandeep: Good point, let me try with a generic Map and see how it works.

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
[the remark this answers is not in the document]
moresandeep: Well no bug, I thought you could specify generic client name (i.e. OidcClient and specify oidc.type=azure) in which case the code would have failed,  but looks like you have specify Client name, bummer.

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
risdenk: Well you are probably right since I couldn't find any. Surprising :(

https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/constant-values.html

PR #59, opened 2019-02-25
KNOX-1789 - Refactor RemoteAliasService to use service loading

risdenk:
The RemoteAliasService should use service loading to ensure make the RemoteAliasService extensible. Moved existing Zookeeper impl to its own class and loaded via service loading.

## How was this patch tested?
`mvn -T.5C verify -Ppackage,release`

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/CLIGatewayServices.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/CLIGatewayServices.java
@@ -38,30 +38,14 @@ import java.util.Map;
 
 public class CLIGatewayServices implements GatewayServices {
   private Map<String,Service> services = new HashMap<>();
-  private CLIMasterService ms;
-  private DefaultKeystoreService ks;
-
-  public CLIGatewayServices() {
-    super();
-  }
 
   @Override
   public void init(GatewayConfig config, Map<String,String> options) throws ServiceLifecycleException {
-
-    /* create an instance so that it can be passed to other services */
-    final RemoteAliasService alias = new RemoteAliasService();
-
-    final RemoteConfigurationRegistryClientService registryClientService =
-        RemoteConfigurationRegistryClientServiceFactory.newInstance(config);
-    registryClientService.setAliasService(alias);
-    registryClientService.init(config, options);
-    services.put(REMOTE_REGISTRY_CLIENT_SERVICE, registryClientService);
-
-    ms = new CLIMasterService();
+    CLIMasterService ms = new CLIMasterService();
     ms.init(config, options);
     services.put(MASTER_SERVICE, ms);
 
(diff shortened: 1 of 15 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/CLIGatewayServices.java
```diff
@@ -70,16 +54,21 @@ public void init(GatewayConfig config, Map<String,String> options) throws Servic
     defaultAlias.setKeystoreService(ks);
     defaultAlias.init(config, options);
 
+    final RemoteConfigurationRegistryClientService registryClientService =
+        RemoteConfigurationRegistryClientServiceFactory.newInstance(config);
+    registryClientService.setAliasService(defaultAlias);
```
risdenk: Doesn't make sense for this to be set to the remote alias service since this could be remote itself. This uses the default alias service in case of digest authentication.

If digest auth and using remote alias service, then wouldn't be able to connect to ZK anyway.

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasServiceTest.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.knox.gateway.security.impl;
+package org.apache.knox.gateway.services.security.impl;
```
risdenk: Moved to match ZookeeperRemoteAliasService class package

risdenk:
The existing tests for `ZookeeperRemoteAliasService` don't cover the `RemoteAliasService` service loading. Will need to add a test for that.

gateway-server/src/main/java/org/apache/knox/gateway/services/DefaultGatewayServices.java
```diff
@@ -48,50 +49,40 @@
...
-    final RemoteAliasService alias = new RemoteAliasService();
+    final DefaultAliasService defaultAlias = new DefaultAliasService();
+    defaultAlias.setKeystoreService(ks);
+    defaultAlias.setMasterService(ms);
+    defaultAlias.init(config, options);
 
     final RemoteConfigurationRegistryClientService registryClientService =
         RemoteConfigurationRegistryClientServiceFactory.newInstance(config);
-    registryClientService.setAliasService(alias);
+    registryClientService.setAliasService(defaultAlias);
```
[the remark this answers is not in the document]
risdenk: Ah yea good call. I'll update to add that comment

risdenk:
Pushed change to update the comment.

risdenk:
@moresandeep @lmccay @pzampino any thoughts on this PR?

PR #60, opened 2019-02-26
KNOX-1418 - New KnoxShell command to build truststore using the gateway server's public certificate

smolnar82:

Currently, the KnoxShell setup requires some manual steps to login to the machine where the gateway server is located and execute `knoxcli.sh export-cert –type JKS` then copy it to the current user's home.
To make it easier for our end-users a new KnoxShell command was added to do this work:
`buildTrustStore <knox-gateway-url>`

## How was this patch tested?

Executing JUnit tests (including integration tests):
```
[INFO] BUILD SUCCESS
[INFO] Total time: 17:47 min (Wall Clock)
[INFO] Finished at: 2019-02-26T21:21:54+01:00
[INFO] Final Memory: 267M/1641M
```
Additionally, the following manual test steps were executed:
1. Unzipped the updated version of `knoxshell-1.3.0-SNAPSHOT.zip` locally and removed previously created `~/gateway-client-trust.jks`
2. Checked if the new command is available in KnoxShell's help:
```
$ ./bin/knoxshell.sh help

Apache Knox Client Shell
The client shell facility provide a CLI for establishing and managing Apache Knox Sessions
and executing the Apache Knox groovy-based DSL scripts. It may also be used to enter an
interactive shell where groovy-based DSL and groovy code may be entered and executed in realtime.

knoxshell usage: 
   knoxshell.sh [[buildTrustStore <knox-gateway-url>|init <topology-url>|list|destroy|help] | [<script-file-name>]]
   ----------------------------------------------------------
   buildTrustStore <knox-gateway-url> - downloads the given gateway server's public certificate and builds a trust store to be used by KnoxShell
        example: knoxshell.sh buildTrustStore https://localhost:8443/
   init <topology-url> - requests a session from the knox token service at the url
        example: knoxshell.sh init https://localhost:8443/gateway/sandbox
   list - lists the details of the cached knox session token
        example: knoxshell.sh list
   destroy - removes the cached knox session token
        example: knoxshell.sh destroy
   <script-file-name> - executes the groovy script file
        example: knoxshell.sh ~/bin/ls.groovy
```
3. Invoked `knoxshell.sh buildTrustStore` without the mandatory `<knox-gateway-url>` parameter:
```
$ ./bin/knoxshell.sh buildTrustStore
Illegal number of parameters.

Apache Knox Client Shell
The client shell facility provide a CLI for establishing and managing Apache Knox Sessions
and executing the Apache Knox groovy-based DSL scripts. It may also be used to enter an
interactive shell where groovy-based DSL and groovy code may be entered and executed in realtime.

knoxshell usage: 
   knoxshell.sh [[buildTrustStore <knox-gateway-url>|init <topology-url>|list|destroy|help] | [<script-file-name>]]
   ----------------------------------------------------------
   buildTrustStore <knox-gateway-url> - downloads the given gateway server's public certificate and builds a trust store to be used by KnoxShell
        example: knoxshell.sh buildTrustStore https://localhost:8443/
...
```
4. Tested if trust store was built using a valid gateway server's cert and the trust store is OK to run KnoxShell samples:
```
$ ls -al ~/gateway-client-trust.jks
ls: /Users/smolnar/gateway-client-trust.jks: No such file or directory

$ ./bin/knoxshell.sh buildTrustStore https://c7401.ambari.apache.org:8443/
Opening connection to c7401.ambari.apache.org:8443...
Starting SSL handshake...
SSL exception; found non-trusted certificate
Gateway server's certificate is exported into /Users/smolnar/gateway-client-trust.jks

$ ls -al /Users/smolnar/gateway-client-trust.jks
-rw-r--r--  1 smolnar  staff  674 Feb 26 21:26 /Users/smolnar/gateway-client-trust.jks

$ ./bin/knoxshell.sh samples/ExampleWebHdfsLs.groovy
Enter username: guest
Enter password: 
[app-logs, ats, atsv2, hdp, mapred, mr-history, tmp, user]
```
5.  Tested if trust store was built using another (non-gateway) server's cert and running a KnoxShell sample failed:
```
$ ./bin/knoxshell.sh buildTrustStore https://google.com:443/
Opening connection to google.com:443...
Starting SSL handshake...
SSL exception; found non-trusted certificate
Gateway server's certificate is exported into /Users/smolnar/gateway-client-trust.jks

$ ls -al ~/gateway-client-trust.jks
-rw-r--r--  1 smolnar  staff  2068 Feb 26 21:45 /Users/smolnar/gateway-client-trust.jks

$ ./bin/knoxshell.sh samples/ExampleWebHdfsLs.groovy
Enter username: guest
Enter password: 
Caught: org.apache.knox.gateway.shell.KnoxShellException: javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
org.apache.knox.gateway.shell.KnoxShellException: javax.net.ssl.SSLHandshakeException: sun.security.validator.ValidatorException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target
	at org.apache.knox.gateway.shell.AbstractRequest.now(AbstractRequest.java:81)
	at org.apache.knox.gateway.shell.AbstractRequest$now.call(Unknown Source)
	at ExampleWebHdfsLs.run(ExampleWebHdfsLs.groovy:37)
	at org.apache.knox.gateway.shell.Shell.main(Shell.java:58)
	at org.apache.knox.gateway.launcher.Invoker.invokeMainMethod(Invoker.java:68)
	at org.apache.knox.gateway.launcher.Invoker.invoke(Invoker.java:39)
	at org.apache.knox.gateway.launcher.Command.run(Command.java:99)
	at org.apache.knox.gateway.launcher.Launcher.run(Launcher.java:75)
	at org.apache.knox.gateway.launcher.Launcher.main(Launcher.java:52)
```

Also tested the `init|list|destroy` commands to make sure my bash changes did not screw up anything.


smolnar82:
@risdenk @pzampino @rlevas @lmccay @moresandeep Could you please review my changes?
Thanks!

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +170,81 @@ public boolean validate() {
...
+
+    private static final String USAGE = "buildTrustStore --gateway server-url";
+    private static final String DESC = "Downloads the gateway server's public certificate and builds a trust store.";
+    private static final String CLIENT_TRUST_STORE_FILE_NAME = "gateway-client-trust.jks";
+
+    @Override
+    public void execute() throws Exception {
+      final X509Certificate gatewayServerPublicCert = fetchPublicCertFromGatewayServer();
+      if (gatewayServerPublicCert != null) {
+        final File trustStoreFile = new File(System.getProperty("user.home"), CLIENT_TRUST_STORE_FILE_NAME);
```
risdenk: Can this be `Paths.get` instead?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +170,81 @@ public boolean validate() {
...
+      sslContext.init(null, new TrustManager[] { trustManagerWithCertificateChain }, null);
+
+      final URI uri = URI.create(gateway);
+      out.println("Opening connection to " + uri.getHost() + ":" + uri.getPort() + "...");
+      final SSLSocket socket = (SSLSocket) sslContext.getSocketFactory().createSocket(uri.getHost(), uri.getPort());
+      socket.setSoTimeout(10000);
+      try {
+        out.println("Starting SSL handshake...");
+        socket.startHandshake();
+        socket.close();
```
risdenk: use try-with-resources on the Socket? `try ()` ...

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +170,81 @@ public boolean validate() {
...
+        out.println();
+        out.println("No errors, certificate is already trusted");
+      } catch (SSLException e) {
+        // NOP; this is expected in case the gateway server's certificate is not in the trust store the JVM uses
+        out.println("SSL exception; found non-trusted certificate");
```
risdenk: Do we need this message printed out?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +170,81 @@ public boolean validate() {
...
+      }
+
+      @Override
+      public X509Certificate[] getAcceptedIssuers() {
+        throw new UnsupportedOperationException();
+      }
+
+      @Override
+      public void checkClientTrusted(X509Certificate[] chain, String authType) throws CertificateException {
+        throw new UnsupportedOperationException();
```
risdenk: If Knox is setup to require client auth? Does this break?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +170,81 @@ public boolean validate() {
...
+        socket.startHandshake();
+        socket.close();
+        out.println();
+        out.println("No errors, certificate is already trusted");
+      } catch (SSLException e) {
+        // NOP; this is expected in case the gateway server's certificate is not in the trust store the JVM uses
+        out.println("SSL exception; found non-trusted certificate");
+      }
+
+      return trustManagerWithCertificateChain.certificateChain == null ? null : trustManagerWithCertificateChain.certificateChain[0];
```
risdenk: Will this return just the one public certificate or the chain of trust? The reason the chain is important is because there could be multiple intermediate certificates.

risdenk: Or are we just going to directly trust the final server certificate?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
risdenk: Need to make sure we close this socket on any failures.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +170,81 @@ public boolean validate() {
...
+
+    @Override
+    public void execute() throws Exception {
+      final X509Certificate gatewayServerPublicCert = fetchPublicCertFromGatewayServer();
+      if (gatewayServerPublicCert != null) {
+        final File trustStoreFile = new File(System.getProperty("user.home"), CLIENT_TRUST_STORE_FILE_NAME);
+        X509CertificateUtil.writeCertificateToJks(gatewayServerPublicCert, trustStoreFile);
+        out.println("Gateway server's certificate is exported into " + trustStoreFile.getAbsolutePath());
+      } else {
+        out.println("Could not obtain server certificate chain");
```
risdenk: Will this happen in a non exception case? Assuming an exception is thrown from fetchPublicCertFromGatewayServer it would propagate up and this message wouldn't be printed.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +170,81 @@ public boolean validate() {
     public abstract String getUsage();
   }
 
+  private class KnoxBuildTrustStore extends Command {
+
+    private static final String USAGE = "buildTrustStore --gateway server-url";
+    private static final String DESC = "Downloads the gateway server's public certificate and builds a trust store.";
+    private static final String CLIENT_TRUST_STORE_FILE_NAME = "gateway-client-trust.jks";
```
risdenk: Do we need to make the path configurable? Say there are multiple Knox instances? Also writing to "user.home" could overwrite a file?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: Socket is not closeable (try-with-resources does not work); but you are right; a finally block is needed

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: no

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: Yes; I'll fix it.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: It's enough; right? We only need to have one certificate in the KnoxShell's truststore with 'gateway-identity` alias.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: Fixing...

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: Hm...currently KnoxShell supports the following scenarios when building up KnoxSession:
truststore related attributes are either passed as environment variables (`KNOX_CLIENT_TRUSTSTORE_DIR`, `KNOX_CLIENT_TRUSTSTORE_FILENAME` and `KNOX_CLIENT_TRUSTSTORE_PASS`) or KnoxShell uses default values (`$USER_HOME`, `gateway-client-trust.jks` and `changeit`).
I can modify the new command to consider the same environment variables but I would rather enhance it by letting end-users to pass these attributes as optional program arguments.
If you agree I'd let this now as is and create a new JIRA to address this area. Any objection?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: Sure

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
risdenk: I think we should make it consistent with where the KnoxShell will look for the truststore. Otherwise they could run the new command and then not be able to use it.

Probably makes sense to add helper methods to figure out what the correct paths are using the environment variables or the default values. I think passing into KnoxShell as optional parameters is probably overkill for this.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
risdenk: I'm not sure. The Knox certificate could be externally configured or have a CA signed certificate in `gateway.jks` The server should present all the certificates. It would be good to test this with a locally CA signed certificate.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +171,84 @@ public boolean validate() {
...
+      sslContext.init(null, new TrustManager[] { trustManagerWithCertificateChain }, null);
+
+      final URI uri = URI.create(gateway);
+      out.println("Opening connection to " + uri.getHost() + ":" + uri.getPort() + "...");
+      try (Socket socket = sslContext.getSocketFactory().createSocket(uri.getHost(), uri.getPort())) {
+        socket.setSoTimeout(10000);
+        out.println("Starting SSL handshake...");
+        ((SSLSocket) socket).startHandshake();
+        out.println("No errors, certificate is already trusted");
+        trustManagerWithCertificateChain.serverCertificateChain = null; //we already trust the given site's certs; it does not make sense to build a new truststore
```
risdenk: We could return here right instead of setting to null and then doing the null check later?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +171,84 @@ public boolean validate() {
...
+      String result = GATEWAY_CERT_NOT_EXPORTED;
+      try {
+        final X509Certificate gatewayServerPublicCert = fetchPublicCertFromGatewayServer();
+        if (gatewayServerPublicCert != null) {
+          final File trustStoreFile = Paths.get(System.getProperty("user.home"), KnoxSession.GATEWAY_CLIENT_TRUST).toFile();
+          X509CertificateUtil.writeCertificateToJks(gatewayServerPublicCert, trustStoreFile);
+          result = GATEWAY_CERT_EXPORTED_MESSAGE_PREFIX + trustStoreFile.getAbsolutePath();
+        }
+      } catch(Exception e) {
+        //NOP
```
risdenk: I think in this case, we aren't printing anything on failure. Previously the else part of the if/else would not have been reached since the exception would be thrown by `execute()`. Do we need the try/catch? Can we let the exception propagate? Maybe we should wrap the exception with a nicer message?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
```diff
@@ -155,6 +171,84 @@ public boolean validate() {
...
+    private static final String GATEWAY_CERT_NOT_EXPORTED = "Finished work without building truststore";
+    private static final String GATEWAY_CERT_EXPORTED_MESSAGE_PREFIX = "Gateway server's certificate is exported into ";
+
+    @Override
+    public void execute() throws Exception {
+      String result = GATEWAY_CERT_NOT_EXPORTED;
+      try {
+        final X509Certificate gatewayServerPublicCert = fetchPublicCertFromGatewayServer();
+        if (gatewayServerPublicCert != null) {
+          final File trustStoreFile = Paths.get(System.getProperty("user.home"), KnoxSession.GATEWAY_CLIENT_TRUST).toFile();
```
risdenk: I know there are a few notes about this but wanted to call out that we should make this match with existing behavior with the env variables.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
risdenk: I guess the question I have, is there any harm in returning the whole certificate chain?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: Fine by me; will submit the new patch soon

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: We could; I like to have only one exit (return) point within a method but it's ok for me.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSh.java
smolnar82: Agreed; will submit the new patch soon

risdenk:
Changes look good. Running tests locally before merging.

PR #61, opened 2019-02-26
KNOX-1793 - DefaultKeystoreService should not validate the signing key on initialization

rlevas:

Moved validation to `org.apache.knox.gateway.services.token.impl.DefaultTokenAuthorityService#start`.   

Added call to `org.apache.knox.gateway.services.token.impl.DefaultTokenAuthorityService#start` in `org.apache.knox.gateway.services.DefaultGatewayServices#start`.

Created message resource values in `org.apache.knox.gateway.GatewayResources` to help manage error messages. 

Added new unit tests.

## How was this patch tested?

Ran unit tested

Tested the following scenarios:
* Existing Default signing keystore (gateway.jks) and aliases
* Existing Custom signing keystore and aliaes
* Missing default signing keystore - keystore created on start
* Invalid master key (negative test)
* Missing custom signing keystore (negative test)
* Existing custom signing keystore with invalid password (negative test)
* Existing custom signing keystore with invalid key alias (negative test)
* Existing custom signing keystore with invalid key password (negative test)

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/DefaultGatewayServices.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/DefaultGatewayServices.java
@@ -162,6 +162,8 @@ public class DefaultGatewayServices implements GatewayServices {
     SSLService ssl = (SSLService) services.get(SSL_SERVICE);
     ssl.start();
 
+    (services.get(TOKEN_SERVICE)).start();
+
     ServerInfoService sis = (ServerInfoService) services.get(SERVER_INFO_SERVICE);
     sis.start();
 
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/DefaultGatewayServices.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/DefaultGatewayServices.java
@@ -33,6 +33,7 @@ import org.apache.knox.gateway.services.topology.impl.DefaultTopologyService;
 import org.apache.knox.gateway.services.hostmap.impl.DefaultHostMapperService;
 import org.apache.knox.gateway.services.registry.impl.DefaultServiceRegistryService;
 import org.apache.knox.gateway.services.security.KeystoreServiceException;
+import org.apache.knox.gateway.services.security.SSLService;
 import org.apache.knox.gateway.services.security.impl.DefaultAliasService;
 import org.apache.knox.gateway.services.security.impl.DefaultCryptoService;
 import org.apache.knox.gateway.services.security.impl.DefaultKeystoreService;
@@ -49,33 +50,6 @@ import java.util.Map;
 public class DefaultGatewayServices implements GatewayServices {
 
   private static GatewayMessages log = MessagesFactory.get( GatewayMessages.class );
-
-  private static final String[] SERVICE_START_ORDER = new String[]{
-      MASTER_SERVICE,
-      KEYSTORE_SERVICE,
-      ALIAS_SERVICE,
-      SSL_SERVICE,
-      TOKEN_SERVICE,
-      SERVER_INFO_SERVICE,
-      REMOTE_REGISTRY_CLIENT_SERVICE,
-      CLUSTER_CONFIGURATION_MONITOR_SERVICE,
-      TOPOLOGY_SERVICE,
-      METRICS_SERVICE
-  };
-
-  // TODO: This is the original order of service to stop... however maybe reverse of start should be done instead
-  private static final String[] SERVICE_STOP_ORDER = new String[]{
-      MASTER_SERVICE,
-      KEYSTORE_SERVICE,
-      CLUSTER_CONFIGURATION_MONITOR_SERVICE,
-      ALIAS_SERVICE,
-      SSL_SERVICE,
-      TOKEN_SERVICE,
-      SERVER_INFO_SERVICE,
-      REMOTE_REGISTRY_CLIENT_SERVICE,
-      TOPOLOGY_SERVICE,
-      METRICS_SERVICE
-  };
 
   private Map<String,Service> services = new HashMap<>();
   private DefaultMasterService ms;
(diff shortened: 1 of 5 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -184,7 +191,7 @@ public boolean verifyToken(JWT token, RSAPublicKey publicKey)
     PublicKey key;
     try {
       if (publicKey == null) {
-        key = ks.getSigningKeystore().getCertificate(getSigningKeyAlias(signingKeyAlias)).getPublicKey();
+        key = ks.getSigningKeystore().getCertificate(getSigningKeyAlias(null)).getPublicKey();
```
risdenk: Is `null` correct here? That seems weird...

gateway-server/src/test/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityServiceTest.java
```diff
@@ -327,4 +327,215 @@ public void testTokenCreationCustomSigningKey() throws Exception {
...
+    ks.init(config, new HashMap<>());
+    ks.start();
+
+    DefaultTokenAuthorityService ta = new DefaultTokenAuthorityService();
+    ta.setAliasService(as);
+    ta.setKeystoreService(ks);
+    ta.init(config, new HashMap<>());
+    ta.start();
+
+    EasyMock.verify(config, ms, as);
```
risdenk: Should stop the ks and ts to follow the rest of the lifecycle?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
rlevas: In the original code, `signingKeyAlias` was set to `config.getSigningKeyAlias()`. 

In the new code `getSigningKeyAlias(String signingKeystoreAlias)` is set to return `signingKeystoreAlias` if `signingKeystoreAlias` is not null or `config.getSigningKeyAlias()` if `signingKeystoreAlias` is null.  

This is the same thing.  I don't particularly like the code, but I was trying not to change too much.  Since nothing is broken, I left it.

gateway-server/src/test/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityServiceTest.java
rlevas: This seems unnecessary for the test since we are only testing the start process.  Also both stop operations are no-ops anyway.   Maybe a stop test  (`testServiceStop()`) is in order?

risdenk: Yea I guess thats true. I have experience with Apache Lucene/Solr where they track leaking threads so try to keep an eye out for start/stop patterns.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -205,26 +212,64 @@ public void init(GatewayConfig config, Map<String, String> options)
...
-        if (key == null) {
-          throw new ServiceLifecycleException("Provisioned passphrase cannot be used to acquire signing key.");
-        }
+      if (passphrase == null) {
+        throw new ServiceLifecycleException(RESOURCES.signingKeyPassphraseNotAvailable(config.getSigningKeyAlias()));
       }
-    } catch (AliasServiceException | KeystoreServiceException e) {
-      throw new ServiceLifecycleException("Provisioned signing key passphrase cannot be acquired.", e);
+    } catch (AliasServiceException e) {
+      throw new ServiceLifecycleException(RESOURCES.signingKeyPassphraseNotAvailable(config.getSigningKeyAlias()), e);
```
risdenk: I'm not 100% sure this is correct. I need to pull the code down tomorrow and check. This might be config.getSigningKeyPassphraseAlias?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
risdenk: Ok I'll pull the code down tomorrow into an IDE and take a look to see the flow.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -205,26 +212,64 @@ public void init(GatewayConfig config, Map<String, String> options)
...
     try {
       passphrase = as.getSigningKeyPassphrase();
-      if (passphrase != null) {
-        key = (RSAPrivateKey) ks.getSigningKey(getSigningKeyAlias(signingKeyAlias),
-            passphrase);
-        if (key == null) {
-          throw new ServiceLifecycleException("Provisioned passphrase cannot be used to acquire signing key.");
-        }
+      if (passphrase == null) {
+        throw new ServiceLifecycleException(RESOURCES.signingKeyPassphraseNotAvailable(config.getSigningKeyAlias()));
```
risdenk: should be `config.getSigningKeyPassphraseAlias()`

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
risdenk: Should be `config.getSigningKeyPassphraseAlias()`

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
risdenk: How about something like this:
```
private String getSigningKeyAlias(String signingKeystoreAlias) {
    if(signingKeystoreAlias != null) {
     return signingKeystoreAlias;
    }

    // Fallback to defaults
    return getSigningKeyAlias();
  }

  private String getSigningKeyAlias() {
    String alias = config.getSigningKeyAlias();
    return (alias == null) ? GatewayConfig.DEFAULT_SIGNING_KEY_ALIAS : alias;
  }
```

The places where `getSigningKeyAlias(null)` can be changed to `getSigningKeyAlias()`

gateway-server/src/main/java/org/apache/knox/gateway/services/DefaultGatewayServices.java
```diff
@@ -162,6 +162,8 @@ public void start() throws ServiceLifecycleException {
     SSLService ssl = (SSLService) services.get(SSL_SERVICE);
     ssl.start();
 
+    (services.get(TOKEN_SERVICE)).start();
```
risdenk: If we are going to add this here, we should add the corresponding `stop()` entry.

risdenk:
Sorry @rlevas found a few more things when I pulled this into my IDE.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
rlevas: sure.. I will update

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
rlevas: you are correct.. Nice catch.

gateway-server/src/main/java/org/apache/knox/gateway/services/DefaultGatewayServices.java
rlevas: agreed

gateway-server/src/main/java/org/apache/knox/gateway/services/DefaultGatewayServices.java
```diff
@@ -51,6 +50,33 @@
 
   private static GatewayMessages log = MessagesFactory.get( GatewayMessages.class );
 
+  private static final String[] SERVICE_START_ORDER = new String[]{
```
risdenk: Not sure if it is easy to reverse the array. Might be easier to do `Arrays.asList(...)`. Your comment about reversing the list for stop makes sense to me.

Side note - I've wanted to do this for a long time. Even on the init method. Init is harder since there are dependencies but still seems like it should be possible.

risdenk: Not saying we should change init with this PR, just commenting that I like the cleanup :)

rlevas: I looked at the init method too, but there is a bit more going on.  

As for the reverse, the stop order was different than the start order. So right now stop is not the same as start and would not be the exact reverse as start.  I am not sure what the order really should be, maybe we need to look at the services and declare an official order.  Ideally items will stop in the reverse order of how they were started.  We do not want to stop a service that is a dependent of some running service.

risdenk:
Running through all tests locally and then will merge if they pass.

PR #69, opened 2019-03-11
KNOX-1812 - The Knox Gateway truststore should be configurable

rlevas:

The Knox Gateway truststore should be configurable

Knox relies on a trust store for various purposes
- The *KnoxCLI* (via a KnoxSession) uses a truststore to trust the connection to the Knox Gateway server
  - The truststore is determined by the Java system properties (`javax.net.ssl.trustStore`, `javax.net.ssl.trustStorePassword`) or the JVM's cacerts file

- The *Knox Gateway* server uses a truststore to trust the connections going out to the services (if connecting via TLS/SSL)
 - The truststore is set to be the same keystore as the Gateway's identitiy keystore.
 - See `org.apache.knox.gateway.dispatch.DefaultHttpClientFactory#createHttpClient`.

- The *Knox Gateway* server uses a truststore to trust the user/client connecting to it when clientauth is enabled
  - The truststore is set to be the same keystore as the Gateway's identity keystore unless one is explicitly specified in the gateway-site.xml file (`gateway.truststore.path`, `gateway.truststore.type`). If a truststore is explicitly set, the password for the truststore is looked up, using alias name "`gateway-truststore-password`", from the alias service.
  - See `org.apache.knox.gateway.services.security.impl.JettySSLService#buildSslContextFactory`.

By making the outgoing connection truststore (#2, from above) configurable, it will be possible to use the same truststore for both incoming and outgoing connections, which will be convenient when services communicate with each other via the Knox Gateway. 

To make sure the truststore configuration is flexible and backwards compatible with older versions of Knox,  new properties should be introduced in the gateway-site.xml file:
- `gateway.httpclient.truststore.path`
- `gateway.httpclient.truststore.type`
- `gateway.httpclient.truststore.password.alias`

Note: This naming convention goes along with the following properties used to configure the Gateway's HTTPClient instance: 
- `gateway.httpclient.maxConnections`
- `gateway.httpclient.connectionTimeout`
- `gateway.httpclient.socketTimeout`

If `gateway.httpclient.truststore.path` is not set in the configuration, then Gateway's identity keystore will be used (which is the current implementation); else, the configured truststore details will be used. 
 
Also, to keep things consistent, the password alias name for the _clientauth_ truststore should be configurable using the property name:

- `gateway.truststore.password.alias` (default: "`gateway-truststore-password`")

## How was this patch tested?

Manually tested using various scenarios
- defaults, backwards compatibility
- custom identity keystore, unset truststore
- default identity keystore, custom truststore
- custom identity keystore, custom truststore

Updated and created new unit test cases:
```
...
[INFO] BUILD SUCCESS
[INFO] Total time: 15:35 min (Wall Clock)
[INFO] Finished at: 2019-03-10T11:28:24-04:00
[INFO] Final Memory: 319M/1930M
```

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
@@ -33,6 +33,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.GeneralSecurityException;
 import java.security.Key;
@@ -111,7 +112,7 @@ public class DefaultKeystoreService extends BaseKeystoreService implements
 
   @Override
   public KeyStore getKeystoreForGateway() throws KeystoreServiceException {
-    return getKeystore(config.getIdentityKeystorePath(), config.getIdentityKeystoreType(), config.getIdentityKeystorePasswordAlias(), true);
+    return getKeystore(Paths.get(config.getIdentityKeystorePath()), config.getIdentityKeystoreType(), config.getIdentityKeystorePasswordAlias(), true);
   }
 
   @Override
@@ -121,7 +122,7 @@ public class DefaultKeystoreService extends BaseKeystoreService implements
       // If the trustStorePath is null, fallback to behavior before KNOX-1812
       return getKeystoreForGateway();
     } else {
-      return getKeystore(trustStorePath, config.getHttpClientTruststoreType(), config.getHttpClientTruststorePasswordAlias(), true);
+      return getKeystore(Paths.get(trustStorePath), config.getHttpClientTruststoreType(), config.getHttpClientTruststorePasswordAlias(), true);
     }
   }
 
@@ -132,15 +133,15 @@ public class DefaultKeystoreService extends BaseKeystoreService implements
 
   @Override
   public KeyStore getSigningKeystore(String keystoreName) throws KeystoreServiceException {
-    String keyStoreFile;
+    Path keyStoreFile;
     String keyStoreType;
     String passwordAlias;
     if(keystoreName != null) {
-      keyStoreFile = Paths.get(keyStoreDir, keystoreName + ".jks").toString();
+      keyStoreFile = Paths.get(keyStoreDir, keystoreName + ".jks");
       keyStoreType = "jks";
       passwordAlias = null;
     } else {
-      keyStoreFile = config.getSigningKeystorePath();
+      keyStoreFile = Paths.get(config.getSigningKeystorePath());
       keyStoreType = config.getSigningKeystoreType();
       passwordAlias = config.getSigningKeystorePasswordAlias();
     }
@@ -320,10 +321,9 @@ public class DefaultKeystoreService extends BaseKeystoreService implements
   @Override
   public KeyStore getCredentialStoreForCluster(String clusterName)
       throws KeystoreServiceException {
-    String filename = Paths.get(keyStoreDir, clusterName + CREDENTIALS_SUFFIX).toString();
     // Do not fail getting the credential store if the keystore file does not exist.  The returned
     // KeyStore will be empty.  This seems like a potential bug, but is the behavior before KNOX-1812
-    return getKeystore(filename, "JCEKS", null, false);
+    return getKeystore(Paths.get(keyStoreDir, clusterName + CREDENTIALS_SUFFIX), "JCEKS", null, false);
   }
 
   @Override
@@ -455,18 +455,18 @@ public class DefaultKeystoreService extends BaseKeystoreService implements
    * @return a {@link KeyStore}, or <code>null</code> if the requested keystore cannot be created
    * @throws KeystoreServiceException if an error occurs loading the keystore file
    */
-  private KeyStore getKeystore(String keystorePath, String keystoreType, String alias, boolean failIfNotAccessible) throws KeystoreServiceException {
-    File keystoreFile = new File(keystorePath);
+  private KeyStore getKeystore(Path keystorePath, String keystoreType, String alias, boolean failIfNotAccessible) throws KeystoreServiceException {
+    File keystoreFile = keystorePath.toFile();
 
     if (failIfNotAccessible) {
       if (!keystoreFile.exists()) {
-        LOG.keystoreFileDoesNotExist(keystorePath);
+        LOG.keystoreFileDoesNotExist(keystorePath.toString());
         throw new KeystoreServiceException("The keystore file does not exist: " + keystoreFile.getAbsolutePath());
       } else if (!keystoreFile.isFile()) {
-        LOG.keystoreFileIsNotAFile(keystorePath);
+        LOG.keystoreFileIsNotAFile(keystorePath.toString());
         throw new KeystoreServiceException("The keystore file is not a file: " + keystoreFile.getAbsolutePath());
       } else if (!keystoreFile.canRead()) {
-        LOG.keystoreFileIsNotAccessible(keystorePath);
+        LOG.keystoreFileIsNotAccessible(keystorePath.toString());
         throw new KeystoreServiceException("The keystore file cannot be read: " + keystoreFile.getAbsolutePath());
       }
     }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
@@ -64,8 +64,8 @@ public class DefaultKeystoreService extends BaseKeystoreService implements
   private static final String CERT_GEN_MODE = "hadoop.gateway.cert.gen.mode";
   private static final String CERT_GEN_MODE_LOCALHOST = "localhost";
   private static final String CERT_GEN_MODE_HOSTNAME = "hostname";
-  private static GatewayMessages LOG = MessagesFactory.get( GatewayMessages.class );
-  private static GatewayResources RES = ResourcesFactory.get( GatewayResources.class );
+  private static GatewayMessages LOG = MessagesFactory.get(GatewayMessages.class);
+  private static GatewayResources RES = ResourcesFactory.get(GatewayResources.class);
 
   private GatewayConfig config;
   private Map<String, Map<String, String>> cache = new ConcurrentHashMap<>();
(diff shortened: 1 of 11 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -448,6 +437,48 @@ public String getKeystorePath() {
...
+   * reads on the file.
+   *
+   * @param keystorePath        the path to the keystore file
+   * @param keystoreType        the type of keystore file
+   * @param alias               the alias for the password to the keystore file (see {@link #getKeystorePassword(String)})
+   * @param failIfNotAccessible <code>true</code> to ensure the keystore file exists and is readable; <code>false</code> to not check
+   * @return a {@link KeyStore}, or <code>null</code> if the requested keystore cannot be created
+   * @throws KeystoreServiceException if an error occurs loading the keystore file
+   */
+  private KeyStore getKeystore(String keystorePath, String keystoreType, String alias, boolean failIfNotAccessible) throws KeystoreServiceException {
```
risdenk: Can we pass in a `Path` instead of a `String` for `keystorePath`?

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
```diff
@@ -368,13 +375,14 @@ public void TestBuildSslContextFactoryExplicitTrustStoreNullPassword() throws Ex
...
+    String truststorePasswordAlias = "trust_store_password";
 
-    GatewayConfig config = createGatewayConfig(true, true, identityKeystorePath, identityKeystoreType, identityKeyAlias, truststorePath, truststoreType);
+    GatewayConfig config = createGatewayConfig(true, true, identityKeystorePath, identityKeystoreType, identityKeyAlias, truststorePath, truststoreType, truststorePasswordAlias);
 
     AliasService aliasService = createMock(AliasService.class);
     expect(aliasService.getGatewayIdentityKeystorePassword()).andReturn(identityKeystorePassword).atLeastOnce();
     expect(aliasService.getGatewayIdentityPassphrase()).andReturn(identityKeyPassphrase).atLeastOnce();
-    expect(aliasService.getPasswordFromAliasForGateway(eq("gateway-truststore-password"))).andReturn(null).atLeastOnce();
+    expect(aliasService.getPasswordFromAliasForGateway(eq("trust_store_password"))).andReturn(null).atLeastOnce();
```
risdenk: I think this can be truststorePasswordAlias instead of the string?

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
```diff
@@ -333,13 +339,14 @@ public void TestBuildSslContextFactoryExplicitTrustStoreErrorGettingPassword() t
...
+    String truststorePasswordAlias = "trust_store_password";
 
-    GatewayConfig config = createGatewayConfig(true, true, identityKeystorePath, identityKeystoreType, identityKeyAlias, truststorePath, truststoreType);
+    GatewayConfig config = createGatewayConfig(true, true, identityKeystorePath, identityKeystoreType, identityKeyAlias, truststorePath, truststoreType, truststorePasswordAlias);
 
     AliasService aliasService = createMock(AliasService.class);
     expect(aliasService.getGatewayIdentityKeystorePassword()).andReturn(identityKeystorePassword).atLeastOnce();
     expect(aliasService.getGatewayIdentityPassphrase()).andReturn(identityKeyPassphrase).atLeastOnce();
-    expect(aliasService.getPasswordFromAliasForGateway(eq("gateway-truststore-password"))).andThrow(new AliasServiceException(null)).atLeastOnce();
+    expect(aliasService.getPasswordFromAliasForGateway(eq("trust_store_password"))).andThrow(new AliasServiceException(null)).atLeastOnce();
```
risdenk: I think this can be truststorePasswordAlias instead of the string?

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
```diff
@@ -281,13 +286,14 @@ public void TestBuildSslContextFactoryExplicitTrustStore() throws Exception {
...
+    String truststorePasswordAlias = "trust_store_password";
 
-    GatewayConfig config = createGatewayConfig(true, true, identityKeystorePath, identityKeystoreType, identityKeyAlias, truststorePath, truststoreType);
+    GatewayConfig config = createGatewayConfig(true, true, identityKeystorePath, identityKeystoreType, identityKeyAlias, truststorePath, truststoreType, truststorePasswordAlias);
 
     AliasService aliasService = createMock(AliasService.class);
     expect(aliasService.getGatewayIdentityKeystorePassword()).andReturn(identityKeystorePassword).atLeastOnce();
     expect(aliasService.getGatewayIdentityPassphrase()).andReturn(identityKeyPassphrase).atLeastOnce();
-    expect(aliasService.getPasswordFromAliasForGateway(eq("gateway-truststore-password"))).andReturn(truststorePassword).atLeastOnce();
+    expect(aliasService.getPasswordFromAliasForGateway(eq("trust_store_password"))).andReturn(truststorePassword).atLeastOnce();
```
risdenk: I think this can be truststorePasswordAlias instead of the string?

gateway-server/src/test/java/org/apache/knox/gateway/config/impl/GatewayConfigImplTest.java
```diff
@@ -348,4 +348,43 @@ public void testCustomIdentityKeystoreOptions() {
...
+  public void testHttpClientTruststoreOptions() {
+    GatewayConfigImpl config = new GatewayConfigImpl();
+
+    // Validate default options (backwards compatibility)
+    assertEquals("gateway-httpclient-truststore-password", config.getHttpClientTruststorePasswordAlias());
+    assertEquals("JKS", config.getHttpClientTruststoreType());
+    assertNull(config.getHttpClientTruststorePath());
+
+    // Validate changed options
+    config.set("gateway.httpclient.truststore.password.alias", "custom_password_alias");
```
risdenk: Can we use the gatewayconfig constants here?

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreServiceTest.java
```diff
@@ -0,0 +1,178 @@
...
+  public void testGetTruststoreForHttpClientCustomTrustStore() throws Exception {
+    final File dataDir = testFolder.newFolder();
+    final File truststoreFile = testFolder.newFile();
+    final String truststoreType = "jks";
+    final String truststorePasswordAlias = "password-alias";
+    final char[] truststorePassword = "truststore_password".toCharArray();
+
+    GatewayConfigImpl config = new GatewayConfigImpl();
+    config.set("gateway.data.dir", dataDir.getAbsolutePath());
+    config.set("gateway.httpclient.truststore.path", truststoreFile.getAbsolutePath());
```
risdenk: Can we use the gatewayconfig constants here?

rlevas: I didn't want to do that since I wanted to test that the constants (used internally) were correct.

gateway-server/src/test/java/org/apache/knox/gateway/config/impl/GatewayConfigImplTest.java
rlevas: I didn't want to do that since I wanted to test that the constants (used internally) were correct.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
rlevas: I will change this...

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
rlevas: Fixing...

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
rlevas: Fixing...

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
rlevas: Fixing...

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
```diff
@@ -79,26 +77,21 @@ public HttpClient createHttpClient(FilterConfig filterConfig) {
...
+        // The trustKeystore will be the same as the identityKeystore if a truststore was not explicitly
+        // configured in gateway-site (gateway.truststore.password.alias, gateway.truststore.path, gateway.truststore.type)
+        // This was the behavior before KNOX-1812
+        KeyStore trustKeystore = ks.getTruststoreForHttpClient();
+
         sslcontext = SSLContexts.custom()
-            .loadTrustMaterial(keystoreForGateway, new TrustSelfSignedStrategy())
-            .loadKeyMaterial(keystoreForGateway, keypass)
+            .loadTrustMaterial(trustKeystore, new TrustSelfSignedStrategy())
+            .loadKeyMaterial(identityKeystore, identityKeyPassphrase)
```
risdenk: The below comment has nothing to do with this change, but this change is in this area so commenting about it.

So I'm not sure the logic behind `loadTrustMaterial` and `loadKeyMaterial` is sound. Here are the few scenarios that I'm not sure work:

* If we have a chain of certificates in the truststore, what happens? I am not sure but this might only trust if there is a single certificate in the chain?
* If there are multiple certificates in the keystore, do we need to specify an alias to use? I think if you have multiple aliases that match it picks the first one in the keystore.

Again the above behavior didn't change as part of this PR, but would be good to follow up to make sure the scenarios aren't a problem.

rlevas: I agree this seems to be a little confusing.  Maybe we need to create a Jira to make this more explicit... maybe create our own TrustStrategy and PrivateKeyStrategy implementations.

PR #70, opened 2019-03-12
KNOX-1817 - Fix XSS issues with Alias API

moresandeep:
The Alias API was passing user input back in some cases as response without encoding, this was when an error was thrown or when a response message saying 'alias' for a 'topology' was created. This opens up the API for XSS attacks. The PR:

1. Adds encoding to the data that is going out as response. 
2. Decodes the user inputs - since the api uses application/x-www-form-urlencoded

## How was this patch tested?
The patch was tested manually e.g.

`curl -iku admin:admin-password -H "Content-Type: application/json"  -d "value=mysecret" -X PUT  'https://localhost:8443/gateway/admin/api/v1/aliases/sandbox/somelongreallylongalias<>'
HTTP/1.1 201 Created
Date: Tue, 12 Mar 2019 19:54:00 GMT
Set-Cookie: KNOXSESSIONID=node0tb9bz05vhh6k1xpp0ti2p0vqh2.node0;Path=/gateway/admin;Secure;HttpOnly
Set-Cookie: rememberMe=deleteMe; Path=/gateway/admin; Max-Age=0; Expires=Mon, 11-Mar-2019 19:54:00 GMT
Content-Length: 85
Server: Jetty(9.4.15.v20190215)

{ "created" : { "topology": "sandbox", "alias": "somelongreallylongalias&lt;&gt;" } }`

### DIFF
--- a/gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
+++ b/gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.knox.gateway.services.GatewayServices;
 import org.apache.knox.gateway.services.security.AliasService;
 import org.apache.knox.gateway.services.security.AliasServiceException;
+import org.owasp.encoder.Encode;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.Consumes;
@@ -35,7 +36,11 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_FORM_URLENCODED;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
@@ -87,9 +92,9 @@ public class AliasResource {
   @Produces({ APPLICATION_JSON })
   @Consumes(APPLICATION_FORM_URLENCODED)
   @Path(ALIAS_API_PATH)
-  public Response postAlias(@PathParam("topology") final String topology,
-      @PathParam("alias") final String alias,
-      @FormParam("value") final String value) {
+  public Response postAlias(@PathParam("topology") String topology,
+      @PathParam("alias") String alias,
+      @FormParam("value") String value) {
 
     if (value == null || value.isEmpty()) {
       return status(BAD_REQUEST).
@@ -100,15 +105,20 @@ public class AliasResource {
     final AliasService as = getAliasService();
 
     try {
+      /* check if the string is properly encoded and expect encoded values */
+      topology = URLDecoder.decode(topology, StandardCharsets.UTF_8.name());
+      alias = URLDecoder.decode(alias, StandardCharsets.UTF_8.name());
+      value = URLDecoder.decode(value, StandardCharsets.UTF_8.name());
+
       as.addAliasForCluster(topology, alias, value);
       return status(Response.Status.CREATED).
-          entity("{ \"created\" : { \"topology\": \"" + topology
-              + "\", \"alias\": \"" + alias + "\" } }").
+          entity("{ \"created\" : { \"topology\": \"" + Encode.forHtml(topology)
+              + "\", \"alias\": \"" + Encode.forHtml(alias) + "\" } }").
           type(MediaType.APPLICATION_JSON_TYPE).
           build();
-    } catch (AliasServiceException e) {
+    } catch (AliasServiceException | UnsupportedEncodingException e) {
       return status(INTERNAL_SERVER_ERROR).
-          entity("Error adding alias " + alias + " for cluster " + topology
+          entity("Error adding alias " + Encode.forHtml(alias) + " for cluster " + Encode.forHtml(topology)
               + ", reason: " + e.toString()).
           type(MediaType.APPLICATION_JSON_TYPE).build();
     }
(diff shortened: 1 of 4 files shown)


moresandeep:
@risdenk and @lmccay would love to get your inputs on this patch.

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
```diff
@@ -100,15 +105,20 @@ public Response postAlias(@PathParam("topology") final String topology,
...
+          entity("{ \"created\" : { \"topology\": \"" + Encode.forHtml(topology)
+              + "\", \"alias\": \"" + Encode.forHtml(alias) + "\" } }").
           type(MediaType.APPLICATION_JSON_TYPE).
           build();
-    } catch (AliasServiceException e) {
+    } catch (AliasServiceException | UnsupportedEncodingException e) {
       return status(INTERNAL_SERVER_ERROR).
-          entity("Error adding alias " + alias + " for cluster " + topology
+          entity("Error adding alias " + Encode.forHtml(alias) + " for cluster " + Encode.forHtml(topology)
               + ", reason: " + e.toString()).
```
risdenk: Will the e.toString() potentially print the value of the alias from line 111 on a decoding error?

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
```diff
@@ -100,15 +105,20 @@ public Response postAlias(@PathParam("topology") final String topology,
     final AliasService as = getAliasService();
 
     try {
+      /* check if the string is properly encoded and expect encoded values */
+      topology = URLDecoder.decode(topology, StandardCharsets.UTF_8.name());
+      alias = URLDecoder.decode(alias, StandardCharsets.UTF_8.name());
+      value = URLDecoder.decode(value, StandardCharsets.UTF_8.name());
```
risdenk: Does this limit the values that can be used for an alias?

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/TopologiesResource.java
```diff
@@ -360,7 +361,8 @@ public Response uploadProviderConfiguration(@PathParam("name") String name, @Con
         }
       } catch (URISyntaxException e) {
         log.invalidResourceURI(e.getInput(), e.getReason(), e);
-        response = status(Response.Status.BAD_REQUEST).entity("{ \"error\" : \"Failed to deploy provider configuration " + name + "\" }").build();
+        response = status(Response.Status.BAD_REQUEST).entity("{ \"error\" : \"Failed to deploy provider configuration " + Encode
+            .forHtml(name) + "\" }").build();
```
risdenk: Seems like a weird break? Put `Encode.forHtml(...` on the same line?

pom.xml
```diff
@@ -230,6 +230,7 @@
         <xml-matchers.version>0.10</xml-matchers.version>
         <zip4j.version>1.3.2</zip4j.version>
         <zookeeper.version>3.4.13</zookeeper.version>
+        <owasp-java-encoder>1.2.2</owasp-java-encoder>
```
risdenk: Should be in alphabetical order, this is out of place

risdenk: Property should also end in `.version` so something like `owasp-java-encoder.version`

rlevas:
Since the content type is JSON, I think it would be an issue if the return JSON document contained characters encoded for HTML.  This may be confusing to a client using the JSON document since HTML-encoded characters may be re-encoded if being displayed in an HTML document. 

`<script>...</script>` means something in an HTML document, but has no meaning (other than the literal string) in a JSON document.   Therefore I would expect a consumer of a JSON document to properly encode the data for the target viewer.  If the target was an HTML document, I would expect that the JSON to HTML translation code would perform the encoding of the string as needed.

moresandeep:
> One change required for sure - new version property not in the right place and doesn't follow same naming convention.

Oh right, this needs to be fixed.

> 
> Also I don't know if we need to url decode the parameters. jax-rs should decode them automatically?
> 
> * https://docs.jboss.org/resteasy/docs/1.0.2.GA/userguide/html/_Encoded_and_encoding.html
> * https://www.concretepage.com/webservices/resteasy-3/rest-web-service-jax-rs-encoded-annotation-example-with-resteasy-3
> 
> If the path params are decoded automatically, not sure we need to add the url decoding and the added catch exceptions.

You are right, jax-rs does automatically decode params so we don't this.

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
moresandeep: Interesting, there is a chance it could, I will remove it.

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
moresandeep: values ? I did not follow the comment. I will be taking out this piece of code because of your earlier comment.

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/TopologiesResource.java
moresandeep: yeah, I fixed one weird break and introduced another :(

pom.xml
moresandeep: I did not realize we were doing props in alphabetical order, previously we used to add new properties at the bottom, I'll change it to make it in order.

Missed the .version, forgot to clean up, I'll fix this.

moresandeep:
> was

Thanks @rlevas for using the <script> example. I was in two minds about using the HTML encoder (which keeps the <script> tag), the argument I had in my mind was, something like Angular can pull out the value from JSON and use it. I get your point that HTML encoding is the responsibility of the app that does the conversion and <script> have no meaning as far as JSON is concerned. I am inclined towards taking the HTML encoding out for the JSON responses now.

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
risdenk: Yea sorry wasn't clear. I meant does the url decoding mess up what alias values can be stored. ie: `kevin%20password` but guess it doesn't matter since this gets removed anyway :)

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
risdenk: I think if the URLDecoder stuff is removed then there shouldn't be a reason for the alias to be printed. Not sure though.

pom.xml
risdenk: Yea alphabetical after I cleaned up all the poms. Makes it easier to find later :)

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/AliasResource.java
moresandeep: Ahh, I see, righto, I thought I was doing it wrong by not decoding it and including the 'application/x-www-form-urlencoded' but I was wrong.

moresandeep:
After more discussion with @rlevas offline with reference to his previous comments it appears that this would not be an issue since the API will always return a JSON response.

Thanks @rlevas and @risdenk for the inputs !

PR #74, opened 2019-03-15
KNOX-1111 - 2-way SSL Truststore and Keystore Improvements

rlevas:

Most of the improvements needed for [KNOX-1111](https://issues.apache.org/jira/browse/KNOX-1111) have been completed using various Jiras.  However, the ability to set a custom trust store when two-way SSL was not enabled for a service was not implemented.  This patch implements that feature.  

## How was this patch tested?

Added and ran unit test:
```
...
[INFO] BUILD SUCCESS
[INFO] Total time: 15:59 min (Wall Clock)
[INFO] Finished at: 2019-03-14T20:56:51-04:00
[INFO] Final Memory: 292M/1907M
```
Manually tested various scenarios:
- two-way SSL **with no** trust store configured, used the configured identity keystore instead
- two-way SSL **with** a trust store configured, used the configured trust store
- one-way SSL **with no** trust store configured, no custom SSL context was created (assume cacerts was used)
- one-way SSL **with** a trust store configured, used the configured trust store


gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
```diff
@@ -76,28 +77,14 @@ public HttpClient createHttpClient(FilterConfig filterConfig) {
...
-            .loadTrustMaterial(trustKeystore, new TrustSelfSignedStrategy())
-            .loadKeyMaterial(identityKeystore, identityKeyPassphrase)
-            .build();
-      } catch (Exception e) {
-        throw new IllegalArgumentException("Unable to create SSLContext", e);
-      }
-      builder.setSSLSocketFactory(new SSLConnectionSocketFactory(sslcontext));
+    // Conditionally set a custom SSLContext
+    boolean useTwoWaySsl = Boolean.parseBoolean(filterConfig.getInitParameter("useTwoWaySsl"));
+    SSLContext sslContext = createSSLContext(services, useTwoWaySsl);
```
risdenk: So to be more "future proof", can we send in the filterConfig and move the boolean into the createSSLContext? This way we can use other filterconfig items later if needed to configure the SSLContext

rlevas: Sure... I can do that.  

But if we need to make changes to `createSSLContext` in the future, changing the args to passing the filterConfig could be done then.  `createSSLContext` is basically a private method, so changing its signature later should be a minor task.  If `createSSLContext` was part of an interface (or just public) that may be another story.

risdenk:
Tests passed locally. Waiting on Travis to merge.

PR #75, opened 2019-03-17
KNOX-1828 Fix Websocket Message Size

shawnweeks:
Resolves issue with Zeppelin and websocket messages being capped at 65,536 bytes.

Set websocket container max text and binary message sizes to Integer.MAX_VALUE.

## How was this patch tested?

Create workbook in zeppelin that returns more than 65,536 bytes of data. I used a sql query from Phoenix but anything that makes a big message should work.
After proxying through Knox page never completes loading and Knox spams error logs with message too large.
(If this patch involves UI changes, please attach a screen-shot; otherwise, remove this)

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
@@ -81,6 +81,10 @@ public class ProxyWebSocketAdapter extends WebSocketAdapter {
      * plumbing takes place
      */
     container = ContainerProvider.getWebSocketContainer();
+    container.setDefaultMaxTextMessageBufferSize(frontEndSession.getPolicy().getMaxTextMessageBufferSize());
+    container.setDefaultMaxBinaryMessageBufferSize(frontEndSession.getPolicy().getMaxBinaryMessageBufferSize());
+    container.setAsyncSendTimeout(frontEndSession.getPolicy().getAsyncWriteTimeout());
+    container.setDefaultMaxSessionIdleTimeout(frontEndSession.getPolicy().getIdleTimeout());
 
     final ProxyInboundClient backendSocket = new ProxyInboundClient(getMessageCallback());
 
--- a/gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyInboundClient.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyInboundClient.java
@@ -55,10 +55,6 @@ public class ProxyInboundClient extends Endpoint {
     this.session = backendSession;
     this.config = config;
 
-    /* Set the max message size */
-    session.setMaxBinaryMessageBufferSize(Integer.MAX_VALUE);
-    session.setMaxTextMessageBufferSize(Integer.MAX_VALUE);
-
     /* Add message handler for binary data */
     session.addMessageHandler(new MessageHandler.Whole<byte[]>() {
 
--- a/gateway-server/src/test/java/org/apache/knox/gateway/websockets/BigEchoSocketHandler.java
+++ b/gateway-server/src/test/java/org/apache/knox/gateway/websockets/BigEchoSocketHandler.java
@@ -36,8 +36,8 @@ class BigEchoSocketHandler extends WebSocketHandler implements WebSocketCreator 
 
   @Override
   public void configure(WebSocketServletFactory factory) {
-    factory.getPolicy().setMaxTextMessageSize(10);
-    factory.getPolicy().setMaxBinaryMessageSize(10);
+    factory.getPolicy().setMaxTextMessageSize(66000);
+    factory.getPolicy().setMaxTextMessageBufferSize(66000);
     factory.setCreator(this);
   }
 
--- a/gateway-server/src/test/java/org/apache/knox/gateway/websockets/MessageFailureTest.java
+++ b/gateway-server/src/test/java/org/apache/knox/gateway/websockets/MessageFailureTest.java
@@ -17,6 +17,7 @@
  */
 package org.apache.knox.gateway.websockets;
 
+import org.apache.commons.lang.RandomStringUtils;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.handler.ContextHandler;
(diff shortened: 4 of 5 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
```diff
@@ -87,6 +87,8 @@ public void onWebSocketConnect(final Session frontEndSession) {
      * plumbing takes place
      */
     container = ContainerProvider.getWebSocketContainer();
+    container.setDefaultMaxBinaryMessageBufferSize(Integer.MAX_VALUE);
+    container.setDefaultMaxTextMessageBufferSize(Integer.MAX_VALUE);
```
risdenk: Would be great to use the configs for this instead of hard coding to `Integer.MAX_VALUE`. I haven't looked to see how easy it is to access the gateway config here.

risdenk:
@moresandeep would appreciate your thoughts on this.

gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
shawnweeks: I can grab those from the frontEndSession. Originally I was trying to figure out how to just set the policy on the container but as far as I can tell Jetty doesn't provide a good way to do that. The backendSession is still use Integer.MAX_VALUE from another effort. Should I try to set it as well?

shawnweeks:
@moresandeep Changing that won't set it for the container as far as I can tell because it's not being created by ProxyWebSocketAdapter(). That would fix the backend though.

shawnweeks:
@moresandeep That will fix the backendsession but to fix the container I'll have to explicity cast it ClientContainer and modify the policy because ContainerProvider.getWebSocketContainer() and the Jetty code behind it sets the defaults.

gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
moresandeep: The backendSession has hardcoded Integer.MAX_VALUE unfortunately which needs to be fixed as well, feel free to open a bug and I will fix this. It appears that this value has no effect.

moresandeep:
> @moresandeep That will fix the backendsession but to fix the container I'll have to explicity cast it ClientContainer and modify the policy because ContainerProvider.getWebSocketContainer() and the Jetty code behind it sets the defaults.

Sorry, I did not get it, what I mean is this, in the ProxyWebSocketAdapter class, pass a map of gateway config variables, something like

<pre>
public ProxyWebSocketAdapter(final URI backend, final ExecutorService pool, final ClientEndpointConfig clientConfig, final Map<String, Long> config) {
    super();
    this.backend = backend;
    this.pool = pool;
    this.clientConfig = clientConfig;
    this.gatewayConfig = config;
  }
</pre>

then in `public void onWebSocketConnect(final Session frontEndSession)` 
you can have

`container.setAsyncSendTimeout(gatewayConfig.get(DEFAULT_WEBSOCKET_MAX_TEXT_MESSAGE_BUFFER_SIZE));`
.....

shawnweeks:
@moresandeep updated ProxyWebSocketAdapter to copy configuration from front end session. There a couple of ways to do this and this seemed like the least amount of change.

shawnweeks:
Not completely sure how to do a unit test for this since we'd have to have a websocket endpoint to talk to. I'll look at the other unit tests for the websocket stuff for ideas.

shawnweeks:
Squashed and pushed, modified existing test case to include boundary checking around the defaults. We now verify that messages under the limit work, messages over the default limit but under the specified limit work and that messages over the limit fail. I'm able to run tests on my side so I'm not sure why the previous build was failing.

gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
```diff
@@ -81,6 +81,10 @@ public void onWebSocketConnect(final Session frontEndSession) {
      * plumbing takes place
      */
     container = ContainerProvider.getWebSocketContainer();
```
shawnweeks: I might should be casting this to ClientContainer so I can actually set the profile completely. I checked the jetty code and set max buffer size also sets max size.

shawnweeks:
Travis is unhappy...

`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-remote-resources-plugin:1.5:process (process-resource-bundles) on project gateway-server: Error resolving project artifact: Could not transfer artifact commonj.sdo:commonj.sdo:pom:2.1.1.v201112051852 from/to central (http://repo.maven.apache.org/maven2): repo.maven.apache.org: Temporary failure in name resolution for project commonj.sdo:commonj.sdo:jar:2.1.1.v201112051852: Unknown host repo.maven.apache.org: Temporary failure in name resolution -> [Help 1]`

risdenk:
@shawnweeks I restarted the JDK 8 travis job since it was a network failure. For future reference, for any travis job if you close and then reopen the PR, the check jobs will run again. (I think its overkill but it seems to be the only way to retrigger the jobs)

gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
```diff
@@ -87,6 +87,10 @@ public void onWebSocketConnect(final Session frontEndSession) {
...
+    container.setDefaultMaxTextMessageBufferSize(frontEndSession.getPolicy().getMaxTextMessageBufferSize());
```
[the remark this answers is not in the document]
moresandeep: Sorry about the delay, meant to submit it yesterday but the review was in "pending" state, apparently I don't git well :(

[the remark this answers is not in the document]
shawnweeks: @moresandeep Yes, the values that is being set for the servlet factory get applied to any session for the frontEndSession. I'm just copying them to client container. Since the container is the generic type I can't actually set the max text message size policy but I've looked at the Jetty code and setting the buffer also sets the message size.

shawnweeks:
@risdenk I'll go and fix the permissions. Normally I'm either working out of WSL/Ubuntu or OSX but I was having trouble getting everything rebased yesterday.

PR #78, opened 2019-03-20
KNOX-1584 - YARN v2 UI - Application - History link fix

deepashreeraghu:
Updated the rewrite.xml file to have the Application Master/History link which comes under YARN UI's application url.

## How was this patch tested?
I have created a HDP 3.1 cluster and enabled knox rewrite rules for YARN and below is the snapshot of the same.
Fixed_Link_Snapshot

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
@@ -171,12 +171,34 @@
     </content>
   </filter>
    
-  <rule dir="IN" name="YARNUIV2/yarnuiv2/inbound/proxy" pattern="*://*:*/**/yarnuiv2/proxy/{**}?{**}">
-      <rewrite template="{$serviceUrl[YARNUIV2]}/proxy/{**}?{**}"/>
+  <filter name="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job">
+    <content type="application/x-http-headers">
+      <apply path="Location" rule="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job/location"/>
+    </content>
+   </filter>
+
+  <rule dir="OUT" name="YARNUIV2/yarnuiv2/outbound/proxy" pattern="*://*:*/proxy/{**}">
+      <rewrite template="{$frontend[url]}/yarnuiv2/proxy/{**}"/>
    </rule>
 
-   <rule dir="OUT" name="YARNUIV2/yarnuiv2/outbound/proxy" pattern="*://*:*/proxy/{**}">
-      <rewrite template="{gateway.scheme}://{gateway.host}:{gateway.port}/gateway/default/sparkhistory/history/{**}?{**}"/>
-   </rule>
+
+   <rule flow="OR" dir="OUT" name="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job/location">
+      <match pattern="{scheme}://{host}:{port}/jobhistory/job/{**}">
+          <rewrite template="{$frontend[url]}/jobhistory/job/{**}?{scheme}?{host}?{port}"/>
+      </match>
+      <match pattern="{scheme}://{host}:{port}/jobhistory/logs/{**}?{**}">
+          <rewrite template="{$frontend[url]}/jobhistory/logs/{**}"/>
+      </match>        
+      <match pattern="*://*:*/history/{**}?{**}">
+          <rewrite template="{$frontend[url]}/sparkhistory/history/{**}?{**}"/>
+      </match>
+      <match pattern="*://*:*/cluster/app/{**}">
+          <rewrite template="{$frontend[url]}/yarnuiv2/cluster/app/{**}"/>
+      </match>
+      <match pattern="*://*:*/cluster/apps/{**}">
+          <rewrite template="{$frontend[url]}/yarnuiv2/cluster/apps/{**}"/>
+      </match>
+    </rule>
+   
  </rules>
   
--- a/gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
@@ -170,5 +170,13 @@
       <apply path="Location" rule="YARNUIV2/yarnuiv2/outbound/headers/index/location"/>
     </content>
   </filter>
+   
+  <rule dir="IN" name="YARNUIV2/yarnuiv2/inbound/proxy" pattern="*://*:*/**/yarnuiv2/proxy/{**}?{**}">
+      <rewrite template="{$serviceUrl[YARNUIV2]}/proxy/{**}?{**}"/>
+   </rule>
+
+   <rule dir="OUT" name="YARNUIV2/yarnuiv2/outbound/proxy" pattern="*://*:*/proxy/{**}">
+      <rewrite template="{gateway.scheme}://{gateway.host}:{gateway.port}/gateway/default/sparkhistory/history/{**}?{**}"/>
+   </rule>
  </rules>
   
--- a/gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
@@ -175,30 +175,30 @@
     <content type="application/x-http-headers">
       <apply path="Location" rule="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job/location"/>
     </content>
-   </filter>
+  </filter>
 
   <rule dir="OUT" name="YARNUIV2/yarnuiv2/outbound/proxy" pattern="*://*:*/proxy/{**}">
-      <rewrite template="{$frontend[url]}/yarnuiv2/proxy/{**}"/>
-   </rule>
+    <rewrite template="{$frontend[url]}/yarnuiv2/proxy/{**}"/>
+  </rule>
 
 
-   <rule flow="OR" dir="OUT" name="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job/location">
-      <match pattern="{scheme}://{host}:{port}/jobhistory/job/{**}">
-          <rewrite template="{$frontend[url]}/jobhistory/job/{**}?{scheme}?{host}?{port}"/>
-      </match>
-      <match pattern="{scheme}://{host}:{port}/jobhistory/logs/{**}?{**}">
-          <rewrite template="{$frontend[url]}/jobhistory/logs/{**}"/>
-      </match>        
-      <match pattern="*://*:*/history/{**}?{**}">
-          <rewrite template="{$frontend[url]}/sparkhistory/history/{**}?{**}"/>
-      </match>
-      <match pattern="*://*:*/cluster/app/{**}">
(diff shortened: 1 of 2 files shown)


gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
```diff
@@ -170,5 +170,13 @@
       <apply path="Location" rule="YARNUIV2/yarnuiv2/outbound/headers/index/location"/>
     </content>
   </filter>
+   
+  <rule dir="IN" name="YARNUIV2/yarnuiv2/inbound/proxy" pattern="*://*:*/**/yarnuiv2/proxy/{**}?{**}">
+      <rewrite template="{$serviceUrl[YARNUIV2]}/proxy/{**}?{**}"/>
+   </rule>
+
+   <rule dir="OUT" name="YARNUIV2/yarnuiv2/outbound/proxy" pattern="*://*:*/proxy/{**}">
+      <rewrite template="{gateway.scheme}://{gateway.host}:{gateway.port}/gateway/default/sparkhistory/history/{**}?{**}"/>
```
risdenk: This isn't correct for a few reasons.
* This assumes the default topology due to `gateway/default`
* This will cause all proxying to go to Spark history server and will break mapreduce history server links.

moresandeep: Right, also the rule "YARNUIV2/yarnuiv2/outbound/proxy" does not seem to be references anywhere, how is this getting triggered ? you might also want to try a more specific pattern for this rule, currently it is  `*://*:*/proxy/{**}`

deepashreeraghu: Ok, Let me check on both and get back.

deepashreeraghu:
Hi @risdenk ,

I have updated both rewrite.xml and service.xml, which are now generic and works for MR job, Spark app and tested a Tez job as well. 
Please review.

Thanks and Regards,
Deepa

deepashreeraghu:
Hi @risdenk 

I am not able to understand as to CI build failure from the log. Could you please help on this?

Thanks and Regards,
Deepa

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/service.xml
```diff
@@ -89,6 +89,14 @@
       <rewrite apply="YARNUIV2/yarnuiv2/inbound/config" to="request.url"/>
     </route>
 
+    <route path="/yarnuiv2/proxy/**">
+      <rewrite apply="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job" to="response.headers"/>
```
risdenk: This doesn't seem right? proxy path pointing to jobhistory?

risdenk: Also the proxy rewrite rule doesn't operate on headers like the jobhistory one does?

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
```diff
@@ -170,5 +170,35 @@
       <apply path="Location" rule="YARNUIV2/yarnuiv2/outbound/headers/index/location"/>
     </content>
   </filter>
+   
+  <filter name="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job">
+    <content type="application/x-http-headers">
+      <apply path="Location" rule="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job/location"/>
+    </content>
+   </filter>
```
risdenk: Indentation seems off

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
```diff
@@ -170,5 +170,35 @@
...
+
+  <rule dir="OUT" name="YARNUIV2/yarnuiv2/outbound/proxy" pattern="*://*:*/proxy/{**}">
+      <rewrite template="{$frontend[url]}/yarnuiv2/proxy/{**}"/>
+   </rule>
```
risdenk: Indentation seems off

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/service.xml
deepashreeraghu: It can be updated to response.body, but the rule will still be jobhistory one as that is what is defined in the rewrite.xml.

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
```diff
@@ -171,12 +171,34 @@
...
-      <rewrite template="{$serviceUrl[YARNUIV2]}/proxy/{**}?{**}"/>
-   </rule>
+  <filter name="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job">
+    <content type="application/x-http-headers">
+      <apply path="Location" rule="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job/location"/>
+    </content>
+  </filter>
+
+  <rule dir="OUT" name="YARNUIV2/yarnuiv2/outbound/proxy" pattern="*://*:*/proxy/{**}">
+    <rewrite template="{$frontend[url]}/yarnuiv2/proxy/{**}"/>
```
moresandeep: Should be using {$serviceUrl[YARNUIV2]} instead of {$frontend[url]}

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
```diff
@@ -171,12 +171,34 @@
...
 
-   <rule dir="OUT" name="YARNUIV2/yarnuiv2/outbound/proxy" pattern="*://*:*/proxy/{**}">
-      <rewrite template="{gateway.scheme}://{gateway.host}:{gateway.port}/gateway/default/sparkhistory/history/{**}?{**}"/>
-   </rule>
+  <rule flow="OR" dir="OUT" name="YARNUIV2/yarnuiv2/outbound/headers/jobhistory/job/location">
+    <match pattern="{scheme}://{host}:{port}/jobhistory/job/{**}">
+      <rewrite template="{$frontend[url]}/jobhistory/job/{**}?{scheme}?{host}?{port}"/>
+    </match>
+    <match pattern="{scheme}://{host}:{port}/jobhistory/logs/{**}?{**}">
+      <rewrite template="{$frontend[url]}/jobhistory/logs/{**}"/>
```
moresandeep: Shouldn't this be `{$frontend[url]}/jobhistory/logs/{**}?{**}`

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
```diff
@@ -171,12 +171,34 @@
...
+    </match>        
+    <match pattern="*://*:*/history/{**}?{**}">
+      <rewrite template="{$frontend[url]}/sparkhistory/history/{**}?{**}"/>
+    </match>
+    <match pattern="*://*:*/cluster/app/{**}">
+      <rewrite template="{$frontend[url]}/yarnuiv2/cluster/app/{**}"/>
```
moresandeep: You should be using {$serviceUrl[YARNUIV2]}

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
```diff
@@ -171,12 +171,34 @@
...
+    </match>
+    <match pattern="*://*:*/cluster/apps/{**}">
+      <rewrite template="{$frontend[url]}/yarnuiv2/cluster/apps/{**}"/>
```
moresandeep: You should be using {$serviceUrl[YARNUIV2]}

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
deepashreeraghu: I see references to frontendurl in the existing rewrite file itself? Any specific reason that, this needs to be changed?

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
deepashreeraghu: Updated this in the latest commit

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
deepashreeraghu: I see references to frontendurl in the existing rewrite file itself? Any specific reason that, this needs to be changed?

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
moresandeep: {$serviceUrl[YARNUIV2]} is the preferred way. You want to include the specific URL that was supplied in the topology file, that way changing the URL in topology will update URL's all the rules.

deepashreeraghu: @moresandeep : I tried changing {$frontend[url]} with {$serviceUrl[YARNUIV2]} and I am not sure why, but the fix stopped working. So I am leaving it with frontendurl itself.

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/service.xml
deepashreeraghu: @risdenk : Let me know if there is any other better way for this. I can't remove the rule, I can only change response.headers to response.body in case of proxy rule. But it is needed without which it fails.

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/rewrite.xml
moresandeep: Interesting, is it because  {$serviceUrl[YARNUIV2]} adds the absolute url ? some more information on the failure would be good to have.
Thanks.

gateway-service-definitions/src/main/resources/services/yarnuiv2/3.0.0/service.xml
deepashreeraghu: Hi @risdenk @moresandeep ,

Please let me know what else needs to be done?

Thanks and Regards,
Deepa

risdenk: This still seems wrong. That applying /proxy/** is rewrting some jobhistory stuff.

risdenk:
@moresandeep is there more left here?

moresandeep:
Well, the rules look OK to me nothing obvious stands out, Yarnui rules are somewhat unique in what they do.

risdenk:
@deepashreeraghu looks like KNOX-1588 handled this in a more succinct way and handled the Spark history links as well. That was merged to master today. Can you confirm that it solves the same issues you are seeing?

Looks like you have been updating this same PR with other changes making it hard to merge. It helps to keep changes separate and not push to the same PR each time.

PR #80, opened 2019-03-29
KNOX-1836 - Migrate from commons-lang to commons-lang3

risdenk:

Remove commons-lang usage and replace with commons-lang3

## How was this patch tested?

`mvn -T.5C verify -Ppackage,release`


risdenk:
It looks like I missed removing from the top level pom. Need to check that next.

risdenk:
Updated to remove remaining references to commons-lang.

PR #82, opened 2019-04-01
KNOX-1819 - Ensure services are started and stopped in the correct order

rlevas:

Ensure services are started and stopped in the correct order in `org.apache.knox.gateway.services.DefaultGatewayServices` and `org.apache.knox.gateway.services.CLIGatewayServices`.

The following order is being enforced for starting services:

1. MasterService
1. KeystoreService
1. AliasService
1. SslService
1. TokenService
1. ServerInfoService
1. RemoteRegistryClientService
1. ClusterConfigurationMonitorervice
1. TopologyService
1. MetricsService
1. CryptoService
1. HostMappingService
1. ServiceDefinitionRegistry
1. ServiceRegistryService

The order is reversed for stopping. 

A new class, `org.apache.knox.gateway.services.ServiceType`, was added to enumerate the different service types.

A new class, `org.apache.knox.gateway.services.AbstractGatewayServices`, was added to implement the starting and stopping order listed above. Only _registered_ services will be started and stopped. An associated unit test class was added as well -  `org.apache.knox.gateway.services.AbstractGatewayServicesTest`

The rest of the changes are basically converting `GatewayServices` constants to `ServiceType` enums... for example, `GatewayServices.MASTER_SERVICE` to `ServiceType.MASTER_SERVICE`

## How was this patch tested?

Added new unit tests and executed them:
```
...
[INFO] BUILD SUCCESS
[INFO] Total time: 16:22 min (Wall Clock)
[INFO] Finished at: 2019-04-01T11:19:15-04:00
[INFO] Final Memory: 274M/1472M
```

Started server to see new logged messages:
```
2019-04-01 12:14:52,720 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: MasterService
2019-04-01 12:14:52,720 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: KeystoreService
2019-04-01 12:14:52,720 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: AliasService
2019-04-01 12:14:52,720 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: SSLService
2019-04-01 12:14:52,720 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: TokenService
2019-04-01 12:14:52,722 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: ServerInfoService
2019-04-01 12:14:52,722 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: RemoteConfigRegistryClientService
2019-04-01 12:14:52,722 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: ClusterConfigurationMonitorService
2019-04-01 12:14:52,723 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: TopologyService
2019-04-01 12:14:52,723 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: MetricsService
2019-04-01 12:14:52,723 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: CryptoService
2019-04-01 12:14:52,723 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: HostMappingService
2019-04-01 12:14:52,723 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: ServiceDefinitionRegistry
2019-04-01 12:14:52,723 INFO  knox.gateway (AbstractGatewayServices.java:start(67)) - Starting service: ServiceRegistryService
```

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/AbstractGatewayServices.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/AbstractGatewayServices.java
@@ -17,16 +17,14 @@
  */
 package org.apache.knox.gateway.services;
 
-import static org.apache.knox.gateway.services.ServiceType.PREFERRED_START_ORDER;
-
 import org.apache.knox.gateway.GatewayMessages;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.EnumMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -48,6 +46,7 @@ public abstract class AbstractGatewayServices implements GatewayServices {
   private static final GatewayMessages LOG = MessagesFactory.get(GatewayMessages.class);
 
   private final Map<ServiceType, Service> services = new EnumMap<>(ServiceType.class);
+  private final List<Service> orderedServices = new ArrayList<>();
   private final String role;
   private final String name;
 
@@ -58,29 +57,21 @@ public abstract class AbstractGatewayServices implements GatewayServices {
 
   @Override
   public void start() throws ServiceLifecycleException {
-    for (ServiceType serviceType : PREFERRED_START_ORDER) {
-      Service service = services.get(serviceType);
-      // If the the service has not been registered, skip it.
-      if (service != null) {
-        LOG.startingService(serviceType.getServiceTypeName());
-        service.start();
-      }
+    for (Service service : orderedServices) {
+      LOG.startingService(service.getClass().getName());
+      service.start();
     }
   }
 
   @Override
   public void stop() throws ServiceLifecycleException {
     // Reverse the preferred startup order
-    ArrayList<ServiceType> copy = new ArrayList<>(Arrays.asList(PREFERRED_START_ORDER));
+    ArrayList<Service> copy = new ArrayList<>(orderedServices);
     Collections.reverse(copy);
 
-    for (ServiceType serviceType : copy) {
-      Service service = services.get(serviceType);
-      // If the the service has not been registered, skip it.
-      if (service != null) {
-        LOG.stoppingService(serviceType.getServiceTypeName());
-        service.stop();
-      }
+    for (Service service : copy) {
+      LOG.stoppingService(service.getClass().getName());
+      service.stop();
     }
   }
 
@@ -115,6 +106,13 @@ public abstract class AbstractGatewayServices implements GatewayServices {
    * @param service     the {@link Service} implementation
    */
   protected void addService(ServiceType serviceType, Service service) {
+
+    // Ensure the service is not null
+    if (service == null) {
+      throw new NullPointerException("A null service may not be added.");
+    }
+
+    orderedServices.add(service);
     services.put(serviceType, service);
   }
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/AbstractGatewayServices.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/AbstractGatewayServices.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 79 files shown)


gateway-spi/src/main/java/org/apache/knox/gateway/services/ServiceType.java
```diff
@@ -0,0 +1,65 @@
...
+  METRICS_SERVICE("MetricsService"),
+  REMOTE_REGISTRY_CLIENT_SERVICE("RemoteConfigRegistryClientService"),
+  SERVER_INFO_SERVICE("ServerInfoService"),
+  SERVICE_DEFINITION_REGISTRY("ServiceDefinitionRegistry"),
+  SERVICE_REGISTRY_SERVICE("ServiceRegistryService"),
+  SSL_SERVICE("SSLService"),
+  TOKEN_SERVICE("TokenService"),
+  TOPOLOGY_SERVICE("TopologyService");
+
+  public static final ServiceType[] PREFERRED_START_ORDER = {
```
risdenk: Do we need this? Shouldn't services be started in the order they are added? I think this would simplify the start/stop as well since wouldn't need to worry about checking if the service is registered.

gateway-server/src/main/java/org/apache/knox/gateway/services/AbstractGatewayServices.java
```diff
@@ -0,0 +1,122 @@
+/*
```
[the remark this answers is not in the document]
rlevas: Yikes.. my IDE (IntellJ IDEA) had extra lines in the _copyright_ tag.

gateway-server/src/test/java/org/apache/knox/gateway/services/AbstractGatewayServicesTest.java
```diff
@@ -0,0 +1,159 @@
+/*
```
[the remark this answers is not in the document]
rlevas: Yikes.. my IDE (IntellJ IDEA) had extra lines in the copyright tag.

gateway-spi/src/main/java/org/apache/knox/gateway/services/ServiceType.java
rlevas: If services were started in the order which which they were added, then any implementation of GatewayServices would need to make sure the services were added in the correct order in the event there were dependencies.  For example, a programmer would need to  know that the Alias Service and Keystore service rely on the Master Service. And therefore the Master Service should added/started before the Alias Service and Keystore Service.  Also, the Alias Service is dependent on the Keystore service, so they need to be started in the proper order. 

Since the Gateway does not have dynamic injection, these dependencies are manually set and a programmer would be tipped off on the correct order.  However there will be no guarantee that the services will be added or started in the expected order.  

If we think that this is acceptable, then I will remove the preferred order stuff and rely on the order in which the services are added for starting and reversed for stopping.

risdenk: So in init() is where the services get added. Init requires the programmer to know how to initialize each service. I think we should do more to require each service to be setup as necessary before a dependent service. Say use constructors instead of the `.setAliasService()` as done today. It shouldn't be the responsibility to the ServiceType to know what order is required.

rlevas: ok

gateway-server/src/main/java/org/apache/knox/gateway/services/AbstractGatewayServices.java
```diff
@@ -0,0 +1,118 @@
...
+   * @param service     the {@link Service} implementation
+   */
+  protected void addService(ServiceType serviceType, Service service) {
+
+    // Ensure the service is not null
+    if (service == null) {
+      throw new NullPointerException("A null service may not be added.");
+    }
+
+    orderedServices.add(service);
```
risdenk: I think we should make sure we don't add the same service twice. This would case the same service to be started twice which could cause issues potentially.

rlevas: It may be possible for the same class to implement multiple services.  So it is conceivable that a class instance could be added for multiple types.  That said, may be lifecycle methods (init, start, stop) should be smarter and require a certain state to exist before being allowed to be executed.  

For example:
* you should not be able to `start` a service that was previously started (warn, and ignore?)
* you should not be able to `init` a started service.  (throw exception?)

risdenk: Hmm thats a good point. That would be a bigger change though.

PR #85, opened 2019-04-05
KNOX-1849 - Start the Java process with 'exec' when running the app in the foreground

smolnar82:

Running the Knox Gateway and the LDAP server using exec in the foreground gives us the opportunity to execute the underlying Java command that completely replaces the current bash process (there won't be any parent process which makes graceful stops easier by clients).

Note: without this change CM will consider the Gateway stopped even if the JAva process is still running in the background.

## How was this patch tested?

Running Knox Gateway and the LDAP server in the foreground after I built the project.

### DIFF
--- a/gateway-release-common/home/bin/knox-functions.sh
+++ b/gateway-release-common/home/bin/knox-functions.sh
@@ -163,7 +163,7 @@ function getPID {
 
 function appStart {
    if [ "$APP_RUNNING_IN_FOREGROUND" == true ]; then
-      $JAVA $APP_JAVA_OPTS -jar $APP_JAR $@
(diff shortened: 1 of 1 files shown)


smolnar82:
@risdenk @rlevas @moresandeep 
Could you please review this change? Thanks!

smolnar82:
Unit tests passed locally:
```
$ mvn clean -T1C verify -Prelease,package
...
[INFO] BUILD SUCCESS
[INFO] Total time: 20:27 min (Wall Clock)
[INFO] Finished at: 2019-04-05T10:03:42+02:00
[INFO] Final Memory: 359M/2471M
```

@risdenk Could you please re-trigger the job?

PR #90, opened 2019-05-17
KNOX-1858 - Add service name to X-Forwarded-Context header

moresandeep:
Add a configurable way to append service name to the X-Forward-Context header

## How was this patch tested?
Manually  tested using netcat to check headers.

### DIFF
--- a/gateway-release/home/conf/gateway-site.xml
+++ b/gateway-release/home/conf/gateway-site.xml
@@ -139,5 +139,10 @@ limitations under the License.
         <description>The whitelist to be applied for dispatches associated with the service roles specified by gateway.dispatch.whitelist.services.
         If the value is DEFAULT, a domain-based whitelist will be derived from the Knox host.</description>
     </property>
+    <property>
+        <name>gateway.header.x-forward-context.append.servicename</name>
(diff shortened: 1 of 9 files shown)


gateway-release/home/conf/gateway-site.xml
```diff
@@ -139,5 +139,10 @@ limitations under the License.
         <description>The whitelist to be applied for dispatches associated with the service roles specified by gateway.dispatch.whitelist.services.
         If the value is DEFAULT, a domain-based whitelist will be derived from the Knox host.</description>
     </property>
+    <property>
+        <name>gateway.header.x-forward-context.append.servicename</name>
```
risdenk: `private static final String XFORWARDED_ENABLED = GATEWAY_CONFIG_FILE_PREFIX + ".xforwarded.enabled";`

There is an existing config that is `xforwarded`. I think it would be better for the config to be:

`gateway.xforwarded.header.context.append.servicename`

or something like that to make it clear these are related.

moresandeep: Good point, that looks much better, thanks !

PR #91, opened 2019-05-21
KNOX-1864: Implement PATCH method

bvanhaute2:

the PATCH http method was not implemented and is necessary for certain REST api's (e.g. Apache Metron) to function.

## How was this patch tested?

It was tested by manually replacing the gateway-spi jar by the newly built one and validating if the expected functionality works properly.

Before the patch, Knox would return a 405 http error when a patch request was submitted. After the patch, a 200 OK is returned as expected.


risdenk:
Checked and this looks similar to the HEAD patch that fixed that issue. Planning to merge.

PR #92, opened 2019-05-22
KNOX-1856 - Using the appropriate error messages when throwing exception in case there is no public cert for the gateway or the cert is not an X509Certificate

smolnar82:

Until now Knox threw exceptions with misleading error messages in case:
- there was no public cert for the gateway or the cert
- the cert is not an X509Certificate

## How was this patch tested?

Running JUnit tests


smolnar82:
@rlevas @risdenk @moresandeep 
Could you please review my change?

PR #93, opened 2019-05-22
KNOX-1848 - Default to 'zookeeper' as remote alias configuration type in case it is not set in gateway-site.xml

smolnar82:

Make sure to default `gateway.remote.alias.service.config.type` to `zookeeper` in case it was not set in `gateway-site.xml`

## How was this patch tested?

Running JUnit tests.


smolnar82:
@risdenk @moresandeep 
Could you please review this PR? Thanks!

PR #94, opened 2019-05-24
KNOX-1866 Fix HBase UI proxying.

joshelser:

Current HBase UI was extremely broken, tons of dead/incorrect links. This corrects all of them but the LogLevel servlet.

## How was this patch tested?

Deployed Knox 1.3.0-SNAPSHOT locally with HBase 2.1.5-SNAPSHOT and validated that the UI can be traversed in full.

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml
@@ -16,27 +16,26 @@
    limitations under the License.
 -->
 <rules>
-  <rule dir="IN" name="HBASEUI/hbase/inbound/master/root" pattern="*://*:*/**/hbase/webui/">
-    <!-- Visiting http://master-host:16010/ triggers a meta-refresh to /master-status
-         which doesn't work when inside the proxy. Just avoid hitting that endpoint. -->
-    <rewrite template="{$serviceUrl[HBASEUI]}/master-status"/>
-  </rule>
   <rule dir="IN" name="HBASEUI/hbase/inbound/master/root/qualified" pattern="*://*:*/**/hbase/webui/?{host}?port}">
     <rewrite template="{$serviceScheme[HBASEUI]}://{host}:{port}/master-status"/>
   </rule>
   <!-- Make sure that redirects back to the master landing page work -->
   <rule dir="OUT" name="HBASEUI/hbase/outbound/master/root" pattern="/">
-    <rewrite template="{gateway.url}/hbase/webui/"/>
+    <rewrite template="{$frontend[url]}/hbase/webui/"/>
   </rule>
-  <!-- Generic pages on the default serviceUrl -->
-  <rule dir="IN" name="HBASEUI/hbase/inbound/master/home" pattern="*://*:*/**//hbase/webui/master?{host}?{port}">
-    <rewrite template="{$serviceScheme[HBASEUI]}://{host}:{port}/master-status"/>
-  </rule>
+  <!-- Master convenience URLs -->
   <rule dir="IN" name="HBASEUI/hbase/inbound/master/home" pattern="*://*:*/**//hbase/webui/master?{host}?{port}?{**}">
     <rewrite template="{$serviceScheme[HBASEUI]}://{host}:{port}/master-status?{**} "/>
   </rule>
   <rule dir="OUT" name="HBASEUI/hbase/outbound/master/home" pattern="//{host}:{port}/master-status?{**}">
-    <rewrite template="{gateway.url}/hbase/webui/master/master-status?host={$hostmap(host)}?{port}?{**}"/>
+    <rewrite template="{$frontend[url]}/hbase/webui/master/master-status?host={$hostmap(host)}?{port}?{**}"/>
+  </rule>
+  <!-- RS convenience URLs -->
+  <rule dir="IN" name="HBASEUI/hbase/inbound/regionserver/home" pattern="*://*:*/**//hbase/webui/regionserver?{host}?{port}?{**}">
+    <rewrite template="{$serviceScheme[HBASEUI]}://{host}:{port}/rs-status?{**} "/>
+  </rule>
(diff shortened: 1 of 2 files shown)


risdenk:
Hmmmm so there are 2 concerning errors in the logs when I tested this:

When visiting https://KNOX_HOST:8443/gateway/cluster_1/hbase/webui/
```
2019-05-24 08:34:43,981 ERROR knox.gateway (UrlRewriteContextImpl.java:resolve(118)) - Failed to find values by parameter name query.param.host: java.lang.NullPointerException
```

And when clicking on any of the links, end up with this:

https://KNOX_HOST:8443/gateway/cluster_1/hbase/webui/master/processMaster.jsp?host&port
```
2019-05-24 08:33:45,376 INFO  knox.gateway (WhitelistUtils.java:getDispatchWhitelist(63)) - Applying a derived dispatch whitelist because none is configured in gateway-site: ^\/.*$;^https?:\/\/(.+\.gce\.cloudera\.com):[0-9]+\/?.*$
2019-05-24 08:33:45,377 ERROR knox.gateway (GatewayDispatchFilter.java:isDispatchAllowed(156)) - The dispatch to http://null:null/procedures.jsp was disallowed because it fails the dispatch whitelist validation. See documentation for dispatch whitelisting.
```

risdenk:
It looks like everything works if you use a URL like:

https://KNOX_HOST:8443/gateway/cluster_1/hbase/webui/?host=HBASE_HOST&port=22002

risdenk:
I think we should remove the plain `hbase/webui/` handling if it only works with the `host` and `port` parameters. Another option might be to redirect/rewrite to the proper endpoint with the host and port information. Right now it seems misleading :/

joshelser:
> I think we should remove the plain hbase/webui/ handling if it only works with the host and port parameters. Another option might be to redirect/rewrite to the proper endpoint with the host and port information

That's fine. My original hope was that, if the user doesn't give us host/port, we can default to the host/port in the topology, but (like you pointed out) that breaks outgoing links.

Would like to document that too, but I'm not sure where to but it to say "You gotta provide `?host=..&port=...`". Maybe the user-manual?

gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml
```diff
@@ -0,0 +1,222 @@
...
+  <rule dir="OUT" name="HBASEUI/hbase/outbound/regionserver/children" pattern="/conf">
+    <rewrite template="{$frontend[url]}/hbase/webui/regionserver/conf?host={$inboundurl[host]}?port={$inboundurl[port]}"/>
+  </rule>
+  <rule dir="OUT" name="HBASEUI/hbase/outbound/regionserver/children" pattern="/tablesDetailed.jsp">
+    <rewrite template="{$frontend[url]}/hbase/webui/regionserver/tablesDetailed.jsp?host={$inboundurl[host]}?port={$inboundurl[port]}"/>
+  </rule>
+  <rule dir="OUT" name="HBASEUI/hbase/outbound/regionserver/children" pattern="/procedures.jsp">
+    <rewrite template="{$frontend[url]}/hbase/webui/regionserver/procedures.jsp?host={$inboundurl[host]}?port={$inboundurl[port]}"/>
+  </rule>
+  <rule dir="OUT" name="HBASEUI/hbase/outbound/regionserver/children" pattern="/dump">
```
risdenk: So `/conf`, `/dump`, `/jmx`, and `/prof` (there might be others) all match both the regionserver and master handling. The redirect urls are written from a RS url back to a master url. It looks like the right data (since the host/port) is on the end - so it might make sense to make these rules not master/regionserver specific?

Starting from:

https://KNOX_HOST:8443/gateway/cluster_1/hbase/webui/regionserver/rs-status?host=HBASE_REGIONSERVER_HOST&port=22102

and clicking "Metrics Dump", I end up with the url:

https://KNOX_HOST:8443/gateway/cluster_1/hbase/webui/master/jmx?host=HBASE_REGIONSERVER_HOST&port=22102

The data coming back starts with 
```
{"beans":[{"name":"Hadoop:service=HBase,name=RegionServer,
```

so it looks like it is pulling the JMX metrics from the RegionServer, but the url is definitely `/hbase/webui/master/jmx`

joshelser: > So /conf, /dump, /jmx, and /prof (there might be others) all match both the regionserver and master handling. The redirect urls are written from a RS url back to a master url. It looks like the right data (since the host/port) is on the end - so it might make sense to make these rules not master/regionserver specific?

Yeah, these should just be non-specific to regionserver/master at this point. Pretty sure there aren't problems in practice (will double-check) but duplicity will be confusing (in a time where I'm trying to make this less confusing).

joshelser:
Alright, I think de223d0 fixes all the stuff you mentioned, Kevin, and then some. Check out the commit-msg on that commit for the full details at your convenience, please!

Thanks for the feedback. The polish helped for sure.

risdenk:
Changes look good. Testing out here shortly and then will look at merging. I'll kick Travis since looks like a transitive failure.

joshelser:
Thanks Kevin!

risdenk:
@joshelser definitely an improvement. It looks like hitting the "home" link multiple times can cause the rewrite to be broken. Either from the master or regionserver the home link eventually ends up being https://KNOX_HOST:8443/gateway/cluster_1/hbase/webui/ - This is a side effect of removing the base handling of just `/webui/` but previously this was broken any way since home would work but any links after that would be broken with out the `host`/`port`.

joshelser:
> This is a side effect of removing the base handling of just /webui/. Previously this was broken later anyway since home would work but any links after that would be broken with out the host/port.

Was your testing done with https://issues.apache.org/jira/browse/HBASE-22467? I think what you're describing would be fixed by that change.

risdenk:
@joshelser probably not :) Thanks for the link. That looks like it would fix the issues I was seeing with the home link.

PR #97, opened 2019-06-03
KNOX-1862 - Add Service Definition for Hue UI

jdesjean:
First version of the Hue UI service definition

## How was this patch tested?
Manual

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/hue/1.0.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/hue/1.0.0/rewrite.xml
@@ -1,0 +1,134 @@
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
+   The ASF licenses this file to You under the Apache License, Version 2.0
+   (the "License"); you may not use this file except in compliance with
+   the License.  You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+<rules>
+  <rule dir="IN" name="HUE/hue/inbound/huerule" flow="OR">
+    <!-- We have exceptions for dashboard/hbase/notebook, because endpoint that ends at the app/lib level need to terminate with a slash and we remove them in generic case. -->
+    <!-- Other solution would be to accept trailing slash on all endpoints, but there's some issue with the filebrowser -->
+    <match pattern="*://*:*/**/hue/hue/dashboard?{**}">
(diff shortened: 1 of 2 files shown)


jdesjean:
@moresandeep Can you take a look?

risdenk:
@jdesjean - I'll take a closer look, but I would guess this will fail the build due to Rat checking license headers. The files are missing the Apache license header. You should be able to copy from any other XML files.

gateway-service-definitions/src/main/resources/services/hue/1.0.0/service.xml
```diff
@@ -0,0 +1,37 @@
...
+      <rewrite apply="HUE/hue/outbound/headers" to="response.headers"/>
+      <rewrite apply="HUE/hue/outbound/huerooturl" to="response.body"/>
+    </route>
+    <route path="/hue/?**">
+      <rewrite apply="HUE/hue/inbound/huerule" to="request.url"/>
+      <rewrite apply="HUE/hue/outbound/headers" to="response.headers"/>
+      <rewrite apply="HUE/hue/outbound/huerooturl" to="response.body"/>
+    </route>
+  </routes>
+  <dispatch classname="org.apache.knox.gateway.dispatch.PassAllHeadersDispatch" />
```
risdenk: This here means that this won't use Knox authentication and instead will rely on the backend to handle auth. Since the basic auth header will be passed through.

gateway-service-definitions/src/main/resources/services/hue/1.0.0/rewrite.xml
```diff
@@ -0,0 +1,134 @@
...
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+<rules>
+  <rule dir="IN" name="HUE/hue/inbound/huerule" flow="OR">
+    <!-- We have exceptions for dashboard/hbase/notebook, because endpoint that ends at the app/lib level need to terminate with a slash and we remove them in generic case. -->
+    <!-- Other solution would be to accept trailing slash on all endpoints, but there's some issue with the filebrowser -->
+    <match pattern="*://*:*/**/hue/hue/dashboard?{**}">
```
risdenk: I think this will be `/hue/hue` in the browser so might be a bit weird. I think I follow the comment, but just calling out in case there is something we can do to make this easier.

risdenk:
Will need to setup Hue and Knox to test this further.

gateway-service-definitions/src/main/resources/services/hue/1.0.0/rewrite.xml
jdesjean: /hue/hue is unfortunate, but I don't see any other solutions. We have endpoints on both /hue/dashboard & /dashboard and this seems the only way to disambiguate both.

gateway-service-definitions/src/main/resources/services/hue/1.0.0/service.xml
jdesjean: Are you certain about this?
Are there any alternatives? We need to pass all cookies as our csrf token is in there.

jdesjean:
@risdenk The CI says pending. Anything I need to do on my side?

PR #98, opened 2019-06-04
KNOX-1872 - Update Ranger service definitions to support trusted proxy

risdenk:

Make Ranger service definitions support trusted proxy by default

## How was this patch tested?

* `mvn -T.5C clean verify -Ppackage,release`
* Tested proxying Ranger UI and API with updated service definitions on Kerberized cluster


risdenk:
Hmmm this is causing a test failure. Not sure why yet.

```
[ERROR] Failures: 
[ERROR]   DeploymentFactoryFuncTest.testServiceAnonAuth:716 
Expected: an XML document with XPath /gateway/resource[1]/filter[1]/role "authentication"
     but: was "federation"
```

risdenk:
Test was assuming that RANGER service definition was anonymous. Switched to Ambari which is anonymous and fixed some parsing of the `gateway.xml`

PR #99, opened 2019-06-07
KNOX-1874. Ignore irrelevant files in service definition directories

toddlipcon:


toddlipcon:
@lmccay @risdenk please take a look. Having this would haved saved me an hour of debugging (and bothering Kevin) yesterday :)

PR #102, opened 2019-06-17
KNOX-1878 - Enforce single version of dependencies

risdenk:

Use maven enforcer plugin to make sure only single version of dependency is included.

## How was this patch tested?

`mvn -T.5C clean verify -Ppackage,release -Dshellcheck`

### DIFF
--- a/gateway-test-release/pom.xml
+++ b/gateway-test-release/pom.xml
@@ -35,10 +35,7 @@
     </modules>
 
     <properties>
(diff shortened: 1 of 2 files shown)


risdenk:
Expected to fail due to having dependencies that are listed with multiple versions (or at least transitively). Need to fix that to get the build to pass.

smolnar82:
LGTM (it was a **really** good idea to enable this plugin)

PR #104, opened 2019-06-21
KNOX-1881 - DefaultKeystoreService should use Java NIO API locking as well

risdenk:

Use Java NIO locking to ensure that Keystores are not written to by multiple processes. Removes the read/write lock and falls back to synchronized methods for the few interactions with keystores. Reading should be handled by the OS and will wait until the writing lock is released.

## How was this patch tested?

* `mvn -T.5C clean verify -Ppackage,release`
* Will be trying to change aliases with both `knoxcli` and the `gateway` to see if corrupt keystore issue goes away.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
@@ -35,10 +35,13 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.InetAddress;
+import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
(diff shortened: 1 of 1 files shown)


risdenk:
@rlevas - Can you take a look?

risdenk:
Reading a `KeyStore` from disk is done in 2 places:
* `isKeyStoreAvailable`
* `loadKeyStore`

In all both cases, a new `KeyStore` instance is created. In `loadKeyStore` the new `KeyStore` instance is returned and can be used by the callers without synchronization issues.

Writing a `KeyStore` is done in 1 place:
* `writeKeyStoreToFile`

In every case `writeKeyStoreToFile` is used, it is a new `KeyStore` instance. Either from `loadKeyStore` or a brand new `KeyStore` in `createKeyStore`.

PR #108, opened 2019-07-01
KNOX-1740 - Add Trusted Proxy Support to Knox

rlevas:

There are token exchange scenarios where an application may want to acquire a KnoxToken on behalf of a user authenticated by the application. We need to implement a version of the Hadoop Trusted Proxy/Impersonation pattern for Knox at the topology level.

This includes:

- Principal assertion method (possibilities: doAs query param, path segment within an API, HTTP header)
-  Config within topology for trusted principals, groups that they are allowed to impersonate, users that they are allowed to impersonate, ip address from which requests are expected
- Make part of the identity assertion provider since this is the provider that determines which identity to assert to the down stream service
- Config will need to be qualified by service due to the multiple services per topology
- (Please fill in changes proposed in this fix)

## How was this patch tested?

```
...
[INFO] BUILD SUCCESS
[INFO] Total time: 15:53 min (Wall Clock)
[INFO] Finished at: 2019-07-01T16:51:41-04:00
[INFO] Final Memory: 278M/1761M
```

Manually tested....
```
# kinit rlevas
rlevas@EXAMPLE.COM's password:
# curl -k --negotiate -u: -X GET "https://knox.local:8443/gateway/dt/knoxtoken/api/v1/token?doAs=userA"
{"access_token":"eyJhbGciOiJSUzI...","endpoint_public_cert":"MIIDaD...","token_type":"Bearer","expires_in":1562008370648}%
```
```
# kinit jqpublic
jqpublic@EXAMPLE.COM's password:
# curl -k --negotiate -u: -X GET "https://knox.local:8443/gateway/dt/knoxtoken/api/v1/token?doAs=userA"
{
  "RemoteException" : {
    "message" : "User: jqpublic is not allowed to impersonate userA",
    "exception" : "AuthorizationException",
    "javaClassName" : "org.apache.hadoop.security.authorize.AuthorizationException"
  }
}%
```

### DIFF
--- a/gateway-provider-security-hadoopauth/pom.xml
+++ b/gateway-provider-security-hadoopauth/pom.xml
@@ -66,5 +66,16 @@
             <artifactId>gateway-test-utils</artifactId>
             <scope>test</scope>
         </dependency>
-    </dependencies>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-provider-security-pac4j</artifactId>
+            <scope>test</scope>
+        </dependency>
+      </dependencies>
 </project>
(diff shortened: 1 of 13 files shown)


risdenk:
@rlevas were there any major differences here? I can always cherry pick the commit from PR #106

rlevas:
> @rlevas were there any major differences here? I can always cherry pick the commit from PR #106

Nope.. there are no differences.  I just applied the patch to this branch and submitted the PR (after re-testing, of course. :) )

risdenk:
Merged via cherry pick

PR #120, opened 2019-07-17
KNOX-1744 - Add rewrite rules to fix executor stdout/stderr links in Spark History Server UI

ankuriitg:
Testing Done:
Deployed on a cluster with Knox and ensured that the URLs are rewritten


ankuriitg:
@moresandeep

PR #121, opened 2019-07-25
KNOX-1948 - If no rules are defined don't rewrite

moresandeep:

This change affects inbound requests with JSON payloads that do not have any rewrite rules defined. Currently Knox will try to apply the *best match* from all the inbound rules which in most cases is problematic. This patch will prevent rewriting inbound JSON payloads unless there a rewrite rule is explicitly defined for that request payload. 

This change will not affect other rewrite functionality.

## How was this patch tested?

This patch was manually tested and unit tests provided.

### DIFF
--- a/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/json/JsonUrlRewriteFilterReader.java
+++ b/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/json/JsonUrlRewriteFilterReader.java
@@ -52,14 +52,25 @@ public class JsonUrlRewriteFilterReader extends JsonFilterReader {
 
   @Override
   protected String filterValueString(String name, String value, String rule ) {
-    try {
-      Template input = Parser.parseLiteral( value );
-      Template output = rewriter.rewrite( resolver, input, direction, rule );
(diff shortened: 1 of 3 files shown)


gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/json/JsonUrlRewriteFilterReader.java
```diff
@@ -52,14 +52,19 @@ public JsonUrlRewriteFilterReader(
 
   @Override
   protected String filterValueString(String name, String value, String rule ) {
-    try {
-      Template input = Parser.parseLiteral( value );
-      Template output = rewriter.rewrite( resolver, input, direction, rule );
-      value = output.getPattern();
-    } catch( URISyntaxException e ) {
-      LOG.failedToParseValueForUrlRewrite( value );
+    if(rule != null || UrlRewriter.Direction.OUT == direction) {
```
risdenk: Might help to add a big comment here about what and why.

moresandeep: sure, good idea !

PR #123, opened 2019-07-26
KNOX-1955 - Admin UI should handle gateway.path change

risdenk:

Use `window.location` instead of hardcoding the API endpoints

## How was this patch tested?

* `mvn -T.5C clean verify -Ppackage,release -Dshellcheck`
* Ran built Knox and updated `gateway.path` in `gateway-site.xml` and checked the AdminUI


risdenk:
@pzampino can you take a look?

PR #127, opened 2019-08-06
KNOX-1967 - Add a service definition for Impala Hiveserver2

twmarshall:

Allows for proxying connections to Impala's Hiveserver2 interface.
The service definition is based on Hive's Hiveserver2 definition.

Impala does not currently support HA.

## How was this patch tested?

Tested by setting up Impala and Knox in a real cluster and verifying that Impala can be connected to over Knox and queries can be run through jdbc and impala-shell

### DIFF
--- a/gateway-service-impala/pom.xml
+++ b/gateway-service-impala/pom.xml
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


gateway-service-impala/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.knox</groupId>
+        <artifactId>gateway</artifactId>
+        <version>1.4.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>gateway-service-impala</artifactId>
+    <name>gateway-service-impala</name>
```
risdenk: Since this is a new module, going to need to make sure to edit:
* top level `pom.xml` to make sure to call this out as a module and make sure it is added like `gateway-service-hive`
* `gateway-release` and make sure it is included in the release like `gateway-service-hive` would be

Without the above changes, the module won't get picked up for building. 

If you run `mvn -T.5C clean package -DskipTests -Ppackage,release` with your change, after building you should see `target/1.4.0-SNAPSHOT/knox-1.4.0-SNAPSHOT.tar.gz` and if you extract it make sure you see `data/services/impala` and `lib/gateway-service-impala*.jar`

twmarshall: Done

PR #134, opened 2019-08-25
KNOX-1934 - Setting the default value of knoxsso.cookie.secure.only based on ssl.enabled flag in gateway-site.xml

smolnar82:

In case the `knoxsso.cookie.secure.only` is not set we use the value of `ssl.enabled` flag (defaults to true). Using this approach and not setting `knoxsso.cookie.secure.only` in our OOTB `knoxsso.xml` makes it easier to access the admin UI without the need to edit the `knoxsso` topology in case SSL is disabled.

## How was this patch tested?

Adding new JUnit tests:
```
...
[INFO] BUILD SUCCESS
[INFO] Total time: 17:33 min (Wall Clock)
[INFO] Finished at: 2019-08-25T17:21:54+02:00
[INFO] Final Memory: 384M/2246M
```
Also tested the deliverable by changing the values in `gateway-site.xml` and `knoxsso.xml` as follows:
```
ssl.enabled=false
knoxsso.cookie.secure.only is not set
----
ssl.enabled=true
knoxsso.cookie.secure.only is not set
----
ssl.enabled=false
knoxsso.cookie.secure.only=false
----
ssl.enabled=false
knoxsso.cookie.secure.only=true
```
It's been working as expected.

### DIFF
--- a/gateway-release/home/conf/topologies/knoxsso.xml
+++ b/gateway-release/home/conf/topologies/knoxsso.xml
@@ -103,10 +103,12 @@
 
     <service>
         <role>KNOXSSO</role>
+        <!--
         <param>
             <name>knoxsso.cookie.secure.only</name>
             <value>true</value>
         </param>
+         -->
         <param>
             <name>knoxsso.token.ttl</name>
             <value>-1</value>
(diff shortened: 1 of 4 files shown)


smolnar82:
@risdenk @moresandeep @pzampino Could any of you please review my changes? Thanks!

PR #138, opened 2019-08-29
KNOX-1998: WebHDFS rewrite.xml does not have rewrite rule for Location field in json

lokeshj1703:
rewrite.xml for webhdfs service definition does not have rewrite rule for Location field in json response. The location field contains a url which must be rewritten.


lokeshj1703:
@moresandeep Can you please review the PR?

PR #144, opened 2019-09-10
KNOX-2011 - Don't block SET-COOKIE response header for Ranger UI

moresandeep:
This PR updates the dispatch for RangerUI from DefaulrDispatch to ConfigurableDispatch. This is done to allow SET-COOKIE response header to pass through Knox.

## How was this patch tested?

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/rangerui/1.0.0/service.xml
+++ b/gateway-service-definitions/src/main/resources/services/rangerui/1.0.0/service.xml
@@ -34,10 +34,6 @@
     </routes>
     <dispatch classname="org.apache.knox.gateway.dispatch.ConfigurableDispatch" ha-classname="org.apache.knox.gateway.dispatch.ConfigurableDispatch">
         <param>
-            <name>requestExcludeHeaders</name>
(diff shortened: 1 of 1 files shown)


gateway-service-definitions/src/main/resources/services/rangerui/1.0.0/service.xml
```diff
@@ -32,5 +32,14 @@
             <rewrite apply="RANGERUI/rangerui/outbound/headers" to="response.headers"/>
         </route>
     </routes>
-    <dispatch classname="org.apache.knox.gateway.dispatch.DefaultDispatch"/>
+    <dispatch classname="org.apache.knox.gateway.dispatch.ConfigurableDispatch" ha-classname="org.apache.knox.gateway.dispatch.ConfigurableDispatch">
+        <param>
+            <name>requestExcludeHeaders</name>
```
risdenk: Not sure this needs to be specified? It should default if not specified?

moresandeep: Right, this list is default and can be removed. I'll update the PR.

PR #145, opened 2019-09-13
KNOX-2011 - Fix an issue where multiple sessions are created during Ranger login

moresandeep:
Change RANGER service definition dispatch to ConfigurableDispatch. It is observed that RANGER service is used alongside RANGERUI service to serve UI.    

## How was this patch tested?
This patch was locally tested.


risdenk:
@moresandeep  - I think you need to rebase this on top of current apache master? Also looks like KNOX-2011 was already resolved?

moresandeep:
Yup, let me create a new JIRA

PR #146, opened 2019-09-13
KNOX-2012 Fix an issue where multiple sessions are created during Ranger login

moresandeep:

Change RANGER service definition dispatch to ConfigurableDispatch. It is observed that RANGER service is used alongside RANGERUI service to serve UI.

## How was this patch tested?
This patch was locally tested.


moresandeep:
Merging the patch with failed builds, builds failing for unrelated (and different) reasons.

PR #149, opened 2019-09-19
KNOX-2017 - Making Cloudera repository available in Knox's parent POM so that Cloudera dependencies are available in every children project

smolnar82:

Fixing build issue of MAven cannot find Cloudera dependencies in all Knox Maven projects (other than `gateway-discovery-cm`.

## How was this patch tested?

Running JUnit tests and checked manually if the packaged deliverable has the `cloudera-manager-api-swagger:jar`

### DIFF
--- a/gateway-discovery-cm/pom.xml
+++ b/gateway-discovery-cm/pom.xml
@@ -26,11 +26,6 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
-    <properties>
-        <cloudera-manager.version>6.3.0</cloudera-manager.version>
-        <cloudera-manager-api-swagger.version>${cloudera-manager.version}</cloudera-manager-api-swagger.version>
-    </properties>
-
     <artifactId>gateway-discovery-cm</artifactId>
     <dependencies>
         <dependency>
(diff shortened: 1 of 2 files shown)


smolnar82:
@risdenk @pzampino @moresandeep Can any of you please review my changes and merge it ASAP (so I can rebase my previous commit to PR 147 and resubmit)?

pom.xml
```diff
@@ -238,6 +238,8 @@
         <xmltool.version>3.3</xmltool.version>
         <xml-matchers.version>0.10</xml-matchers.version>
         <zookeeper.version>3.4.14</zookeeper.version>
+        <cloudera-manager.version>6.3.0</cloudera-manager.version>
```
risdenk: These should be alphabetical like the rest of the list.

gateway-discovery-cm/pom.xml
```diff
@@ -50,26 +45,16 @@
...
         </dependency>
         <dependency>
             <groupId>com.squareup.okhttp</groupId>
             <artifactId>okhttp</artifactId>
-            <version>${okhttp.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpcore</artifactId>
             <scope>compile</scope>
```
risdenk: Remove `<scope>compile</scope>` since that is the default and not needed. (I know its not directly related but its the same pom file)

pom.xml
smolnar82: Done.

gateway-discovery-cm/pom.xml
smolnar82: Done.

PR #150, opened 2019-09-20
KNOX-2014 - Make sure ATLASSESSIONID cookie is set

moresandeep:
This PR contains two changes
1. Update Atlas dispatch from DefaultDispatch to ConfigurableDispatch so set-cookie header is not blocked.
2. DefaultHADispatch now extends ConfigurableDispatch. It is possible to create a new dispatch ConfigurableHADispatch but that would be unnecessary code duplication. If anyone thinks this is a better way let me know.

## How was this patch tested?
Patch was tested locally.

### DIFF
--- a/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/AtlasTrustedProxyHaDispatch.java
+++ b/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/AtlasTrustedProxyHaDispatch.java
@@ -25,7 +25,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 
-public class AtlasTrustedProxyHaDispatch extends DefaultHaDispatch {
+public class AtlasTrustedProxyHaDispatch extends ConfigurableHADispatch {
   public AtlasTrustedProxyHaDispatch() {
     setServiceRole("ATLAS");
   }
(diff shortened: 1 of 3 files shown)


gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/DefaultHaDispatch.java
```diff
@@ -39,7 +39,7 @@
 /**
  * Default HA dispatch class that has a very basic failover mechanism
  */
-public class DefaultHaDispatch extends DefaultDispatch {
+public class DefaultHaDispatch extends ConfigurableDispatch {
```
risdenk: This is a pretty drastic change. Should be separate from this.

moresandeep: Agreed, but the change does not affect the existing features it just adds the configurability. So, in theory there is really no change. Creating a separate dispatch, like I mentioned in the PR, would simply duplicate the code.

moresandeep:
There is a bug in my service.xml, wil upload a new patch.

risdenk:
@moresandeep does this need to touch the atlas api service definition as well?

moresandeep:
I don't think this needs to touch the atlas api, atlas api service appears to be exclusively used for APIs unlike the case with Ranger Service defs.

PR #151, opened 2019-09-22
KNOX-2018 - KnoxShellTable Filtering needs greaterThan and lessThan, equals methods

lmccay4:

It would be useful to have greaterThan and lessThan filtering for KnoxShellTable.

Currently we have a regex match to filter matches.

Some easier methods for greaterThan, lessThan, greaterThanOrEqualTo, lessThanOrEqualTo, equalTo, equalToIgnoreCase, etc would be very helpful as regex for those things may be possible but more complicated than necessary.

## How was this patch tested?

Unit and manual tests

### DIFF
--- a/gateway-discovery-cm/pom.xml
+++ b/gateway-discovery-cm/pom.xml
@@ -26,11 +26,6 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
-    <properties>
-        <cloudera-manager.version>6.3.0</cloudera-manager.version>
-        <cloudera-manager-api-swagger.version>${cloudera-manager.version}</cloudera-manager-api-swagger.version>
-    </properties>
-
     <artifactId>gateway-discovery-cm</artifactId>
     <dependencies>
         <dependency>
@@ -40,36 +35,23 @@
         <dependency>
             <groupId>org.apache.knox</groupId>
             <artifactId>gateway-i18n</artifactId>
-            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.knox</groupId>
             <artifactId>gateway-util-configinjector</artifactId>
-            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>com.cloudera.api.swagger</groupId>
             <artifactId>cloudera-manager-api-swagger</artifactId>
-            <version>${cloudera-manager-api-swagger.version}</version>  <!-- or CM version 6.0 and above -->
         </dependency>
         <dependency>
             <groupId>com.squareup.okhttp</groupId>
             <artifactId>okhttp</artifactId>
-            <version>${okhttp.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpcore</artifactId>
-            <scope>compile</scope>
         </dependency>
     </dependencies>
-
-    <repositories>
-        <repository>
-            <id>Cloudera</id>
-            <name>Cloudera</name>
-            <url>https://repository.cloudera.com/artifactory/cloudera-repos</url>
-        </repository>
-    </repositories>
 </project>
 
--- a/pom.xml
+++ b/pom.xml
@@ -154,6 +154,8 @@
         <buildnumber-maven-plugin.version>1.4</buildnumber-maven-plugin.version>
         <cglib.version>3.2.12</cglib.version>
         <checkstyle.version>8.23</checkstyle.version>
+        <cloudera-manager.version>6.3.0</cloudera-manager.version>
+        <cloudera-manager-api-swagger.version>${cloudera-manager.version}</cloudera-manager-api-swagger.version>
         <commons-beanutils.version>1.9.3</commons-beanutils.version>
         <commons-cli.version>1.4</commons-cli.version>
         <commons-codec.version>1.13</commons-codec.version>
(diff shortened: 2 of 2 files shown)


gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
```diff
@@ -654,5 +655,146 @@ public KnoxShellTable regex(String regex) {
       }
       return table;
     }
+
+    public KnoxShellTable filter(Predicate p) {
+        KnoxShellTable table = new KnoxShellTable();
+        table.headers.addAll(headers);
+        for (List<String> row : rows) {
+          if (p.test(row.get(index))) {
+            table.row();
```
risdenk: What does this line do without any arguments? Seems like it isn't necessary?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
```diff
@@ -654,5 +655,146 @@ public KnoxShellTable regex(String regex) {
...
+            row.forEach(value -> {
+              table.value(value);
+            });
+          }
+        }
+        return table;
+      }
+
+      public KnoxShellTable greaterThan(String comparator) {
```
risdenk: I'm not sure you need multiple copies of each method? `compareTo` should be on the Java `Object` class so should be able to compare any object together without trying to figure out the type.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
```diff
@@ -654,5 +655,146 @@ public KnoxShellTable regex(String regex) {
...
+        }
+
+      public KnoxShellTable equalToIgnoreCase(String comparator) {
+          Predicate<String> p = (s)->s.equalsIgnoreCase(comparator);
+
+          return filter(p);
+        }
+
+      public KnoxShellTable greaterThanOrEquals(String comparator) {
+          Predicate<String> p = (s)-> s.compareTo(comparator) > 0 || s.compareTo(comparator) == 0;
```
risdenk: You might be able to reuse the `greaterThan` and `equals` predicates so you don't have to repeat the logic here. Basically make a new predicate the combination of the other predicates.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
risdenk: If it is necessary, would be good to add a comment saying what it is for.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
smolnar82: +1 for `return greaterThan(comparator) || equalTo(comparator)`

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
smolnar82: You may consider changing the method signature to `greaterThan(Comparable combarable), lessThan(Comparable comparable),...` to achieve what @risdenk suggested.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
lmccay4: After spending a while looking at this I can't think of a way to do this that would decrease redundancy given the type-specific methods in use - making calls to greaterThan() and equalTo() would return actual table objects where booleans would be needed. Do you have any suggestions?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
lmccay4: I don't think this can be done due to the type-specific conversion occurring in each predicate, where the data point from the table is of type String being converted into the method-specific argument type. I might be missing something here, let me know if I am.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxShellTable.java
lmccay4: I'll add a comment to clarify that.

smolnar82: Where does `index` come from?

lmccay4: index is a member variable in KnoxShellTableFilter

risdenk:
@lmccay4 so I think a lot of the concern from the comments come from dealing with strings directly. This ends up requiring a lot of type specific methods. Is there a reason that strings are being passed around and filtered on directly instead of the underlying datatype?

lmccay4:
> @lmccay4 so I think a lot of the concern from the comments come from dealing with strings directly. This ends up requiring a lot of type specific methods. Is there a reason that strings are being passed around and filtered on directly instead of the underlying datatype?

The underlying data type is String for all of the values in the table.

smolnar82:
@lmccay4 - I've just submitted a new PR with `KnoxShellTable` refactoring that allows us to use `Comparable` instead of `String`.
I believe this PR should be closed and unmerged and a new one should be created with the new code.

PR #153, opened 2019-09-23
KNOX-2020 - AWS federation support added to hadoop-jwt cookie

sharad-oss:
This contribution enhances the Hadoop JWT cookie to contain AWS credentials to interact with the AWS ecosystem. This is done via an optional configuration when using SAML for authentication. The account admin can set up AWS federation with SAML (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_enable-console-saml.html). The SAML Response in such a case will contain the Identity Provider (IDP) injected Role. Simple Token Service (STS) is invoked for AssumeRoleWithSaml using this SAML Response for credentials, and on  success a cookie is added to the browser. The value for this cookie is the Base 64 encoded credentials. The TTL for this cookie matches the TTL for AWS credentials. From there on the usual Knox flow continues.

Knox will add the hadoop-jwt cookie that is understood by the Hadoop ecosystem.  A JWT provides standard info about the subject, TTL and claims. We enhance the claims to contain the AWS credentials by using the info in the previously stashed aws cookie. This way the Hadoop ecosystem services can now interact with AWS services too. The lifetime for the cookies matches the life of AWS credentials.

The Pac4j filter has been modified to optionally do AWS federation. This change has no effect if the configuration is not enabled.
A new security module gateway-provider-security-aws handles all the logic to do SAML federation and adding an AWS cookie.
This module receives the SAML Response and federates into AWS, and assume the first role in the assertion. This could be enhanced in future to determine which Role to pick. AWS credentials are stashed in a cookie for further usage (after the redirect happens in the filter). At the end it returns back to the pac4j filter with a cookie added to the response.

Methods to issue JWT token have been enhanced to take custom claims as parameters. 

In KnoxSSO we enhance the hadoop-jwt cookie with the contents from the AWS cookie.
This is done by using custom claims. The procured AWS credentials are added to the JWT claims. Please look at the testing section for how the cookie looks like after the change. Note the cookie is unaffected if the option to do AWS SAML federation  is not turned on.

Knox token has also been modified to reflect the new token format.

## How was this patch tested?

Ran all unit tests
Tested on an EC2 deployment (AWS EMR). The knoxtoken (/knoxtoken/api/v1/token) looks like
```
{
  "sub": "[sharad]",
  "aud": "tokenbased",
  "iss": "KNOXSSO",
  "aws-session": "eyJ1c2VybmFtZSI6InNoYXJrYWxhQGdtYWlsLmNvbSIsIkFXU0FjY2Vzc0tleUlkIjoiQVNJQVNTRkFaVFBBV0tGWTdVNFMiLCJBV1NTZWNyZXRLZXkiOiJYT01yS0FHdFBNakdaWi91SlJibzM2eFhzRFlwT0RzU1IvWlA1N2RQIiwic2Vzc2lvblRva2VuIjoiRlFvR1pYSXZZWGR6RUZFYURDNE8vS0NVZ2pNNWFYY0luaUtsQWh1dTNvK2NkTEoyNmwxMzdKNVI1dFNROUZKUHIzQ2Y4SGJhdHMzc1pWSzJZeUlGTFZGcjJUTXBCZjBxVjZTQ2ZYSzZwUzZ2L3hHeFZvZGl2L1VOVzZLclBEeUhkbEtPT3lKdVBncGdPRkdoakdSZUw3YTBqSmNpVUhZalVhRmo4UVd5R2lkM05xRVFzeTVQRlEwalM0SmFZUkU0K04yczlycWhxSWF4K1hreDBEb0gzWGl1Y2RDVUtBQVcwK0NNaHB3eGV1d2paZDIwZGJNOTFOTTVsR2dNbmxZdzBiNjdPdXlwUDI3T0NoeU9TYkdGVVJtbC9DY3YrbjNwMFA2RVY0Q3pJbnZlUjBRN28zL0tDQmRsdGF2Mi9zWXUzSDZ0N0pXVE9ZYUJvUHhralZzTENlZmN2OXRJc2Q3bGx4cmxsUEhmRUZBMmJIOXRaa0xOQlY2alE4NmRzL1NlS09ZTlJBTnNaTmJOckxZOTlJekR2NmdPQklaVWlTNXZ4aUY4ajZMVEs2VjBLSnV2NitzRiIsImV4cGlyYXRpb24iOjE1NjgzMzUyNzUwMDB9",
  "exp": 1568367676
}
The AWS credentials are in Base 64 encoded format, and the exp matches the AWS credentials exp.
```


gateway-provider-security-aws/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+            <artifactId>lombok</artifactId>
+            <version>${lombok-version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-core</artifactId>
+            <version>${aws-sdk-version}</version>
+            <scope>compile</scope>
```
risdenk: `<scope>` not needed since compile is default

gateway-provider-security-aws/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+            <version>${aws-sdk-version}</version>
+        </dependency>
+
+        <!-- ********** ********** ********** ********** ********** ********** -->
+        <!-- ********** Test Dependencies                           ********** -->
+        <!-- ********** ********** ********** ********** ********** ********** -->
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>1.10.19</version>
```
risdenk: Shouldn't need the explicit version. If mockito isn't included, please don't add it. I think we tried to consolidate to one mocking library.

gateway-provider-security-aws/src/main/java/org/apache/knox/gateway/aws/AwsMessages.java
```diff
@@ -0,0 +1,40 @@
...
+package org.apache.knox.gateway.aws;
+
+import org.apache.knox.gateway.i18n.messages.Message;
+import org.apache.knox.gateway.i18n.messages.MessageLevel;
+import org.apache.knox.gateway.i18n.messages.Messages;
+
+/**
+ * Logging messages for the pac4j provider.
+ *
+ * @since 0.8.0
```
risdenk: Don't need the `@since` tag

gateway-provider-security-aws/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+        
+        <dependency>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.opensaml</groupId>
+            <artifactId>opensaml-core</artifactId>
+            <version>${opensaml-version}</version>
```
risdenk: Remove the `<version>` tags from child poms and declare dependencies in the parent pom with version in one place.

pom.xml
```diff
@@ -150,6 +151,7 @@
         <apache-rat-plugin.version>0.13</apache-rat-plugin.version>
         <ant-nodeps.version>1.8.1</ant-nodeps.version>
         <asm.version>7.1</asm.version>
+        <aws-sdk-version>1.11.562</aws-sdk-version>
```
risdenk: nit - Can you use `.version` like the other properties?

pom.xml
```diff
@@ -214,6 +216,7 @@
         <libpam4j.version>1.11</libpam4j.version>
         <log4j.version>1.2.17</log4j.version>
         <log4j2.version>2.12.0</log4j2.version>
+        <lombok-version>1.16.20</lombok-version>
```
risdenk: Do we need lombok? Looks like it adds some helper annotations that aren't strictly required. It just ends up being another library to have to maintain if we don't need it.

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
```diff
@@ -269,7 +289,7 @@ private Response getAuthenticationToken(int statusCode) {
       } catch (IOException e) {
         log.unableToCloseOutputStream(e.getMessage(), Arrays.toString(e.getStackTrace()));
       }
-    } catch (TokenServiceException| AliasServiceException e) {
+    } catch (TokenServiceException| AliasServiceException | UnsupportedEncodingException e) {
```
risdenk: Assuming `UnsupportedEncodingException` comes from the `new String`?

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
```diff
@@ -241,6 +249,18 @@ private Response getAuthenticationToken(int statusCode) {
     Principal p = request.getUserPrincipal();
 
     try {
+      Map<String, String> claims = new HashMap<>();
+      int cookieAge = maxAge;
+      Optional<String> awsCookie =  CookieUtils.getCookieValue(request, AwsConstants.AWS_COOKIE_NAME);
+      AwsSamlCredentials awsSamlCredentials = null;
+      if (awsCookie.isPresent()) {
+        String cookieJson = new String(Base64.decodeBase64(awsCookie.get().getBytes("UTF-8")), "UTF-8");
```
risdenk: Should probably use `StandardCharsets...` instead of "UTF-8" if possible.

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
```diff
@@ -99,6 +106,7 @@
   private String signatureAlgorithm = "RS256";
   private List<String> ssoExpectedparams = new ArrayList<>();
   private String clusterName;
+  private Gson gson = new Gson();
```
risdenk: Do we need gson here when we already have jackson? Looks like we don't include gson anywhere else in the project currently.

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
```diff
@@ -47,6 +50,10 @@
 import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.audit.log4j.audit.Log4jAuditor;
 import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.commons.codec.binary.Base64;
```
risdenk: Since we are on Java 8 minimum, should be able to use Base64 from the JDK and not need commons-codec.

gateway-service-knoxsso/pom.xml
```diff
@@ -81,5 +91,12 @@
             <artifactId>gateway-test-utils</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.8.5</version>
```
risdenk: See other comment about removing gson altogether if possible. Otherwise don't need version and scope on this.

gateway-service-knoxsso/pom.xml
```diff
@@ -50,11 +50,21 @@
...
+        </dependency>
+
         <dependency>
             <groupId>com.nimbusds</groupId>
             <artifactId>nimbus-jose-jwt</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
```
risdenk: See other comment about removing base64 so commons-codec isn't necessary here.

gateway-server/pom.xml
```diff
@@ -358,7 +357,12 @@
             <groupId>org.apache.knox</groupId>
             <artifactId>gateway-util-configinjector</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
```
risdenk: See comment about removing gson if possible.

risdenk:
@sharad-oss looks like the jdk11 Travis failure might be real as well:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project gateway-provider-security-aws: Fatal error compiling: java.lang.ExceptionInInitializerError: com.sun.tools.javac.code.TypeTags -> [Help 1]
```

gateway-provider-security-aws/pom.xml
sharad-oss: Actually I had federation tests with some other services other than STS. Sure, will remove if I don't add them.

gateway-provider-security-aws/pom.xml
sharad-oss: Sure, not needed.

pom.xml
sharad-oss: Sure, will stick to the current nomenclature.

pom.xml
sharad-oss: [Lombok](https://projectlombok.org/) avoids a lot of boilerplate code - setters, getters, constructors. Many other modules can be so much concise if using Lombok - also avoids human errors.

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
sharad-oss: Right

sharad-oss:
@risdenk I wasn't aware Knox was being built on JDK-11 too. My current tests were all done using JDK-8. I will invest some time to build with JDK-11.

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
sharad-oss: Done

gateway-provider-security-aws/pom.xml
sharad-oss: Done

lmccay:
From my comment on JIRA KNOX-2020:

@sharad-oss - this looks interesting!
I'd actually like to see a one-pager type doc that describes the usecases, the design and security considerations. Please attach it to the JIRA itself.
One thing that I am concerned about is the inclusion of sensitive credentials in the JWT based cookie.
The cookie from KnoxSSO is intended for browsers and generally represents the authenticated user but doesn't include credentials. It is essentially in clear text since the JWT is merely base64 encoded. This is not sufficient protection for credentials that can be used outside of the scope of Knox itself.
In terms of usecases, I'd like to understand the full flow including how/where the credentials are actually used and what consumer will be provided in Knox for the credentials.

sharad-oss:
@lmccay sure will work on the one pager doc soon, and update the JIRA.

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
sharad-oss: Will switch to Jackson.

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
sharad-oss: Sure, can use the java.util.Base64.

gateway-service-knoxsso/pom.xml
sharad-oss: Done

gateway-service-knoxsso/pom.xml
sharad-oss: Done

gateway-server/pom.xml
sharad-oss: Done

sharad-oss:
@risdenk Addressed your feedback. I am seeing some test failures but they don't seem related to my changes.

I am also working on a one pager as requested by @lmccay.

pom.xml
```diff
@@ -209,15 +211,19 @@
         <libpam4j.version>1.11</libpam4j.version>
         <log4j.version>1.2.17</log4j.version>
         <log4j2.version>2.12.1</log4j2.version>
+        <lombok-version>1.18.4</lombok-version>
+        <log4j2.version>2.12.1</log4j2.version>
```
risdenk: Looks like merge/copy paste error? Duplicate `log4j2.version` property

pom.xml
```diff
@@ -209,15 +211,19 @@
...
         <maven-checkstyle-plugin.version>3.1.0</maven-checkstyle-plugin.version>
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
         <maven-dependency-plugin.version>3.1.1</maven-dependency-plugin.version>
         <maven-enforcer-plugin.version>3.0.0-M2</maven-enforcer-plugin.version>
         <maven-pmd-plugin.version>3.12.0</maven-pmd-plugin.version>
         <metrics.version>4.1.0</metrics.version>
+        <mockito.version>1.10.19</mockito.version>
```
risdenk: Do we need `mockito`? We use `easymock` everywhere else for mocking. I don't have anything really against `mockito` we just tried to "standardize" not too long ago on `easymock`. It should be easy to replace `mockito` usage with `easymock` where applicable. For reference here is the jira that did that: https://issues.apache.org/jira/browse/KNOX-1479

pom.xml
```diff
@@ -209,15 +211,19 @@
         <libpam4j.version>1.11</libpam4j.version>
         <log4j.version>1.2.17</log4j.version>
         <log4j2.version>2.12.1</log4j2.version>
+        <lombok-version>1.18.4</lombok-version>
```
risdenk: `lombok.version` instead

pom.xml
```diff
@@ -209,15 +211,19 @@
...
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
         <maven-dependency-plugin.version>3.1.1</maven-dependency-plugin.version>
         <maven-enforcer-plugin.version>3.0.0-M2</maven-enforcer-plugin.version>
         <maven-pmd-plugin.version>3.12.0</maven-pmd-plugin.version>
         <metrics.version>4.1.0</metrics.version>
+        <mockito.version>1.10.19</mockito.version>
         <mina.version>2.0.20</mina.version>
         <nimbus-jose-jwt.version>7.8</nimbus-jose-jwt.version>
         <okhttp.version>2.7.5</okhttp.version>
+        <opensaml-version>3.3.0</opensaml-version>
```
risdenk: `opensaml.version` instead

pom.xml
```diff
@@ -1667,6 +1678,43 @@
...
+            <dependency>
+                <groupId>org.opensaml</groupId>
+                <artifactId>opensaml-saml-impl</artifactId>
+                <version>${opensaml-version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <scope>provided</scope>
```
risdenk: `<scope>provided</scope>` seems weird here?

pom.xml
lmccay: Why are we pulling in lombok here? This isn't used across the other features within Knox and I don't see any reason to add it here making this code different from the rest.

We do need to see a contributed consumer of this within the PR that makes it usable out of the box for the Knox community.

sharad-oss: Sure, will remove Lombok to keep the code consistent.

pom.xml
sharad-oss: Removed Lombok.

pom.xml
sharad-oss: Done

pom.xml
sharad-oss: Deleted the Lombok altogether.

pom.xml
sharad-oss: Moved to EasyMock to maintain code consistency.

smolnar82:

PR #156, opened 2019-09-27
KNOX-2024 - KnoxShellTable - Case Insensitive Operations with Col Names

lmccay4:

Case sensitivity for column names in table filtering

## How was this patch tested?

Manual and unit tests


lmccay4:
In regards to the failed check, it builds fine on my machine, unsure why it's failing on here but it doesn't seem to be from my change.

risdenk:
Most likely going to want to rebase this change on top of PR #157

PR #158, opened 2019-09-28
KNOX-2018 - KnoxShellTable Filtering needs greaterThan and lessThan, equals Methods

lmccay4:

Added lessThan, greaterThanOrEqualTo, lessThanOrEqualTo, and equalTo methods for KnoxShellTable. 

## How was this patch tested?

Unit and manual tests

### DIFF
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
@@ -41,7 +41,7 @@ public class KnoxShellTableFilter {
     return this;
   }
 
-  //TODO: use Predicate to evaluate the Pattern.matches 
+  //TODO: use Predicate to evaluate the Pattern.matches
   // for regular expressions: startsWith, endsWith, contains,
   // doesn't contain, etc
   public KnoxShellTable regex(String regex) {
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
@@ -91,15 +91,14 @@ public class KnoxShellTableFilter {
 
   @SuppressWarnings("unchecked")
   public KnoxShellTable greaterThanOrEqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
-    return filter(s -> s.compareTo(comparable) > 0 || s.compareTo(comparable) == 0);
+    return filter(s -> s.compareTo(comparable) > 0 || s.equals(comparable));
   }
 
   @SuppressWarnings("unchecked")
   public KnoxShellTable lessThanOrEqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
-    return filter(s -> s.compareTo(comparable) < 0 || s.compareTo(comparable) == 0);
+    return filter(s -> s.compareTo(comparable) < 0 || s.equals(comparable));
   }
 
-  @SuppressWarnings("unchecked")
   public KnoxShellTable equalTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
     return filter(s -> s.equals(comparable));
   }
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
@@ -100,7 +100,7 @@ public class KnoxShellTableFilter {
   }
 
   @SuppressWarnings("unchecked")
-  public KnoxShellTable EqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
-    return filter(s -> s.compareTo(comparable) == 0);
+  public KnoxShellTable equalTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
+    return filter(s -> s.equals(comparable));
   }
 }
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
@@ -41,7 +41,9 @@ public class KnoxShellTableFilter {
     return this;
   }
 
-  //TODO: this method can be eliminated when the 'equalTo' method is implemented: regex(regex) = equalTo(anyString)
+  //TODO: use Predicate to evaluate the Pattern.matches 
+  // for regular expressions: startsWith, endsWith, contains,
+  // doesn't contain, etc
   public KnoxShellTable regex(String regex) {
     final Pattern pattern = Pattern.compile(regex);
     final KnoxShellTable filteredTable = new KnoxShellTable();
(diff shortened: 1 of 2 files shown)


gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
```diff
@@ -82,4 +84,23 @@ public KnoxShellTable greaterThan(Comparable<? extends Object> comparable) throw
...
+    return filter(s -> s.compareTo(comparable) > 0 || s.compareTo(comparable) == 0);
+  }
+
+  @SuppressWarnings("unchecked")
+  public KnoxShellTable lessThanOrEqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
+    return filter(s -> s.compareTo(comparable) < 0 || s.compareTo(comparable) == 0);
+  }
+
+  @SuppressWarnings("unchecked")
+  public KnoxShellTable EqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
```
risdenk: should be `equalTo`

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
```diff
@@ -82,4 +84,23 @@ public KnoxShellTable greaterThan(Comparable<? extends Object> comparable) throw
     return filter(s -> s.compareTo(comparable) > 0);
```
risdenk: I think you can store the predicate as a variable name and reuse it.

Something like:

`Predicate greaterThan = s -> s.compareTo(comparable) > 0;`

Might have to fix it up a bit but it would make the new methods simpler since you could say `greaterThan(s, comparable) || equalTo(s, comparable)`

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
```diff
@@ -82,4 +84,23 @@ public KnoxShellTable greaterThan(Comparable<? extends Object> comparable) throw
...
+  }
+
+  @SuppressWarnings("unchecked")
+  public KnoxShellTable lessThanOrEqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
+    return filter(s -> s.compareTo(comparable) < 0 || s.compareTo(comparable) == 0);
+  }
+
+  @SuppressWarnings("unchecked")
+  public KnoxShellTable EqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
+    return filter(s -> s.compareTo(comparable) == 0);
```
risdenk: Is there any reason not to use `.equals()` here?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
lmccay4: I like the idea but I can't find away to implement this in a fashion that would be worth the effort.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
lmccay4: Changed to `.equals()` in the updated patch.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
lmccay4: Addressed this in the updated patch.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
```diff
@@ -82,4 +84,23 @@ public KnoxShellTable greaterThan(Comparable<? extends Object> comparable) throw
...
   }
 
+  @SuppressWarnings("unchecked")
+  public KnoxShellTable lessThan(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
+    return filter(s -> s.compareTo(comparable) < 0);
+  }
+
+  @SuppressWarnings("unchecked")
+  public KnoxShellTable greaterThanOrEqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
+    return filter(s -> s.compareTo(comparable) > 0 || s.compareTo(comparable) == 0);
```
risdenk: Should be `.equals()` as well for gt and lt.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
```diff
@@ -82,4 +84,23 @@ public KnoxShellTable greaterThan(Comparable<? extends Object> comparable) throw
...
+  }
+
+  @SuppressWarnings("unchecked")
+  public KnoxShellTable lessThanOrEqualTo(Comparable<? extends Object> comparable) throws KnoxShellTableFilterException {
+    return filter(s -> s.compareTo(comparable) < 0 || s.compareTo(comparable) == 0);
```
risdenk: Should be `.equals()` as well for gt and lt

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
lmccay4: Updated in latest patch.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
lmccay4: Updated in latest patch.

gateway-shell/src/test/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilterTest.java
```diff
@@ -50,18 +51,86 @@ public void testRegexFilter() throws KnoxShellTableFilterException {
...
+    assertEquals(TABLE.filter().name("Column Integer").equalTo(50).getRows().size(), 1);
+
+    assertEquals(TABLE.filter().name("Column Double").equalTo(40d).getRows().size(), 1);
+    assertEquals(TABLE.filter().name("Column Double").equalTo(90d).getRows().size(), 0);
+
+    assertEquals(TABLE.filter().name("Column String").equalTo("text90").getRows().size(), 0);
+    assertEquals(TABLE.filter().name("Column String").equalTo("text30").getRows().size(), 1);
+
+    synchronized (TEST_DATE_FORMAT) {
+      assertEquals(TABLE.filter().name("Column Date").equalTo(TEST_DATE_FORMAT.parse("2012-02-01")).getRows().size(), 1);
```
smolnar82: Aren't there 2 rows with `2012-01-01`?

lmccay4: The date on line 42 is `2012-02-01` which is what I'm filtering for on line 132.

smolnar82: Ahh...sorry...my mistake

PR #159, opened 2019-09-30
KNOX-2027 - Need a reverse order sort on a table based on a column.

lmccay4:

Ability to sort ascending and descending a table based on a column.

## How was this patch tested?

Manual and unit tests.

### DIFF
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
@@ -21,6 +21,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import javax.swing.SortOrder;
+
 import org.apache.knox.gateway.util.JsonUtils;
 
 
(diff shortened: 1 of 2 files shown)


gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -122,6 +124,10 @@ public KnoxShellTable select(String cols) {
   }
 
   public KnoxShellTable sort(String colName) {
+    return sort(colName, SortOrder.ASCENDING);
+  }
+
+  public KnoxShellTable sort(String colName, SortOrder order) {
```
risdenk: Does public make sense here? Can SortOrder be passed in from the shell or will it always need to be a string?

lmccay4: Thanks for the review! It does need to be public as other external applications such as KnoxLine will be using it. SortOrder would be passed in from the shell as SortOrder.DESCENDING.

PR #160, opened 2019-09-30
KNOX-2024 - KnoxShellTable - Case Insensitive Operations with Col Names

lmccay4:

Ability to filter, sort, and select based on a case insensitive column name.

## How was this patch tested?

Manual and unit tests

### DIFF
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
@@ -32,7 +32,15 @@ public class KnoxShellTableFilter {
   }
 
   public KnoxShellTableFilter name(String name) {
-    index = tableToFilter == null ? -1 : tableToFilter.headers.indexOf(name);
+    for (int i = 0; i < tableToFilter.headers.size(); i++) {
+      if (tableToFilter.headers.get(i).equalsIgnoreCase(name)) {
+        index = i;
+        break;
+      }
+      else {
+        index = -1;
+      }
+    }
     return this;
   }
 
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
@@ -24,22 +24,22 @@ import java.util.regex.Pattern;
 public class KnoxShellTableFilter {
 
   private KnoxShellTable tableToFilter;
-  private int index;
+  private int index = -1;
 
   public KnoxShellTableFilter table(KnoxShellTable table) {
     this.tableToFilter = table;
     return this;
   }
 
-  public KnoxShellTableFilter name(String name) {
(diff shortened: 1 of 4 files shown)


gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
```diff
@@ -32,7 +32,15 @@ public KnoxShellTableFilter table(KnoxShellTable table) {
   }
 
   public KnoxShellTableFilter name(String name) {
-    index = tableToFilter == null ? -1 : tableToFilter.headers.indexOf(name);
+    for (int i = 0; i < tableToFilter.headers.size(); i++) {
+      if (tableToFilter.headers.get(i).equalsIgnoreCase(name)) {
+        index = i;
```
risdenk: What is index used for outside of this method? Could there be an issue that index is being updated and changed to -1 on no matches before switched back to the right index?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
```diff
@@ -32,7 +32,15 @@ public KnoxShellTableFilter table(KnoxShellTable table) {
   }
 
   public KnoxShellTableFilter name(String name) {
-    index = tableToFilter == null ? -1 : tableToFilter.headers.indexOf(name);
+    for (int i = 0; i < tableToFilter.headers.size(); i++) {
+      if (tableToFilter.headers.get(i).equalsIgnoreCase(name)) {
```
risdenk: Is it possible to store the headers in a case insensitive way? ie: lowercase on insert? This would avoid having to deal with case insensitive headers other places.

lmccay4:
@risdenk my responses don't seem to showing up for whatever reason.

Re: headers being stored lowercase on insert: I'm not sure the headers should be changed on insert as they are likely stored in specific ways by the table creators and intended to be displayed that way.

And as for the index issue I've updated that in the latest patch to resolve any issues that could have caused.

risdenk:
> @risdenk my responses don't seem to showing up for whatever reason.

@lmccay4 if you are in "review mode" - you need to make sure you submit the comments otherwise they just sit in draft basically.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTableFilter.java
risdenk: With the recent change - won't this case only ever be hit once? Since once you set `index` to something else it won't have the "magic" value of -1 to compare against again.

I think this method requires some more thought and some explanation of what the `index` variable is used for. Global state tends to be hard to reason about and if possible avoided.

lmccay4: I see your original concern, my bad. `index` is a member variable of the filter class and only ever used in that one filter instance, not a global variable. As for -1, that is the returned index when no matches exist, which throws an exception.

PR #161, opened 2019-10-03
KNOX-2026: Accept Impala's authentication cookies

twmarshall:
This patch modifies HadoopAuthCookieStore to accept cookies with
Impala's cookie name, "impala.auth".

It also updates a check that is used to ensure the cookie belongs to
Knox - previously, this check parsed the cookie according to the
specific format that Hadoop uses for its cookies and ensures that the
Knox principal appears in the expected location.

Impala uses a similar cookie format, but with a few changes such as
fields being in a different order. The check is made more permissive
such that it will accept any cookie that contains the Knox principal
anywhere in it.

Testing:
- Deployed in a cluster and verified that Knox accepts and returns
  Impala's cookies as expected.


twmarshall:
@pzampino could you possible take a look? You appear to have implemented the check in the first place in KNOX-1341. Thanks

risdenk:
Change looks good to me. Limited to only the 3 cookies that Knox will check anyway. Waiting on @pzampino to review as well

PR #163, opened 2019-10-05
KNOX-2052 - KnoxShellTable mean, median, and mode methods

lmccay4:

Added mean, median, mode, sum, max, min, and toDoubleArray methods so to enable usage of the Apache Commons math StatUtils library (methods require double[]). 

## How was this patch tested?

Manual and unit tests.

### DIFF
--- a/gateway-shell/pom.xml
+++ b/gateway-shell/pom.xml
@@ -32,6 +32,10 @@
 
     <dependencies>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math3</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.knox</groupId>
             <artifactId>gateway-util-common</artifactId>
         </dependency>
--- a/gateway-shell/pom.xml
+++ b/gateway-shell/pom.xml
@@ -32,10 +32,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-math3</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.apache.knox</groupId>
             <artifactId>gateway-util-common</artifactId>
         </dependency>
@@ -97,6 +93,10 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-math3</artifactId>
         </dependency>
         <dependency>
             <groupId>commons-io</groupId>
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
@@ -24,8 +24,8 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.swing.SortOrder;
-
 import com.fasterxml.jackson.annotation.JsonFilter;
+import org.apache.commons.math3.stat.StatUtils;
 
 
 /**
@@ -35,6 +35,16 @@ import com.fasterxml.jackson.annotation.JsonFilter;
  */
 @JsonFilter("knoxShellTableFilter")
 public class KnoxShellTable {
+
+    public enum Conversions {
+        DOUBLE,
+        INTEGER,
+        FLOAT,
+        BYTE,
+        SHORT,
+        LONG
+    }
+
   private static final String LINE_SEPARATOR = System.getProperty("line.separator");
 
   List<String> headers = new ArrayList<String>();
@@ -85,6 +95,117 @@ public class KnoxShellTable {
     rows.forEach(row -> col.add(row.get(colIndex)));
     return col;
   }
+
+  public Conversions getConversion(Comparable<? extends Object> colIndex) {
+      Conversions type = null;
+
+      if (colIndex instanceof Double) {
+        type = Conversions.DOUBLE;
+      }
+      else if (colIndex instanceof Integer) {
+        type = Conversions.INTEGER;
+      }
+      else if (colIndex instanceof Float) {
+        type = Conversions.FLOAT;
+      }
+      else if (colIndex instanceof Byte) {
+        type = Conversions.BYTE;
+      }
+      else if (colIndex instanceof Short) {
+        type = Conversions.SHORT;
+      }
+      else if (colIndex instanceof Long) {
+        type = Conversions.LONG;
+      }
+      return type;
+    }
+
+    public double[] toDoubleArray(String colName) throws IllegalArgumentException {
+      List<Comparable<? extends Object>> col = values(colName);
+      double[] colArray = new double[col.size()];
+      Conversions conversionMethod = null;
+      for (int i = 0; i < col.size(); i++) {
+        if (i == 0) {
+          conversionMethod = getConversion(col.get(i));
+        }
+
+        switch (conversionMethod) {
+          case DOUBLE:
+            colArray[i] = (double) ((Double) col.get(i));
+            break;
+
+          case INTEGER:
+            colArray[i] = (double) ((Integer) col.get(i)).intValue();
+            break;
+
+          case FLOAT:
+            colArray[i] = (double) ((Float) col.get(i)).floatValue();
+            break;
+
+          case BYTE:
+            colArray[i] = (double) ((Byte) col.get(i)).byteValue();
+            break;
+
+          case SHORT:
+            colArray[i] = (double) ((Short) col.get(i)).shortValue();
+            break;
+
+          case LONG:
+            colArray[i] = (double) ((Long) col.get(i)).longValue();
+            break;
+        }
+      }
+      return colArray;
+    }
+
+    public double mean(String colName) {
+      return StatUtils.mean(toDoubleArray(colName));
+    }
+
+    public double mean(int colIndex) {
+      return mean(headers.get(colIndex));
+    }
+
+    public double median(String colName) {
+        return StatUtils.percentile(toDoubleArray(colName), 50);
+    }
+
+    public double median(int colIndex) {
+      return median(headers.get(colIndex));
+    }
+
+    public double mode(String colName) {
+        return (double) StatUtils.mode(toDoubleArray(colName))[0];
+    }
+
+    public double mode(int colIndex) {
+      return mode(headers.get(colIndex));
+    }
+
+    public double sum(String colName) {
+        return StatUtils.sum(toDoubleArray(colName));
+    }
+
+    public double sum(int colIndex) {
+      return sum(headers.get(colIndex));
+    }
+
+    public double max(String colName) {
+        return StatUtils.max(toDoubleArray(colName));
+    }
+
+    public double max(int colIndex) {
+      return max(headers.get(colIndex));
+    }
+
+    public double min(String colName) {
+        return StatUtils.min(toDoubleArray(colName));
+    }
+
+    public double min(int colIndex) {
+      return min(headers.get(colIndex));
+    }
+
 
   public KnoxShellTable apply(KnoxShellTableCell<? extends Comparable<? extends Object>> cell) {
     if (!headers.isEmpty()) {
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
@@ -36,7 +36,7 @@ import org.apache.commons.math3.stat.StatUtils;
 @JsonFilter("knoxShellTableFilter")
 public class KnoxShellTable {
 
-    public enum Conversions {
+    private enum Conversions {
         DOUBLE,
         INTEGER,
         FLOAT,
@@ -96,9 +96,8 @@ public class KnoxShellTable {
     return col;
   }
 
-  public Conversions getConversion(Comparable<? extends Object> colIndex) {
+  private Conversions getConversion(Comparable<? extends Object> colIndex) {
       Conversions type = null;
-
       if (colIndex instanceof Double) {
         type = Conversions.DOUBLE;
       }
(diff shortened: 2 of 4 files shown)


gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -76,6 +80,98 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+        colArray[i] = (double) ((Float) col.get(i)).floatValue();
+      }
+    }
+    return colArray;
+  }
+
+  public double mean(String colName) {
+    double[] colArray = toDoubleArray(colName);
+    double arrayMean = StatUtils.mean(colArray);
+    return arrayMean;
```
smolnar82: `return StatUtils.mean(toDoubleArray(colName))` directly?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -76,6 +80,98 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+  }
+
+  public double mean(int colIndex) {
+    return mean(headers.get(colIndex));
+  }
+
+  public double median(String colName) {
+    double[] colArray = toDoubleArray(colName);
+    double arrayMedian = StatUtils.percentile(colArray, 50);
+    return arrayMedian;
```
smolnar82: `return StatUtils. percentile(toDoubleArray(colName), 50)` directly?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -76,6 +80,98 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+  }
+
+  public double mode(int colIndex) {
+    return mode(headers.get(colIndex));
+  }
+
+  public double sum(String colName) {
+    double[] colArray = toDoubleArray(colName);
+    double arraySum = StatUtils.sum(colArray);
+    return arraySum;
```
smolnar82: `return StatUtils. sum(toDoubleArray(colName))` directly?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -76,6 +80,98 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+  }
+
+  public double sum(int colIndex) {
+    return sum(headers.get(colIndex));
+  }
+
+  public double max(String colName) {
+    double[] colArray = toDoubleArray(colName);
+    double arrayMax = StatUtils.max(colArray);
+    return arrayMax;
```
smolnar82: `return StatUtils. max(toDoubleArray(colName))` directly?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -76,6 +80,98 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+  }
+
+  public double max(int colIndex) {
+    return max(headers.get(colIndex));
+  }
+
+  public double min(String colName) {
+    double[] colArray = toDoubleArray(colName);
+    double arrayMin = StatUtils.min(colArray);
+    return arrayMin;
```
smolnar82: `return StatUtils. min(toDoubleArray(colName))` directly?

gateway-shell/pom.xml
```diff
@@ -31,6 +31,10 @@
     <description>Client DSL shell.</description>
 
     <dependencies>
+    	<dependency>
+  			<groupId>org.apache.commons</groupId>
+  			<artifactId>commons-math3</artifactId>
+		</dependency>
```
smolnar82: Indentation is incorrect.

smolnar82:
Another idea: once you added `long` support you may consider `Date` or `Timestamp` columns using `getTime()`. What do you think?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
lmccay4: Thanks, will update these!

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -33,6 +34,9 @@
  */
 public class KnoxShellTable {
 
+  static final int DOUBLE_CONVERSION_METHOD = 0;
+  static final int INTEGER_CONVERSION_METHOD = 1;
+  static final int FLOAT_CONVERSION_METHOD = 2;
```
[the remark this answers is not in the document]
risdenk: An enum might be better here instead of building out an enum manually?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -76,6 +80,98 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
   }
 
+  public double[] toDoubleArray(String colName) throws IllegalArgumentException {
+    List<Comparable<? extends Object>> col = values(colName);
+    double[] colArray = new double[col.size()];
+    int conversionMethod = 0;
+    for (int i = 0; i < col.size(); i++) {
+      if (i == 0) {
+        if (col.get(i) instanceof Integer) {
+          conversionMethod = INTEGER_CONVERSION_METHOD;
```
risdenk: Do we need to do this check and then later do a complicated convert to `double`? Is it possible to just do something like `Double.valueOf(col.get(i))`?

risdenk: https://repl.it/repls/GigaLavenderInternalcommand

Shows that it looks like this works just using `Double.valueOf(...)`

lmccay4: The check is needed as `valueOf` can't be called on a Comparable. You would need to convert to something like string before doing that. I had something similar to this originally (my comment on the jira) where it was `Double.parseDouble(col.get(i).toString());`, which is not efficient. With this though we're just going from whichever underlying datatype it happens to be out to double.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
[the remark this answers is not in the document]
lmccay4: Agreed, updated in new patch. Also added a switch statement, easier to read.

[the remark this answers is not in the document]
lmccay4: > I also miss the support for other number types, such as `byte`, `short` and `long` as a minimum.

Added support for these types in new patch.

gateway-shell/pom.xml
lmccay4: Corrected.

lmccay4:
> Another idea: once you added `long` support you may consider `Date` or `Timestamp` columns using `getTime()`. What do you think?

@smolnar82 I could see this being useful with finding `mode,` but not quite sure how the rest of the methods here would lend themselves to a date/timestamp. `Max` and `min` could also be useful but I would think that would be more of `sort` thing where you would want the table itself displayed in order with max/min on respective ends of the table. Are there any specific use cases you had in mind? This might be going over my head. Filing a separate JIRA might be a good idea for this.

smolnar82:
> > Another idea: once you added `long` support you may consider `Date` or `Timestamp` columns using `getTime()`. What do you think?
> 
> @smolnar82 I could see this being useful with finding `mode,` but not quite sure how the rest of the methods here would lend themselves to a date/timestamp. `Max` and `min` could also be useful but I would think that would be more of `sort` thing where you would want the table itself displayed in order with max/min on respective ends of the table. Are there any specific use cases you had in mind? This might be going over my head. Filing a separate JIRA might be a good idea for this.

I believe selecting the `min` and `max` date/timestamp is very useful and frequently used in applications dealing with those data types (e.g. give me the last time John logged in). It's true that the same can be achieved with sorting (ascending/descending) and selecting the first element from the result. I'll create a new JIRA.

gateway-shell/pom.xml
smolnar82: I still see it wrong, but it should not prevent merging this change.

lmccay4: Odd, it appeared fine for me in my IDE both times, I'll try to play around with it some more, not sure why it is doing that.

risdenk: tabs vs spaces?

pom.xml
```diff
@@ -758,6 +758,11 @@
 
     <dependencyManagement>
         <dependencies>
+        	<dependency>
+  				<groupId>org.apache.commons</groupId>
+  				<artifactId>commons-math3</artifactId>
+  				<version>3.6.1</version>
```
risdenk: Version needs to be a property. See other examples in the `pom.xml` where it would be `commons-math3.version`

lmccay4: Fixed this in the updated patch

gateway-shell/pom.xml
lmccay4: That looks like what it was, I've replaced all whitespace there with spaces in the newest patch so hopefully that is all it was.

pom.xml
```diff
@@ -236,6 +236,7 @@
         <xmltool.version>3.3</xmltool.version>
         <xml-matchers.version>0.10</xml-matchers.version>
         <zookeeper.version>3.5.5</zookeeper.version>
+        <commons-math3.version>3.6.1</commons-math3.version>
```
smolnar82: I really don't want to say no, but dependencies are listed in alphabetical order...

pom.xml
```diff
@@ -759,6 +760,11 @@
 
     <dependencyManagement>
         <dependencies>
+        	<dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-math3</artifactId>
+                <version>${commons-math3.version}</version>
+			</dependency>
```
smolnar82: 3 tabs here

pom.xml
```diff
@@ -759,6 +760,11 @@
 
     <dependencyManagement>
         <dependencies>
+        	<dependency>
```
smolnar82: Spaces and tabs here

pom.xml
lmccay4: Placed it alphabetically in the updated patch.

pom.xml
lmccay4: Corrected in updated patch.

pom.xml
lmccay4: Ah I see. Corrected in updated patch.

lmccay4:
Ignore latest patch, I messed it up.

lmccay4:
Latest push should find all reviews addressed and free of merge conflicts, hopefully.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -86,6 +96,117 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+      else if (colIndex instanceof Byte) {
+        type = Conversions.BYTE;
+      }
+      else if (colIndex instanceof Short) {
+        type = Conversions.SHORT;
+      }
+      else if (colIndex instanceof Long) {
+        type = Conversions.LONG;
+      }
+      return type;
```
risdenk: Should have some default case?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -86,6 +96,117 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+    }
+
+    public double[] toDoubleArray(String colName) throws IllegalArgumentException {
+      List<Comparable<? extends Object>> col = values(colName);
+      double[] colArray = new double[col.size()];
+      Conversions conversionMethod = null;
+      for (int i = 0; i < col.size(); i++) {
+        if (i == 0) {
+          conversionMethod = getConversion(col.get(i));
```
risdenk: `getConversion` can return `null`? This will then fall through the switch statement and cause an empty double array to be returned?

pom.xml
```diff
@@ -759,6 +760,11 @@
 
     <dependencyManagement>
         <dependencies>
+            <dependency>
```
risdenk: Please don't throw this in a random location. The `commons-*` dependencies are all together so place with them.

gateway-shell/pom.xml
```diff
@@ -31,6 +31,10 @@
     <description>Client DSL shell.</description>
 
     <dependencies>
+        <dependency>
```
risdenk: Can you put this with the other `commons-*` dependencies in this module?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -86,6 +96,117 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+      else if (colIndex instanceof Short) {
+        type = Conversions.SHORT;
+      }
+      else if (colIndex instanceof Long) {
+        type = Conversions.LONG;
+      }
+      return type;
+    }
+
+    public double[] toDoubleArray(String colName) throws IllegalArgumentException {
```
risdenk: Can you add some javadocs here since its a public method?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -35,6 +35,16 @@
  */
 @JsonFilter("knoxShellTableFilter")
 public class KnoxShellTable {
+
+    public enum Conversions {
```
risdenk: Does this need to be public?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -86,6 +96,117 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+      List<Comparable<? extends Object>> col = values(colName);
+      double[] colArray = new double[col.size()];
+      Conversions conversionMethod = null;
+      for (int i = 0; i < col.size(); i++) {
+        if (i == 0) {
+          conversionMethod = getConversion(col.get(i));
+        }
+
+        switch (conversionMethod) {
+          case DOUBLE:
```
risdenk: nit: Above you use `Conversions.DOUBLE` but here you use just the enum `DOUBLE` - it would be better to be consistent. I personally prefer `Conversions.DOUBLE` so it doesn't look like its a built in constant.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -86,6 +96,117 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
     return col;
   }
 
+  public Conversions getConversion(Comparable<? extends Object> colIndex) {
```
risdenk: This doesn't need to be public

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
lmccay4: Doesn't seem to compile that way: The qualified case label KnoxShellTable.Conversions.DOUBLE must be replaced with the unqualified enum constant DOUBLE

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
lmccay4: Changed that to throw an `IllegalArgumentException` in an else on line 120 should any type not specified in the enum be passed in to `getConversion.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
lmccay4: Changed to private (no need to be public) and added javadocs to the other public methods (mean, median, etc).

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
lmccay4: No - changed to private.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
lmccay4: I'm not entirely sure a default case is needed since the updated patch now has an else that throws an `IllegalArgumentException`, as per the response to the review below.

pom.xml
lmccay4: Changed in most recent patch.

lmccay4:
@risdenk New patch is available for review

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -86,6 +96,172 @@ public KnoxShellTable value(Comparable<? extends Object> value) {
...
+        type = Conversions.BYTE;
+      }
+      else if (colIndex instanceof Short) {
+        type = Conversions.SHORT;
+      }
+      else if (colIndex instanceof Long) {
+        type = Conversions.LONG;
+      }
+      else {
+          throw new IllegalArgumentException();
```
risdenk: nit: would be good to include some String here as to why

lmccay4: Thanks for the review! Will follow up with this on another change.

lmccay4: Will address this in KNOX-2060

lmccay:
@smolnar82 - Since you were involved in the review of previous revisions of this patch can you take a look and make sure your concerns have been resolved or have follow up JIRAs?

PR #162, opened 2019-10-04
KNOX-2023 - Recording KnoxShellTable builder/filter chain and providing rollback/replay capabilities using the call history as well as allowing end-users to export JSON without data (in this case only the call history will be serialized)

smolnar82:
## What changes were proposed in this pull request?

The following features have been added with this change in `knoxshell`:
- recording KnoxShellTable builder/filter chain
- providing rollback/replay capabilities using the call history
- allowing end-users to export JSON without data (in this case, only the call history will be serialized)

## How was this patch tested?
Executing a full build (where I added/updated new/existing JUnit tests):
```
$ mvn clean -Dshellcheck=true -T1C verify -Prelease,package
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 17:58 min (Wall Clock)
[INFO] Finished at: 2019-10-04T18:08:41+02:00
[INFO] Final Memory: 389M/1837M
[INFO] ------------------------------------------------------------------------

```

In addition to running JUnit tests the following manual test steps were executed:

1.) Building a table with the JDBC builder:
```
locations = KnoxShellTable.builder().jdbc().driver("org.apache.derby.jdbc.EmbeddedDriver").connectTo("jdbc:derby:/Users/smolnar/test/derbyDb").sql("SELECT * FROM sample.locations where zip < 15")
===> LOCATIONS
+--------+------------+----------+----------+--------------+
|  ZIP   |  COUNTRY   |  STATE   |   CITY   |  POPULATION  |
+--------+------------+----------+----------+--------------+
|   1    |     US     |    NY    |  City1   |    100000    |
|   2    |     US     |    NY    |  City2   |    100000    |
|   3    |     US     |    NY    |  City3   |    100000    |
|   4    |     US     |    NY    |  City4   |    100000    |
|   5    |     US     |    NY    |  City5   |    100000    |
|   6    |     US     |    NY    |  City6   |    100000    |
|   7    |     US     |    NY    |  City7   |    100000    |
|   8    |     US     |    NY    |  City8   |    100000    |
|   9    |     US     |    NY    |  City9   |    100000    |
|   10   |     US     |    NY    |  City10  |    100000    |
|   11   |     US     |    NY    |  City11  |    100000    |
|   12   |     US     |    NY    |  City12  |    100000    |
|   13   |     US     |    NY    |  City13  |    100000    |
|   14   |     US     |    NY    |  City14  |    100000    |
[...]

PR #165, opened 2019-10-15
KNOX-2057 - Using unique table IDs through different test cases even if they are in different test classes (JVM is shared in during parallel execution)

smolnar82:

The Maven `surefire-plugin` reuse JVM forks by default which might end-up overwriting the KnoxShellTable call history with undesired value(s) during parallel execution. To avoid this we need to make sure to use unique table IDs in our tests.

## How was this patch tested?

Ran JUnit tests locally (note: I did it previously many times but I couldn't reproduce this issue).

Executed the same tests as described in #162 .

### DIFF
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
@@ -42,13 +42,12 @@ public class KnoxShellTable {
   String title;
   long id;
 
-  public KnoxShellTable title(String title) {
-    this.title = title;
-    return this;
+  KnoxShellTable() {
+    this.id = getUniqueTableId();
   }
 
-  public KnoxShellTable id(long id) {
-    this.id = id;
+  public KnoxShellTable title(String title) {
+    this.title = title;
     return this;
   }
 
@@ -113,7 +112,7 @@ public class KnoxShellTable {
   }
 
   public static KnoxShellTableBuilder builder() {
-    return new KnoxShellTableBuilder(getUniqueTableId());
+    return new KnoxShellTableBuilder();
   }
 
   static long getUniqueTableId() {
--- a/gateway-shell/src/test/java/org/apache/knox/gateway/shell/table/KnoxShellTableTest.java
+++ b/gateway-shell/src/test/java/org/apache/knox/gateway/shell/table/KnoxShellTableTest.java
@@ -362,11 +362,9 @@ public class KnoxShellTableTest {
   @Test
   public void shouldReturnDifferentCallHistoryForDifferentTables() throws Exception {
     final KnoxShellTable table1 = new KnoxShellTable();
-    table1.id(1);
-    KnoxShellTableCallHistory.getInstance().saveCall(1, new KnoxShellTableCall("class1", "method1", true, Collections.singletonMap("param1", String.class)));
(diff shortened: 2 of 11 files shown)


smolnar82:
@risdenk Could you please review this one? Thanks!

gateway-shell/src/test/java/org/apache/knox/gateway/shell/table/KnoxShellTableTest.java
```diff
@@ -362,11 +362,13 @@ public Object answer() {
   @Test
   public void shouldReturnDifferentCallHistoryForDifferentTables() throws Exception {
     final KnoxShellTable table1 = new KnoxShellTable();
-    table1.id(1);
-    KnoxShellTableCallHistory.getInstance().saveCall(1, new KnoxShellTableCall("class1", "method1", true, Collections.singletonMap("param1", String.class)));
+    final long id1 = KnoxShellTable.getUniqueTableId();
+    table1.id(id1);
```
risdenk: I haven't looked at the implementation but shouldn't a table get a unique ID automatically? Do we need to set an id on the table? We should be able to ask the table for the id?

smolnar82: I've to double-check but there might be a chicken/egg issue between the builder and the table object...I'll let you know.

smolnar82:
@risdenk I did a minor refactor to have the `KnoxShellTable` generate its own ID instead of setting it by the filter/builder. Please review. Thanks!

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -113,7 +112,7 @@ public long getId() {
   }
 
   public static KnoxShellTableBuilder builder() {
-    return new KnoxShellTableBuilder(getUniqueTableId());
+    return new KnoxShellTableBuilder();
   }
 
   static long getUniqueTableId() {
```
risdenk: Probably doesn't need to be accessible anymore? Also would make more sense as a UUID instead of some long based on time? Doesn't really matter for this PR though

smolnar82: What do you mean by `doesn't need to be accessible`? The `builder` method? That must be `public`; that is the entry point...one usually builds a table like this:
`locations = KnoxShellTable.builder().jdbc().driver("org.apache.derby.jdbc.EmbeddedDriver").connectTo("jdbc:derby:/Users/smolnar/test/derbyDb").sql("SELECT * FROM sample.locations where zip < 10")`

risdenk: `static long getUniqueTableId()` but looks like it is used in tests.

smolnar82: Yes, I just wanted to reuse code...

smolnar82:
Reopened to trigger a new Travis CI build (the previous error was unrelated).

smolnar82:
Created https://issues.apache.org/jira/browse/KNOX-2062 to address `RemoteConfigurationMonitorTest` issues in Travis CI.

smolnar82:
@risdenk - can this be merged?

PR #168, opened 2019-10-18
KNOX-2060 - Extend KnoxShellTable statistics methods to work with columns of Strings

lmccay4:

Added ability to convert columns of type String to double[] for operations with StatUtils (mean, median, mode, etc). 

## How was this patch tested?

Manual and unit tests.

### DIFF
--- a/gateway-shell/src/test/java/org/apache/knox/gateway/shell/table/KnoxShellTableTest.java
+++ b/gateway-shell/src/test/java/org/apache/knox/gateway/shell/table/KnoxShellTableTest.java
@@ -385,9 +385,9 @@ public class KnoxShellTableTest {
       Long long20 = (long) 200000000;
       Long long30 = (long) 300000000;
 
-      String string10 = "-10";
-      String string20 = "-20";
-      String string30 = "-30.00";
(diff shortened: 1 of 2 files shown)


gateway-shell/src/test/java/org/apache/knox/gateway/shell/table/KnoxShellTableTest.java
```diff
@@ -385,11 +385,15 @@ public void testMean() throws Exception {
       Long long20 = (long) 200000000;
       Long long30 = (long) 300000000;
 
+      String string10 = "-10";
+      String string20 = "-20";
+      String string30 = "-30.00";
```
risdenk: Need to test positives? Also would do more and less than two digits. Probably makes sense to randomize this. Random and then multiply to make a double. Same for the other values

lmccay4: Ah right, my bad, meant to change that test back to positives and negatives before pushing it up. Thanks for catching that. See updated patch.

PR #172, opened 2019-10-29
KNOX-2025 - KnoxShellTable - Join Builder on Method should accept Col Names

lmccay4:

Overloaded method to join on columns of the same data with different names (ex. ID vs Student_ID)

## How was this patch tested?

Manual and unit tests.


gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/JoinKnoxShellTableBuilder.java
```diff
@@ -52,6 +52,12 @@ public KnoxShellTable on(String columnName) {
     return on(leftIndex, rightIndex);
   }
 
+  public KnoxShellTable on(String columnName, String columnName2) {
```
smolnar82: nit: if I were you I'd use `columnNameLeft` and `columnNameRight` argument names

PR #173, opened 2019-10-31
KNOX-2068 - Let end-users add a new service definition from scratch

smolnar82:

From now on end-users can submit their own service definition from scratch (a simple template is provided).

## How was this patch tested?

See [this video](https://streamable.com/r09t4) about my integration tests.


smolnar82:
@risdenk @lmccay @moresandeep @pzampino Can any of you review my changes? Thanks!

PR #175, opened 2019-11-03
KNOX-2075: Druid coordinator ui is broken

mackong:

Add new rewrite rule for druid coordinator ui.

## How was this patch tested?

Before:
Screenshot from 2019-11-03 17-48-12
After:
Screenshot from 2019-11-03 17-49-52


risdenk:
Change looks good. These are new rules so won't affect older versions. The new rules are specific and won't conflict most likely.

PR #176, opened 2019-11-05
KNOX-1842 - Upgrade httpclient to 4.5.10

risdenk:

HttpClient 4.5.7 broke url normalization. Knox
didn't have any tests for this case and so we
had to revert after the fact. HttpClient 4.5.8
fixed a lot of the url normalization and some
libraries decided to turn url normalization off.

This commit does the following:
* Adds a test for %2F - KNOX-1005
    * This test passes under HttpClient 4.5.6 and 4.5.8+
    * It breaks as expected under HttpClient 4.5.7
* Adds an explicit config enabling url normalization
    * Ensures that we are in control of url normalization
    * Adds a test for this configuration as well
* Test with both HttpClient normalization enabled and disabled
    * `rest-assured` doesn't expose `RequestConfig` to disable
url normalization
    * Shows how to use HttpClient in `GatewayBasicFuncTest`

All the url safe characters like %2F are fixed by HTTPCLIENT-1968.

The case of `/abc///def` is normalized to `/abc/def` the same
way that Knox does internally with `getPathInfo` and Java URI.

## How was this patch tested?

* Tested against multiple versions of HttpClient (4.5.6, 4.5.7, and 4.5.10)
* `mvn -T.5C clean verify -Ppackage,release -Dshellcheck`

### DIFF
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
@@ -192,7 +192,7 @@ public class DefaultHttpClientFactory implements HttpClientFactory {
     }
   }
 
-  private static RequestConfig getRequestConfig( FilterConfig config ) {
+  static RequestConfig getRequestConfig( FilterConfig config ) {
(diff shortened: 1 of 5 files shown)


risdenk:
Ping @smolnar82 as well since you were asking about this Jira

PR #177, opened 2019-11-05
[WIP] KNOX-2095 - Adding in DefaultDispatch code and tests to handle 504 errors

jameschen1519:

Currently, Knox masks all connection errors as 500 errors, when they may be more accurately described using other error codes, especially 504. A change has been made to display a 504 error in the event of a socket timeout..

## How was this patch tested?

Ran ant verify under Knox 1.2.0. The patch was generated against the Knox 1.2.0 branch, and subsequently applied to Knox 1.4.0. Currently ant verify is failing, but these are most likely transient errors as the same errors appear in master. Still running tests.

### DIFF
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultDispatch.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultDispatch.java
@@ -20,6 +20,9 @@ package org.apache.knox.gateway.dispatch;
 import org.apache.http.Header;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
+import org.apache.http.HttpResponseFactory;
+import org.apache.http.HttpStatus;
+import org.apache.http.HttpVersion;
 import org.apache.http.client.methods.HttpDelete;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpHead;
@@ -29,6 +32,8 @@ import org.apache.http.client.methods.HttpPut;
 import org.apache.http.client.methods.HttpPatch;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.entity.ContentType;
+import org.apache.http.impl.DefaultHttpResponseFactory;
+import org.apache.http.message.BasicStatusLine;
 import org.apache.knox.gateway.SpiGatewayMessages;
 import org.apache.knox.gateway.SpiGatewayResources;
 import org.apache.knox.gateway.audit.api.Action;
@@ -49,6 +54,7 @@ import javax.servlet.http.HttpServletResponse;
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
@@ -138,6 +144,10 @@ public class DefaultDispatch extends AbstractGatewayDispatch {
         }
       }
       auditor.audit( Action.DISPATCH, outboundRequest.getURI().toString(), ResourceType.URI, ActionOutcome.SUCCESS, RES.responseStatus( statusCode ) );
+    } catch( SocketTimeoutException e ) {
+        //Set a 504 instead of throwing an IO exception in the event of a socket timeout,
+        //as an IO exception forces a 500 to be displayed.
+        inboundResponse = generateInboundResponseOverride(HttpStatus.SC_GATEWAY_TIMEOUT);
     } catch( Exception e ) {
       // We do not want to expose back end host. port end points to clients, see JIRA KNOX-58
       auditor.audit( Action.DISPATCH, outboundRequest.getURI().toString(), ResourceType.URI, ActionOutcome.FAILURE );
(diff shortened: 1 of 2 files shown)


jameschen1519:
Hmm. It looks like the check at https://api.travis-ci.org/v3/job/607855611/log.txt failed, but the logs suggest that it may be running plugins that aren't thread safe. Assuming that this is a threading issue, is there any way to rerun the tests, possibly without parallelism?

risdenk:
The error was 

```
[ERROR] Failures: 
[ERROR]   GatewayCorrelationIdTest.testTestService:209 
Expected: is <46>
     but: was <45>
```

Not an error I've seen before. I retriggered the JDK 11 job. We have some flaky tests related to ZK but not that test.

jameschen1519:
Hi Kevin, sorry for the late followup; was out yesterday. Seems like it failed on gateway-service-remoteconfig again; was it still the testTestService test? I don't have JDK 11 installed on this machine and I'd like to confirm before trying to pursue the error.
(On that note, is there any good way to check the Surefire reports/view build artifacts?)

jameschen1519:
Grasping at straws here, but looking through the test case at https://github.com/apache/knox/blob/89caa5feeed706abc8d7ce1407830ae00d97d405/gateway-test/src/test/java/org/apache/knox/gateway/GatewayCorrelationIdTest.java, is it possible that the reduced timeout might be causing the issue? I'm not completely sure how the test works, but with the change in this PR, all connection attempts that experience a socket timeout are automatically given a 403, whereas without the change, there would at least be an attempt to contact the failover nodes.

...then again, I suppose this wouldn't explain the successes in JDK8. It's a bit difficult to tell without looking at the reports unfortunately.

risdenk:
@jameschen1519 I don't think the test failures are related to your change. 

> (On that note, is there any good way to check the Surefire reports/view build artifacts?)

The Travis build details are linked in the pr and then you go to the specific build and build log. That has the same output if you were to run locally.

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultDispatch.java
```diff
@@ -138,6 +144,10 @@ protected HttpResponse executeOutboundRequest( HttpUriRequest outboundRequest )
         }
       }
       auditor.audit( Action.DISPATCH, outboundRequest.getURI().toString(), ResourceType.URI, ActionOutcome.SUCCESS, RES.responseStatus( statusCode ) );
+    } catch( SocketTimeoutException e ) {
+        //Set a 504 instead of throwing an IO exception in the event of a socket timeout,
+        //as an IO exception forces a 500 to be displayed.
```
risdenk: So we need an audit and log line here just like the generic `Exception` catch block. 

The other thing we need to do is make sure this doesn't break the HA case. There a failover needs to happen.

jameschen1519: Thanks for the feedback. I'll add in the audit and log lines, but just to confirm, when does a socket timeout usually occur? So far all of the socket timeouts that have occurred to me have been caused artificially by lowering the socketTimeout parameter (so I don't have any information on this), but I've been told that if a socket timeout occurs on one node, it is likely to occur on other nodes. Are there situations in which this isn't the case?

risdenk: Yes one server could be dying or overloaded where the second one isn't. That would cause socket timeout on one server but failover would work correctly.

There can definitely be socket timeouts in active/passive and active/active case.

active passive - active server gets overloaded
active/active - could be something wrong on one of the two hosts

Yes you could failover only to have to wait again but that is what failover is designed to do. You can't just skip failover in some cases since you don't have enough information to say whether the failover will succeed or not. The fix for this would be some sort of active list of backend hosts, but Knox doesn't have that today. That is a bigger feature for sure.

shanyu: @risdenk , for active/passive, it doesn't make sense to failover from active to passive in the case of socket timeout. For active/active, it may have value to failover. However, I have 2 arguments against this:
1) Usually the client side that talks to knox has its own timeout (think about a Jupyter notebook), so letting service failover in the case of socket timeout multiple times would eventually cause client side timeout.
2) Most likely the timeout is because the request itself took too much time exceeding the socket timeout config (considering Livy session creation), so failover to another server would also cause a socket timeout as well.

And for active-active type of service, if really needed we can always retry at the client side. 

I think letting client side know that the request caused timeout is very important, instead of Letting them face a generic 500 error without knowing what to do next.

Or we could

risdenk: @shanyu you are arguing for changing the semantics of failover when this change is all about just handling `SocketTimeoutException` as a 504 instead of a 500.

Failover is a completely separate mechanism. If the topology is not configured for failover then Knox won't failover. The `SocketTimeoutException` handling shouldn't be any different in the case of a failover. It needs to tell Knox that a failure happened. Not just assume it knows best and ignore the possibility to failover.

Whether the current failover logic is correct or not is a completely separate issue from handling `SocketTimeoutException`. 

> I think letting client side know that the request caused timeout is very important, instead of Letting them face a generic 500 error without knowing what to do next.

As originally proposed, yes handling `SocketTimeoutException` and setting the response to 504 makes sense, but it must do so in the framework that exists for failover today.

smolnar82:

PR #178, opened 2019-11-06
KNOX-2071 - Configurable maximum token lifetime for TokenStateService

pzampino:

Added the ability to specify the maximum token lifetime (i.e., token renewal limit) via topology service params.

## How was this patch tested?

Ran automated unit/integration tests and manual tests.
Manual tests included configuration a short maximum token lifetime, attempting to renew a token (via curl), and verifying the error response.

### DIFF
--- a/gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
+++ b/gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
@@ -49,6 +49,7 @@ import java.security.PrivilegedAction;
 import java.security.cert.X509Certificate;
 import java.security.interfaces.RSAPrivateKey;
 import java.security.interfaces.RSAPublicKey;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -680,6 +681,40 @@ public class TokenServiceResourceTest {
   }
 
   @Test
+  public void testTokenRenewal_Enabled_WithDefaultMaxTokenLifetime() throws Exception {
+    final String caller = "yarn";
+
+    // Max lifetime duration is 10ms
+    Map.Entry<TestTokenStateService, Response> testResult =
+                  doTestTokenRenewal(true, caller, null, createTestSubject(caller));
+
+    TestTokenStateService tss = testResult.getKey();
+    assertEquals(1, tss.issueTimes.size());
+    String token = tss.issueTimes.keySet().iterator().next();
+
+    // Verify that the configured max lifetime was honored
+    assertEquals(tss.getDefaultMaxLifetimeDuration(), tss.getMaxLifetime(token) - tss.getIssueTime(token));
+  }
+
+
+  @Test
+  public void testTokenRenewal_Enabled_WithConfigurableMaxTokenLifetime() throws Exception {
+    final String caller = "yarn";
+
+    // Max lifetime duration is 10ms
+    Map.Entry<TestTokenStateService, Response> testResult =
+                                              doTestTokenRenewal(true, caller, 10L, createTestSubject(caller));
+
+    TestTokenStateService tss = testResult.getKey();
+    assertEquals(1, tss.issueTimes.size());
+    String token = tss.issueTimes.keySet().iterator().next();
+
+    // Verify that the configured max lifetime was honored
+    assertEquals(10L, tss.getMaxLifetime(token) - tss.getIssueTime(token));
+  }
+
+
+  @Test
   public void testTokenRevocation_ServerManagedStateNotConfigured() throws Exception {
     Response renewalResponse = doTestTokenRevocation(null, null, null);
     validateRevocationResponse(renewalResponse,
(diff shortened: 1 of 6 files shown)


gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
```diff
@@ -23,6 +23,7 @@
 import com.nimbusds.jose.JWSVerifier;
 import com.nimbusds.jose.crypto.RSASSASigner;
 import com.nimbusds.jose.crypto.RSASSAVerifier;
+import org.apache.commons.lang3.tuple.Pair;
```
risdenk: Java has a Pair class already? Don't think we need commons-lang3.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/TokenStateService.java
```diff
@@ -27,22 +27,41 @@
...
+
   /**
    * Add state for the specified token.
    *
    * @param token     The token.
    * @param issueTime The time the token was issued.
    */
-  void addToken(JWTToken token, long issueTime);
-
+  void addToken(JWTToken token, Long issueTime);
```
risdenk: Is there a reason that all the `long` primitives were changed to `Long` objects? I didn't seem an obvious reason why this was necessary. The `long` primitive should be fine?

risdenk: I see this happened basically everywhere but only commenting once here.

pzampino: I made that change before subsequently exposing the methods for getting the default renewal interval and maximum lifetime values. The Long vs long was so that a null could be passed in, indicating that the default should be applied. Now, clients have to pass in a value, which they can get from the TokenStateService itself, so the params could be reverted to the long type again.

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
pzampino: javafx.util.Pair ?

risdenk: Hmm yea, I thought JDK 8 had a "Pair" class. 

Maybe https://docs.oracle.com/javase/8/docs/api/java/util/AbstractMap.SimpleImmutableEntry.html?

JDK 9+ has https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#entry(K,V)


https://stackoverflow.com/questions/24328679/does-java-se-8-have-pairs-or-tuples

pzampino: commons-lang3 is used in other modules already, and it is only a test dependency in this module.

risdenk: Right but would be better to get rid of it in other modules instead of just using that as an argument to keep adding it. Once it is added as a test module that opens it up to other usages. The only way to prevent usage is to make sure its not included in the pom.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/TokenStateService.java
risdenk: The only concern I had was performance based since `Long` is heavier than `long`. I don't know how often these apis are called, but if possible primitive seemed to make more sense when reading through.

pzampino: I agree, and as I've mentioned, it was an iterative change which is no longer needed. I'll revert that, and submit an update to this PR.

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
pzampino: I'm just going to replace it with Map.Entry for this test

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
```diff
@@ -679,6 +680,40 @@ public void testTokenRenewal_Enabled_WithRenewersWithValidSubject() throws Excep
...
+    assertEquals(1, tss.issueTimes.size());
+    String token = tss.issueTimes.keySet().iterator().next();
+
+    // Verify that the configured max lifetime was honored
+    assertEquals(tss.getDefaultMaxLifetimeDuration(), tss.getMaxLifetime(token) - tss.getIssueTime(token));
+  }
+
+
+  @Test
+  public void testTokenRenewal_Enabled_WithConfigurableMaxTokenLifetime() throws Exception {
```
smolnar82: Do we need the underscores in the method name (see in another new method above too)?

pzampino: It's test code, and it helps me to quickly distinguish test methods that vary only slightly in name. So, are the needed? No. Does it matter? It's test code.

PR #180, opened 2019-11-07
KNOX-2105 - KnoxShell support for token renewal and revocation

pzampino:

Adds KnoxShell support for Knox token renewal and revocation.

## How was this patch tested?

Automated unit/integration tests, manual tests. Manual testing included the groovy script attached to the JIRA issue, invoked with different user credentials (some in the configured whitelist, others not).

### DIFF
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/ErrorResponse.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/ErrorResponse.java
@@ -19,7 +19,7 @@ package org.apache.knox.gateway.shell;
 
 import org.apache.http.HttpResponse;
 
-class ErrorResponse extends RuntimeException {
+public class ErrorResponse extends RuntimeException {
 
   HttpResponse response;
 
@@ -28,7 +28,7 @@ class ErrorResponse extends RuntimeException {
     this.response = response;
   }
 
-  public HttpResponse getReponse() {
+  public HttpResponse getResponse() {
     return response;
   }
 
--- a/gateway-shell/src/test/java/org/apache/knox/gateway/shell/knox/token/TokenTest.java
+++ b/gateway-shell/src/test/java/org/apache/knox/gateway/shell/knox/token/TokenTest.java
@@ -129,47 +129,21 @@ public class TokenTest {
   }
 
   private void testRenewToken(boolean setDoAsUser, String doAsUser) throws Exception {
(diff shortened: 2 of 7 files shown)


gateway-shell/src/test/java/org/apache/knox/gateway/shell/knox/token/TokenTest.java
```diff
@@ -76,4 +127,100 @@ private void testToken(boolean setDoAsUser, String doAsUser) {
...
+    assertEquals("http://localhost/base/knoxtoken/api/v1/token/renew" + (shouldHaveDoAs ? ("?doAs=" + doAsUser) : ""),
+                 request.getRequestURI().toString());
+
+    assertEquals(testToken, request.getToken());
+
+    Callable<TokenLifecycleResponse> callable = request.callable();
+    try {
+      callable.call();
+    } catch (Exception e) {
+      e.printStackTrace();
```
risdenk: Should this fail the test? (same with other catch locations)

gateway-shell/src/main/java/org/apache/knox/gateway/shell/ErrorResponse.java
```diff
@@ -28,7 +28,7 @@
     this.response = response;
   }
 
-  public HttpResponse getReponse() {
```
risdenk: This change didn't need require changes anywhere else?

pzampino: Actually, I should remove the printing of the stack trace. The call should fail, but is required to setup the actual underlying request (so the setting of the entity can be verified). Short answer: no, it should not fail the test in either case.

gateway-shell/src/test/java/org/apache/knox/gateway/shell/knox/token/TokenTest.java
pzampino: Actually, I should remove the printing of the stack trace. The call should fail, but is required to setup the actual underlying request (so the setting of the entity can be verified). Short answer: no, it should not fail the test in either case.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/ErrorResponse.java
pzampino: No, it couldn't even be used elsewhere because the entire class was package-private before this change.

PR #191, opened 2019-11-12
KNOX-2122 - Code cleanup from static code analysis

risdenk:

* Diamond operator
* Static method calls instead of on instances
* Remove null check before instanceof
* Use computeIfAbsent where possible
* Remove simple nested if statements
* Remove redundant cast
* Use entrySet for iteration of maps

## How was this patch tested?

`mvn -T.75C verify -U -Ppackage,release -Dshellcheck`


smolnar82:
Let me know if you need another hand here!

PR #195, opened 2019-11-15
KNOX-2126. Add an API to retrieve the public key used for Knox JWT verification.

sidseth:

## How was this patch tested?
Not tested yet. Looking for some feedback on client side changes, if any are required.

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -19,6 +19,7 @@ package org.apache.knox.gateway.service.knoxtoken;
 
 import java.security.KeyStoreException;
 import java.security.Principal;
+import java.security.PublicKey;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.X509Certificate;
@@ -40,6 +41,7 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.security.SubjectUtils;
 import org.apache.knox.gateway.services.ServiceType;
@@ -64,6 +66,9 @@ public class TokenResource {
   private static final String TARGET_URL = "target_url";
   private static final String ENDPOINT_PUBLIC_CERT = "endpoint_public_cert";
   private static final String BEARER = "Bearer";
+  // VisibleForTesting
+  static final String PUBLIC_KEY = "public_key";
+  static final String ALGORITHM = "algorithm";
   private static final String TOKEN_TTL_PARAM = "knox.token.ttl";
   private static final String TOKEN_AUDIENCES_PARAM = "knox.token.audiences";
   private static final String TOKEN_TARGET_URL = "knox.token.target.url";
@@ -78,6 +83,7 @@ public class TokenResource {
   static final String RESOURCE_PATH = "knoxtoken/api/v1/token";
   static final String RENEW_PATH = "/renew";
   static final String REVOKE_PATH = "/revoke";
+  static final String PUBLIC_KEY_PATH = "/publickey";
   private static final String TARGET_ENDPOINT_PULIC_CERT_PEM = "knox.token.target.endpoint.cert.pem";
   private static TokenServiceMessages log = MessagesFactory.get(TokenServiceMessages.class);
   private long tokenTTL = TOKEN_TTL_DEFAULT;
@@ -276,6 +282,39 @@ public class TokenResource {
     }
 
     return resp;
+  }
+
+  @GET
+  @Path(PUBLIC_KEY_PATH)
+  @Produces ({APPLICATION_JSON})
+  public Response publicKey() {
+    GatewayServices services = (GatewayServices) request.getServletContext()
+        .getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
+
+    KeystoreService ks = services.getService(ServiceType.KEYSTORE_SERVICE);
+
+    try {
+      PublicKey publicKey = ks.getSigningKeystore().getCertificate(getSigningKeyAlias()).getPublicKey();
+      publicKey.getAlgorithm();
+      byte[] pubKeyBytes = publicKey.getEncoded();
+      String base64Key = Base64.encodeBase64String(pubKeyBytes);
+      Map<String, String> map = new HashMap<>();
+      map.put(PUBLIC_KEY, base64Key);
+      map.put(ALGORITHM, publicKey.getAlgorithm());
+
+      String jsonResponse = JsonUtils.renderAsJsonString(map);
+      return Response.ok().entity(jsonResponse).build();
+    } catch (KeyStoreException | KeystoreServiceException e) {
+      return Response.serverError().build();
+      // TODO: Look at serverError(). ... (Ref: ServiceDefinitionResource for more info)
+    }
+  }
+
+  private String getSigningKeyAlias() {
+    GatewayConfig config = (GatewayConfig) request.getServletContext()
+        .getAttribute(GatewayConfig.GATEWAY_CONFIG_ATTRIBUTE);
+    String alias = config.getSigningKeyAlias();
+    return (alias == null) ? GatewayConfig.DEFAULT_SIGNING_KEY_ALIAS : alias;
   }
 
   private X509Certificate extractCertificate(HttpServletRequest req) {
--- a/gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
+++ b/gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
@@ -23,11 +23,17 @@ import com.nimbusds.jose.JWSSigner;
 import com.nimbusds.jose.JWSVerifier;
 import com.nimbusds.jose.crypto.RSASSASigner;
 import com.nimbusds.jose.crypto.RSASSAVerifier;
+import java.security.KeyStore;
+import java.security.cert.Certificate;
+import javax.ws.rs.core.Response.Status;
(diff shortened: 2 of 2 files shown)


gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -278,6 +284,39 @@ public Response revoke(String token) {
...
+
+      String jsonResponse = JsonUtils.renderAsJsonString(map);
+      return Response.ok().entity(jsonResponse).build();
+    } catch (KeyStoreException | KeystoreServiceException e) {
+      return Response.serverError().build();
+      // TODO: Look at serverError(). ... (Ref: ServiceDefinitionResource for more info)
+    }
+  }
+
+  private String getSigningKeyAlias() {
```
risdenk: I think the code already accounts for the `alias == null` case?

https://github.com/apache/knox/blame/master/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java#L749

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -278,6 +284,39 @@ public Response revoke(String token) {
...
+    KeystoreService ks = services.getService(ServiceType.KEYSTORE_SERVICE);
+
+    try {
+      PublicKey publicKey = ks.getSigningKeystore().getCertificate(getSigningKeyAlias()).getPublicKey();
+      publicKey.getAlgorithm();
+      byte[] pubKeyBytes = publicKey.getEncoded();
+      String base64Key = Base64.encodeBase64String(pubKeyBytes);
+      Map<String, String> map = new HashMap<>();
+      map.put(PUBLIC_KEY, base64Key);
+      map.put(ALGORITHM, publicKey.getAlgorithm());
```
risdenk: Do we need to expose the algorithm here? Wouldn't it be better to normalize? Thinking about consumers of this they would need to be able to deal with different algorithms before reading the public key?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -278,6 +284,39 @@ public Response revoke(String token) {
...
+    GatewayServices services = (GatewayServices) request.getServletContext()
+        .getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
+
+    KeystoreService ks = services.getService(ServiceType.KEYSTORE_SERVICE);
+
+    try {
+      PublicKey publicKey = ks.getSigningKeystore().getCertificate(getSigningKeyAlias()).getPublicKey();
+      publicKey.getAlgorithm();
+      byte[] pubKeyBytes = publicKey.getEncoded();
+      String base64Key = Base64.encodeBase64String(pubKeyBytes);
```
risdenk: Is this base64 encoded pem format then? Might be worth a separate method that takes the `PublicKey` and describes exactly what is being returned in what format. PEM/DER/base64/etc. 

I don't think it changes the functionality but would make it more clear.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -278,6 +284,39 @@ public Response revoke(String token) {
...
+  @Produces ({APPLICATION_JSON})
+  public Response publicKey() {
+    GatewayServices services = (GatewayServices) request.getServletContext()
+        .getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
+
+    KeystoreService ks = services.getService(ServiceType.KEYSTORE_SERVICE);
+
+    try {
+      PublicKey publicKey = ks.getSigningKeystore().getCertificate(getSigningKeyAlias()).getPublicKey();
+      publicKey.getAlgorithm();
```
risdenk: Nit: looks like an extra line?

pzampino: The result of this (publicKey.getAlgorithm()) is ignored? Is it necessary?

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
```diff
@@ -778,6 +785,32 @@ public void testTokenRevocation_Enabled_WithRenewersWithValidSubject() throws Ex
...
+        Base64.decodeBase64(responseMap.get(TokenResource.PUBLIC_KEY)), publicKey.getEncoded());
+
+  }
+
+  @Test
+  public void testGetPublicKey_error() throws Exception {
+    TokenResource tr = createTokenResourceForPublicKeyTests(true);
+
+    Response response = tr.publicKey();
+    assertEquals("Unexpoected response code", Status.INTERNAL_SERVER_ERROR.getStatusCode(),
```
pzampino: Typo: Unexpoected

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
```diff
@@ -778,6 +785,32 @@ public void testTokenRevocation_Enabled_WithRenewersWithValidSubject() throws Ex
...
+  public void testGetPublicKey() throws Exception {
+    TokenResource tr = createTokenResourceForPublicKeyTests(false);
+
+    Response response = tr.publicKey();
+    assertEquals("Unexpected response code", Status.OK.getStatusCode(), response.getStatus());
+    String retString = response.getEntity().toString();
+    Map<String, String> responseMap = parseJSONResponse(retString);
+    assertNotNull("PublicKey is null", responseMap.get(TokenResource.PUBLIC_KEY));
+    assertNotNull("Algorithm is null", responseMap.get(TokenResource.ALGORITHM));
+    assertEquals("Mismatched algorithm", responseMap.get(TokenResource.ALGORITHM),
```
pzampino: Nitpick: Generally, the expected value is first in assertEquals(expected, actual)

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -278,6 +284,39 @@ public Response revoke(String token) {
...
+      byte[] pubKeyBytes = publicKey.getEncoded();
+      String base64Key = Base64.encodeBase64String(pubKeyBytes);
+      Map<String, String> map = new HashMap<>();
+      map.put(PUBLIC_KEY, base64Key);
+      map.put(ALGORITHM, publicKey.getAlgorithm());
+
+      String jsonResponse = JsonUtils.renderAsJsonString(map);
+      return Response.ok().entity(jsonResponse).build();
+    } catch (KeyStoreException | KeystoreServiceException e) {
+      return Response.serverError().build();
```
pzampino: We should at least log the exception message here in the gateway log.
I'm always conflicted about how much detail we should return to the client in such cases.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
sidseth: Removed the check.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
sidseth: Good question. Let me evaluate this a little more, and get back.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
sidseth: Removed. Left over from testing.

sidseth: Removed. Left over from testing.

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
sidseth: Fixed

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
sidseth: Won't count that as a nitpick. Expected should go first per the API. Fixed.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
sidseth: Will add a log line.
Yup - what level of information goes back to the client is always problematic. In this particular case, I doubt the client learns much from the error going back (There's no parameters etc for the client to get wrong which they could try fixing), and this requires a server side fix.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
sidseth: @risdenk - could you please elaborate a bit. Are you also suggesting including this information in the response (i.e. will the client need this to interpret the response correctly - that is why I had initially included the algorithm. I'd imagine the format is actually more important).

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
pzampino: I agree with that assessment of the response content.

risdenk:
@sidseth  what is the status of this change?

smolnar82:

PR #202, opened 2019-11-18
KNOX-2127 - ZooKeeperAliasService mishandles mixed-case alias keys pr…

pzampino:
…operly

ZookeeperRemoteAliasService is modified to treat alias keys consistently wrt case, lower-casing them when aliases are added (used to only lower-case when values were looked up).

## How was this patch tested?

Modified ZookeeperRemoteAliasServiceTest to include a mixed-case alias key.


pzampino:
> Will this break on upgrade of Knox?
> 
> ie: Knox won't be able to get mixed case aliases from ZK anymore if they are in use today?

Knox could not get mixed-case aliases from ZK before this change. The only real difference here is that the implementation won't add mixed-case aliases (which are immediately unreachable via the API) any more.

PR #205, opened 2019-11-22
KNOX-2134 - Add caching to ZookeeperRemoteAliasService

pzampino:

Added a cache to ZookeeperRemoteAliasService to improve the performance of alias lookups.

## How was this patch tested?

Added test to ZookeeperRemoteAliasServiceTest.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
@@ -39,6 +39,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * An {@link AliasService} implementation based on zookeeper remote service registry.
@@ -48,8 +49,7 @@ public class ZookeeperRemoteAliasService implements AliasService {
 
   public static final String PATH_KNOX = "/knox";
   public static final String PATH_KNOX_SECURITY = PATH_KNOX + "/security";
-  public static final String PATH_KNOX_ALIAS_STORE_TOPOLOGY =
-      PATH_KNOX_SECURITY + "/topology";
+  public static final String PATH_KNOX_ALIAS_STORE_TOPOLOGY = PATH_KNOX_SECURITY + "/topology";
   public static final String PATH_SEPARATOR = "/";
 
   private static final GatewayMessages LOG = MessagesFactory.get(GatewayMessages.class);
@@ -81,6 +81,8 @@ public class ZookeeperRemoteAliasService implements AliasService {
           return true;
         }
       };
+
+  private Map<String, Map<String, String>> aliasCache = new ConcurrentHashMap<>();
 
   private final AliasService localAliasService;
   private final MasterService ms;
@@ -169,19 +171,23 @@ public class ZookeeperRemoteAliasService implements AliasService {
   public List<String> getAliasesForCluster(final String clusterName) throws AliasServiceException {
     List<String> remoteAliases = new ArrayList<>();
 
-    /* If we have remote registry configured, query it */
-    if (remoteClient != null) {
-      remoteAliases = remoteClient
-          .listChildEntries(buildClusterEntryName(clusterName));
+    // First, check the alias cache
+    if (aliasCache.containsKey(clusterName)) {
+      remoteAliases.addAll(aliasCache.get(clusterName).keySet());
+    } else { // If it's not available locally, then check ZK
+      /* If we have remote registry configured, query it */
+      if (remoteClient != null) {
+        remoteAliases = remoteClient.listChildEntries(buildClusterEntryName(clusterName));
+      }
     }
 
     return remoteAliases;
   }
 
   @Override
-  public void addAliasForCluster(final String clusterName,
-      final String alias, final String value)
+  public void addAliasForCluster(final String clusterName, final String alias, final String value)
       throws AliasServiceException {
+
     if (remoteClient != null) {
       final String aliasEntryPath = buildAliasEntryName(clusterName, alias);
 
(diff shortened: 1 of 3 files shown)


gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasMonitorTest.java
```diff
@@ -234,13 +233,19 @@ public void testListener() throws Exception {
                 + ZookeeperRemoteAliasService.PATH_SEPARATOR + expectedAliasDev,
             zkAlias.encrypt(expectedPasswordDev).getBytes(StandardCharsets.UTF_8));
 
+    try {
+      Thread.sleep(1000);
```
risdenk: Why do we need to do this? Why 1 second? What are we sleeping for?

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasMonitorTest.java
```diff
@@ -234,13 +233,19 @@ public void testListener() throws Exception {
...
+    } catch (InterruptedException e) {
+      //
```
risdenk: Reset the interrupted thread status. Fixed a bunch of that here - https://issues.apache.org/jira/browse/KNOX-2130 couldn't figure out how to fail the build sadly :(

`Thread.currentThread().interrupt();`

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasServiceTest.java
```diff
@@ -193,61 +302,123 @@ public void testAliasForCluster() throws Exception {
...
+        assertEquals("ZNode value mismatch.", expectedAliases.get(cluster).get(alias), decrypted);
+      }
+    }
+  }
+
+  private void waitForCallback() {
+    try {
+      Thread.sleep(1000);
+    } catch (InterruptedException e) {
+      //
```
risdenk: Same as above - `Thread.currentThread().interrupt();`

Also again why 1 second?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
```diff
@@ -199,12 +205,17 @@ public void addAliasForCluster(final String clusterName,
             "Failed to store alias %s for cluster %s in remote registry", alias,
             clusterName));
       }
+    } else {
+      // Add it to the local cache only
+      Map<String, String> clusterAliases = aliasCache.computeIfAbsent(clusterName, m -> new ConcurrentHashMap<>());
```
risdenk: These 2 lines are used a few times. Might help to make it a "add to alias cache" method.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
```diff
@@ -438,6 +457,9 @@ public void childEvent(final RemoteConfigurationRegistryClient client,
           client.removeEntryListener(path);
           if (paths.length > 1) {
             localAliasService.removeAliasForCluster(paths[0], paths[1]);
+            if (aliasCache.containsKey(paths[0])) {
+              aliasCache.get(paths[0]).remove(paths[1]);
+            }
```
risdenk: This could be another helper method for alias cache. Remove from aliasCache. Not sure if there is a possible race condition here between the check and remove.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
```diff
@@ -228,35 +239,46 @@ public void removeAliasForCluster(final String clusterName, final String alias)
...
+            generateAliasForCluster(clusterName, alias);
+            password = getPasswordFromAliasForCluster(clusterName, alias);
+          }
+
+        } else {
+          try {
+            password = decrypt(encrypted).toCharArray();
+          } catch (final Exception e) {
+            throw new AliasServiceException(e);
+          }
```
risdenk: I need to look at this in more detail - probably in an IDE. Just noting to myself to look closer.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
```diff
@@ -169,19 +171,23 @@ private static void checkPathsExist(
...
-      remoteAliases = remoteClient
-          .listChildEntries(buildClusterEntryName(clusterName));
+    // First, check the alias cache
+    if (aliasCache.containsKey(clusterName)) {
+      remoteAliases.addAll(aliasCache.get(clusterName).keySet());
+    } else { // If it's not available locally, then check ZK
+      /* If we have remote registry configured, query it */
+      if (remoteClient != null) {
+        remoteAliases = remoteClient.listChildEntries(buildClusterEntryName(clusterName));
+      }
```
risdenk: I don't think I follow this logic. In my head, the cache should be up to date. We should never just straight return remove data without updating the cache.

risdenk: If the cache isn't up to date for whatever reason, we should update the cache and return that data. Now have to be careful about a race condition here.

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasMonitorTest.java
pzampino: It takes some time for the znode changes to propagate to the listener. Back when I had done the origin remote registry work, I had used much shorter delay intervals in the tests, but then the tests would intermittently fail. The delay in those tests was bumped to 1s, so I've done the same here, to avoid the same intermittent failures.

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasMonitorTest.java
pzampino: Well, this is test-only code, and no one of importance is checking the interrupted status further up the stack in this case. So, I'm not convinced there is much value in setting the interrupted status here.

pzampino:
In light of the related discussion, closing this PR to work on the more complete solution.

PR #211, opened 2019-12-04
KNOX-2144 - Alias API KnoxShell support should provide response types…

pzampino:

Improve response handling of KnoxShell classes for interacting with the Alias Service API, such that every client thereof need not parse JSON.

## How was this patch tested?

Unit tests, and manual tests. Attached ExampleAliases.groovy to the JIRA issue.

### DIFF
--- a/gateway-shell/src/test/java/org/apache/knox/gateway/shell/alias/AbstractResponseTest.java
+++ b/gateway-shell/src/test/java/org/apache/knox/gateway/shell/alias/AbstractResponseTest.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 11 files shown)


gateway-shell/src/test/java/org/apache/knox/gateway/shell/alias/AbstractResponseTest.java
```diff
@@ -0,0 +1,158 @@
...
+    } catch (Exception e) {
+      fail(e.getMessage());
+    }
+  }
+
+  @Test
+  public void testInvalidResponseContentType() {
+    StringEntity entity = null;
+    try {
+      entity = new StringEntity("{ \"topology\": \"testInvalidJSONCluster\", \"aliases\" : [ ");
```
moresandeep: The json object is invalid.

pzampino: It does not matter since the response content type is invalid, the content never gets processed.

moresandeep: ah I see.

PR #213, opened 2019-12-12
KNOX-2136 - Caching credentials in DefaultKeystoreService when an alias is being added or loaded from keystore and using a different cache implementation

smolnar82:

The following changes we made in `DefaultKeystoreService`:

- make sure we cache a credential of a given alias right after we saved it in the keystore (until my change it was cached after the first time the credential was fetched from the keystore)
- changing cache implementation from a simple `ConcurrentHashMap` to [Caffeine Cache](https://github.com/ben-manes/caffeine/wiki). This change allows us to configure appropriate eviction policy (as of now, policy attributes are hard-coded)

## How was this patch tested?

Added a new JUnit test and executed a full build:
```
...
[INFO] BUILD SUCCESS
[INFO] Total time: 18:41 min (Wall Clock)
[INFO] Finished at: 2019-12-12T12:51:54+01:00
[INFO] Final Memory: 414M/2402M
```

In addition to unit testing, I've executed a performance test to check how much do we gain when adding/fetching 200/500 credentials (repeated 5 times) after my changes were in place:

| Num of credentials | Before Change (ms) | After Change (ms) |
|--|--|--|
| 200 | 44686<br>51197<br>50423<br>50734<br>50207 | 23011<br>28117<br>27075<br>26996<br>27070 |
|500| 139908<br>188967<br>184465<br>175496<br>182561 | 77049<br>105465<br>105407<br>105488<br>105559 |

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
@@ -19,6 +19,8 @@ package org.apache.knox.gateway.services.security.impl;
 
 import static org.apache.knox.gateway.services.security.AliasService.NO_CLUSTER_NAME;
 
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.knox.gateway.GatewayMessages;
 import org.apache.knox.gateway.GatewayResources;
 import org.apache.knox.gateway.config.GatewayConfig;
@@ -30,6 +32,9 @@ import org.apache.knox.gateway.services.security.KeystoreService;
 import org.apache.knox.gateway.services.security.KeystoreServiceException;
 import org.apache.knox.gateway.services.security.MasterService;
 import org.apache.knox.gateway.util.X509CertificateUtil;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -54,10 +59,9 @@ import java.security.cert.Certificate;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.text.MessageFormat;
-import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
 
 import javax.crypto.spec.SecretKeySpec;
 
@@ -71,8 +75,9 @@ public class DefaultKeystoreService implements KeystoreService, Service {
   private static GatewayMessages LOG = MessagesFactory.get(GatewayMessages.class);
   private static GatewayResources RES = ResourcesFactory.get(GatewayResources.class);
 
+  //let's configure the cache with hard-coded attributes now; we can introduce new gateway configuration later on if needed
+  private final Cache<CacheKey, String> cache = Caffeine.newBuilder().expireAfterAccess(10, TimeUnit.MINUTES).maximumSize(1000).build();
   private GatewayConfig config;
-  private Map<String, Map<String, String>> cache = new ConcurrentHashMap<>();
 
   private MasterService masterService;
   private Path keyStoreDirPath;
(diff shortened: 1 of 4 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -333,6 +336,7 @@ public void removeCredentialForCluster(String clusterName, String alias) throws
 
         final Path keyStoreFilePath = keyStoreDirPath.resolve(clusterName + CREDENTIALS_SUFFIX);
         writeKeyStoreToFile(ks, keyStoreFilePath, masterService.getMasterSecret());
+        removeFromCache(clusterName, alias);
```
risdenk: We want this to happen no matter what right? Maybe in the finally block or leave it at the beginning?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -290,6 +294,7 @@ public void addCredentialForCluster(String clusterName, String alias, String val
 
         final Path keyStoreFilePath = keyStoreDirPath.resolve(clusterName + CREDENTIALS_SUFFIX);
         writeKeyStoreToFile(ks, keyStoreFilePath, masterService.getMasterSecret());
+        addToCache(clusterName, alias, value);
```
risdenk: We want this to happen no matter what right? Maybe in the finally block or leave it at the beginning?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
smolnar82: Why? Should not we only remove it from cache only when it was removed from the keystore successfully?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
smolnar82: I see this the same way as in the `remove` method: I would not save it in the cache unless it was saved in the keystore successfully.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -71,8 +74,9 @@
   private static GatewayMessages LOG = MessagesFactory.get(GatewayMessages.class);
   private static GatewayResources RES = ResourcesFactory.get(GatewayResources.class);
 
+  //let's configure the cache with hard-coded attributes now; we can introduce new gateway configuration later on if needed
+  private final Cache<String, Map<String, String>> cache = Caffeine.newBuilder().expireAfterWrite(10, TimeUnit.MINUTES).maximumSize(1000).build();
```
pzampino: Is it possible to have entries evicted after some duration has passed without a read?

smolnar82: I believe it is possible. That rings the bell that we may want to keep the line which adds back an entry into the cache when the credential was loaded from the keystore.
Is this you'd like to point out?

smolnar82: So the following should be implemented:

1. add an entry: into keystore and cache
2. 10 minutes passed w/o reading that entry -> got evicted from cache
3. fetch it -> we load it from the keystore but not adding into the cache

The least should be changed to add it back to the cache, right?

pzampino: At least that much. I'm thinking that it will be more efficient to keep the most-recently used entries in the cache, and evict those which have not been referenced for some time. So, if we could evict those entries which have not been read after n minutes, I think that would be preferable.

smolnar82: This could be done easily (see `expireAfterAccess` [here](https://github.com/ben-manes/caffeine/wiki/Eviction)). Let me submit a new patch soon.

smolnar82: Done.

PR #215, opened 2019-12-12
KNOX-1962 - CM discovery - Avoid reading krb5 login config multiple t…

pzampino:
…imes

Load JAAS config file for Kerberos login once, rather than every time the Subject is needed.

## How was this patch tested?
Manual testing, and added AuthUtilsTest to validate the one-time loading of the configuration.

### DIFF
--- a/gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/auth/AuthUtilsTest.java
+++ b/gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/auth/AuthUtilsTest.java
@@ -1,0 +1,90 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.knox.gateway.topology.discovery.cm.auth;
+
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.junit.Test;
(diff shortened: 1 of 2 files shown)


gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/auth/AuthUtilsTest.java
```diff
@@ -0,0 +1,90 @@
...
+    File loginConfigFile = createTestKrb5LoginConfigFile();
+    Configuration conf1 = null;
+    Configuration conf2 = null;
+    try {
+      System.setProperty(GatewayConfig.KRB5_LOGIN_CONFIG, loginConfigFile.getAbsolutePath());
+      conf1 = AuthUtils.getKerberosJAASConfiguration();
+      conf2 = AuthUtils.getKerberosJAASConfiguration();
+    } catch (Throwable e) {
+      e.printStackTrace();
+      fail(e.getMessage());
```
risdenk: should be able to just `fail(e)` which prints the stacktrace

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/auth/AuthUtilsTest.java
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * KNOX-1962
+   */
+  @Test
+  public void testLoadJAASConfigOnce() {
+    File loginConfigFile = createTestKrb5LoginConfigFile();
+    Configuration conf1 = null;
+    Configuration conf2 = null;
+    try {
+      System.setProperty(GatewayConfig.KRB5_LOGIN_CONFIG, loginConfigFile.getAbsolutePath());
```
risdenk: Do we have to use system properties here? Mostly just asking since multiple tests/modules at once will conflict I think.

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/auth/AuthUtilsTest.java
pzampino: Yes, that got left in when I was debugging my initial bad config file contents. It should go away.

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/auth/AuthUtilsTest.java
pzampino: The class under test uses the system property, so the test must set it. Are you suggesting that the class under test should not be using the system property?

PR #216, opened 2019-12-12
KNOX-2149 - JWTTokenProvider - JWT verification with OIDC provider by invoking JWKS verification url

nxverma:
**What changes were proposed in this pull request?**

change the code to pass the JWKS verification url and if a key is changed - no knox config change is required. Change done to support using JWKS verification url to validate the token :
Class JWTFederationFilter was modified to get an additional parameter (JWKS verification url) and code to use this url to get the public key and then use this to validate the token. This approach will make it easy to manage for key rotation.
Library used is - https://connect2id.com/products/nimbus-jose-jwt/examples/validating-jwt-access-tokens
As part of this patch , we also extracting custom claim from JWT instead of default subject

**How was this patch tested?**

Patch was tested fully manual test cases.
We deployed modified code version of Knox (That build on top of Knox 1.3.0)
Test 1: Able to Invoke Livy endpoint to submit spark job via Knox JWT-Okta authentication
Test Request:
```
curl -kvvvvvv --request POST --url https://localhost:8443/gateway/tokenbased/livy/v1/batches --header 'content-type: application/json' --header 'accept: application/json' --header 'authorization: Bearer eyJra...dfg' --data '
{"file": "s3://aws-bigdata-blog/artifacts/aws-blog-emr-knox/spark-examples.jar", "className": "org.apache.spark.examples.SparkPi", "args": ["100"]}
```

Test 2: Able to Invoke Knox admin rest endpoint to create knox topology and alias with JWT-Okta verification
Test Request:
```
curl -ivk -H "Authorization: Bearer eyJra...dffd" https://localhost:8443/gateway/tokenbased/api/v1/topologies/jwtadmin -X PUT -H 'Content-type: application/xml;charset=UTF-8' --data @payload-knox.xml
```

Sample Knox Topology : 
```
<?xml version="1.0" encoding="UTF-8"?>
<topology>
   <uri>https://localhost:443/gateway/tokenbased</uri>
   <name>tokenbased</name>
   <generated>false</generated>
   <gateway>
      <provider>
         <role>federation</role>
         <name>JWTProvider</name>
         <enabled>true</enabled>
         <param>
            <name>knox.token.jwks.url</name>
            <value>https://test.okta.com/oauth2/aussdfgsg/v1/keys</value>
         </param>
         <param>
            <name>knox.token.principal.claim</name>
            <value>empId</value>
         </param>
         <param>
            <name>jwt.expected.issuer</name>
            <value>https://test.okta.com/oauth2/aussdfgsg</value>
         </param>
         <param>
            <name>jwt.expected.sigalg</name>
            <value>RS256</value>
         </param>
      </provider>
      <provider>
         <role>hostmap</role>
         <name>static</name>
         <enabled>false</enabled>
         <param>
            <name>localhost</name>
            <value>localhost</value>
         </param>
      </provider>
      <provider>
         <name>Default</name>
         <enabled>true</enabled>
      </provider>
   </gateway>
   <service>
      <role>LIVYSERVER</role>
      <url>https://localhost:8998</url>
   </service>
   <service>
      <role>HIVE</role>
      <url>http://localhost:10001/cliservice</url>
   </service>
  <service>
      <role>KNOX</role>
   </service>
</topology>
```


gateway-util-common/pom.xml
```diff
@@ -1,96 +1,123 @@
...
+	OF ANY KIND, either express or implied. See the License for the specific 
+	language governing permissions and limitations under the License. -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.apache.knox</groupId>
+		<artifactId>gateway</artifactId>
+		<version>1.3.0</version>
```
risdenk: This needs to be fixed

gateway-util-common/pom.xml
```diff
@@ -1,96 +1,123 @@
...
-        <dependency>
-            <groupId>org.apache.directory.server</groupId>
-            <artifactId>apacheds-jdbm</artifactId>
-        </dependency>
+		<dependency>
+			<groupId>org.apache.knox</groupId>
+			<artifactId>gateway-test-utils</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<!-- <dependency> <groupId>com.auth0</groupId> <artifactId>jwks-rsa</artifactId>
```
risdenk: Is there a reason the existing `nimbus-jose-jwt` dependency couldn't be used? I looked through the nimbus does and looks like it does JWT verification. I don't think we need to pull in from okta if we can avoid it. Especially not in the gateway-util-common.

gateway-util-common/pom.xml
```diff
@@ -1,96 +1,123 @@
...
+			<artifactId>gateway-test-utils</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<!-- <dependency> <groupId>com.auth0</groupId> <artifactId>jwks-rsa</artifactId> 
+			<version>0.9.0</version> </dependency> <dependency> <groupId>com.auth0</groupId> 
+			<artifactId>java-jwt</artifactId> <version>3.8.3</version> </dependency> -->
+		<dependency>
+			<groupId>com.okta.jwt</groupId>
+			<artifactId>okta-jwt-verifier</artifactId>
+			<version>0.4.0</version>
```
risdenk: All dependencies need to be declared at the top level pom.xml with versions as properties.

gateway-util-common/pom.xml
risdenk: From a quick look all the changes should be in `gateway-provider-security-jwt` module and not `gateway-util-common`

gateway-util-common/pom.xml
nxverma: this is fixed

gateway-util-common/pom.xml
nxverma: WIP for nimbus-jose-jwt .. looking if nimbus-jose-jwt can  supports JWKS url  (to get keys dynamically )or not . 

Remove all changes  from gateway-util-common

gateway-util-common/pom.xml
nxverma: this is fixed

gateway-util-common/pom.xml
nxverma: nimbus-jose-jwt doesn't support JWKS Url based key fetch . We have to write custom code to fetch keys with httpclient url and do caching  of keys to avoid performance and also need to do invalidating keys in cache .   All these features automatically comes with Okta  JWT . Please suggest .

risdenk: https://connect2id.com/products/nimbus-jose-jwt/examples/jwk-retrieval

^^^ potentially?

nxverma: Thanks @risdenk This API just fetch  list of Keys , still i see we have to do remaining work like converting each key to RSA publickey and do JWT verify  in  loop . I don't have direct nimbus jose api .. still Digging on it

nxverma: looks like https://connect2id.com/products/nimbus-jose-jwt/examples/validating-jwt-access-tokens promising ... i will update as soon as able to implement it . Thanks @risdenk

nxverma: removed okta specific apis . used nimbus-jose-jwt api

smolnar82:
I agree with @risdenk: way too much of formatting changes were added with this change.
@nxverma, Could you please first format your newly added and/or modified code so that it'd be easier to review?

gateway-util-common/pom.xml
nxverma: @risdenk @smolnar82  is there any code formatter template specific for Knox  that i can import in my eclipse

risdenk: Not currently no. Sorry. The formatting of the code in Knox is definitely all over the map. It would be nice to get a handle on that, but needs to be done separately from a functional change.

nxverma: all fixes are done please review  @risdenk

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -187,71 +190,75 @@ protected boolean validateAudiences(JWT jwtToken) {
...
   protected Subject createSubjectFromToken(JWT token) {
-    final String principal = token.getSubject();
+    String principal = token.getSubject();
+    String claimvalue = null;
+    if (expectedPrincipalClaim != null) {
+      claimvalue = token.getClaim(expectedPrincipalClaim);
+    }
 
+    if (claimvalue != null) {
+      principal = claimvalue.toLowerCase(Locale.ENGLISH);
```
risdenk: Is `Locale.ENGLISH` correct here? `Locale.ROOT`?

risdenk: I think this is the only substantial change in this file? Everything else is formatting?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -75,41 +102,92 @@ public void destroy() {
   public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
       throws IOException, ServletException {
     String header = ((HttpServletRequest) request).getHeader("Authorization");
+    String header_hive = ((HttpServletRequest) request).getHeader("HiveAuthToken");
```
risdenk: Seems weird to have a special case for Hive here.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -75,41 +102,92 @@ public void destroy() {
...
+      wireToken = header_hive;
+    } else {
       // check for query param
       wireToken = request.getParameter(paramName);
     }
 
-    if (wireToken != null) {
-      try {
-        JWT token = new JWTToken(wireToken);
-        if (validateToken((HttpServletRequest)request, (HttpServletResponse)response, chain, token)) {
```
risdenk: why a new `validateJWTToken` method instead of making use of the existing `validateToken` method?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -75,41 +102,92 @@ public void destroy() {
...
+  }
+
+  /**
+   * @param wireToken
+   * @param request
+   * @param response
+   * @param chain
+   * @throws IOException
+   */
+  private void validateJWTtoken(String wireToken, ServletRequest request, ServletResponse response, FilterChain chain)
```
risdenk: This whole method should not be here and should be combined with `validateToken` somehow instead of delegating to it.

gateway-util-common/pom.xml
```diff
@@ -6,9 +6,7 @@
   The ASF licenses this file to You under the Apache License, Version 2.0
```
risdenk: This file is just formatting changes. Please remove the changes.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
nxverma: fixed

nxverma: Moved all JWT Token validation code in ValidToken method

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
nxverma: Moved all JWT Token validation code in ValidToken method

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
nxverma: Moved all JWT Token validation code in ValidToken method

gateway-util-common/pom.xml
nxverma: removed changes

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
nxverma: WIP for it .. As beeline doesn't support space  like http.header.Authorization=Bearer eyJraWQiOi..     Thats why we used Hive specific header for  JWT ..

nxverma:
@risdenk @smolnar82 please review ... done all refactor of code

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -223,38 +239,60 @@ else if (t instanceof ServletException) {
...
+        JWSKeySelector<SecurityContext> keySelector = new JWSVerificationKeySelector<>(expectedJWSAlg, keySource);
+        // Create a JWT processor for the access tokens
+        ConfigurableJWTProcessor<SecurityContext> jwtProcessor = new DefaultJWTProcessor<>();
+        jwtProcessor.setJWSKeySelector(keySelector);
+        JWTClaimsSetVerifier<SecurityContext> claimsVerifier = new DefaultJWTClaimsVerifier<>();
+        jwtProcessor.setJWTClaimsSetVerifier(claimsVerifier);
+        // Process the token
+        SecurityContext ctx = null; // optional context parameter, not required here
+        jwtProcessor.process(token.toString(), ctx);
+        verified = true;
```
risdenk: So I think there is a misunderstanding here between my last comment and current set of code.

I would have expected something like the following:

`verified = authority.verifyToken(token, url);`

Where the specifics are delegated down to the `JWTokenAuthority` instead of trying to have the `AbstractJWTFilter` handle all of this. I would expect very little to be verified directly here.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWKSFederationFilter.java
```diff
@@ -0,0 +1,110 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.knox.gateway.services.security.token.impl.JWT;
+import org.apache.knox.gateway.services.security.token.impl.JWTToken;
+
+public class JWKSFederationFilter extends AbstractJWTFilter {
```
risdenk: Maybe I am missing something but why do we need a new filter? The existing filter can just have the new parameters potentially?

nxverma: @risdenk  I don't want to modify existing filter as it may cause confusion for JWTProvider parameters , if someone put both JWKS url  and Publiccert in topology , it is tough to explain . this provider clearly dedicated to JWKS url.. no public cert . its my point of view but if community suggest to modify  same filter i will update same. Please let me know

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
nxverma: I can do this change

nxverma: verifyToken(JWT token, RSAPublicKey publicKey) and authority.verifyToken(token, url); method will conflict as we saw verifyToken(token, null); called  in DefaultTokenAuthorityService

nxverma: it is fixed

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWKSFederationFilter.java
risdenk: So right now that logic of what is chosen is in AbstractJwtFilter. Its pretty clear that if a public key is provided it will be used over URL. Having both is definitely in valid. Not sure why we would need a separate filter. The URL is a separate alternative option to the public key.

nxverma: removed  new filter , modified existing JWTfilter . Need to add another http header  for Hive JDBC where we can't pass token with Bearer  keyword.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -80,6 +92,10 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
       // what follows the bearer designator should be the JWT token being used to request or as an access token
       wireToken = header.substring(BEARER.length());
     }
+    else if((((HttpServletRequest) request).getHeader(HEADER_TOKEN_NAME))!= null) {
```
risdenk: Sorry - This is going to be super nitpicky - can this be a separate Jira/PR for this change for Hive? I want to call it out separately since its not really the main part of this change. I also wonder why a header with a space can't be passed. That seems like a weird Hive JDBC restriction that I haven't seen before.

nxverma: okay , Let me remove as part of this Pull request . for Hive I  will work  different PR

nxverma:
@risdenk  raised bug for Hive JWT https://issues.apache.org/jira/browse/KNOX-2150

nxverma:
@pzampino @lmccay Please review

risdenk:
Pulling this locally to cleanup a few formatting things before merging.

PR #218, opened 2019-12-13
KNOX-2134 - Checking if password is available via local alias service before going to fetch it from remote ZK server

smolnar82:

Formatted the code and did a minor cleanup to make it easier to read and understand the code.
Apart from code cleanup, I changed the way how passwords are fetched in `ZookeeperRemoteAliasService`: before going to the remote ZK server we first check if the desired password is available through the local alias service. If so; we skip the remote check. This came with a separate commit to make it easier for review (see [here](https://github.com/apache/knox/pull/218/commits/2703684c0d4900340ac18741a731869c4199b451)).

## How was this patch tested?

Added a JUnit test to measure performance (ignoring it from automatic execution, but can be executed manually by commenting out the `Ignore` annotation): added and fetched 200 aliases/passwords with and without my changes in place with the following results

|Without cache (ms) |  With Cache (ms) |
|--|--|
| 30841<br>31335<br>32762<br>31646<br>31736<br>31641<br>31204<br>34219<br>33720<br>33297 |  16715<br>16265<br>17469<br>17312<br>17482<br>20990<br>17343<br>19637<br>19240<br>17705 |

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
@@ -44,461 +44,403 @@ import java.util.Map;
  * An {@link AliasService} implementation based on zookeeper remote service registry.
  */
 public class ZookeeperRemoteAliasService implements AliasService {
-  public static final String TYPE = "zookeeper";
+    public static final String TYPE = "zookeeper";
+    public static final String PATH_KNOX = "/knox";
+    public static final String PATH_KNOX_SECURITY = PATH_KNOX + "/security";
+    public static final String PATH_KNOX_ALIAS_STORE_TOPOLOGY = PATH_KNOX_SECURITY + "/topology";
+    public static final String PATH_SEPARATOR = "/";
 
-  public static final String PATH_KNOX = "/knox";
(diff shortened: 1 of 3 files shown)


smolnar82:
@pzampino Please review my changes when you have time. Thanks!

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
```diff
@@ -44,461 +44,403 @@
...
+    /**
+     * Get a list of all aliases for a given cluster. Remote aliases are preferred over local.
+     *
+     * @param clusterName
+     *            cluster name
+     * @return List of all the aliases
+     */
+    @Override
+    public List<String> getAliasesForCluster(final String clusterName) throws AliasServiceException {
+        return remoteClient == null ? new ArrayList<>() : remoteClient.listChildEntries(buildClusterEntryName(clusterName));
```
pzampino: So, this method doesn't check the local alias service at all?

smolnar82: This is correct; I was only focusing on retrieving passwords as quickly as possible. Let me fix this and submit a new patch soon.

smolnar82: Fixed.

PR #221, opened 2019-12-16
KNOX-1921 - CM discovery - Hue Load balancer HTTP/HTTPS scheme

pzampino:
Added support for SSL-enabled Hue load balancer URLs.

## How was this patch tested?
Ran existing CM discovery tests, added HueURLCollectorTest#testLBSSL(), and manually tested CM service discovery against a cluster.


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/collector/HueURLCollector.java
```diff
@@ -28,12 +28,27 @@
  */
 public class HueURLCollector extends AbstractURLCollector {
 
+  private static final String SCHEME_HTTP  = "http";
+  private static final String SCHEME_HTTPS = "https";
```
risdenk: Might make sense to put on abstract url collector since will be used a bunch?

PR #226, opened 2019-12-19
KNOX-2154 - Allow KNOX service during topology generation without URLs and parameters

smolnar82:

The `KNOX` service usually does not come with any `url` or `parameter` so that we should allow this service to be added in the generated topology without these attributes.

## How was this patch tested?

Updated and ran JUnit tests.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
@@ -47,6 +47,7 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.TreeSet;
 
 /**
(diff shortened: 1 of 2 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
```diff
@@ -84,6 +85,8 @@
 
     private static Map<String, ServiceDiscovery> discoveryInstances = new HashMap<>();
 
+    private static final Set<String> ALLOWED_SERVICES_WITHOUT_URLS_AND_PARAMS = Collections.singleton("KNOX");
```
risdenk: I'm not sure I like this hardcoded, but overall change looks good.

risdenk: I don't have any better ideas right now. Could technically be a config, but that seems overkill to so eh.

PR #227, opened 2019-12-20
KNOX-2155 - KnoxSSO should handle multiple cookies with the same name

risdenk:
This commit moves getting cookies by name to a
new utility class. It forces callers to look
through multiple cookies returned and handle
that case.

### DIFF
--- a/gateway-adapter/src/main/java/org/apache/hadoop/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
+++ b/gateway-adapter/src/main/java/org/apache/hadoop/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
@@ -39,18 +39,6 @@ public class SSOCookieFederationFilter
   }
 
   /**
-   * Encapsulate the acquisition of the JWT token from HTTP cookies within the
-   * request.
-   *
-   * @param req servlet request to get the JWT token from
-   * @return serialized JWT token
-   */
-  @Override
-  protected String getJWTFromCookie(HttpServletRequest req) {
-    return super.getJWTFromCookie(req);
-  }
-
-  /**
    * Create the URL to be used for authentication of the user in the absence of
    * a JWT token within the incoming request.
    *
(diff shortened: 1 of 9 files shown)


gateway-adapter/src/main/java/org/apache/hadoop/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
```diff
@@ -38,18 +38,6 @@ protected void handleValidationError(HttpServletRequest request,
...
-   * request.
-   *
-   * @param req servlet request to get the JWT token from
-   * @return serialized JWT token
-   */
-  @Override
-  protected String getJWTFromCookie(HttpServletRequest req) {
-    return super.getJWTFromCookie(req);
-  }
-
```
risdenk: Will this break backwards compat if someone upgrades? I don't quite understand how these adapter classes work.

risdenk: I'm assuming no since there is no reason this method gets called anymore since its deleted everywhere?

gateway-util-common/src/main/java/org/apache/knox/gateway/util/CookieUtils.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class CookieUtils {
+  private static final GatewayUtilCommonMessages LOGGER = MessagesFactory.get(GatewayUtilCommonMessages.class);
+
+  private CookieUtils() {
+  }
+
+  public static List<Cookie> getCookiesForName(HttpServletRequest request, String name) {
```
risdenk: I need to add tests for this.

risdenk: done. Added tests.

gateway-adapter/src/main/java/org/apache/hadoop/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
moresandeep: It should not break backwards compatibility, looks good.

risdenk:
Merged in 1b019617d408312ac33bf8f07f0ec862f7b33382

PR #228, opened 2019-12-20
KNOX-2156 - CM discovery - KUDUUI discovery

pzampino:

Added support for discovering URLs for the Apache Kudu UI.

## How was this patch tested?

Added tests to ClouderaManagerServiceDiscoveryTest, and manually tested.

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/kudu/KuduUIServiceModelGenerator.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/kudu/KuduUIServiceModelGenerator.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/kudu/KuduUIServiceModelGenerator.java
```diff
@@ -0,0 +1,66 @@
...
+
+  @Override
+  public ServiceModel generateService(ApiService service, ApiServiceConfig serviceConfig, ApiRole role, ApiConfigList roleConfig) throws ApiException {
+    String hostname = role.getHostRef().getHostname();
+
+    boolean sslEnabled = Boolean.parseBoolean(getRoleConfigValue(roleConfig, "ssl_enabled"));
+    String scheme = sslEnabled ? "https" : "http";
+
+    String port = getRoleConfigValue(roleConfig, "webserver_port");
+    return createServiceModel(String.format(Locale.getDefault(), "%s://%s:%s/", scheme, hostname, port));
```
risdenk: I'm not sure if this is in others and I missed it. I think `Locale.ROOT` is probably better here. I don't think we need to use localization for any of the parameters here. We should just plop in as is.

Note 99% sure we won't easily run into this since there are only a few locales where there is a differences.

pzampino: Local.getDefault() is used in the others. We can revisit that for all of them.

PR #230, opened 2020-01-02
KNOX-2157 - Verifying the server's state in addition to PID check at gateway start and registering shutdown hook in order to stop the server gracefully.

smolnar82:

Apart from the already existing PID check, we are going to verify if the Jetty server is in `STARTED` state. We can achieve it by implementing Jetty's `LifeCylce.Listener` interface and write out the status (STARTING, STARTED, FAILURE, STOPPING, STOPPED) into `$DATA_DIR/gatewayServer.status` file. The startup script will return 0 when this file contains `STARTED`; 1 otherwise (the file is overwritten in every status change).

Additionally, two new command-line options are introduced to the `start` command:
* `--test-gateway-retry-attempts`: indicates the number of tries the startup script should execute before it fails. Defaults to 5.
* `--test-gateway-retry-sleep`: the amount of time that the test process will wait or sleep before a retry is issued. Defaults to 2s.

Additionally, in order to make the server stopped gracefully from `gateway.sh`, I registered a shutdown hook in `GatewayServer` which invokes the desired stop commands on services/monitor/Jetty and persists the appropriate states (`STOPPING` and `STOPPED`) in the new status file.

## How was this patch tested?

Executed a full build:
```
...
[INFO] BUILD SUCCESS
[INFO] Total time: 18:24 min (Wall Clock)
[INFO] Finished at: 2020-01-02T18:50:47+01:00
[INFO] Final Memory: 410M/1798M
```

Tested manually:
* confirmed that the new status file became created/removed when started/stopped the server. All the statuses were written out as expected. For instance
```
2020-01-07T10:55:10.684+0100
STARTING

2020-01-07T10:55:11.570+0100
STARTED

2020-01-07T10:55:46.599+0100
STOPPING

2020-01-07T10:55:46.620+0100
STOPPED
```
* tried the `start` command with different options (`--printEnv`, `--test-gateway-retry-attempts`, `--test-gateway-retry-sleep`)
* made sure (using debug messages in the shell scripts) that the status check took place (in my environment it needed 3 re-tries)

### DIFF
--- a/gateway-release-common/home/bin/knox-functions.sh
+++ b/gateway-release-common/home/bin/knox-functions.sh
@@ -41,6 +41,9 @@ APP_JAVA_LIB_PATH=${KNOX_GATEWAY_JAVA_LIB_PATH:-$DEFAULT_JAVA_LIB_PATH}
 # JAVA options used by the JVM
 declare -a APP_JAVA_OPTS
 
+#status-based test related variables
+DEFAULT_APP_STATUS_TEST_RETRY_ATTEMPTS=5
+DEFAULT_APP_STATUS_TEST_RETRY_SLEEP=2s
 
 ############################
 ##### common functions #####
@@ -163,7 +166,7 @@ function buildAppJavaOpts {
     # echo "APP_JAVA_OPTS =" "${APP_JAVA_OPTS[@]}"
 }
 
-function appIsRunning {
+function appIsRunningByPID {
    if [ "$1" -eq 0 ]; then return 0; fi
 
    ps -p "$1" > /dev/null
@@ -173,6 +176,52 @@ function appIsRunning {
    else
      return 1
    fi
+}
+
+function appIsRunningByStatus {
+   retryAttempts=${APP_STATUS_TEST_RETRY_ATTEMPTS:-$DEFAULT_APP_STATUS_TEST_RETRY_ATTEMPTS}
+   retrySleep=${APP_STATUS_TEST_RETRY_SLEEP:-$DEFAULT_APP_STATUS_TEST_RETRY_SLEEP}
+
+   #echo "Retry attempts = $retryAttempts"
+   #echo "Retry sleep = $retrySleep"
+
+   statusCheck=0
+   for ((i=1; i<=retryAttempts; i++))
+   do
+     #echo "$i. try"
+
+     if grep -Fxqs "STARTED" "$APP_DATA_DIR"/gatewayServer.status; then
+       statusCheck=1
+       break
+     fi
+
+     sleep "$retrySleep"
+   done
+
+   return $statusCheck
+}
+
+#returns 0 if not running and 1 if running
+function appIsRunning {
+   appIsRunningByPID "$1"
+   if [ $? -eq 1 ]; then
+     #echo "PID check succeeded"
+     if [[ "$TEST_APP_STATUS" = "true" ]]; then
+       #echo "Checking status..."
+       appIsRunningByStatus
+       if [ $? -eq 1 ]; then
+         #echo "Status check passed"
+         return 1;
+       else
+         #echo "Status check NOT passsed"
+         return 0;
+       fi
+     else
+       return 1;
+     fi
+   fi;
+
+   return 0
 }
 
 # Returns 0 if the app is running and sets the $PID variable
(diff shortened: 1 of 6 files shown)


risdenk:
So just to make sure I understand the problem, the issue is that the launcher runs the gateway process and doesn't wait for the initialization of the server to complete before returning and saying this is up and running.

Is it possible to maybe use the listener or another method in the GatewayServer to wait until initialization is complete? Is the launcher returning too soon? 

Ideally after `GatewayServer` starts the server would be completely ready. It is weird that we are saying that everything is good before that :/

I am not sure I am a fan of the writing a file and checking the output to avoid what looks to be a fundamental gateway starting issue. I wonder how we avoid this in tests currently. There must be an existing method to make sure that the topology is deployed or something along those lines or we would see issues like this in the tests. The solution in this PR wouldn't help the tests since it would be in the shell script only.

smolnar82:
The main difference between our tests and the way how we start it from the startup script is:
* within the tests we directly call `GatewayServer.startServer` (usually in methods annotated with `@BeforeClass`) and wait until it returns
* whereas in the script - where the application is configured to run in the background by default - creates a Java process with `nohup` which invokes `GatewayServer.main()` with the given argument and let the process live its own life in the background. Right after the Java call the script continues to check if the PID exists. Usually, it takes some time for the server to start up but the script indicates it's up&running because the PID is there.

It's not _easy_ to add hook in a shell script to check the 2nd case:
1. either the server persists its state somewhere which the script can read (this is what I implemented here)
2. or the script tries to ping the server somehow
2.1. The REST API would be ideal, but this would require at least one servlet which does not require authentication.
2.2. Or one might try to check if the port is open. For instance:
```
$ nc -z localhost 8443

$ echo $?
1

$ nc -z localhost 8443

$ echo $?
1

$ nc -z localhost 8443
Connection to localhost port 8443 [tcp/pcsync-https] succeeded!

$ echo $?
0
```

gateway-release-common/home/bin/knox-functions.sh
```diff
@@ -250,6 +302,11 @@ function appStop {
      exit 1
    else
      rm -f "$APP_PID_FILE"
+
+     #TODO: due to the current way of shutting down the server the status has never been changed to STOPPED (STARTING seems to be not reliable either which is weird).
```
risdenk: This TODO seems relevant if this isn't being written then this will cause issues. Were you able to figure out why lifecycle isn't writing this file?

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServerLifecyleListener.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private static final GatewayMessages log = MessagesFactory.get(GatewayMessages.class);
+
+  private static final ThreadLocal<DateFormat> DATE_FORMAT = ThreadLocal.withInitial(() -> new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ", Locale.getDefault()));
+
+  private enum Status {
+    STARTING, STARTED, FAILURE, STOPPING, STOPPED
+  };
+
+  private final Path lifeCylceFilePath;
```
risdenk: Should be `lifeCycleFilePath`

smolnar82: Will change it as soon as I can.

gateway-release-common/home/bin/knox-functions.sh
smolnar82: Currently, the JVM is "stopped" with a `kill` command but there is no [shutdown hook](https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#addShutdownHook-java.lang.Thread-) registered in `GatewayServer`. Adding a shutdown hook would help in most of the cases but it may still send a SIGKILL when stopping a server 'normally' did not succeed.

smolnar82: Added the above-mentioned shutdown hook and it's working like a charm.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServerLifecyleListener.java
smolnar82: Done.

PR #231, opened 2020-01-03
KNOX-2128 - Custom DataSource and SQL Commands for KnoxShell and KnoxShellTable

lmccay:

As described in KIP-14 , GroovyShell allows for the extension of the shell itself with custom commands. By providing commands for the management of DataSource configuration and use in SQL queries, we can simplify the interaction with SQL engines and databases for JDBC-based database access in KnoxShell to make for a powerful scripting and shell environment.

DataSource and SQL commands really go hand in hand here and will be done at the same time.

    * Datasources (:datasource|:ds) CRUD and select operations for a set of JDBC datasources that are persisted to disk in the user home directory
    * SQL (:SQL|:sql) SQL query execution with persisted SQL history per datasource

In order to add these custom commands and depend on third-party jars like JDBC drivers, I've had to add a lib directory to the gateway-shell home and add all jars within that directory to the shell classpath. This required moving away from the executable jar and changing the knoxshell.sh to include the main class on the command line rather than executing the jar.

There are also two things persisted as session state to the users' home directory in hidden directories.

    * knoxdatasources.json - allowing the configured datasources to be available across shell sessions
    * knoxsqlhistories.json - allowing sql queries per datasource to be available across shell sessions with 20 queries

A CSV custom command was also added in order to create KnoxShellTable's from CSV files.

    * example - :csv withHeaders https://url-to-csv-file

## How was this patch tested?
 Primarily tested new custom groovysh commands manually as writing tests is a bit challenging due to the lack of the groovysh environment within the tests. Will follow up with a patch for mocking that environment if possible.

In order to test there are a few required steps that will need to be documented for general use as well:

    * download JDBC drivers and place in the {gateway-shell-home}/lib directory
    * add datasources for JDBC endpoints using ":ds add ds-name connectStr driverClass authn_type" where authn_type is "none" or "basic"
    * selecting the desired datasource is required only when there are more than one defined and is accomplished via ":ds select ds-name"
    * submitting a SQL query is done via the :SQL command ":sql". You may also include an "assign table-name" pair in the command line to indicate what variable name to set the resulting table to within the environment. This will launch a swing-based SQL Dialog for submitting the Query. This is necessary due to how GroovySh Custom Commands parse params from the command line. Where clauses or anything else that would require quotes don't work from the command line alone as the quotes are stripped and each word ends up being a separate param.
    * After submitting the query, the datasource authn_type is interrogated and the user is challenged for username and password when it is set as "basic" with another swing-based dialog box due to similar environment issues.

NOTES FOR REVIEW:

* I'm printing stacktraces for exceptions in many instances since they will print in the knoxshell environment directly and provide dev level of troubleshooting that would be provided as a result of manually doing what the commands are simplifying for us
* Inside the execute methods of the commands sometimes an error message is returned as a String others a KnoxShellTable or other object is returned. These are interchangeable due to the Object return value and rendered by the knoxshell/groovysh environment naturally.
* I've added methods for persisting certain things to the KnoxSession class as these persisted items are intended to live across sessions and this seemed natural to do.


gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/JDBCKnoxShellTableBuilder.java
```diff
@@ -109,12 +107,15 @@ public KnoxShellTable sql(String sql) throws IOException, SQLException {
...
-      return DriverManager.getConnection(connectionUrl);
+  public Connection createConnection() throws SQLException {
+    Connection con = null;
+    if (username != null && pass != null) {
+      con = DriverManager.getConnection(connectionUrl, username, pass);
+    }
+    else {
+      con = DriverManager.getConnection(connectionUrl);
     }
+    return con;
```
risdenk: I think the older version checked empty string and this change removes that check on the username?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/JDBCKnoxShellTableBuilder.java
```diff
@@ -42,16 +40,16 @@ public JDBCKnoxShellTableBuilder username(String username) {
...
     this.pass = pass;
     return this;
   }
 
-  public String password() {
+  public String username() {
+    return username;
+  }
+
+  public String pwd() {
```
risdenk: `password()` was changed to `pwd`? Any reason?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/commands/SelectCommand.java
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  @Override
+  public void keyReleased(KeyEvent event) {
+    // TODO Auto-generated method stub
+  }
+
+  @Override
+  public void keyTyped(KeyEvent event) {
+    // TODO Auto-generated method stub
```
risdenk: Remove TODO

gateway-shell/src/main/java/org/apache/knox/gateway/shell/commands/SelectCommand.java
```diff
@@ -0,0 +1,177 @@
...
+      if (setFromHistory) {
+        sqlField.setText(sqlHistory.get(historyIndex));
+        sqlField.invalidate();
+      }
+    }
+  }
+
+  @Override
+  public void keyReleased(KeyEvent event) {
+    // TODO Auto-generated method stub
```
risdenk: Remove TODO

gateway-shell/src/main/java/org/apache/knox/gateway/shell/commands/KnoxLoginDialog.java
```diff
@@ -0,0 +1,141 @@
...
+  @Override
+  public void setPrompt(String prompt) {
+  }
+
+  @Override
+  public void setName(String name) {
+    this.name = name;
+  }
+
+  private void workAroundFocusIssue(JTextField field) {
```
risdenk: Can a comment be added to this method on what the workaround is for?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/commands/DataSourceCommand.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+    else if (args.get(0).equalsIgnoreCase("remove")) {
+      if (dataSources == null || dataSources.isEmpty()) {
+        return "No datasources to remove.";
+      }
+      // if the removed datasource is currently selected, unselect it
+      dataSources.remove(args.get(1));
+      if (((String)getVariables().get(KNOXDATASOURCE)).equals(args.get(1))) {
+        System.out.println("unselecting datasource.");
+//        getVariables().remove(KNOXDATASOURCE);
```
risdenk: Remove commented out code

gateway-shell/src/main/java/org/apache/knox/gateway/shell/commands/CSVCommand.java
```diff
@@ -0,0 +1,67 @@
...
+
+    try {
+      if (withHeaders) {
+        table = KnoxShellTable.builder().csv().withHeaders().url(url);
+      }
+      else {
+        table = KnoxShellTable.builder().csv().url(url);
+      }
+    } catch (IOException e) {
+      // TODO Auto-generated catch block
```
risdenk: Remove TODO

gateway-shell/src/main/java/org/apache/knox/gateway/shell/commands/AbstractKnoxShellCommand.java
```diff
@@ -0,0 +1,76 @@
...
+                    @Override
+                    public void windowGainedFocus(WindowEvent e) {
+                        c.requestFocus();
+                    }
+                });
+            }
+        }
+    });
+    // end workaround
+  }
```
risdenk: Is this the same workaround as later? Can this logic be in one spot with a comment about why?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSession.java
```diff
@@ -570,6 +580,89 @@ public String toString() {
...
+    try {
+      FileUtils.write(new File(
+          home + File.separator +
+          ".knoxshell" + File.separator + fileName),
+          s, StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      // TODO Auto-generated catch block
+      e.printStackTrace();
+    }
+  }
```
risdenk: So these read/write methods for histories scare me a bit. Could there be multiple knoxshells reading/writing at once? Can we lock the file via the Java NIO methods? Should these methods be synchronized to prevent multiple usage at once from the same JVM. 

Is the directory ".knoxshell" assumed to exist already?

This applies to reading as well.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSession.java
```diff
@@ -570,6 +580,89 @@ public String toString() {
...
+  public static <T> void persistToKnoxShell(String fileName, Map<String, List<T>> map) {
+    String s = JsonUtils.renderAsJsonString(map);
+    String home = System.getProperty("user.home");
+    try {
+      FileUtils.write(new File(
+          home + File.separator +
+          ".knoxshell" + File.separator + fileName),
+          s, StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      // TODO Auto-generated catch block
```
risdenk: Remove TODO

gateway-shell-release/home/bin/knoxshell.sh
```diff
@@ -81,7 +81,7 @@ function main {
 
    checkJava
    buildAppJavaOpts
-   $JAVA "${APP_JAVA_OPTS[@]}" -javaagent:"$APP_BIN_DIR"/../lib/aspectjweaver.jar -jar "$APP_JAR" "$@" || exit 1
+$JAVA "${APP_JAVA_OPTS[@]}" -Dlog4j.configuration=conf/knoxshell-log4j.properties -javaagent:"$APP_BIN_DIR"/../lib/aspectjweaver.jar -cp "$APP_JAR":lib/* org.apache.knox.gateway.shell.Shell "$@" || exit 1
```
risdenk: Indentation

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/JDBCKnoxShellTableBuilder.java
lmccay: hmmm, I wonder why it only had it for username and not pass.
Maybe an empty password is valid?
I don't think that I wrote that but that sort of makes sense.
I'll change it.

risdenk: Yea @smolnar82 wrote it not too long ago. I had commented that empty pass MAY be valid where empty username definitely isn't valid.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/JDBCKnoxShellTableBuilder.java
lmccay: A result of recovering from IDE local history after a problematic merge.
Changing back to password().

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSession.java
lmccay: I can't say it is impossible to have multiple knoxshells reading/writing at once, so I'll add those protections. Thanks. It should create .knoxshell if it doesn't already exist.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/JDBCKnoxShellTableBuilder.java
smolnar82: The `public` visibility is an issue with how AspectJ is configured as of now. If this goes in, I need to figure out how to make it work. I haven't checked it yet, but package-private would not be enough?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSession.java
```diff
@@ -570,6 +584,121 @@ public String toString() {
...
+  public static Map<String, KnoxDataSource> loadDataSources() throws IOException {
+    Map<String, KnoxDataSource> datasources = null;
+    String home = System.getProperty("user.home");
+    String json = null;
+
+    File dsFile = new File(
+        home + File.separator +
+        ".knoxshell" + File.separator + KNOXDATASOURCES_JSON);
+    synchronized(KnoxSession.class) {
+      json = readFileToString(dsFile, "UTF8");
```
risdenk: nit: might make sense to make readFileToString synchronized. to avoid calling that method without synchronization. Or at least move the synchronized block into the readFileToString method.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/KnoxSession.java
```diff
@@ -570,6 +584,121 @@ public String toString() {
...
+   * @param <T>
+   * @param fileName of persisted file
+   * @param map to persist
+   */
+  public static <T> void persistToKnoxShell(String fileName, Map<String, List<T>> map) {
+    String s = JsonUtils.renderAsJsonString(map);
+    String home = System.getProperty("user.home");
+    synchronized(KnoxSession.class) {
+      try {
+        write(new File(
```
risdenk: nit: might make sense to make write synchronized. to avoid calling that method without synchronization. Or at least move the synchronized block into the write method.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/commands/AbstractKnoxShellCommand.java
risdenk: Looks like this duplication still exists after recent commits?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/JDBCKnoxShellTableBuilder.java
risdenk: Looks like this is still `pwd()` instead of `password()` in the recent changes?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/JDBCKnoxShellTableBuilder.java
lmccay: package protected will not be sufficient since it is required to called as part of the fluent API for "applications" that want to own the management of the connection rather than rely on it to be opened and closed within the call itself.

See this sort of code where the connection is externalized:

  protected Connection getConnection(KnoxDataSource ds, String user, String pass) throws SQLException, Exception {
    Connection conn = getConnectionFromSession(ds);
    if (conn == null) {
      if (user != null && pass != null) {
        conn = KnoxShellTable.builder().jdbc()
          .connectTo(ds.getConnectStr())
          .driver(ds.getDriver())
          .username(user)
          .password(pass)
          .createConnection();
      }
      else {
        conn = KnoxShellTable.builder().jdbc()
          .connectTo(ds.getConnectStr())
          .driver(ds.getDriver())
          .createConnection();
      }
      HashMap<String, Connection> connections =
          (HashMap<String, Connection>) getVariables()
          .getOrDefault(KNOXDATASOURCE_CONNECTIONS,
              new HashMap<String, Connection>());
      connections.put(ds.getName(), conn);
      getVariables().put(KNOXDATASOURCE_CONNECTIONS, connections);
    }
    return conn;
  }

lmccay: It is also noting that this configuration of aspectj that limits anyone adding new public methods is too brittle and needs to be addressed if possible.

lmccay: I guess we could completely externalize the creation of the connection if it is going to be externally managed but that still doesn't address the brittleness aspectj is imposing on the builders.

PR #239, opened 2020-01-17
KNOX-2153 - CM discovery - Monitor Cloudera Manager

pzampino:

This change adds the ability for Knox to monitor ClouderaManager-managed clusters for configuration changes, and respond by re-generating topologies based on the most recent configuration.

When Knox performs service discovery against a CM-managed cluster to generate a topology, it registers that cluster data with the monitor, which will periodically check that cluster for service restart events. In CM, config changes are meaningless unless the associated service is restarted.
When Knox identifies a restart event, it will query the current config for the restarted service, and compare it with its previous record from the last discovery thereof. If there are changes, it will trigger rediscovery of the affected topologies.

## How was this patch tested?

So far, this has been tested manually. There will be a subsequent commit with tests added to this PR prior to merging. Given the magnitude of this change, I've created the PR without tests to facilitate early review.

Update: Added tests for the ServiceModel metadata changes used by the monitor, and the configuration model and associated serialization/deserialization support.

### DIFF
--- a/gateway-discovery-cm/pom.xml
+++ b/gateway-discovery-cm/pom.xml
@@ -30,6 +30,10 @@
     <dependencies>
         <dependency>
             <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.knox</groupId>
             <artifactId>gateway-spi</artifactId>
         </dependency>
         <dependency>
@@ -49,8 +53,28 @@
             <artifactId>okhttp</artifactId>
         </dependency>
         <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpcore</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
@@ -28,12 +28,17 @@ import com.cloudera.api.swagger.model.ApiRoleList;
 import com.cloudera.api.swagger.model.ApiService;
 import com.cloudera.api.swagger.model.ApiServiceConfig;
 import com.cloudera.api.swagger.model.ApiServiceList;
+import org.apache.knox.gateway.GatewayServer;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
+import org.apache.knox.gateway.services.GatewayServices;
+import org.apache.knox.gateway.services.ServiceType;
 import org.apache.knox.gateway.services.security.AliasService;
-import org.apache.knox.gateway.topology.discovery.GatewayService;
+import org.apache.knox.gateway.topology.ClusterConfigurationMonitorService;
+import org.apache.knox.gateway.topology.discovery.ClusterConfigurationMonitor;
 import org.apache.knox.gateway.topology.discovery.ServiceDiscovery;
 import org.apache.knox.gateway.topology.discovery.ServiceDiscoveryConfig;
+import org.apache.knox.gateway.topology.discovery.cm.monitor.ClouderaManagerClusterConfigurationMonitor;
 
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -75,8 +80,9 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
 
   private boolean debug;
 
-  @GatewayService
   private AliasService aliasService;
+
+  private ClouderaManagerClusterConfigurationMonitor configChangeMonitor;
 
 
   ClouderaManagerServiceDiscovery() {
@@ -84,7 +90,12 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
   }
 
   ClouderaManagerServiceDiscovery(boolean debug) {
+    GatewayServices gwServices = GatewayServer.getGatewayServices();
+    if (gwServices != null) {
+      this.aliasService = gwServices.getService(ServiceType.ALIAS_SERVICE);
+    }
     this.debug = debug;
+    this.configChangeMonitor = getConfigurationChangeMonitor();
   }
 
   @Override
@@ -105,6 +116,29 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
     return client;
   }
 
+  /**
+   * Get the ClouderaManager configuration change monitor from the associated gateway service.
+   */
+  private ClouderaManagerClusterConfigurationMonitor getConfigurationChangeMonitor() {
+    ClouderaManagerClusterConfigurationMonitor cmMonitor = null;
+
+    try {
+      GatewayServices gwServices = GatewayServer.getGatewayServices();
+      if (gwServices != null) {
+        ClusterConfigurationMonitorService clusterMonitorService =
+            GatewayServer.getGatewayServices().getService(ServiceType.CLUSTER_CONFIGURATION_MONITOR_SERVICE);
+        ClusterConfigurationMonitor monitor =
+            clusterMonitorService.getMonitor(ClouderaManagerClusterConfigurationMonitor.getType());
+        if (monitor != null && ClouderaManagerClusterConfigurationMonitor.class.isAssignableFrom(monitor.getClass())) {
+          cmMonitor = (ClouderaManagerClusterConfigurationMonitor) monitor;
+        }
+      }
+    } catch (Exception e) {
+      log.errorAccessingConfigurationChangeMonitor(e);
+    }
+    return cmMonitor;
+  }
+
   @Override
   public Map<String, Cluster> discover(GatewayConfig gatewayConfig, ServiceDiscoveryConfig discoveryConfig) {
     Map<String, Cluster> clusters = new HashMap<>();
@@ -115,7 +149,7 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
       String clusterName = apiCluster.getName();
       log.discoveredCluster(clusterName, apiCluster.getFullVersion());
 
-      Cluster cluster = discover(gatewayConfig, discoveryConfig, clusterName, client);
+      ClouderaManagerCluster cluster = discover(gatewayConfig, discoveryConfig, clusterName, client);
       clusters.put(clusterName, cluster);
     }
 
@@ -123,15 +157,17 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
   }
 
   @Override
-  public Cluster discover(GatewayConfig gatewayConfig, ServiceDiscoveryConfig discoveryConfig, String clusterName) {
+  public ClouderaManagerCluster discover(GatewayConfig          gatewayConfig,
+                                         ServiceDiscoveryConfig discoveryConfig,
+                                         String                 clusterName) {
     return discover(gatewayConfig, discoveryConfig, clusterName, getClient(discoveryConfig));
   }
 
-  protected Cluster discover(GatewayConfig          gatewayConfig,
-                             ServiceDiscoveryConfig discoveryConfig,
-                             String                 clusterName,
-                             DiscoveryApiClient     client) {
-    ServiceDiscovery.Cluster cluster = null;
+  protected ClouderaManagerCluster discover(GatewayConfig          gatewayConfig,
+                                            ServiceDiscoveryConfig discoveryConfig,
+                                            String                 clusterName,
+                                            DiscoveryApiClient     client) {
+    ClouderaManagerCluster cluster = null;
 
     if (clusterName == null || clusterName.isEmpty()) {
       log.missingDiscoveryCluster();
@@ -140,6 +176,11 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
 
     try {
       cluster = discoverCluster(client, clusterName);
+
+      if (configChangeMonitor != null) {
+        // Notify the cluster config monitor about these cluster configuration details
+        configChangeMonitor.addServiceConfiguration(cluster, discoveryConfig);
+      }
     } catch (ApiException e) {
       log.clusterDiscoveryError(clusterName, e);
     }
@@ -161,8 +202,9 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
     return clusters;
   }
 
-  private static Cluster discoverCluster(DiscoveryApiClient client, String clusterName) throws ApiException {
-    ClouderaManagerCluster cluster = null;
+  private static ClouderaManagerCluster discoverCluster(DiscoveryApiClient client, String clusterName)
+      throws ApiException {
+    ClouderaManagerCluster cluster;
 
     ServicesResourceApi servicesResourceApi = new ServicesResourceApi(client);
     RolesResourceApi rolesResourceApi = new RolesResourceApi(client);
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/hive/HiveOnTezServiceModelGenerator.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/hive/HiveOnTezServiceModelGenerator.java
@@ -16,11 +16,19 @@
  */
 package org.apache.knox.gateway.topology.discovery.cm.model.hive;
 
+import com.cloudera.api.swagger.client.ApiException;
 import com.cloudera.api.swagger.model.ApiConfigList;
+import com.cloudera.api.swagger.model.ApiRole;
+import com.cloudera.api.swagger.model.ApiService;
+import com.cloudera.api.swagger.model.ApiServiceConfig;
+import org.apache.knox.gateway.topology.discovery.cm.ServiceModel;
 
 public class HiveOnTezServiceModelGenerator extends HiveServiceModelGenerator {
 
   public static final String SERVICE_TYPE = "HIVE_ON_TEZ";
+
+  static final String HIVEONTEZ_TRANSPORT_MODE = TRANSPORT_MODE.replaceAll("\\.", "_");
+  static final String HIVEONTEZ_HTTP_PORT      = HTTP_PORT.replaceAll("\\.", "_");
 
   @Override
   public String getServiceType() {
@@ -29,7 +37,28 @@ public class HiveOnTezServiceModelGenerator extends HiveServiceModelGenerator {
 
   @Override
   protected boolean checkHiveServer2HTTPMode(ApiConfigList roleConfig) {
-    return TRANSPORT_MODE_HTTP.equals(getRoleConfigValue(roleConfig, "hive_server2_transport_mode"));
+    return TRANSPORT_MODE_HTTP.equals(getRoleConfigValue(roleConfig, HIVEONTEZ_TRANSPORT_MODE));
+  }
+
+  @Override
+  protected String getHttpPort(ApiConfigList roleConfig) {
+    return getRoleConfigValue(roleConfig, HIVEONTEZ_HTTP_PORT);
+  }
+
+  @Override
+  public ServiceModel generateService(ApiService       service,
+                                      ApiServiceConfig serviceConfig,
+                                      ApiRole          role,
+                                      ApiConfigList    roleConfig) throws ApiException {
+    ServiceModel model = super.generateService(service, serviceConfig, role, roleConfig);
+    model.addRoleProperty(getRoleType(),
+                          HIVEONTEZ_HTTP_PORT,
+                          getRoleConfigValue(roleConfig, HIVEONTEZ_HTTP_PORT));
+    model.addRoleProperty(getRoleType(),
+                          HIVEONTEZ_TRANSPORT_MODE,
+                          getRoleConfigValue(roleConfig, HIVEONTEZ_TRANSPORT_MODE));
+
+    return model;
   }
 
 }
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/oozie/OozieServiceModelGenerator.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/oozie/OozieServiceModelGenerator.java
@@ -31,6 +31,10 @@ public class OozieServiceModelGenerator extends AbstractServiceModelGenerator {
   public static final String SERVICE_TYPE = "OOZIE";
   public static final String ROLE_TYPE    = "OOZIE_SERVER";
 
+  static final String USE_SSL    = "oozie_use_ssl";
+  static final String HTTP_PORT  = "oozie_http_port";
+  static final String HTTPS_PORT = "oozie_https_port";
+
   @Override
   public String getService() {
     return SERVICE;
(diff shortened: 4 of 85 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,940 @@
...
+  private void loadServiceConfiguration() {
+    File persistenceDir = getPersistenceDir();
+    if (persistenceDir != null) {
+      Collection<File> persistedConfigs = FileUtils.listFiles(persistenceDir, new String[]{"ver"}, false);
+      for (File persisted : persistedConfigs) {
+        try (ObjectInputStream in = new ObjectInputStream(Files.newInputStream(persisted.toPath()))) {
+          String clusterName      = (String) in.readObject();
+          String discoveryAddress = (String) in.readObject();
+          Map<String, ServiceConfigurationProperties> configs =
+              (Map<String, ServiceConfigurationProperties>) in.readObject();
```
risdenk: Does this need to be Objects? Can't this be a JSON file or something else?

build-tools/src/main/resources/build-tools/spotbugs-filter.xml
```diff
@@ -71,4 +71,9 @@ limitations under the License.
     <Bug pattern="NM_SAME_SIMPLE_NAME_AS_SUPERCLASS" />
   </Match>
 
+  <Match>
+    <Class name="org.apache.knox.gateway.topology.discovery.cm.ClouderaManagerClusterConfigurationMonitor" />
+    <Bug pattern="OBJECT_DESERIALIZATION" />
```
risdenk: This is super concerning. If this is required (I'm not sure it is based on looking at this class), can this be limited to a single method instead of the whole class?

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,940 @@
...
+      }
+    }
+
+    @Override
+    public void run() {
+      isActive = true;
+
+      // Initial delay to allow other gateway services (e.g., Topology Service) to be fully ready, so restart events
+      // which occurred while Knox was down are not missed/mishandled.
+      waitFor(20);
```
risdenk: The 20 is what in seconds? This should be waiting on some specific event not just some time.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,940 @@
...
+        default:
+      }
+    }
+  }
+
+
+  /**
+   * Service configuration properties container
+   */
+  private static final class ServiceConfigurationProperties implements Serializable {
```
risdenk: I don't think we should have a serializable class. Maybe a jackson pojo for json?

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,940 @@
...
+  private void persistServiceConfiguration(final String                                      discoveryAddress,
+                                           final String                                      cluster,
+                                           final Map<String, ServiceConfigurationProperties> configs) {
+    File persistenceDir = getPersistenceDir();
+    if (persistenceDir != null) {
+      File persistenceFile = getServiceConfigsPersistenceFile(discoveryAddress, cluster);
+      try ( ObjectOutputStream out = new ObjectOutputStream(Files.newOutputStream(persistenceFile.toPath())) ) {
+        out.writeObject(cluster);
+        out.writeObject(discoveryAddress);
+        out.writeObject(configs);
```
risdenk: This is same as readObject, should use JSON or something to avoid serializing Java classes directly.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
```diff
@@ -63,6 +68,9 @@
   static final String DEFAULT_USER_ALIAS = "cm.discovery.user";
   static final String DEFAULT_PWD_ALIAS  = "cm.discovery.password";
 
+  private static final String GATEWAY_SERVICES_ACCESSOR_CLASS  = "org.apache.knox.gateway.GatewayServer";
```
risdenk: Use `GatewayServer.class...` to get the name instead of hardcoding the string.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
```diff
@@ -105,6 +116,34 @@ private DiscoveryApiClient getClient(ServiceDiscoveryConfig discoveryConfig) {
     return client;
   }
 
+  /**
+   * Get the ClouderaManager configuration change monitor from the associated gateway service.
+   */
+  private ClouderaManagerClusterConfigurationMonitor getConfigurationChangeMonitor() {
```
risdenk: I thought GatewayServices is already a public static object that is accessible from GatewayServer or elsewhere. I don't think we need to go through all this effort to try to use reflection here.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/hive/HiveOnTezServiceModelGenerator.java
```diff
@@ -16,20 +16,45 @@
...
+import com.cloudera.api.swagger.model.ApiRole;
+import com.cloudera.api.swagger.model.ApiService;
+import com.cloudera.api.swagger.model.ApiServiceConfig;
+import org.apache.knox.gateway.topology.discovery.cm.ServiceModel;
 
 public class HiveOnTezServiceModelGenerator extends HiveServiceModelGenerator {
 
   public static final String SERVICE_TYPE = "HIVE_ON_TEZ";
 
+  private static final String HOT_TRANSPORT_MODE = TRANSPORT_MODE.replaceAll("\\.", "_");
```
risdenk: `HOT` must be Hive on Tez, but definitely isn't clear from reading.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
pzampino: Good point. I suppose I was rushing a bit to get this finished. It would be better to use an alternative serialization format.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/hive/HiveOnTezServiceModelGenerator.java
pzampino: Fair enough

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,940 @@
...
+                                          final ServiceConfigurationProperties current) {
+    boolean hasChanged = false;
+
+    // Compare the service configuration properties first
+    Properties previousProps = previous.getServiceProps();
+    Properties currentProps  = current.getServiceProps();
+    for (String name : previousProps.stringPropertyNames()) {
+      String prevValue = previousProps.getProperty(name);
+      String currValue = currentProps.getProperty(name);
+      if (!currValue.equals(prevValue)) {
```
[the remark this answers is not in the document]
pzampino: Good catch.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,940 @@
...
+    }
+
+    init();
+  }
+
+  @Override
+  @SuppressWarnings("PMD.DoNotUseThreads")
+  public void start() {
+    log.startingClouderaManagerConfigMonitor();
+    (new Thread(internalMonitor, "ClouderaManagerConfigurationMonitor")).start();
```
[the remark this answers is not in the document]
smolnar82: One more advantage: using an executor does not require creating a Thread -> the PMD warning suppression would not be necessary

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
pzampino: This module doesn't depend on gateway-server...I think the services would have to be passed down to avoid this

smolnar82:
Just an idea you may consider: a new DAO interface should be declared with one implementation (e.g. `ClouderaManagerClusterConfigurationFileSystemDAO implements ClouderaManagerClusterConfigurationDAO`) and all the persist/load methods are moved into the DAO layer for better separation of the business logic and file system operations.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,940 @@
...
+
+      log.startedClouderaManagerConfigMonitor(interval);
+
+      while (isActive) {
+        for (Map.Entry<String, List<String>> entry : monitor.getClusterNames().entrySet()) {
+          String address = entry.getKey();
+          for (String clusterName : entry.getValue()) {
+            log.checkingClusterConfiguration(clusterName, address);
+
+            // Configuration changes don't mean anything without corresponding service restarts. Therefore, monitor
```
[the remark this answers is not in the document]
pzampino: I would like to consider that as another JIRA (e.g., improvement) beyond the scope of this one.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
pzampino: Added JSON serialization format instead

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
pzampino: Leveraged the separation of the init() and start() lifecycle events in the gateway services to achieve what I was trying to accomplish.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
pzampino: Added a dependency on gateway-server to facilitate the proposed change.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerClusterConfigurationMonitor.java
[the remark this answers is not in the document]
pzampino: Agreed

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/oozie/OozieServiceModelGenerator.java
```diff
@@ -59,15 +63,22 @@ public ServiceModel generateService(ApiService       service,
...
+    String sslEnabled = getServiceConfigValue(serviceConfig, USE_SSL);
+    if(Boolean.parseBoolean(sslEnabled)) {
       scheme = "https";
-      port = getRoleConfigValue(roleConfig, "oozie_https_port");
+      port = getRoleConfigValue(roleConfig, HTTPS_PORT);
     } else {
       scheme = "http";
-      port = getRoleConfigValue(roleConfig, "oozie_http_port");
+      port = getRoleConfigValue(roleConfig, HTTP_PORT);
     }
```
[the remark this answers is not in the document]
pzampino: Since that is an existing condition, I would like to defer that kind of change to a subsequent JIRA.

pzampino:
> The biggest concern I have is the java serialization that doesn't seem entirely necessary.
> 
> The other item that seems confusing is there a bunch of `model/*` cleanup as far as constants and fixing variables. It would be cleaner to separate this so its easier to tell exactly what changed for monitoring the CM API.

I believe I've addressed the object serialization concerns, and the ServiceModel changes are intended to facilitate the determination of relevant config changes. The constants were added because the values are now being referenced in multiple places.

pzampino:
Still working on the tests...

gateway-discovery-cm/pom.xml
```diff
@@ -48,10 +52,29 @@
...
+            <artifactId>commons-io</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpcore</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <scope>compile</scope>
```
risdenk: `<scope>compile</scope>` not needed

gateway-discovery-cm/pom.xml
```diff
@@ -48,10 +52,29 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <scope>compile</scope>
```
risdenk: `<scope>compile</scope>` not needed

gateway-discovery-cm/pom.xml
```diff
@@ -48,10 +52,29 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <scope>compile</scope>
+        </dependency>
```
risdenk: We should REALLY try to avoid guava. I'll try to see what we can do about it. I've tried to limit its usage where possible since it tends to be a PITA dependency.

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryTest.java
```diff
@@ -331,6 +331,17 @@ public void testLivyDiscoverySSL() {
     assertEquals("https://" + hostName + ":" + port, livyURLs.get(0));
   }
 
+  @Test
+  public void testLivyDiscoveryWhenTransportModeIsBinary() {
+    final String hostName    = "livy-host";
+    final String port        = "8998";
+    ServiceDiscovery.Cluster cluster = doTestLivyDiscovery(hostName, port, false, "binary");
```
risdenk: This doesn't make much sense to me. Livy discovery should be on the HTTP endpoint. Binary vs HTTP would be on the thrift server not on the HTTP endpoint....

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryTest.java
```diff
@@ -988,8 +999,17 @@ private void doTestSolrDiscovery(final boolean isSSL) {
...
+  }
+
+
+  private ServiceDiscovery.Cluster doTestLivyDiscovery(final String  hostName,
+                                                       final String  port,
+                                                       final Boolean isSSL,
+                                                       final String  transportMode) {
     // Configure the role
     Map<String, String> roleProperties = new HashMap<>();
+    roleProperties.put("livy_server_thrift_transport_mode", transportMode);
```
risdenk: We shouldn't care about this. Livy has a thrift server which is completely separate from the HTTP API.

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ServiceConfigurationRecordObjectMapperTest.java
```diff
@@ -0,0 +1,240 @@
...
+    try (ByteArrayInputStream in = new ByteArrayInputStream(serialized.getBytes(StandardCharsets.UTF_8))) {
+      deserialized =
+          (new ServiceConfigurationRecordObjectMapper()).readValue(in, ServiceConfigurationRecord.class);
+    } catch (Exception e) {
+      fail(e.getMessage());
+    }
+    return deserialized;
+  }
+
+  @SuppressWarnings({"PMD.AppendCharacterWithChar", "PMD.ConsecutiveLiteralAppends"})
```
risdenk: Not sure why these are suppressed? It would be better to actually do what it says. One of the simple ones is use `'` around literals. ie: `'{'` or `'"'`

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,876 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * ClusterConfigurationMonitor implementation for clusters managed by ClouderaManager.
+ */
+public class ClouderaManagerClusterConfigurationMonitor implements ClusterConfigurationMonitor {
+
+  private static final String TYPE = "CM";
+
+  private static final String CLUSTERS_DATA_DIR_NAME = TYPE.toLowerCase(Locale.getDefault()) + "-clusters";
```
risdenk: Most likely want `Locale.ROOT` to avoid weirdness in the TYPE lowercasing.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,876 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * ClusterConfigurationMonitor implementation for clusters managed by ClouderaManager.
+ */
+public class ClouderaManagerClusterConfigurationMonitor implements ClusterConfigurationMonitor {
```
risdenk: So overall I think this class is doing too much. It would probably be best to split out the classes that are in here to separate classes. 

The locking is weird too it would be better to split out the required locking concerns into its own class and encapsulate what is happening. I think this would be easier to check for correctness.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,876 @@
...
+    return TYPE;
+  }
+
+  ClouderaManagerClusterConfigurationMonitor(final GatewayConfig config, final AliasService aliasService) {
+    this.gatewayConfig = config;
+    this.aliasService = aliasService;
+
+    eventQueryTimestampFormat.get().setTimeZone(TimeZone.getTimeZone("UTC"));
+
+    ThreadFactory tf = (new ThreadFactoryBuilder()).setNameFormat("ClouderaManagerConfigurationMonitor-%d").build();
```
risdenk: Looks like this is the only usage of Guava. We already have a pattern for doing this with commons-lang3 instead?

https://github.com/apache/knox/blob/master/gateway-cm-integration/src/main/java/org/apache/knox/gateway/cm/descriptor/ClouderaManagerDescriptorMonitor.java#L59

Another option is potentially a custom thread factory.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ServiceConfigurationRecordObjectMapper.java
```diff
@@ -0,0 +1,111 @@
...
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.databind.ser.std.StdSerializer;
+
+import java.io.IOException;
+import java.util.Iterator;
+
+/**
+ * Custom ObjectMapper implementation for serializing/deserializing ServiceConfigurationRecord objects.
+ */
+class ServiceConfigurationRecordObjectMapper extends ObjectMapper {
```
risdenk: So this could be ignorance regarding Jackson, but I thought we could just annotate the class instead of having to write custom serialize/deserialize stuff. We use Jackson elsewhere and no where else do we `extends ObjectMapper`

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ClouderaManagerClusterConfigurationMonitor.java
```diff
@@ -0,0 +1,876 @@
...
+      if (!clustersDir.exists()) {
+        if (!clustersDir.mkdirs()) {
+          log.failedToCreatePersistenceDirectory(clustersDir.getAbsolutePath());
+        }
+      }
+      persistenceDir = clustersDir;
+    }
+
+    return persistenceDir;
+  }
```
risdenk: Nit: It might be cleaner to use `Path` and Java NIO here. This would cascade to the rest of the `File` usages.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ClouderaManagerClusterConfigurationMonitor.java
pzampino: +1

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ClouderaManagerClusterConfigurationMonitor.java
pzampino: +1

gateway-discovery-cm/pom.xml
pzampino: +1

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ClouderaManagerClusterConfigurationMonitor.java
smolnar82: I'm with Kevin here (see my previous comment on creating a DAO to load/persist data into the disk) as well as separating at least one other class out.

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryTest.java
pzampino: Ok

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryTest.java
pzampino: Understood

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ServiceConfigurationRecordObjectMapper.java
pzampino: I had encountered some difficulty employing annotations, which led me to the current solution. This could be my ignorance of Jackson.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/ClouderaManagerClusterConfigurationMonitor.java
pzampino: I've made some changes along these lines.

PR #243, opened 2020-01-27
KNOX-2200 - DefaultKeystoreService can lose entries under concurrent access

risdenk:

Add synchronization to `addCredentialForCluster` and `removeCredentialForCluster` to prevent read then write synchronization issues. Added a test that failed without this change for each method and then confirmed passed with fix.

## How was this patch tested?

* Checked new unit test fails without change
* Checked new unit test doesn't fail with change
* `mvn -T.75C verify -Ppackage,release -Dshellcheck`

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
@@ -76,7 +76,8 @@ public class DefaultKeystoreService implements KeystoreService, Service {
   private static GatewayResources RES = ResourcesFactory.get(GatewayResources.class);
 
   //let's configure the cache with hard-coded attributes now; we can introduce new gateway configuration later on if needed
-  private final Cache<CacheKey, String> cache = Caffeine.newBuilder().expireAfterAccess(10, TimeUnit.MINUTES).maximumSize(1000).build();
+  // visible for testing
+  final Cache<CacheKey, String> cache = Caffeine.newBuilder().expireAfterAccess(10, TimeUnit.MINUTES).maximumSize(1000).build();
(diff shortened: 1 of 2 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -286,18 +287,21 @@ public KeyStore getCredentialStoreForCluster(String clusterName)
...
-        final Key key = new SecretKeySpec(value.getBytes(StandardCharsets.UTF_8), "AES");
-        ks.setKeyEntry(alias, key, masterService.getMasterSecret(), null);
-
-        final Path keyStoreFilePath = keyStoreDirPath.resolve(clusterName + CREDENTIALS_SUFFIX);
-        writeKeyStoreToFile(ks, keyStoreFilePath, masterService.getMasterSecret());
-        addToCache(clusterName, alias, value);
-      } catch (KeyStoreException | IOException | CertificateException | NoSuchAlgorithmException e) {
-        LOG.failedToAddCredentialForCluster(clusterName, e);
+    // Needed to prevent read then write synchronization issue where alias is not added
+    synchronized (this) {
```
pzampino: Why not just mark the whole method as synchronized?

risdenk: Changes the method signature. There isn't necessarily a need to synchronize the public method. This only comes up because we read the keystore then write to it. Other implementations can just add (ie: if this was a map we could update in place)

pzampino: Sounds good to me

PR #249, opened 2020-01-31
KNOX-2213 - Service Discovery Support for CM UI, API

moresandeep:
Currently when autodiscovery is turned on for CM all the services that are defined in the descriptor are added to the topology except for the CM UI and CM API. They are special cases given CM url is used for service discovery. This PR populates the topology with CM UI and CM API url when they are added to the descriptor.

## How was this patch tested?
This patch was manually Tested.

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
@@ -221,17 +221,15 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
 
     ApiServiceList serviceList = getClusterServices(servicesResourceApi, clusterName);
 
-    /* Since Cloudera Manager does not have a service for itself, we will add a skeleton CM
-      service so that we can add CM service to topology when auto-discovery is
-      turned on and CM service is selected in the descriptor */
-    final ApiService cmUIService = new ApiService();
-    cmUIService.setName(CM_SERVICE_TYPE.toLowerCase(Locale.ROOT));
-    cmUIService.setType(CM_SERVICE_TYPE);
-    final ApiService cmAPIService = new ApiService();
-    cmAPIService.setName(CM_SERVICE_TYPE.toLowerCase(Locale.ROOT));
-    cmAPIService.setType(CM_SERVICE_TYPE);
-    serviceList.addItemsItem(cmUIService);
-    serviceList.addItemsItem(cmAPIService);
+    /*
+    Since Cloudera Manager does not have a service for itself, we will add a skeleton CM
+    service so that we can add CM service to topology when auto-discovery is
+    turned on and CM service is selected in the descriptor
+    */
+    final ApiService cmService = new ApiService();
+    cmService.setName(CM_SERVICE_TYPE.toLowerCase(Locale.ROOT));
+    cmService.setType(CM_SERVICE_TYPE);
+    serviceList.addItemsItem(cmService);
 
     if (serviceList != null) {
       for (ApiService service : serviceList.getItems()) {
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
@@ -221,19 +221,17 @@ public class ClouderaManagerServiceDiscovery implements ServiceDiscovery {
 
     ApiServiceList serviceList = getClusterServices(servicesResourceApi, clusterName);
 
-    /*
-    Since Cloudera Manager does not have a service for itself, we will add a skeleton CM
-    service so that we can add CM service to topology when auto-discovery is
-    turned on and CM service is selected in the descriptor
-    */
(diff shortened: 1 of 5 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
```diff
@@ -202,6 +204,22 @@ private static Cluster discoverCluster(DiscoveryApiClient client, String cluster
...
+        for (ServiceModelGenerator serviceModelGenerator : smgList) {
+          if (serviceModelGenerator != null) {
+            serviceModelGenerator.setApiClient(client);
+            ServiceModel serviceModel = serviceModelGenerator
+                .generateService(null, null, null, null);
+            serviceModels.add(serviceModel);
+          }
+        }
+      }
+
```
smolnar82: I believe it would be easier to read and understand later on if you add the CM service above to `serviceList` and modify `getRoles(rolesResourceApi, clusterName, serviceName)` to fetch the roles using the CM REST API for other services and add CM API/UI role manually. Then you would not have to copy-paste the logic which checks service model generators

pzampino: +1

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryTest.java
```diff
@@ -562,6 +565,30 @@ public void testNiFiRegistryDiscoverySSL() {
     doTestNiFiRegistryDiscovery(true);
   }
 
+  @Test
+  public void testCMDiscovery() {
```
pzampino: I think it would be better to separate these, as has been done with other services (e.g., testCMDiscoveryAPI(), testCMDiscoveryUI())

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
moresandeep: Sure, I'll update the PR with this change.

gateway-discovery-cm/src/test/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryTest.java
moresandeep: I see, sure, I'll update this.

moresandeep:
@pzampino @smolnar82 I addressed your comments, let me know what you think.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
```diff
@@ -216,18 +220,38 @@ private static ClouderaManagerCluster discoverCluster(DiscoveryApiClient client,
...
+    /*
+    Since Cloudera Manager does not have a service for itself, we will add a skeleton CM
+    service so that we can add CM service to topology when auto-discovery is
+    turned on and CM service is selected in the descriptor
+    */
+    final ApiService cmService = new ApiService();
+    cmService.setName(CM_SERVICE_TYPE.toLowerCase(Locale.ROOT));
+    cmService.setType(CM_SERVICE_TYPE);
+    serviceList.addItemsItem(cmService);
```
smolnar82: `serviceList` can be `null` -> possible NPE here

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
```diff
@@ -280,7 +304,17 @@ private static ApiRoleList getRoles(RolesResourceApi rolesResourceApi,
                                       String serviceName) {
     ApiRoleList roles = null;
     try {
-      roles = rolesResourceApi.readRoles(clusterName, serviceName, "", VIEW_SUMMARY);
+      /* Populate roles for CM Service since they are not discoverable */
+      if(CM_SERVICE_TYPE.equalsIgnoreCase(serviceName)) {
+        final ApiRoleList roleList = new ApiRoleList();
```
smolnar82: There is already an `ApiRoleList` type local variable - `roles` - which should be re-used and initiated here.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
moresandeep: Thanks, updated the PR.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscovery.java
moresandeep: Thanks, updated the PR.

PR #250, opened 2020-01-31
KNOX-2209 - Improve logging for Knox token handling

pzampino:

Added logging to the Token and TokenState services.

## How was this patch tested?

Manually tested by way of visual inspection of logs.

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -160,6 +160,9 @@ public class TokenResource {
 
     // If server-managed token expiration is configured, set the token store service
     if (Boolean.valueOf(context.getInitParameter(TokenStateService.CONFIG_SERVER_MANAGED))) {
+      String topologyName = getTopologyName();
+      log.serverManagedTokenStateEnabled(topologyName);
+
       GatewayServices services = (GatewayServices) context.getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
       tokenStateService = services.getService(ServiceType.TOKEN_STATE_SERVICE);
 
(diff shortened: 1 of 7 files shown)


gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -384,4 +391,9 @@ private long getExpiry() {
     }
     return expiry;
   }
+
+  private String getTopologyName() {
+    return (String) context.getAttribute("org.apache.knox.gateway.gateway.cluster");
```
risdenk: nit: Would be good to make this a constant. I ran into this looking into something else. It was NOT clear that this attribute was the topology name.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceMessages.java
```diff
@@ -22,7 +22,7 @@
 import org.apache.knox.gateway.i18n.messages.Messages;
 import org.apache.knox.gateway.i18n.messages.StackTrace;
 
-@Messages(logger="org.apache.knox.gateway.service.knoxsso")
+@Messages(logger="org.apache.knox.gateway.service.token")
```
risdenk: Might make sense to match the package name? `knoxtoken`

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
pzampino: This is why I put it in a method that makes it more clear (aside from the obvious benefit of re-use). I can create a constant for it. Are you suggesting a constant in this class, or somewhere else for broader consumption?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceMessages.java
pzampino: I can buy that. I was surprised to see that it had been knoxsso :-O

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
risdenk: Broader consumption - doesn't have to be as part of this change. The string `"org.apache.knox.gateway.gateway.cluster"` is being used to insert the topology value into the context.

pzampino: +1

PR #251, opened 2020-01-31
KNOX-2215 - Token service should return a 401 response when the renew…

pzampino:
…er is not white-listed

The Knox Token service has been modified to respond to renew/revoke requests with a HTTP 401 Unauthorized status when the "renewer" is either unknown or has not been white-listed in the service configuration instead of the more generic HTTP 400 Bad Request.

## How was this patch tested?

Modified existing TokenServiceResourceTest methods for the affected renew/revoke scenarios.

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -215,7 +215,9 @@ public class TokenResource {
     Response resp;
 
     long expiration = 0;
-    String  error   = "";
+
+    String          error       = "";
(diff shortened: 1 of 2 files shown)


pzampino:
@risdenk, my rational for choosing 401 is the definition from https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html:

401: "If the request already included Authorization credentials, then the 401 response indicates that **authorization** has been refused for those credentials."

403: "The server understood the request, but is refusing to fulfill it. Authorization will not help"

So, in this case, the renewing/revoking user is _AUTHENTICATED_, but **NOT AUTHORIZED** to perform the requested operation.

lmccay:
I believe 403 is more appropriate but it should align with what hadoop does
either way.

PR #252, opened 2020-02-04
KNOX-2207 - TokenStateService revocation should remove persisted token state

moresandeep:

Do not maintain state for revoked tokens. 

## How was this patch tested?

This patch was manually tested.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
@@ -105,14 +105,8 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService {
   @Override
   public void revokeToken(final String token) {
     /* no reason to keep revoked tokens around */
-    removeRevokedExpiredToken(token);
+    removeToken(token);
     log.revokedToken(getTokenDisplayText(token));
-  }
-
-  @Override
-  protected boolean isRevoked(final String token) {
-    /* since we no longer maintain revoked tokens, revoked tokens are unknown tokens */
-    return isUnknown(token);
   }
 
   @Override
@@ -127,11 +121,8 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService {
   }
 
   @Override
-  protected void removeRevokedExpiredToken(final String token) {
-    if (isUnknown(token)) {
-      log.unknownToken(getTokenDisplayText(token));
-      throw new IllegalArgumentException("Unknown or revoked token.");
-    }
+  protected void removeToken(final String token) {
+    validateToken(token);
 
     try {
       aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME, token);
@@ -146,7 +137,7 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService {
   protected void updateExpiration(final String token, long expiration) {
     if (isUnknown(token)) {
       log.unknownToken(getTokenDisplayText(token));
-      throw new IllegalArgumentException("Unknown or revoked token.");
+      throw new IllegalArgumentException("Unknown token.");
     }
 
     try {
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
@@ -104,14 +104,15 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService {
 
   @Override
   public void revokeToken(final String token) {
-    // Record the revocation by setting the expiration to -1
-    updateExpiration(token, -1L);
+    /* no reason to keep revoked tokens around */
+    removeRevokedExpiredToken(token);
     log.revokedToken(getTokenDisplayText(token));
   }
 
   @Override
   protected boolean isRevoked(final String token) {
-    return (getTokenExpiration(token) < 0);
+    /* since we no longer maintain revoked tokens, revoked tokens are unknown tokens */
+    return isUnknown(token);
   }
 
   @Override
@@ -126,10 +127,26 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService {
   }
 
   @Override
+  protected void removeRevokedExpiredToken(final String token) {
+    if (isUnknown(token)) {
+      log.unknownToken(getTokenDisplayText(token));
+      throw new IllegalArgumentException("Unknown or revoked token.");
+    }
+
+    try {
+      aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME, token);
+      aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME,token + "--max");
+    } catch (AliasServiceException e) {
+      log.failedToUpdateTokenExpiration(e);
+    }
+
+  }
+
+  @Override
   protected void updateExpiration(final String token, long expiration) {
     if (isUnknown(token)) {
       log.unknownToken(getTokenDisplayText(token));
-      throw new IllegalArgumentException("Unknown token.");
+      throw new IllegalArgumentException("Unknown or revoked token.");
     }
 
     try {
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
@@ -167,15 +167,16 @@ public class DefaultTokenStateService implements TokenStateService {
 
   @Override
   public boolean isExpired(final String token) {
-    boolean isUnknownToken;
-
-    isUnknownToken = isUnknown(token); // Check if the token exist
-    if (!isUnknownToken) {
-      // If it not unknown, check its expiration
-      isUnknownToken = (getTokenExpiration(token) <= System.currentTimeMillis());
+    try {
+      validateToken(token);
+    } catch (final IllegalArgumentException e) {
+      if (e.toString().contains("Token data cannot be null.") || e.toString().contains("Unknown token")) {
+        return true;
+      } else {
+        throw e;
+      }
     }
-
-    return isUnknownToken;
+    return (getTokenExpiration(token) <= System.currentTimeMillis());
   }
 
   protected void setMaxLifetime(final String token, long issueTime, long maxLifetimeDuration) {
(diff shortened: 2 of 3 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -104,14 +104,15 @@ public long getTokenExpiration(final String token) {
...
   public void revokeToken(final String token) {
-    // Record the revocation by setting the expiration to -1
-    updateExpiration(token, -1L);
+    /* no reason to keep revoked tokens around */
+    removeRevokedExpiredToken(token);
     log.revokedToken(getTokenDisplayText(token));
   }
 
   @Override
   protected boolean isRevoked(final String token) {
```
pzampino: I'm wondering if we need this method at all since there is no need to distinguish revoked tokens from unknown tokens.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -125,11 +126,27 @@ protected boolean isUnknown(final String token) {
     return isUnknown;
   }
 
+  @Override
+  protected void removeRevokedExpiredToken(final String token) {
```
pzampino: I think the more simply-named removeToken(String) would be better. Who cares why it's being removed?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -125,11 +126,27 @@ protected boolean isUnknown(final String token) {
...
+    }
+
+  }
+
   @Override
   protected void updateExpiration(final String token, long expiration) {
     if (isUnknown(token)) {
       log.unknownToken(getTokenDisplayText(token));
-      throw new IllegalArgumentException("Unknown token.");
+      throw new IllegalArgumentException("Unknown or revoked token.");
```
pzampino: I think we can continue treating them simply as "unknown", rather than giving the illusion of distinction.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -125,11 +126,27 @@ protected boolean isUnknown(final String token) {
     return isUnknown;
   }
 
+  @Override
+  protected void removeRevokedExpiredToken(final String token) {
+    if (isUnknown(token)) {
+      log.unknownToken(getTokenDisplayText(token));
+      throw new IllegalArgumentException("Unknown or revoked token.");
```
pzampino: I think we can continue treating them simply as "unknown", rather than giving the illusion of distinction.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -125,11 +126,27 @@ protected boolean isUnknown(final String token) {
     return isUnknown;
   }
 
+  @Override
+  protected void removeRevokedExpiredToken(final String token) {
+    if (isUnknown(token)) {
```
pzampino: Would it be better to employ validateToken() here?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -160,7 +156,8 @@ public void revokeToken(final JWTToken token) {
   @Override
   public void revokeToken(final String token) {
     validateToken(token);
```
pzampino: If removeToken() invokes validateToken(), then we can skip the validateToken() call here.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -208,6 +205,18 @@ protected void updateExpiration(final String token, long expiration) {
     }
   }
 
+  protected void removeRevokedExpiredToken(final String token) {
```
pzampino: Again, removeToken(String) should be a sufficiently-clear name.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -208,6 +205,18 @@ protected void updateExpiration(final String token, long expiration) {
...
+    if (!isValidIdentifier(token)) {
```
pzampino: Could just call validateToken() here?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -222,7 +231,7 @@ protected long getMaxLifetime(final String token) {
   }
 
   protected boolean isRevoked(final String token) {
```
pzampino: May not need this method at all.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateServiceMessages.java
```diff
@@ -33,7 +33,7 @@
   @Message(level = MessageLevel.DEBUG, text = "Revoked token {0}")
   void revokedToken(String tokenDisplayText);
 
-  @Message(level = MessageLevel.DEBUG, text = "Unknown token {0}")
+  @Message(level = MessageLevel.DEBUG, text = "Unknown or revoked token {0}")
```
pzampino: I think we can continue treating them as unknown.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
moresandeep: Sure, updated.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
moresandeep: Addressed in new update.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
moresandeep: Addressed in new update.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: Addressed in new update.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: Addressed in new update.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateServiceMessages.java
moresandeep: Addressed in new update.

moresandeep:
@pzampino  addressed the review comments in new PR, let me know what you think.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -168,15 +167,15 @@ public boolean isExpired(final JWTToken token) {
 
   @Override
   public boolean isExpired(final String token) {
-    boolean isExpired;
+    boolean isUnknownToken;
```
pzampino: Should still be isExpired ?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -168,15 +167,15 @@ public boolean isExpired(final JWTToken token) {
...
 
-    isExpired = isRevoked(token); // Check if it has been revoked first
-    if (!isExpired) {
-      // If it has not been revoked, check its expiration
-      isExpired = (getTokenExpiration(token) <= System.currentTimeMillis());
+    isUnknownToken = isUnknown(token); // Check if the token exist
```
pzampino: Should this be using validateToken() for the unknown check?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: It's just a variable name change, I changed it because it was similar to the method name (nothing wrong with it but just a habit from other programming language)

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: Yup, I was trying to avoid refactoring other parts, but this simplifies things.

moresandeep: Actually, we cannot use validateToken(). validateToken() method throws an exception and does not return a boolean value. So calls to isExpired() will now start throwing exceptions. I could wrap validateToken() with try catch to get better reuse the code.

pzampino: Ok, just use isUnknown() as isRevoked() was used originally.

moresandeep: Ok, reverted to older change.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -125,6 +120,19 @@ protected boolean isUnknown(final String token) {
...
 
+  @Override
+  protected void removeToken(final String token) {
+    validateToken(token);
+
+    try {
+      aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME, token);
+      aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME,token + "--max");
+    } catch (AliasServiceException e) {
+      log.failedToUpdateTokenExpiration(e);
```
smolnar82: This seems a misleading log message

moresandeep: Right, I'll get it fixed, thanks for letting me know @smolnar82

PR #256, opened 2020-02-06
KNOX-2189 - KnoxShellTable.select() must handle whitespace

lmccay4:

Ability to use whitespace as a delimiter in the select method

## How was this patch tested?

Manual and unit tests

### DIFF
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
@@ -356,6 +356,7 @@ public class KnoxShellTable {
   public KnoxShellTable select(String cols) {
     KnoxShellTable table = new KnoxShellTable();
     List<List<Comparable<? extends Object>>> columns = new ArrayList<>();
+    cols = cols.trim();
     String[] colnames = cols.split(",\\s*");
     for (String colName : colnames) {
       table.header(colName);
--- a/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
+++ b/gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
@@ -357,7 +357,7 @@ public class KnoxShellTable {
     KnoxShellTable table = new KnoxShellTable();
     List<List<Comparable<? extends Object>>> columns = new ArrayList<>();
     cols = cols.trim();
-    String[] colnames = cols.split(",\\s*");
+    String[] colnames = cols.split("\\s*,\\s*");
     for (String colName : colnames) {
       table.header(colName);
       columns.add(values(headers.indexOf(colName)));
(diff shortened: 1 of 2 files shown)


gateway-shell/src/main/java/org/apache/knox/gateway/shell/table/KnoxShellTable.java
```diff
@@ -356,7 +356,7 @@ public KnoxShellTableFilter filter() {
   public KnoxShellTable select(String cols) {
     KnoxShellTable table = new KnoxShellTable();
     List<List<Comparable<? extends Object>>> columns = new ArrayList<>();
-    String[] colnames = cols.split(",");
+    String[] colnames = cols.split(",\\s*");
```
risdenk: Does this work if `Column A` or `Column C` has trailing whitespace? It might be better to split on comma and then do a `colName.trim()` or something along those lines.

lmccay: It seems to me that it will "work" depending on what your expectations are. I don't believe that anyone would be surprised that a comma before a comma is left on the end of the previous string or if you have a trailing slash after the last name. It seems more of an explicit action to match the actual colName than a mistake to me.

On the other hand, if your colName does have a trailing space for some reason there would be no way to use select to get it if you strip out all trailing spaces.

I think this change should be for accommodating spaces following a comma as an extended delimiter. Of course, if you have colNames that have a leading slash will not work this way either. I guess we could add a new select method as well which takes a String[] where you can include all the spaces you want.

For this particular method, I think this approach is fine.

lmccay: It may be a good idea to add tests for those explicit scenarios to show what is expected behavior though.

risdenk: Its not just trailing space either - leading whitespace on the first column will be still there. So only leading whitespace after columns is removed. So as this is doesn't really work consistently.

lmccay: I think the only way to make it work consistently would be to trim leading and trailing spaces in the header names before adding to the table then doing the same here would work. I can't imagine that there is a valid reason to have leading or trailing spaces there but assuming that will mean that we might be changing the actual data when coming from CSV files and the like. Thoughts?

lmccay: @lmccay4 - can you add a trim() for the whole string before doing the split to remove and leading or trailing spaces in the comma separated list? Then your initial regex for the split will work properly now that KNOX-2227 is pushed.

lmccay4: @lmccay See the newest patch for the trim()

risdenk: what happens in this case:

`"         Column A         ,            Column B            ,        Column C               "`

Right now you are just testing the case where you are looking for. You just happen to have the correct regex and trim. Any trailing space after a column name before a comma looks like it will fail.

lmccay4: Please see newest patch for updated test

PR #257, opened 2020-02-07
KNOX-2214 - Periodic job to evict expired tokens

moresandeep:
This PR adds a periodic task that looks for expired tokens and then evicts them. The aim of this task to run at a low priority and be able to gracefully handle failures if there are issues. 

## How was this patch tested?
Local testing

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
@@ -248,7 +248,9 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   private static final long DEFAULT_CLOUDERA_MANAGER_ADVANCED_SERVICE_DISCOVERY_CONF_MONITOR_INTERVAL = 30000L;
 
   private static final String KNOX_TOKEN_EVICTION_INTERVAL = GATEWAY_CONFIG_FILE_PREFIX + ".knox.token.eviction.interval";
+  private static final String KNOX_TOKEN_EVICTION_GRACE_PERIOD = GATEWAY_CONFIG_FILE_PREFIX + ".knox.token.eviction.grace.period";
   private static final long KNOX_TOKEN_EVICTION_INTERVAL_DEFAULT = TimeUnit.MINUTES.toSeconds(5);
+  private static final long KNOX_TOKEN_EVICTION_GRACE_PERIOD_DEFAULT = TimeUnit.MINUTES.toSeconds(5);
 
   public GatewayConfigImpl() {
     init();
@@ -1124,5 +1126,10 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   @Override
   public long getKnoxTokenEvictionInterval() {
     return getLong(KNOX_TOKEN_EVICTION_INTERVAL, KNOX_TOKEN_EVICTION_INTERVAL_DEFAULT);
+  }
+
+  @Override
+  public long getKnoxTokenEvictionGracePeriod() {
+    return getLong(KNOX_TOKEN_EVICTION_GRACE_PERIOD, KNOX_TOKEN_EVICTION_GRACE_PERIOD_DEFAULT);
   }
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
@@ -247,6 +247,9 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   private static final String CLOUDERA_MANAGER_ADVANCED_SERVICE_DISCOVERY_CONF_MONITOR_INTERVAL = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.advanced.service.discovery.config.monitor.interval";
   private static final long DEFAULT_CLOUDERA_MANAGER_ADVANCED_SERVICE_DISCOVERY_CONF_MONITOR_INTERVAL = 30000L;
 
+  private static final String KNOX_TOKEN_EVICTION_INTERVAL = GATEWAY_CONFIG_FILE_PREFIX + ".knox.token.eviction.interval";
+  private static final long KNOX_TOKEN_EVICTION_INTERVAL_DEFAULT = TimeUnit.MINUTES.toSeconds(5);
+
   public GatewayConfigImpl() {
     init();
   }
@@ -1117,4 +1120,9 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
     return getBoolean(TOKEN_STATE_SERVER_MANAGED, false);
   }
 
+
+  @Override
+  public long getKnoxTokenEvictionInterval() {
+    return getLong(KNOX_TOKEN_EVICTION_INTERVAL, KNOX_TOKEN_EVICTION_INTERVAL_DEFAULT);
+  }
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
@@ -21,7 +21,10 @@ import org.apache.knox.gateway.services.ServiceLifecycleException;
 import org.apache.knox.gateway.services.security.AliasService;
 import org.apache.knox.gateway.services.security.AliasServiceException;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * A TokenStateService implementation based on the AliasService.
@@ -29,6 +32,7 @@ import java.util.Map;
 public class AliasBasedTokenStateService extends DefaultTokenStateService {
 
   private AliasService aliasService;
+  private static final String TOKEN_MAX_LIFETIME_POSTFIX = "--max";
 
   public void setAliasService(AliasService aliasService) {
     this.aliasService = aliasService;
@@ -62,7 +66,7 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService {
   protected void setMaxLifetime(final String token, long issueTime, long maxLifetimeDuration) {
     try {
       aliasService.addAliasForCluster(AliasService.NO_CLUSTER_NAME,
-                                      token + "--max",
+                                      token + TOKEN_MAX_LIFETIME_POSTFIX,
                                       String.valueOf(issueTime + maxLifetimeDuration));
     } catch (AliasServiceException e) {
       log.failedToSaveTokenState(e);
@@ -74,7 +78,7 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService {
     long result = 0;
     try {
       char[] maxLifetimeStr =
-                      aliasService.getPasswordFromAliasForCluster(AliasService.NO_CLUSTER_NAME, token + "--max");
+                      aliasService.getPasswordFromAliasForCluster(AliasService.NO_CLUSTER_NAME, token + TOKEN_MAX_LIFETIME_POSTFIX);
       if (maxLifetimeStr != null) {
         result = Long.parseLong(new String(maxLifetimeStr));
       }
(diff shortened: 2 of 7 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -245,6 +245,9 @@
   private static final String CLOUDERA_MANAGER_ADVANCED_SERVICE_DISCOVERY_CONF_MONITOR_INTERVAL = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.advanced.service.discovery.config.monitor.interval";
   private static final long DEFAULT_CLOUDERA_MANAGER_ADVANCED_SERVICE_DISCOVERY_CONF_MONITOR_INTERVAL = 30000L;
 
+  private static final String KNOX_TOKEN_EVICTION_INTERVAL = GATEWAY_CONFIG_FILE_PREFIX + "knox.token.eviction.interval";
```
risdenk: missing dot on before `knox`. all other configs have `GATEWAY_CONFIG_FILE_PREFIX + ".abcdef"`

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -126,9 +130,9 @@ protected void removeToken(final String token) {
 
     try {
       aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME, token);
-      aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME,token + "--max");
+      aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME,token + "TOKEN_MAX_LIFETIME_POSTFIX");
     } catch (AliasServiceException e) {
-      log.failedToUpdateTokenExpiration(e);
+      log.errorAccessingTokenState(e);
```
risdenk: Why did this log message change?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -43,9 +48,19 @@
 
   private final Map<String, Long> maxTokenLifetimes = new HashMap<>();
 
+  /* token eviction interval in seconds, default is 5 mins */
+  private long tokenEvictionInterval;
+
+  private final ScheduledExecutorService evictionScheduler = Executors.newScheduledThreadPool(1);
```
risdenk: Where do we shut this down?

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
moresandeep: good catch, I'll fix it.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
moresandeep: This was an error in my previous commit that @smolnar82 pointed after the commit was merged.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: Yes, this needs to be shutdown in stop method(), thanks for catching it.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -43,9 +48,19 @@
...
+
 
   @Override
   public void init(final GatewayConfig config, final Map<String, String> options) throws ServiceLifecycleException {
+    tokenEvictionInterval = config.getKnoxTokenEvictionInterval();
+    if (tokenEvictionInterval > 0) {
```
moresandeep: This probably should go in start() method.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -126,9 +130,9 @@ protected void removeToken(final String token) {
 
     try {
       aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME, token);
-      aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME,token + "--max");
+      aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME,token + "TOKEN_MAX_LIFETIME_POSTFIX");
```
risdenk: This shouldn't be a string - it should be the variable reference?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -147,4 +151,18 @@ protected void updateExpiration(final String token, long expiration) {
...
     }
   }
+
+  @Override
+  protected List<String> getTokens() {
+    List<String> allAliases = new ArrayList();
+    try {
+      allAliases = aliasService.getAliasesForCluster(AliasService.NO_CLUSTER_NAME);
+      /* only get the aliases that represent tokens and extract the current list of tokens */
+      allAliases.stream().filter(a -> a.contains("TOKEN_MAX_LIFETIME_POSTFIX")).map(a -> a.substring(0, a.indexOf("TOKEN_MAX_LIFETIME_POSTFIX")))
```
risdenk: This shouldn't be a string - it should be the variable reference?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
moresandeep: I don't follow, this is a variable reference, I took out the string and created a variable "TOKEN_MAX_LIFETIME_POSTFIX".

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
moresandeep: I don't follow, this is a variable reference, I took out the string and created a variable "TOKEN_MAX_LIFETIME_POSTFIX".

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
risdenk: Yea you have `TOKEN_MAX_LIFETIME_POSTFIX` as a string with `"TOKEN_MAX_LIFETIME_POSTFIX"`. Instead of just using `TOKEN_MAX_LIFETIME_POSTFIX`.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
risdenk: Yea you have TOKEN_MAX_LIFETIME_POSTFIX as a string with "TOKEN_MAX_LIFETIME_POSTFIX". Instead of just using TOKEN_MAX_LIFETIME_POSTFIX.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
risdenk: Yea agreed should go in start.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
moresandeep: <facepalm> and I was wondering my *--max tokens are not getting deleted. Thanks @risdenk !

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -264,4 +280,29 @@ protected String getTokenDisplayText(final String token) {
     return String.format(Locale.ROOT, "%s...%s", token.substring(0, 10), token.substring(token.length() - 3));
   }
 
+  /**
+   * Method that deletes expired tokens based on the token timestamp.
+   */
+  protected void evictExpiredTokens() {
+    try {
+      for (final String token : getTokens()) {
+        if (isExpired(token)) {
```
pzampino: I've been wondering if we should allow some "grace period" beyond the expiration if renewal is still possible for a given token before evicting it.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -264,4 +280,29 @@ protected String getTokenDisplayText(final String token) {
...
+          log.evictToken(getTokenDisplayText(token));
+          removeToken(token);
+        }
+      }
+    } catch(final Exception e) {
```
pzampino: Would it not be better to catch the error for each eviction attempt, and continue trying the others when one fails, rather than failing the whole attempt because of a single error?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: Hmm, wouldn't that make the token expiration time a bit ambiguous? are you suggesting that renew token should be possible until (expiry + x) seconds?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: Sure, that can be done.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: Yes, as we've discussed, while renewers likely will request renewals prior to expiration, they are not currently required to do so. Delaying the eviction avoids potential "unknown token" issues for renewers, and does not affect the expiration time.

moresandeep: Thanks for explaining Phil, fixed it in new PR.

moresandeep:
@pzampino Addressed your review comments with the new PR.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -278,18 +281,34 @@ protected String getTimestampDisplay(long timestamp) {
...
 
+  /**
+   * Method that checks if an expired token is ready to be evicted
+   * by adding configured grace period to the expiry time.
+   * @param token
+   * @return
+   */
+  protected boolean needsEviction(final String token) {
+    boolean needsEviction;
+    needsEviction = isUnknown(token); // Check if the token exist
```
pzampino: If the token is unknown in this case, you're going to get an IllegalArgumentException when the subsequent removal is attempted because this statement says an unknown token -> needsEviction. I would argue that an unknown token means it has probably already been evicted.
The following call to getTokenExpiration will already invoked validateToken(), so the up-front isUnknownToken check here is unnecessary (IMO) and the conclusion is incorrect.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -51,13 +51,16 @@
 
   /* token eviction interval in seconds, default is 5 mins */
   private long tokenEvictionInterval;
+  /* grace period (in seconds) after which an expired token should be evicted, default 5 mins */
```
pzampino: nit: This comment won't change when the default value changes in the future. I try not to include values in comments

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -278,18 +281,34 @@ protected String getTimestampDisplay(long timestamp) {
...
+   * by adding configured grace period to the expiry time.
+   * @param token
+   * @return
+   */
+  protected boolean needsEviction(final String token) {
+    boolean needsEviction;
+    needsEviction = isUnknown(token); // Check if the token exist
+    if (!needsEviction) {
+      /* If it not unknown, check if it is expired and within grace period before evicting */
+      needsEviction = (getTokenExpiration(token) + TimeUnit.SECONDS.toMillis(tokenEvictionGracePeriod)) <= System.currentTimeMillis();
```
pzampino: I think this whole method could be simply this single statement:
return ((getTokenExpiration(token) + TimeUnit.SECONDS.toMillis(tokenEvictionGracePeriod)) <= System.currentTimeMillis());

moresandeep:
@pzampino addressed the review comments

PR #260, opened 2020-02-10
KNOX-2228 - JWTFilter should handle unknown token exception from toke…

pzampino:
…n state service

Modified handling of "Unknown token" exception in JWTFilters, such that it no longer results in a HTTP 500 response. Also, improved some token-related log messages.

## How was this patch tested?

Augmented CommonJWTFilterTest.

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
@@ -41,7 +41,10 @@ public interface JWTMessages {
   @Message( level = MessageLevel.INFO, text = "Unable to verify token: {0}" )
   void unableToVerifyToken(@StackTrace( level = MessageLevel.ERROR) Exception e);
 
-  @Message( level = MessageLevel.ERROR, text = "Unable to verify token: {0}" )
+  @Message( level = MessageLevel.WARN, text = "Unable to verify token expiration: {0}" )
+  void unableToVerifyExpiration(@StackTrace( level = MessageLevel.DEBUG) Exception e);
+
+  @Message( level = MessageLevel.ERROR, text = "Unable to issue token: {0}" )
   void unableToIssueToken(@StackTrace( level = MessageLevel.DEBUG) Exception e);
 
   @Message( level = MessageLevel.DEBUG, text = "Sending redirect to: {0}" )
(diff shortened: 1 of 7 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -170,7 +170,13 @@ protected void configureExpectedParameters(FilterConfig filterConfig) {
   protected boolean tokenIsStillValid(JWT jwtToken) {
     Date expires;
     if (tokenStateService != null) {
-      expires = new Date(tokenStateService.getTokenExpiration(jwtToken.toString()));
+      long timestamp = 0;
+      try {
+        timestamp = tokenStateService.getTokenExpiration(jwtToken.toString());
+      } catch (Exception e) {
```
risdenk: Do we not have a specific exception coming from getTokenExpiration? or tokenStateService?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
```diff
@@ -41,6 +41,9 @@
   @Message( level = MessageLevel.INFO, text = "Unable to verify token: {0}" )
   void unableToVerifyToken(@StackTrace( level = MessageLevel.ERROR) Exception e);
 
+  @Message( level = MessageLevel.ERROR, text = "Unable to verify token expiration: {0}" )
```
risdenk: Do we need this at an error level?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
pzampino: I've been debating the need for an UnknownTokenException, and your comment may have just convinced me to implement it.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
pzampino: I went back and forth with this, and could go either way really. Do you think INFO is sufficient?

risdenk: INFO or WARN should be good I think. I just saw the message above about unable to verify being INFO and was like hmmm that seems worse than not being able to verify the expiration.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
pzampino: I would rather change the method signatures in a separate commit, so I've created https://issues.apache.org/jira/browse/KNOX-2230 to address this concern.

PR #262, opened 2020-02-11
KNOX-2231 - Fix KnoxSSO OIDC integration

moresandeep:
Fix broken OIDC KnoxSSO integration

## How was this patch tested?
This patch was locally tested

### DIFF
--- a/gateway-test/pom.xml
+++ b/gateway-test/pom.xml
@@ -172,6 +172,12 @@
             <groupId>io.rest-assured</groupId>
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.activation</groupId>
+                    <artifactId>activation</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.rest-assured</groupId>
--- a/gateway-test/pom.xml
+++ b/gateway-test/pom.xml
@@ -172,12 +172,6 @@
             <groupId>io.rest-assured</groupId>
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
-            <exclusions>
-                <exclusion>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -2068,12 +2068,8 @@
                 <artifactId>pac4j-oidc</artifactId>
                 <version>${pac4j.version}</version>
                 <exclusions>
-                    <exclusion>
-                        <groupId>com.sun.mail</groupId>
-                        <artifactId>javax.mail</artifactId>
-                    </exclusion>
-		    <exclusion>
-		        <groupId>org.mockito</groupId>
+		            <exclusion>
```
risdenk: nit: spacing is off

gateway-test/pom.xml
```diff
@@ -172,6 +172,12 @@
             <groupId>io.rest-assured</groupId>
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
+            <exclusions>
```
risdenk: Why do we need this? This looks like a test only exclusion?

moresandeep: javax.mail has a different version of javax.activation causing builds to breaks.

risdenk: Exclude it at the top level pom then? Where rest-assured is defined?

moresandeep: That can be done but do we want to exclude this dependency for all the projects ? it should be OK here given that this is for test, it should not matter since we do check for unit tests before builds. Let me try that.

pom.xml
moresandeep: Turns out git UI does not handle spaces and tabs well. My IDE inserted tabs which messed up git UI. I'll change it to spaces.

risdenk:
Looks good.

Future note to myself - it might make sense to avoid excluding `javax.activation` and instead include a specific version in `pom.xml`. I don't think it matters here, but just something to keep in mind.

PR #264, opened 2020-02-12
KNOX-2233 - DefaultKeystoreService getCredentialForCluster uses cache…

pzampino:
… without synchronization

Added synchronization to getCredentialForCluster(String, String) since it uses/updates the cache (e.g., checkCache(String, String), addToCache(String, String, String)). At a minimum, this avoids potential repeated unnecessary keystore file accesses for the same alias.

## How was this patch tested?
Performed some manual testing, but since the overall behavior should not change, ran all existing tests as validation.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
@@ -309,25 +309,30 @@ public class DefaultKeystoreService implements KeystoreService, Service {
   @Override
   public char[] getCredentialForCluster(String clusterName, String alias)
       throws KeystoreServiceException {
-    char[] credential = checkCache(clusterName, alias);
-    if (credential == null) {
-      KeyStore ks = getCredentialStoreForCluster(clusterName);
-      if (ks != null) {
-        try {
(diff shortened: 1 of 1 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -309,25 +309,30 @@ public void addCredentialForCluster(String clusterName, String alias, String val
...
-          char[] masterSecret = masterService.getMasterSecret();
-          Key credentialKey = ks.getKey( alias, masterSecret );
-          if (credentialKey != null) {
-            byte[] credentialBytes = credentialKey.getEncoded();
-            String credentialString = new String( credentialBytes, StandardCharsets.UTF_8 );
-            credential = credentialString.toCharArray();
-            addToCache(clusterName, alias, credentialString);
+    char[] credential;
+
+    synchronized (this) {
```
moresandeep: Instead of synchronizing the whole block wouldn't it be better to add something like a ReadWriteLock ?

pzampino: The intention here is to complete the existing synchronization pattern. More drastic changes merit a separate JIRA IMO.

PR #267, opened 2020-02-19
KNOX-2238 CM discovery - Add TLS support to Phoenix auto discovery

stoty:

Add TLS support for Phoenix Query Server for the CM service discovery module

The TLS support was omitted from the original discovery module, as PQS didn't have TLS support yet. 
Now I have re-added the TLS support. The code is a little different from most similar modules, because the ssl_enabled property is not present in older CM releases, and we have to handle that as well.

## How was this patch tested?

* Wrote and run new unit test
* Tested against live Cloudera Manager instance which had the the TLS for PQS support
* Tested against live older Cloudera Manager instance which did not have the TLS for PQS support

In both cases I was able to connect to the auto discovered PQS service.

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/phoenix/PhoenixServiceModelGenerator.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/phoenix/PhoenixServiceModelGenerator.java
@@ -31,7 +31,8 @@ public class PhoenixServiceModelGenerator extends AbstractServiceModelGenerator 
   public static final String SERVICE_TYPE = "PHOENIX";
   public static final String ROLE_TYPE    = "PHOENIX_QUERY_SERVER";
 
-  static final String QUERY_SERVER_PORT = "phoenix_query_server_port";
+  static final String SSL_ENABLED         = "ssl_enabled";
+  static final String QUERY_SERVER_PORT   = "phoenix_query_server_port";
 
   @Override
   public String getService() {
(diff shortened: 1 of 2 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/phoenix/PhoenixServiceModelGenerator.java
```diff
@@ -58,14 +59,24 @@ public ServiceModel generateService(ApiService       service,
                                       ApiServiceConfig serviceConfig,
                                       ApiRole          role,
                                       ApiConfigList    roleConfig) {
+
     String hostname = role.getHostRef().getHostname();
-    // Phoenix Query Server does not support https
+    String sslEnabledRaw = getRoleConfigValue(roleConfig, SSL_ENABLED);
```
risdenk: This can be simplified. No need to store `sslEnabledRaw`. just do something like:

```
if (Boolean.parseBoolean(getRoleConfigValue(roleConfig, SSL_ENABLED))) {
    scheme = "https";
}
```

`Boolean.parseBoolean` will handle null correctly and return false.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/phoenix/PhoenixServiceModelGenerator.java
```diff
@@ -58,14 +59,24 @@ public ServiceModel generateService(ApiService       service,
...
     String scheme = "http";
+
+    if (sslEnabledRaw != null) {
+      if (Boolean.parseBoolean(getRoleConfigValue(roleConfig, SSL_ENABLED))) {
```
risdenk: If you are going to keep `sslEnabledRaw` - then no need to `null` check here and should use `sslEnabledRaw` inside the `parseBoolean`.

PR #269, opened 2020-02-21
KNOX-2239 - Websocket use the configured truststore in gateway-site config file

moresandeep:

In Knox wss support was broken, the jetty implementation was picking up system keystore and not the keystore configured in gateway-site.xml file. This PR fixes it.

## How was this patch tested?
This was tested locally with ws and wss protocols.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/websockets/GatewayWebsocketHandler.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/websockets/GatewayWebsocketHandler.java
@@ -20,11 +20,13 @@ package org.apache.knox.gateway.websockets;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
-import org.apache.knox.gateway.services.ServiceType;
 import org.apache.knox.gateway.services.GatewayServices;
+import org.apache.knox.gateway.services.ServiceType;
 import org.apache.knox.gateway.services.registry.ServiceDefEntry;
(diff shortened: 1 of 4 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/websockets/GatewayWebsocketHandler.java
```diff
@@ -118,14 +120,28 @@ public Object createWebSocket(ServletUpgradeRequest req,
       LOG.debugLog("Generated backend URL for websocket connection: " + backendURL);
 
       /* Upgrade happens here */
-      return new ProxyWebSocketAdapter
-              (URI.create(backendURL), pool, getClientEndpointConfig(req), config);
+      final ClientEndpointConfig clientConfig = getClientEndpointConfig(req);
+      clientConfig.getUserProperties().put("org.apache.knox.gateway.websockets.truststore", getTruststore());
```
smolnar82: nit: You may have wanted to introduce a constant for the new property name

moresandeep: Yup, I'll try to update it.

risdenk:
So we have to get and set the truststore on every single connection? We can't setup the HTTP client or websocket client once and reuse it? Seems a bit weird to me.

moresandeep:
> So we have to get and set the truststore on every single connection? We can't setup the HTTP client or websocket client once and reuse it? Seems a bit weird to me.

The issue is the `javax.net` api does not support adding custom truststore as a result jetty does not support it either. New instance of WebSocketContainer is created on every connection (along with the underlying httpclient) this is the only entry point i could find where we could update truststore.

PR #270, opened 2020-02-24
KNOX-2249 Add Spark 3 History Server definition

vladhlinsky:

Add Spark 3 History Server definition. Spark 3 HS could be run alongside Spark 2 HS, so separate service definition is required.

## How was this patch tested?

Manually and with unit tests.


vladhlinsky:
cc @risdenk

PR #276, opened 2020-03-03
KNOX-2258 - Add new rewrite rule in Livy service for handling redirect requests to /ui

abhi2196:

Added a new rewrite rule in Livy service for handling redirects to /ui endpoint

## How was this patch tested?

Manually tested in a Cluster with Knox and Livy Services.
1. Edited the rewrite/service.xml for Livy service with the changes.
2. Verified that `https://<KNOX_HOST:KNOX_PORT>/gateway/default/livy` gets redirected to `https://<KNOX_HOST:KNOX_PORT>/gateway/default/livy/ui`

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
@@ -45,4 +45,8 @@
     <rewrite template="{$frontend[url]}/yarn/nodemanager/node/containerlogs/{**}?{scheme}?host={$hostmap(host)}?{port}"/>
   </rule>
 
+  <rule dir="OUT" name="LIVYSERVER/livy/outbound/headers/ui" pattern="{scheme}://{host}:{port}/ui">
+    <rewrite template="{$frontend[url]}/livy/ui"/>
+  </rule>
+
 </rules>
--- a/gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
@@ -45,8 +45,8 @@
     <rewrite template="{$frontend[url]}/yarn/nodemanager/node/containerlogs/{**}?{scheme}?host={$hostmap(host)}?{port}"/>
   </rule>
 
-  <rule dir="OUT" name="LIVYSERVER/livy/outbound/headers/ui" pattern="{scheme}://{host}:{port}/ui">
-    <rewrite template="{$frontend[url]}/livy/ui"/>
+  <rule dir="OUT" name="LIVYSERVER/livy/outbound/headers/ui" pattern="{scheme}://{host}:{port}/ui/">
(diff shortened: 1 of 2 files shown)


abhi2196:
cc @risdenk @wypoon

gateway-service-definitions/src/main/resources/services/livy/0.4.0/service.xml
```diff
@@ -24,7 +24,9 @@
     </metadata>
   <routes>
     <route path="/livy/**?**"/>
-    <route path="/livy"/>
+    <route path="/livy">
+        <rewrite apply="LIVYSERVER/livy/outbound/headers/ui" to="response.headers"/>
```
risdenk: I think you want this to apply to the route after this as well. `/livy/`

gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
```diff
@@ -45,4 +45,8 @@
     <rewrite template="{$frontend[url]}/yarn/nodemanager/node/containerlogs/{**}?{scheme}?host={$hostmap(host)}?{port}"/>
   </rule>
 
+  <rule dir="OUT" name="LIVYSERVER/livy/outbound/headers/ui" pattern="{scheme}://{host}:{port}/ui">
+    <rewrite template="{$frontend[url]}/livy/ui"/>
```
risdenk: Probably best to have trailing slash on this path.

gateway-service-definitions/src/main/resources/services/livy/0.4.0/service.xml
abhi2196: Added change to apply to `/livy/` path.

gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
abhi2196: Added required changes.

gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
```diff
@@ -45,4 +45,8 @@
     <rewrite template="{$frontend[url]}/yarn/nodemanager/node/containerlogs/{**}?{scheme}?host={$hostmap(host)}?{port}"/>
   </rule>
 
+  <rule dir="OUT" name="LIVYSERVER/livy/outbound/headers/ui" pattern="{scheme}://{host}:{port}/ui/">
```
risdenk: Hmm I guess I wasn't completely clear and was talking about only the `rewrite template` not the pattern here. Can you double check that looking for `/ui/` as the pattern is correct still? I want to make sure the header we are looking for matches in all the cases.

abhi2196: Yes I verified with these changes that looking for `/ui/` as the pattern is correctly redirecting to livy ui page.

PR #277, opened 2020-03-04
KNOX-2258 - Add filter for Location header

abhi2196:

Follow up for #276 
Added a new filter in Livy rewrite.xml for replacing Location header
Removed rewrite rules from routes path in service.xml

## How was this patch tested?

Manually tested in a Cluster with Knox and Livy Services.

1. Edited the rewrite/service.xml for Livy service with the changes.
2. Verified that `https://<KNOX_HOST:KNOX_PORT>/gateway/default/livy` gets redirected to `https://<KNOX_HOST:KNOX_PORT>/gateway/default/livy/ui`
3. Verified that `https://<KNOX_HOST:KNOX_PORT>/gateway/default/livy/` gets redirected to `https://<KNOX_HOST:KNOX_PORT>/gateway/default/livy/ui`

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
@@ -45,4 +45,8 @@
     <rewrite template="{$frontend[url]}/yarn/nodemanager/node/containerlogs/{**}?{scheme}?host={$hostmap(host)}?{port}"/>
   </rule>
 
+  <rule dir="OUT" name="LIVYSERVER/livy/outbound/headers/ui" pattern="{scheme}://{host}:{port}/ui">
+    <rewrite template="{$frontend[url]}/livy/ui"/>
+  </rule>
+
 </rules>
--- a/gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
@@ -49,4 +49,10 @@
     <rewrite template="{$frontend[url]}/livy/ui/"/>
   </rule>
 
+  <filter name="LIVYSERVER/livy/outbound/headers">
+    <content type="application/x-http-headers">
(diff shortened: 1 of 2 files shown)


gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
```diff
@@ -49,4 +49,10 @@
     <rewrite template="{$frontend[url]}/livy/ui/"/>
   </rule>
 
+  <filter name="LIVYSERVER/livy/outbound/headers">
```
moresandeep: Where is this filter getting called from ? we need a corresponding entry in service.xml
```xml
    <route path="/livy">
        <rewrite apply="LIVYSERVER/livy/outbound/headers" to="response.headers"/>
    </route>
```

gateway-service-definitions/src/main/resources/services/livy/0.4.0/service.xml
```diff
@@ -24,12 +24,8 @@
     </metadata>
   <routes>
     <route path="/livy/**?**"/>
-    <route path="/livy">
```
moresandeep: Like mentioned previously we need an entry to call the filter.

gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
abhi2196: Sure Added the relevant change.
But not sure why it is working in my test cluster without calling this filter explicitly in route path,  are these filters/rules applied by default based on matched pattern ?

gateway-service-definitions/src/main/resources/services/livy/0.4.0/service.xml
abhi2196: Sure Added the relevant change.

gateway-service-definitions/src/main/resources/services/livy/0.4.0/rewrite.xml
moresandeep: They should not be, that is weird.

PR #278, opened 2020-03-04
KNOX-2262 Accessing hbase logs through knox exposes hbase endpoint ur…

stoty:
…l instead of routing through knox

* Add redirect filter for HBase webu /logs location
* Perform Request URL trailing space correction in HBase dispatch
* Remove deprecation from HBase dispatch
* Add HA dispatch for HBase ( identical functionality to non-HA)
* Refer HA dispatch in service.xml
* Refactor common dispatch logic into HBaseRequestUtil.java

## How was this patch tested?

Deployed patched Knox on a live HBase cluster
checked that I can access the directories, and the files in the subdirectories
checked that the links and redirects all go to Knox proxied URLs

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml
@@ -174,6 +174,15 @@
   <rule dir="IN" name="HBASEUI/hbase/inbound/logs/files" pattern="*://*:*/**/hbase/webui/logs/{**}?{host}?{port}">
     <rewrite template="{$serviceScheme[HBASEUI]}://{$hostmap(host)}:{port}/logs/{**}"/>
   </rule>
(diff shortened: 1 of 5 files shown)


risdenk:
Sorry @stoty this definitely fell off my radar after I initially looked at it.

stoty:
Thanks for the review @risdenk, it happens.

Actually I don't think that the decodeUrl() method does anything useful (anymore) either, considering that a lot of folks were running without the old dispatch, since it didn't apply to the HA setup.

However that was not a can of worms I wanted to open, so I left it alone.

risdenk:
Thanks @stoty I made some slight changes locally, but planning to commit this today. 

I created HBaseUIDispatch and HBaseUIHADispatch so we can avoid the deprecated HBaseDispatch. I made the newer UI dispatches not use the decodeUrl since that isn't needed.

stoty:
Thanks for the review and cleanup @risdenk

PR #282, opened 2020-03-09
KNOX-2259 KNOX-2260 KNOX-2261 - Fixed Impala/Kudu/HBase UI context path in service metadata

smolnar82:

Corrected the contexts for Impala/Kudu/HBase UI services and fixed the code which selects the correct service definition when looking for metadata (the version was not considered before).

## How was this patch tested?

Tested manually: added a sample descriptor (with IMPALAUI, KUDUUI, HBASEUI services) pointing to a remote cluster (where valid URLs are found). Knox generated the topology properly and the correct URLs are displayed on the Home page.

Screen Shot 2020-03-09 at 9 57 13 AM

### DIFF
--- a/gateway-service-metadata/src/main/java/org/apache/knox/gateway/service/metadata/ServiceModel.java
+++ b/gateway-service-metadata/src/main/java/org/apache/knox/gateway/service/metadata/ServiceModel.java
@@ -17,6 +17,8 @@
  */
 package org.apache.knox.gateway.service.metadata;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.Locale;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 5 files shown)


gateway-service-metadata/src/main/java/org/apache/knox/gateway/service/metadata/ServiceModel.java
```diff
@@ -101,14 +103,27 @@ public String getContext() {
...
+        final URL backendUrl = new URL(backendUrlString);
+        if (context.indexOf("{{SCHEME}}") > -1) {
+          context = context.replace("{{SCHEME}}", backendUrl.getProtocol());
+        }
+        if (context.indexOf("{{HOST}}") > -1) {
+          context = context.replace("{{HOST}}", backendUrl.getHost());
+        }
+        if (context.indexOf("{{PORT}}") > -1) {
+          context = context.replace("{{PORT}}", String.valueOf(backendUrl.getPort()));
+        }
```
risdenk: Not sure we need all the double checks? Can't we just do `context.replace(...)` and if it doesn't match it won't replace? The strings here are super short. It might even be slower to do a check and then replace anyway.

smolnar82: We definetly need the first if (backend URL may start with unknown protocol, such as dfs, ws, etc..) -> result in an error.
I agree with the small ifs.

smolnar82: Fixed; I'll wait till Travis finishes and merge this one.

smolnar82:
> Looks good.
> Suggestion: It might be useful if you update KNOX-2259 with additional fixes done here like for HBase and Metadata code.

Done (HBase and KUDU has its own JIRA; I referenced this PR for them too)

PR #283, opened 2020-03-09
KNOX-2267 - Ambari/CM discovery - Needs to point to configured truststore

risdenk:

Currently Ambari and CM discovery doesn't use the configured HTTP truststore. This change adds the HTTP truststore to both the Ambari and CM discovery HTTP clients.

## How was this patch tested?

* `mvn -U -T.75C clean verify -Ppackage,release -Dshellcheck`
* Spun up a cluster with CM AutoTLS and checked that Knox didn't get handshake errors talking to CM

### DIFF
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/util/TruststoreSSLContextUtils.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/util/TruststoreSSLContextUtils.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


gateway-spi/src/main/java/org/apache/knox/gateway/util/TruststoreSSLContextUtils.java
```diff
@@ -0,0 +1,55 @@
...
+public class TruststoreSSLContextUtils {
+  private static final GatewaySpiMessages LOGGER = MessagesFactory.get(GatewaySpiMessages.class);
+
+  private TruststoreSSLContextUtils() {
+  }
+
+  public static SSLContext getTruststoreSSLContext(KeystoreService keystoreService) {
+    SSLContext sslContext = null;
+    try {
+      SSLContextBuilder sslContextBuilder = SSLContexts.custom();
```
smolnar82: nit: I'd create this instance only if it's really needed (`keystoreService != null && truststore != null`)

risdenk: hmm yea good point.

PR #284, opened 2020-03-09
KNOX-2266 - Tokens Should Include a Unique Identifier

pzampino:

JWTs issued and validated by Knox now include a unique identifier as a private claim. This is mostly to guarantee token uniqueness, even for multiple requests within the same second.
Further, the TokenStateService has been updated to leverage this unique identifier as the key for handling token state. This identifier is less susceptible to the nuances of some storage mechanisms.

## How was this patch tested?

Mulitple existing tests were modified to accommodate this change while ensuring the maintenance of existing behavior. TokenServiceResourceTest#testConcurrentGetToken was added to validate these changes. I've also done a bit of manual testing.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
@@ -25,13 +25,10 @@ import org.apache.knox.gateway.services.security.token.UnknownTokenException;
 import org.apache.knox.gateway.services.security.token.impl.JWT;
 import org.apache.knox.gateway.services.security.token.impl.JWTToken;
 
-import java.text.ParseException;
 import java.time.Instant;
-import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.OptionalLong;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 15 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -97,48 +94,62 @@ public long getDefaultMaxLifetimeDuration() {
...
-      /* if token permissiveness is enabled we check JWT token expiration when the token state is unknown */
-      if (permissiveValidationEnabled && getJWTTokenExpiration(token).isPresent()) {
-        return getJWTTokenExpiration(token).getAsLong();
-      } else {
+      expiration = getTokenExpiration(TokenUtils.getTokenId(token));
+    } catch (UnknownTokenException e) {
+      if (permissiveValidationEnabled) {
+        String exp = token.getExpires();
+        if (exp != null) {
+          log.permissiveTokenHandling(TokenUtils.getTokenId(token), e.getMessage());
```
moresandeep: e.getMessage() can throw a NPE in case there is no message, e.toString() is safer.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -97,48 +94,62 @@ public long getDefaultMaxLifetimeDuration() {
...
+          expiration = Long.parseLong(exp);
+        }
+      }
+
+      if (expiration == -1) {
```
moresandeep: -1 seems like a valid value for expiration (might not be now but in the future and for different types of tokens) , Azure uses it https://github.com/apache/hadoop/pull/1872/files#diff-c20196161db7a6b5e0f9d390ab924209R397.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -184,33 +195,22 @@ public long renewToken(final String token, long renewInterval) throws UnknownTok
...
-  @Override
-  public boolean isExpired(final String token) throws UnknownTokenException {
-    boolean isExpired;
-    isExpired = isUnknown(token); // Check if the token exist
-    if (!isExpired) {
-      // If it not unknown, check its expiration
-      isExpired = (getTokenExpiration(token) <= System.currentTimeMillis());
-    }
-    return isExpired;
+    return getTokenExpiration(token) <= System.currentTimeMillis();
```
moresandeep: This method would throw `UnknownTokenException` if a token is not found. Exceptions are expensive (filling up stack trace and unwinding it every time it is passed to a new method) so should `getTokenExpiration(token)` return some negative value instead?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -246,9 +249,14 @@ public Response renew(String token) {
...
         try {
+          JWTToken jwt = new JWTToken(token);
           // If renewal fails, it should be an exception
-          expiration = tokenStateService.renewToken(token,
+          expiration = tokenStateService.renewToken(jwt,
                                                     renewInterval.orElse(tokenStateService.getDefaultRenewInterval()));
+          log.renewedToken(getTopologyName(), TokenUtils.getTokenDisplayText(token), TokenUtils.getTokenId(jwt));
+        } catch (ParseException e) {
+          log.invalidToken(getTopologyName(), TokenUtils.getTokenDisplayText(token), e);
+          error = e.getMessage();
```
moresandeep: e.getMessage() will throw NPE in case message is null.

pzampino: Yes, getMessage() may return null. I will add logic to handle that, but I don't believe replacing it with toString() is equivalent.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: Since I think the UnknownTokenException scenario should be exceptional, I think it's ok as it is. It's not intended to be a regularly occurring logic path. Does that make sense?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: In terms of milliseconds since the epoch, I don't know what -1 could mean as a valid expiration time. I'm interested to understand the Azure treatment though.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: I don't agree that toString() is a replacement for getMessage(), but I will account for the possibility of a null result from getMessage().

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: If that code means (-1 = now), that seems quite strange to me, and I don't think Knox tokens require such a convention.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: Yup, it does.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
moresandeep: Sometimes -1 is treated as a special value, like unlimited expiry. I do agree that in context of Knox token (which we control) -1 = expired token

PR #289, opened 2020-03-15
KNOX-2295 - UI services improvement on Knox Home page

smolnar82:

Removed the data table when displaying UI services and using the services' logos with Angular's material grid. Long descriptions are shown by hovering over the short descriptions in the footer of each tile.
Services are available when you click on their logos.

## How was this patch tested?

Tested manually; see these screenshots:
Screen Shot 2020-03-15 at 11 05 03 PM

Screen Shot 2020-03-15 at 11 12 25 PM


smolnar82:
One comment: the Hue logo you are seeing in the above screenshot is replaced with one with white background.

smolnar82:
Thanks, @moresandeep !
Trying to answer your comments:
1. I'll try to find a better CM logo
2. If there is no logo -> a JIRA should be submitted. I tried to add logos for all the actively used services.
3. Hue logo is replaced already (see my comment above)

knox-homepage-ui/package.json
```diff
@@ -10,11 +10,13 @@
   },
   "private": true,
   "dependencies": {
+    "@angular/cdk": "^5.2.5",
```
risdenk: All the other angular versions is 5.2.0 but this is 5.2.5?

smolnar82: I haven't tested with `5.2.0` as only `5.2.5` was listed as an officially supported version of Angular's material lib:

Screen Shot 2020-03-16 at 3 57 49 PM

smolnar82:
Regarding @risdenk 's comment on licensing: all service logos are logos of Apache projects except for the CM-[UI|API]. I understand his concerns about it and it's a valid question if we want to copy it here.

smolnar82:
To address the logo stuff I figured out the following approach should work: I'll add a `general_logo.png` (with the text of 'Service Logo comes here') and display it for
- any service that does not have its own logo uploaded by this patch (anyone is free to upload their owns logos going forward as a separate patch)
- CM API/UI

smolnar82:
This is how it looks like now:
Screen Shot 2020-03-16 at 5 09 48 PM

risdenk:
@smolnar82  instead of "service logo comes here" can it instead be the text of the service? It would just be plain text on say a white background. That would probably look better.

smolnar82:
> @smolnar82 instead of "service logo comes here" can it instead be the text of the service? It would just be plain text on say a white background. That would probably look better.

So, we either use a static image (as of now; I'm open to change the text though) or try to find out if the referenced image exists and act differently (e.g. have a text on white background).

I don't think the general logo will appear as much; once someone realizes his logo is not displayed I'd imagine that he uploads one quickly -> the 2nd approach just not worth implementing.

smolnar82:
Newest version: 
Screen Shot 2020-03-16 at 6 16 24 PM

PR #297, opened 2020-03-20
KNOX-2301 and KNOX-2302

pzampino:

These are two closely-related issues. Added (actually, removed restrictions on) performing re-discovery for every applicable descriptor on gateway start-up. To prevent unnecessary redeployments of unchanged topologies, also added a check to skip the persistence of generated topologies if there exists an equivalent deployed topology of the same name.

## How was this patch tested?

A great deal of manual testing was performed, and a number of existing tests were modified.  Also added some new tests.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
@@ -556,6 +556,11 @@ public class DefaultTopologyService extends FileAlterationListenerAdaptor implem
         log.remoteConfigurationMonitorStartFailure(remoteMonitor.getClass().getTypeName(), e.getLocalizedMessage());
       }
     }
+
+    // Trigger descriptor discovery (KNOX-2301)
+    for (File descriptor : getDescriptors()) {
+      descriptor.setLastModified(System.currentTimeMillis()); // 'Touch' the descriptor
+    }
   }
 
   @Override
@@ -606,7 +611,6 @@ public class DefaultTopologyService extends FileAlterationListenerAdaptor implem
 
   @Override
   public void start() {
-
   }
 
   @Override
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
@@ -17,8 +17,6 @@
  */
 package org.apache.knox.gateway.services.topology.impl;
 
-import org.apache.commons.digester3.Digester;
-import org.apache.commons.digester3.binder.DigesterLoader;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.io.filefilter.TrueFileFilter;
@@ -54,7 +52,6 @@ import org.apache.knox.gateway.topology.TopologyListener;
 import org.apache.knox.gateway.topology.TopologyMonitor;
 import org.apache.knox.gateway.topology.TopologyProvider;
 import org.apache.knox.gateway.topology.Version;
-import org.apache.knox.gateway.topology.builder.TopologyBuilder;
 import org.apache.knox.gateway.topology.discovery.ClusterConfigurationMonitor;
 import org.apache.knox.gateway.topology.discovery.ServiceDiscovery;
 import org.apache.knox.gateway.topology.monitor.RemoteConfigurationMonitor;
@@ -62,9 +59,8 @@ import org.apache.knox.gateway.topology.monitor.RemoteConfigurationMonitorFactor
 import org.apache.knox.gateway.topology.simple.SimpleDescriptor;
 import org.apache.knox.gateway.topology.simple.SimpleDescriptorFactory;
 import org.apache.knox.gateway.topology.validation.TopologyValidator;
-import org.apache.knox.gateway.topology.xml.AmbariFormatXmlTopologyRules;
-import org.apache.knox.gateway.topology.xml.KnoxFormatXmlTopologyRules;
 import org.apache.knox.gateway.util.ServiceDefinitionsLoader;
+import org.apache.knox.gateway.util.TopologyUtils;
 import org.eclipse.persistence.jaxb.JAXBContextProperties;
 import org.xml.sax.SAXException;
 
@@ -74,6 +70,7 @@ import javax.xml.bind.Marshaller;
 import java.io.File;
 import java.io.FileFilter;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -87,8 +84,6 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
-import static org.apache.commons.digester3.binder.DigesterLoader.newLoader;
-
 public class DefaultTopologyService extends FileAlterationListenerAdaptor implements TopologyService, TopologyMonitor,
     TopologyProvider, FileFilter, FileAlterationListener, ServiceDefinitionChangeListener {
 
@@ -101,8 +96,6 @@ public class DefaultTopologyService extends FileAlterationListenerAdaptor implem
   public static final List<String> SUPPORTED_TOPOLOGY_FILE_EXTENSIONS = Collections.unmodifiableList(Arrays.asList("xml", "conf"));
 
   private static GatewayMessages log = MessagesFactory.get(GatewayMessages.class);
-  private static DigesterLoader digesterLoader = newLoader(new KnoxFormatXmlTopologyRules(),
-      new AmbariFormatXmlTopologyRules());
   private Map<String, FileAlterationMonitor> monitors = new ConcurrentHashMap<>();
   private File topologiesDirectory;
   private File sharedProvidersDirectory;
@@ -153,17 +146,21 @@ public class DefaultTopologyService extends FileAlterationListenerAdaptor implem
     return topology;
   }
 
+  @Override
+  public Topology parse(final InputStream content) throws IOException, SAXException {
+    return TopologyUtils.parse(content);
+  }
+
   private Topology loadTopologyAttempt(File file) throws IOException, SAXException {
     Topology topology;
-    Digester digester = digesterLoader.newDigester();
-    TopologyBuilder topologyBuilder = digester.parse(FileUtils.openInputStream(file));
-    if (null == topologyBuilder) {
-      return null;
+    try (InputStream in = FileUtils.openInputStream(file)) {
+      topology = parse(in);
+      if (topology != null) {
+        topology.setUri(file.toURI());
+        topology.setName(FilenameUtils.removeExtension(file.getName()));
+        topology.setTimestamp(file.lastModified());
+      }
     }
-    topology = topologyBuilder.build();
-    topology.setUri(file.toURI());
-    topology.setName(FilenameUtils.removeExtension(file.getName()));
-    topology.setTimestamp(file.lastModified());
     return topology;
   }
 
@@ -558,8 +555,12 @@ public class DefaultTopologyService extends FileAlterationListenerAdaptor implem
     }
 
     // Trigger descriptor discovery (KNOX-2301)
+    triggerDescriptorDiscovery();
+  }
+
+  private void triggerDescriptorDiscovery() {
     for (File descriptor : getDescriptors()) {
-      descriptor.setLastModified(System.currentTimeMillis()); // 'Touch' the descriptor
+      descriptorsMonitor.onFileChange(descriptor);
     }
   }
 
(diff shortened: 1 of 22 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
```diff
@@ -153,17 +146,18 @@ private Topology loadTopology(File file) throws IOException, SAXException, Inter
...
-    Digester digester = digesterLoader.newDigester();
-    TopologyBuilder topologyBuilder = digester.parse(FileUtils.openInputStream(file));
-    if (null == topologyBuilder) {
-      return null;
-    }
-    topology = topologyBuilder.build();
-    topology.setUri(file.toURI());
-    topology.setName(FilenameUtils.removeExtension(file.getName()));
-    topology.setTimestamp(file.lastModified());
+    Topology topology = parse(FileUtils.openInputStream(file));
```
risdenk: I think it would be cleaner to open/close input stream in the parse method. Basically pass a file reference and not an input stream? 

Otherwise I think we need to make sure to close the input stream?

gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
```diff
@@ -153,17 +146,18 @@ private Topology loadTopology(File file) throws IOException, SAXException, Inter
...
+    if (topology != null) {
+      topology.setUri(file.toURI());
+      topology.setName(FilenameUtils.removeExtension(file.getName()));
+      topology.setTimestamp(file.lastModified());
+    }
```
risdenk: If we pass in the file reference, you can move this logic in there too and get the full topology object back.

gateway-server/src/test/java/org/apache/knox/gateway/services/topology/DefaultTopologyServiceTest.java
```diff
@@ -177,6 +181,21 @@ public void testGetTopologies() throws Exception {
...
+   *
+   * @param gws A GatewayServices object, or null.
+   */
+  private void setGatewayServices(final GatewayServices gws) {
+    try {
+      Field gwsField = GatewayServer.class.getDeclaredField("services");
+      gwsField.setAccessible(true);
+      gwsField.set(null, gws);
+    } catch (Exception e) {
+      e.printStackTrace();
```
risdenk: Not sure you need the catch here? Since it's a test probably better to throw it?

gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/TopologyComparisonUtil.java
```diff
@@ -0,0 +1,192 @@
...
+import org.apache.knox.gateway.topology.Topology;
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Utility for comparing Topology objects for equivalence.
+ */
+class TopologyComparisonUtil {
```
risdenk: Is this class necessary? Can't we just ask if the two topology objects are equal? Most of this logic looks like it should be in an topology equals method.

gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
pzampino: That makes sense.

gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/TopologyComparisonUtil.java
smolnar82: Assuming the JSON representation of `topology1` and `topology2` are the same in case they are equal you can simply do a test like:
`org.apache.knox.gateway.util.JsonUtils.renderAsJsonString(topology1).equals(JsonUtils.renderAsJsonString(topology2)`

gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
pzampino: The reason for the TopologyService method taking an InputStream (rather than a File) is that there is not a File available in every case. In one case, it is being used to parse a generated String. There could be two signatures, but I think the InputStream signature supports the two existing cases well enough. The streams must be closed though.

gateway-server/src/test/java/org/apache/knox/gateway/services/topology/DefaultTopologyServiceTest.java
pzampino: +1

gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/TopologyComparisonUtil.java
pzampino: This thought had crossed my mind as I was implementing it; I think the fact that this requirement doesn't care about the URI, default service path or timestamp is what made me opt for this. I think a complete equals() implementation would include at least the URI, but if others disagree, I could be persuaded to move this logic into the Topology equals() method.

pzampino: I'm already converting JSON to a Topology object for this purpose, and would prefer to avoid re-converting them back to JSON for the sake of comparison.

pzampino: Moved to Topology#equals() method.

gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
```diff
@@ -556,6 +553,11 @@ public void startMonitor() throws Exception {
         log.remoteConfigurationMonitorStartFailure(remoteMonitor.getClass().getTypeName(), e.getLocalizedMessage());
       }
     }
+
+    // Trigger descriptor discovery (KNOX-2301)
+    for (File descriptor : getDescriptors()) {
+      descriptor.setLastModified(System.currentTimeMillis()); // 'Touch' the descriptor
+    }
```
smolnar82: I'd move this out to a new method instead of keeping it in `startMonitor` with a good method name. Otherwise, in the future, it'd confuse others why it's here.

gateway-server/src/main/java/org/apache/knox/gateway/util/TopologyUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.StringReader;
+
+import static org.apache.commons.digester3.binder.DigesterLoader.newLoader;
+
+public class TopologyUtils {
+
+  private static DigesterLoader digesterLoader = newLoader(new KnoxFormatXmlTopologyRules(),
```
smolnar82: nit: can be `final`

gateway-server/src/test/java/org/apache/knox/gateway/services/topology/DefaultTopologyServiceTest.java
```diff
@@ -177,6 +181,17 @@ public void testGetTopologies() throws Exception {
...
+   * Set the static GatewayServices field to the specified value.
+   *
+   * @param gws A GatewayServices object, or null.
+   */
+  private void setGatewayServices(final GatewayServices gws) throws Exception {
+    Field gwsField = GatewayServer.class.getDeclaredField("services");
+    gwsField.setAccessible(true);
+    gwsField.set(null, gws);
+  }
+
```
smolnar82: If I understand well this is implemented because EasyMock does not support static method mocking. I'm ok with this change, however, you may consider using [PowerMock](https://github.com/powermock/powermock)

gateway-spi/src/main/java/org/apache/knox/gateway/topology/Provider.java
```diff
@@ -82,4 +84,41 @@ public void setRole( String role ) {
...
+  public int hashCode() {
+    return new HashCodeBuilder().append(name)
+                                .append(role)
+                                .append(params)
+                                .append(enabled)
+                                .build();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
smolnar82: I really like you are using `HashCodeBuilder` above. Why not use `EqualsBuilder` here (this would require to implement `hashCode` and `equals` in `Provider` too)?

gateway-spi/src/main/java/org/apache/knox/gateway/topology/Service.java
```diff
@@ -122,32 +122,61 @@ public boolean equals(Object object) {
...
+      if (!param.getValue().equals(that.params.get(param.getKey()))) {
+        return false;
+      }
+    }
+
     return true;
   }
 
   @Override
   public int hashCode() {
```
smolnar82: Adding `HashCodeBuilder` and `EqualsBuilder` here too?

gateway-spi/src/main/java/org/apache/knox/gateway/topology/Topology.java
```diff
@@ -153,4 +161,106 @@ public void addProvider( Provider provider ) {
...
+  public int hashCode() {
+    return (new HashCodeBuilder(17, 31)).append(name)
+                                        .append(providerList)
+                                        .append(services)
+                                        .append(applications)
+                                        .build();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
smolnar82: If `Provider`, `Service` and `Application` implements `hashCode` and `equals` properly you do not need the private methods below and you can simply use `EqualsBuilder` here too.

gateway-test/src/test/java/org/apache/knox/gateway/AmbariServiceDefinitionTest.java
```diff
@@ -114,6 +114,12 @@ public static void setupGateway() throws Exception {
     File deployDir = new File( config.getGatewayDeploymentDir() );
     deployDir.mkdirs();
 
+    File descDir = new File( config.getGatewayDescriptorsDir() );
```
smolnar82: This is out-of-scope of this JIRA so I filed a new one: https://issues.apache.org/jira/browse/KNOX-2309

gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
```diff
@@ -571,13 +579,49 @@ private static File resolveProviderConfigurationReference(final String reference
...
+            if (topologyDescriptor.exists()) {
+                // If it does, only overwrite it if it has changed (KNOX-2302)
+                // Compare the generated topology with the in-memory topology
+                Topology existing = null;
+                TopologyService topologyService = gwServices.getService(ServiceType.TOPOLOGY_SERVICE);
+                for (Topology t : topologyService.getTopologies()) {
+                    if (topologyFilename.equals(t.getName())) {
+                        existing = t;
+                        break;
+                    }
```
smolnar82: Maybe introducing `topologyService.getTopology(String topologyName)`?

gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
```diff
@@ -571,13 +579,49 @@ private static File resolveProviderConfigurationReference(final String reference
...
 
-            try (OutputStream outputStream = Files.newOutputStream(topologyDescriptor.toPath());
-                 OutputStreamWriter outputStreamWriter = new OutputStreamWriter(outputStream, StandardCharsets.UTF_8);
-                 BufferedWriter fw = new BufferedWriter(outputStreamWriter)) {
-              fw.write(sw.toString());
-              fw.flush();
+            // KNOX-2302
+            boolean persistGenerated = false;
+            // Determine if the topology already exists.
+            if (topologyDescriptor.exists()) {
```
smolnar82: I'd move this new logic to its own private method with a meaningful name (e.g. `shouldRedeployTopology`) to make the code cleaner.

gateway-server/src/test/java/org/apache/knox/gateway/services/topology/DefaultTopologyServiceTest.java
pzampino: Your understanding is correct.

gateway-spi/src/main/java/org/apache/knox/gateway/topology/Topology.java
pzampino: The reason for not using EqualsBuilder here is that equals() for Lists requires the ordering to be the same. However, for the purpose of evaluating topologies for equality, such ordering is not important.

pzampino: Not using EqualsBuilder here because it does not handle potentially-null fields well.

gateway-test/src/test/java/org/apache/knox/gateway/AmbariServiceDefinitionTest.java
pzampino: I had to make these changes to get the tests to pass. It is a side-effect requirement of the changes in this PR.

gateway-spi/src/main/java/org/apache/knox/gateway/topology/Provider.java
pzampino: +1

gateway-spi/src/main/java/org/apache/knox/gateway/topology/Service.java
pzampino: +1

gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
pzampino: I feel this is out of scope. It is not required for this PR.

gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
pzampino: +1

gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
pzampino: This is the reason for the comment, but pulling it out into a separate explicitly-named method is more clear and resilient to future confusion from the comment not being updated.

PR #296, opened 2020-03-19
KNOX-2240 - KnoxShell Custom Command for WEBHDFS Use

lmccay:
(It is very **important** that you created an Apache Knox JIRA for this change and that the PR title/commit message includes the Apache Knox JIRA ID!)

## What changes were proposed in this pull request?

This PR introduces a new custom groovy command for KnoxShell for interaction with hadoop filesystems through the WebHDFS API. This includes the ability to mount filesystems, unmount them and list mounts. Mounts are persisted in the user's home directory for knoxshell so that they live across KnoxShell sessions.

It is limited to the hdfs client classes that are provided in KnoxShell today but with follow up patches to both the client classes and the command will cover more of the Hadoop filesystem capabilities - such as chown and chmod.

## How was this patch tested?

Primarily tested through manual use of the KnoxShell environment. Minor improvements in CSV builder have added unit tests as well.

`knox:000> **:fs mounts**
===> +----------------+------------------------------------------------------------------------+
|  Mount Point   |                              Topology URL                              |
+----------------+------------------------------------------------------------------------+
|     sandor     |  https://knox-host1:8443/gateway/cdp-proxy-api  |
+----------------+------------------------------------------------------------------------+

knox:000> **:fs mount https://knox-host1:8443/gateway/cdp-proxy-api sando**r
===> https://knox-host1:8443/gateway/cdp-proxy-api mounted as sandor
knox:000> 
knox:000> **:fs ls /sandor/tmp**
===> /tmp
+--------------+----------+--------------+----------+--------------------------------+--------------------------------------------+
|  permission  |  owner   |    group     |  length  |            modtime             |                    name                    |
+--------------+----------+--------------+----------+--------------------------------+--------------------------------------------+
|      0       |   hdfs   |  supergroup  |    0     |  Thu Mar 19 20:39:30 EDT 2020  |  .cloudera_health_monitoring_canary_files  |
|     733      |   hive   |  supergroup  |    0     |  Mon Mar 16 12:09:18 EDT 2020  |                    hive                    |
|     1777     |  mapred  |    hadoop    |    0     |  Mon Mar 16 12:09:30 EDT 2020  |                    logs                    |
+--------------+----------+--------------+----------+--------------------------------+--------------------------------------------+

knox:000> **:fs ls /sandor/user/knoxui/**
===> /user/knoxui/
+--------------+----------+----------+----------+--------------------------------+------------------------------------+
|  permission  |  owner   |  group   |  length  |            modtime             |                name                |
[...]

PR #299, opened 2020-03-21
KNOX-2307 - CSVKnoxShellTableBuilder must support quoted strings and …

lmccay:
…embedded commas

Change-Id: I3d2e0dc8144e79c73df2067a7b6320e5d0a84266

Embedded commas within a quoted string/col in CSV files result in separate cols currently. This patch allows for them to be ignored during the split() call via regex for identifying such patterns.

## How was this patch tested?

New unit test added and existing unit tests run.
Manually tested with dataset with such embedded commas in quoted strings.


lmccay:
Fixed to not rely on changes from PR #296

PR #307, opened 2020-04-07
KNOX-2304 - CM discovery cluster config monitor needs to be aware of …

lmccay:
…all relevant CM event types

Change-Id: Ic26ad36fcb110e01d30d636f14d5b383de01ff17

This patch adds support for CM Event commands of type Start in addition to the existing Restart.
This is to account for services that are added and started but not restarted and discovery needing to be aware of these in case the service in question has been enabled in a descriptor.

Updated existing unit test to include Starts and ran all existing unit tests.

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
@@ -72,15 +72,15 @@ public class PollingConfigurationAnalyzer implements Runnable {
 
   private static final String START_COMMAND = "Start";
 
-  // The format of the filter employed when restart events are queried from ClouderaManager
-  private static final String RESTART_EVENTS_QUERY_FORMAT =
+  // The format of the filter employed when start events are queried from ClouderaManager
+  private static final String EVENTS_QUERY_FORMAT =
                                 "category==" + ApiEventCategory.AUDIT_EVENT.getValue() +
                                 ";attributes.cluster==\"%s\"%s";
 
-  // The format of the timestamp element of the restart events query filter
+  // The format of the timestamp element of the start events query filter
   private static final String EVENTS_QUERY_TIMESTAMP_FORMAT = ";timeOccurred=gt=%s";
 
-  // The default amount of time before "now" to check for restart events the first time
+  // The default amount of time before "now" to check for start events the first time
   private static final long DEFAULT_EVENT_QUERY_DEFAULT_TIMESTAMP_OFFSET = (60 * 60 * 1000); // one hour
 
   private static final int DEFAULT_POLLING_INTERVAL = 60;
@@ -110,10 +110,10 @@ public class PollingConfigurationAnalyzer implements Runnable {
   // Cache of ClouderaManager API clients, keyed by discovery address
   private final Map<String, DiscoveryApiClient> clients = new ConcurrentHashMap<>();
 
-  // Timestamp records of the most recent restart event query per discovery address
+  // Timestamp records of the most recent start event query per discovery address
   private Map<String, String> eventQueryTimestamps = new ConcurrentHashMap<>();
 
-  // The amount of time before "now" to will check for restart events the first time
+  // The amount of time before "now" to will check for start events the first time
   private long eventQueryDefaultTimestampOffset = DEFAULT_EVENT_QUERY_DEFAULT_TIMESTAMP_OFFSET;
 
   private boolean isActive;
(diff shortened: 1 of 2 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
```diff
@@ -365,12 +375,32 @@ private DiscoveryApiClient getApiClient(final ServiceDiscoveryConfig discoveryCo
                                                clusterName,
                                                lastTimestamp);
     for (ApiEvent event : events) {
-      restartEvents.add(new RestartEvent(event));
+      List<ApiEventAttribute> attributes = event.getAttributes();
+      Map<String,Object> map = getAttributeMap(attributes);
```
risdenk, line 379: Why not loop through the ApiEventAttribute list directly? There doesn't seem to be any value in converting to a Map. You break the map apart in the same way you would have to interrogate the ApiEventAttribute anyway.

lmccay: That is what I had before and it is much more complicated as I wanted to avoid creating a map. Note that we are checking two correlated attributes. The command issued and its status. Since you don't know what order they are in you have to squirrel them away then check whether you have both then add. If we end up having to add more events that would be even messier to do. Now, we can get the map upfront and do whatever processing of it afterward much more cleanly.

risdenk: Hmm I guess I am now more concerned with the logic here potentially then. I'll comment on the specific section below.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
```diff
@@ -365,12 +375,32 @@ private DiscoveryApiClient getApiClient(final ServiceDiscoveryConfig discoveryCo
...
+    status = (String) ((List<String>) map.get(COMMAND_STATUS)).get(0);
+    if (START_COMMAND.equals(command) || RESTART_COMMAND.equals(command) &&
+        SUCCEEDED_STATUS.equals(status) || STARTED_STATUS.equals(status)) {
+      restartEvents.add(new RestartEvent(event));
+    }
+  }
+
+  private Map<String, Object> getAttributeMap(List<ApiEventAttribute> attributes) {
+    Map<String,Object> map = new HashMap<>();
+    attributes.forEach(attr -> { map.put(attr.getName(), attr.getValues());});
```
risdenk, line 400: This will overwrite in the map correct? The attribute get name - is that unique? I think this is by COMMAND or COMMAND_STATUS based on the logic for eventually checking the map. If there are duplicate commands won't they get overwritten in the map? So its last write wins?

If you do a Start of SERVICEA and then Start of SERVICEB - will you see both events or just collapse it down to one?

Another example is doing start of same service twice with different results. start success then (implicit stop) then start failure. Will that get picked up?

I think the map is totally breaking the uniqueness and ordering of the events.

pzampino: The map is associated with a specific ApiEvent, so I'm not sure that the map is the concern. We may want to sort the ApiEvent collection by _timeOccurred_ to avoid issues associated with what you're describing.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
```diff
@@ -365,12 +375,32 @@ private DiscoveryApiClient getApiClient(final ServiceDiscoveryConfig discoveryCo
...
+      List<ApiEventAttribute> attributes = event.getAttributes();
+      Map<String,Object> map = getAttributeMap(attributes);
+      addIfRelevantEvent(restartEvents, event, map);
     }
 
     return restartEvents;
   }
 
+  @SuppressWarnings("unchecked")
+  private void addIfRelevantEvent(List<RestartEvent> restartEvents, ApiEvent event, Map<String, Object> map) {
```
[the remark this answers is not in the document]
pzampino: Since the map is derived from the ApiEvent, addIfRelevantEvent should not need the attribute Map passed in as a param.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
lmccay: Yes, the map is redone for each event purposely. It is only a map of the attributes for the given event. The order of the events hasn't changed as a result of this patch and is driven by the REST API response and has nothing to do with the creation of this map. It works as intended.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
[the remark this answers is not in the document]
lmccay: I didn't see any point in changing it to reflect Start and Restart the are both "start" events. We can change to Start which is even a semantically meaningless change. I wanted to keep the map external rather than create it inside that method in case we need to add an additional use of the map later. We can move it in and refactor later if needed.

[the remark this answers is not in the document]
lmccay: Latest revision addresses the "restart" terminology and cleans up the method as you suggested offline. Thanks for the review!

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
lmccay: I think the latest revision will make the above more clear, Kevin. Thanks for the review!

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
```diff
@@ -360,19 +370,43 @@ private DiscoveryApiClient getApiClient(final ServiceDiscoveryConfig discoveryCo
...
+  private boolean isRelevantEvent(ApiEvent event) {
+    boolean rc = false;
+    String command = null;
+    String status = null;
+    List<ApiEventAttribute> attributes = event.getAttributes();
+    Map<String,Object> map = getAttributeMap(attributes);
+    command = (String) ((List<String>) map.get(COMMAND)).get(0);
+    status = (String) ((List<String>) map.get(COMMAND_STATUS)).get(0);
+    if (START_COMMAND.equals(command) || RESTART_COMMAND.equals(command) &&
+        SUCCEEDED_STATUS.equals(status) || STARTED_STATUS.equals(status)) {
```
risdenk, line 396: Does this logic need to be grouped somehow? Right now this is mixing && and ||?

pzampino: I could see this being grouped as
`(START_COMMAND.equals(command) && STARTED_STATUS.equals(status)) || (RESTART_COMMAND.equals(command) && SUCCEEDED_STATUS.equals(status))`

pzampino: IIRC, && does have a higher precedence than ||, which could mean the default grouping is `START_COMMAND.equals(command) || (RESTART_COMMAND.equals(command) &&
        SUCCEEDED_STATUS.equals(status)) || STARTED_STATUS.equals(status)`, which probably still works in this case

pzampino: It also may pick up _any_ event type that has a status value of `STARTED_STATUS`

lmccay: Not at this point. They won't be mixed up between the two and even if they were they both mean success.

PR #314, opened 2020-04-13
KNOX-2321 - HDFS UI rewrite rules should handle LogLevel

lokeshj1703:
https://issues.apache.org/jira/browse/KNOX-2321

Add a rewrite rule for LogLevel in hdfs ui.

## How was this patch tested?

The patch was tested on a cluster.

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/hdfsui/3.0.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/hdfsui/3.0.0/rewrite.xml
@@ -133,6 +133,9 @@
   <rule dir="OUT" name="HDFSUI/hdfs/outbound/logshtml" pattern="/logs">
     <rewrite template="{gateway.url}/hdfs/logs/?host={$inboundurl[host]}"/>
   </rule>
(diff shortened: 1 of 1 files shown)


bharatviswa504:
This link is added to DN and explorer page also, do we need some changes for Dn and explorer  rewrite rules also to make it work?

moresandeep:
> This link is added to DN and explorer page also, do we need some changes for Dn and explorer rewrite rules also to make it work?

We probably, should, @lokeshj1703 thoughts?

lmccay:
Would the other UIs be addressed via global rewrite rules by any chance?

moresandeep:
> Would the other UIs be addressed via global rewrite rules by any chance?

Global rewrite rule is a double edge sword, it might break other UIs unknowingly. I think some code duplication might be tolerable in this case.

lmccay:
Okay - I want to cut an RC in the next day or so - so let's get this resolved.

moresandeep:
Ah right, I'll merge this and we will followup with new JIRA if needed.

lokeshj1703:
@bharatviswa504 @moresandeep I think it should be fine. The rewrite rules are added under section.
`<!-- outbound rewrite rules for links under utilities -->`
I think these rewrite rules apply for datanode as well as explorer html.

PR #325, opened 2020-04-22
KNOX-2354 An HBASEJARS service which can proxy HBase jars hosted by t…

joshelser:
…he HBASEUI

In HBase, we can host a "mini" maven repo in hbase itself which a maven
application can point to.

We want this to be in its own service because it has different semantics
than the normal hbaseUI service has (e.g. one would be hooked up to SSO
login, but SSO won't work against that), and Maven cannot function with
the query-string based backend-routing that we do to make the HBaseUI
look like a normal website (instead of a collection of independent
webservers).

Added a new service HBASEJARS

## How was this patch tested?

* Deployed gateway locally with HBase
* Validated that new IN/OUT rules work to traverse the Jetty
* Validated ability to pull resources via Curl directly
* Validated that I can build a Maven application against this (https://github.com/joshelser/knox-hbasejars-exemplar)

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/hbasejars/2.2.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/hbasejars/2.2.0/rewrite.xml
@@ -1,0 +1,49 @@
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
+   contributor license agreements.  See the NOTICE file distributed with
+   this work for additional information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


joshelser:
Rekicked the checks as it seems to have failed because of a failure to pull jars out of maven central (rate-limiting inside github? who knows)

joshelser:
Any feedback from this one, folks? @lmccay @risdenk @moresandeep

PR #326, opened 2020-04-24
KNOX-2357 - Descriptor handler should not default discovery type to A…

pzampino:
…nd Extended Default Grace Period

Prior to this change, if no discovery properties are present in a descriptor, then Ambari discovery was assumed, which would fail due to the lack of configuration. With this change, discovery attempts are skipped altogether unless at least one discovery configuration property is present in a descriptor. If the discovery-type property is NOT present, but there is at least one other such property, then the type will still be defaulted to Ambari.

## How was this patch tested?

Augmented SimpleDescriptorHandlerTest, and tested manually.

### DIFF
--- a/gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
+++ b/gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
@@ -106,9 +106,14 @@ public class SimpleDescriptorHandler {
         Map<String, Map<String, String>> serviceParams = new HashMap<>();
         Map<String, List<String>> serviceURLs = new HashMap<>();
 
-        ServiceDiscovery.Cluster cluster = performDiscovery(config, desc, gatewayServices);
-        if (cluster == null) {
-            log.failedToDiscoverClusterServices(desc.getName());
+        ServiceDiscovery.Cluster cluster = null;
+        if (shouldPerformDiscovery(desc)) {
+            cluster = performDiscovery(config, desc, gatewayServices);
+            if (cluster == null) {
+                log.failedToDiscoverClusterServices(desc.getName());
+            }
+        } else {
+            log.discoveryNotConfiguredForDescriptor(desc.getName());
         }
 
         for (SimpleDescriptor.Service descService : desc.getServices()) {
@@ -195,6 +200,22 @@ public class SimpleDescriptorHandler {
                                 gws);
     }
 
+    /**
+     * Determine whether discovery should be performed for the specified descriptor.
+     *
+     * @param desc A SimpleDescriptor
+     * @return true, if discovery should be performed for the descriptor; Otherwise, false.
+     */
+    private static boolean shouldPerformDiscovery(final SimpleDescriptor desc) {
+        // If there is a discovery type specified, then discovery should be performed
+        final String discoveryType = desc.getDiscoveryType();
+        if (discoveryType != null && !discoveryType.isEmpty()) {
+            return true;
+        }
+        log.missingDiscoveryTypeInDescriptor(desc.getName());
+        return false;
+    }
+
     private static GatewayServices getGatewayServices(Service... services) {
       for (Service service : services) {
         if (service instanceof GatewayServices) {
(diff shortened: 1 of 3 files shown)


gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java
```diff
@@ -195,6 +200,47 @@
...
+     * Determine whether discovery should be performed for the specified descriptor.
+     *
+     * @param desc A SimpleDescriptor
+     * @return true, if discovery should be performed for the descriptor; Otherwise, false.
+     */
+    private static boolean shouldPerformDiscovery(final SimpleDescriptor desc) {
+
+        // If there is a discovery type specified, then discovery should be performed
+        final String discoveryType = desc.getDiscoveryType();
+        if (discoveryType != null && !discoveryType.isEmpty()) {
```
moresandeep, line 213: Instead of relying on all the properties shouldn't we just rely on discoveryType? less confusion with if else logic and also while defining the descriptor.

pzampino: Discovery type has been optional since the beginning, and defaults to Ambari when omitted. I didn't want to break this by adding a requirement that discovery type be specified. What I'm saying here is that any discovery-related properties in a descriptor signal that discovery is expected, and the rest carries on as it has prior to this change.

moresandeep: Okay, I see.

smolnar82:
My two cents:
- checking the `username` and `passwordAlias` here is not enough because in CM discovery Knox defaults to `cm.discovery.user` and `cm.discovery.password` if they are not defined -> following the new logic, performing the emptyness check here does not mean the user does not want service discovery
- checking the `address` here is not enough because in Ambari discovery Knox defaults to the value of `gateway.discovery.default.address` Gateway configuration

What if we simply rely on the discovery type. If none is selected -> no discovery will be performed. Discovery type is still optional (that is, the end-user does not need to select one) and makes the code easier to read and understand:
- `performDiscovery` should only be invoked if discovery type is not `null` [here](https://github.com/apache/knox/blob/v1.4.0/gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java#L109)
- the [default discovery type](https://github.com/apache/knox/blob/v1.4.0/gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java#L71) should be removed
- we may want to add some `WARN` level messages in the log in case discovery type is not selected but any of the other discovery related configuration is present. This check should be implemented in `SimpleDescriptorImpl` instead of in `SimpleDescriptorHandler` where only the log should be placed.

Any thoughts?

pzampino:
> My two cents:
> 
> * checking the `username` and `passwordAlias` here is not enough because in CM discovery Knox defaults to `cm.discovery.user` and `cm.discovery.password` if they are not defined -> following the new logic, performing the emptyness check here does not mean the user does not want service discovery
> * checking the `address` here is not enough because in Ambari discovery Knox defaults to the value of `gateway.discovery.default.address` Gateway configuration
> 
And in this case, the discovery type must be set in the descriptor to trigger discovery (which it will be because CM is not the default type).

> What if we simply rely on the discovery type. If none is selected -> no discovery will be performed. Discovery type is still optional (that is, the end-user does not need to select one) and makes the code easier to read and understand:
> 
My only concern is that existing descriptors that depend on the default type determination may be broken. Perhaps, that is not a significant concern though.

> * `performDiscovery` should only be invoked if discovery type is not `null` [here](https://github.com/apache/knox/blob/v1.4.0/gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java#L109)
> * the [default discovery type](https://github.com/apache/knox/blob/v1.4.0/gateway-topology-simple/src/main/java/org/apache/knox/gateway/topology/simple/SimpleDescriptorHandler.java#L71) should be removed
> * we may want to add some `WARN` level messages in the log in case discovery type is not selected but any of the other discovery related configuration is present. This check should be implemented in `SimpleDescriptorImpl` instead of in `SimpleDescriptorHandler` where only the log should be placed.
> 
I may be misunderstanding, but I think I disagree about the placement of this logic. The handler is interpreting the descriptor state to make the decision to perform discovery or not; I don't think that is an implementation-specific decision, and I don't think the descriptor should be responsible for discovery logic since it's really just a representation of a config file.

pzampino:
I've updated the changes to remove checks for any descriptor property other than discovery-type, making discovery-type a required property in descriptors for which discovery is desired.
I've also fixed the log messages (SimpleDescriptorMessages) so they'll actually be emitted now.

PR #336, opened 2020-05-12
KNOX-2371 - DefaultTopologyService may skip cluster config change pro…

pzampino:
… Restart Events

Corrected DefaultTopologyService.TopologyDiscoveryTrigger#onConfigurationChange to log invalid descriptors, and continue analyzing the remaining descriptors instead of quitting upon hitting the invalid one.

## How was this patch tested?

Augmented DefaultTopologyServiceTest, and manually tested.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
@@ -89,14 +89,14 @@ public class DefaultTopologyService extends FileAlterationListenerAdaptor implem
 
   private static final JAXBContext jaxbContext = getJAXBContext();
 
-  private static Auditor auditor = AuditServiceFactory.getAuditService().getAuditor(
+  private static final Auditor auditor = AuditServiceFactory.getAuditService().getAuditor(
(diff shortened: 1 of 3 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
```diff
@@ -193,7 +193,6 @@ private void redeployTopology(Topology topology) {
             } else {
               Thread.sleep(10);
               elapsed = System.currentTimeMillis() - start;
-              continue;
```
moresandeep, line 196: Just curious, why was continue removed?

moresandeep, line 196: Just curious, why was continue removed?

pzampino: Unnecessary as last statement in loop

pzampino: Just a clean-up thing while I was in there

moresandeep: ah gotcha.

PR #337, opened 2020-06-02
KNOX-2375 - Token state eviction should access the keystore file less…

pzampino:
… frequently

The goal of these changes is to reduce the number of times during a token state reaper iteration that the keystore file is loaded and subsequently written. This change reduces that number from (1 + 2-to-4-per-token) to (1 + 0-to-1-per-token + 1). So, for 100 expired tokens and 100 unexpired tokens, that means going from 601 loads to 2 loads (if all the token state aliases are cached) or up to 202 in the absolute worst case when none of the token state aliases is cached.

Part of the solution includes adding bulk alias/key removal method to the AliasService and KeystoreService respectively. 

On the request side, I've added a delay and bulk creation of token state aliases. Because key pair generation is so slow, doing this in a background thread allows Knox token requests to complete MUCH more quickly. Now, when a token is issued, its state is cached in the _unpersistedState_ collection. A periodic thread converts this unpersisted state into aliases in the keystore.

Finally, I've resolved an issue which was pausing KnoxToken request processing because the DefaultTokenAuthority was blocking on access to aliases for every single request. The values of the aliases being referenced are now being cached after the first access, removing the contention for the KeystoreService locks.

## How was this patch tested?

I've added multiple tests to exercise these changes: AliasBasedTokenStateServiceTest, DefaultTokenStateServiceTest, DefaultKeystoreServiceTest, RemoteAliasServiceTest, and ZookeeperRemoteAliasServiceTest. I'm continuing manual testing, but wanted to get the review process started as soon as possible.


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -237,12 +249,37 @@ protected boolean isUnknown(final String token) {
 
   protected void updateExpiration(final String tokenId, long expiration) {
     synchronized (tokenExpirations) {
-      tokenExpirations.replace(tokenId, expiration);
+      if (tokenExpirations.containsKey(tokenId)) {
+        tokenExpirations.replace(tokenId, expiration);
+      } else {
+        tokenExpirations.put(tokenId, expiration);
+      }
```
smolnar82, line 256: `Map.replace` also runs a `containsKey`:
```
The default implementation is equivalent to, for this map:
 
 if (map.containsKey(key)) {
     return map.put(key, value);
 } else
     return null;
```
I think a simple `put` is enough: you want to make sure you have the given expiration in `tokenExpirations`

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -237,12 +249,37 @@ protected boolean isUnknown(final String token) {
...
+    // Duplicating the logic from removeToken(String) here because this method is supposed to be an optimization for
+    // sub-classes that access an external store, for which bulk token removal performs better than individual removal.
+    // Sub-classes will have implemented removeToken(String), and calling that here will undo any optimizations provided
+    // by the sub-class's implementation of this method.
+    for (String tokenId : tokenIds) {
+      removeTokenState(tokenId);
+    }
+  }
+
+  private void removeTokenState(final String tokenId) {
```
smolnar82, line 282: I'm not sure if I agree that any non-optimized code should be left if we already know that code piece will result in causing performance issues.
My two cents here:

1. in `removeToken` I'd call `removeTokens(Collections.singleton(tokenId))`
2. in `removeTokens` I'd **not** call a synchronized block `N` times. Instead, you may synchronize once, fetch the keyset and call the `removeAll` on that key set -> that collection is 

> backed by the map, so changes to the map are reflected in the set, and vice-versa.

Something like this:
```
    synchronized (tokenExpirations) {
      tokenExpirations.keySet().removeAll(tokenIds);
    }
```

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -319,16 +366,17 @@ protected void evictExpiredTokens() {
...
 
   /**
    * Get a list of tokens
    *
-   * @return List of tokens
+   * @return
    */
   protected List<String> getTokens() {
-    return new ArrayList<>(tokenExpirations.keySet());
+    return tokenExpirations.keySet().stream().collect(Collectors.toList());
```
smolnar82, line 379: Should not we synchronize here too?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -254,7 +291,7 @@ protected void removeToken(final String tokenId) throws UnknownTokenException {
 
   protected boolean hasRemainingRenewals(final String tokenId, long renewInterval) {
     // Is the current time + 30-second buffer + the renewal interval is less than the max lifetime for the token?
-    return ((System.currentTimeMillis() + TimeUnit.SECONDS.toMillis(30) + renewInterval) < getMaxLifetime(tokenId));
+    return ((System.currentTimeMillis() + 30000 + renewInterval) < getMaxLifetime(tokenId));
```
smolnar82, line 294: Why is this 30 seconds buffer added?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -109,30 +133,52 @@ public long getTokenExpiration(final String tokenId) throws UnknownTokenExceptio
...
-
     try {
       aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME, tokenId);
       aliasService.removeAliasForCluster(AliasService.NO_CLUSTER_NAME, tokenId + TOKEN_MAX_LIFETIME_POSTFIX);
       log.removedTokenState(tokenId);
     } catch (AliasServiceException e) {
       log.failedToRemoveTokenState(tokenId, e);
     }
+    super.removeToken(tokenId);
+  }
```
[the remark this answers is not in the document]
pzampino: Yes, that would be better than invoking the individual removal method twice, now that there is the bulk removal method.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: Yes, I had meant to return to this after verifying that the replace invocation that was already there is not needed.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: +1

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: Yes, good catch

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: It is not added as part of this change, and I would have to revisit why it is there.

pzampino: The change to the literal value was unintentional

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -357,6 +359,30 @@ public void removeCredentialForCluster(String clusterName, String alias) throws
...
+    synchronized (this) {
+      KeyStore ks = getCredentialStoreForCluster(clusterName);
+      if (ks != null) {
+        try {
+          // Delete all the entries
+          for (String alias : aliases) {
+            if (ks.containsAlias(alias)) {
+              ks.deleteEntry(alias);
+            }
+            removeFromCache(clusterName, alias);
```
moresandeep, line 373: Instead of invalidating one entry at a time we can bulk invalidate them using [invalidateAll(keys)](https://www.javadoc.io/doc/com.github.ben-manes.caffeine/caffeine/1.2.0/com/github/benmanes/caffeine/cache/Cache.html#invalidateAll-java.lang.Iterable-) API

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
```diff
@@ -237,6 +238,13 @@ public void removeAliasForCluster(final String clusterName, final String alias)
         }
     }
 
+    @Override
+    public void removeAliasesForCluster(String clusterName, Set<String> aliases) throws AliasServiceException {
```
moresandeep, line 242: Not related to review but just a side note: we should make some calls to RemoteAliasService non blocking, it would help a lot.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -143,15 +189,21 @@ protected void updateExpiration(final String tokenId, long expiration) {
...
+      List<String> allAliases = aliasService.getAliasesForCluster(AliasService.NO_CLUSTER_NAME);
+
+      // Filter for the token state aliases, and extract the token ID
+      tokenIds = allAliases.stream()
+                           .filter(a -> a.contains(TOKEN_MAX_LIFETIME_POSTFIX))
+                           .map(a -> a.substring(0, a.indexOf(TOKEN_MAX_LIFETIME_POSTFIX)))
+                           .collect(Collectors.toList());
     } catch (AliasServiceException e) {
-      log.errorEvictingTokens(e);
+      e.printStackTrace(); // TODO: PJZ: Logging
```
moresandeep, line 203: Why are we throwing a stack trace? if this is for debugging can we wrap it up in logger and then throw instead of using e.printStackTrace().

pzampino: This is an oversight. Thanks for catching it.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
pzampino: Is there any benefit in doing it this way?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
pzampino: I'm not sure what you mean here. Can you elaborate?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
pzampino: Done.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
pzampino: Out of scope for this change IMO

moresandeep: yup, this was just a note :)

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -135,6 +138,25 @@ public JWT issueToken(Principal p, List<String> audiences, String algorithm, lon
...
+                                      final char[] signingKeystorePassphrase)
+          throws KeystoreServiceException, TokenServiceException {
+
+    if (signingKeystorePassphrase != null) {
+      char[] passphrase;
+      try {
+        passphrase = getSigningKeyPassphrase(signingKeystorePassphrase);
+      } catch (AliasServiceException e) {
+        throw new TokenServiceException(e);
+      }
```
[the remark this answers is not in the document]
pzampino: +1

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -177,11 +192,7 @@ public JWT issueToken(Principal p, List<String> audiences, String algorithm, lon
   }
 
   private char[] getSigningKeyPassphrase(char[] signingKeyPassphrase) throws AliasServiceException {
```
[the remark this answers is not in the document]
pzampino: +1

pzampino:
> Besides my review comments below I'm not sure if I got the idea of the `unpersistedState` collection in `AliasBasedTokenStateService`. Could you please update the PR description to make it cleaner for me :)

Done.

PR #340, opened 2020-06-08
KNOX-2381 racking UI of flink session is broken in YARNUIV2

morhidi:
A fix for a missing trailing '/' issue in YARNUIV2 that already fixed in YARNUI see [KNOX-2074](https://issues.apache.org/jira/browse/KNOX-2074)

## How was this patch tested?
Manually see the attached screenshot
image


smolnar82:
ing @PrabhuJoseph  since you last worked on the rewrite rules for YARN UI V2

PR #348, opened 2020-06-18
KNOX-2390 - Let end-users configure SAML2 client configuration using Pac4J provider parameters

smolnar82:

From now on, end-users can set the following SAML 2 configurations using provider parameters (in addition to the already supported params):
- useNameQualifier
- forceAuth
- passive
- nameIdPolicyFormat

Additionally, I refactored the special `Azure AD` client handling to its own class (another implementation of the new interface).

## How was this patch tested?

Added and executed JUnit tests:
```
[INFO] BUILD SUCCESS
[INFO] Total time: 18:47 min (Wall Clock)
[INFO] Finished at: 2020-06-18T12:40:14+02:00
[INFO] Final Memory: 432M/2028M
```

Manually tested with different provider parameters set. For instance:
- `useNameQualifier = true`
- `forceAuth = true`

```
<saml2p:AuthnRequest xmlns:saml2p="urn:oasis:names:tc:SAML:2.0:protocol"
                     AssertionConsumerServiceURL="https://www.local.com:8443/gateway/knoxsso/api/v1/websso?pac4jCallback=true&client_name=SAML2Client"
                     Destination="https://$OKTA_HOST/sso/saml"
                     ForceAuthn="true"
                     ID="_cef3baeef567461aae87cd95234d2bc38862368"
                     IsPassive="false"
                     IssueInstant="2020-06-18T09:17:21.672Z"
                     ProtocolBinding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
                     ProviderName="pac4j-saml"
                     Version="2.0"
                     >
    <saml2:Issuer xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion"
                  Format="urn:oasis:names:tc:SAML:2.0:nameid-format:entity"
                  NameQualifier="https://www.local.com:8443/gateway/knoxsso/api/v1/websso?pac4jCallback=true&client_name=SAML2Client"
                  >https://www.local.com:8443/gateway/knoxsso/api/v1/websso?pac4jCallback=true&client_name=SAML2Client</saml2:Issuer>
</saml2p:AuthnRequest>
```

### DIFF
--- a/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/config/AzureADClientConfigurationDecorator.java
+++ b/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/config/AzureADClientConfigurationDecorator.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.pac4j.config;
+
+import java.util.List;
+import java.util.Map;
+
+import org.pac4j.core.client.Client;
+import org.pac4j.core.http.callback.PathParameterCallbackUrlResolver;
+import org.pac4j.oidc.client.AzureAdClient;
+
+public class AzureADClientConfigurationDecorator implements ClientConfigurationDecorator {
+  private static final String AZURE_AD_CLIENT_CLASS_NAME = AzureAdClient.class.getSimpleName();
+
+  @Override
+  public void decorateClients(List<Client> clients, Map<String, String> properties) {
+    for (Client client : clients) {
+      if (AZURE_AD_CLIENT_CLASS_NAME.equalsIgnoreCase(client.getName())) {
+        // special handling for Azure AD, use path separators instead of query params
+        ((AzureAdClient) client).setCallbackUrlResolver(new PathParameterCallbackUrlResolver());
+      }
+    }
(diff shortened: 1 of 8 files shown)


gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/config/AzureADClientConfigurationDecorator.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class AzureADClientConfigurationDecorator implements ClientConfigurationDecorator {
+  private static final String AZURE_AD_CLIENT_CLASS_NAME = AzureAdClient.class.getSimpleName();
+
+  @Override
+  public void decorateClients(List<Client> clients, Map<String, String> properties) {
+    for (Client client : clients) {
+      if (AZURE_AD_CLIENT_CLASS_NAME.equalsIgnoreCase(client.getName())) {
+        // special handling for Azure AD, use path separators instead of query params
+        ((AzureAdClient) client).setCallbackUrlResolver(new PathParameterCallbackUrlResolver());
```
moresandeep, line 35: Looks like we are not passing the properties down to AzureAdClient here, or may be we don't need one here.

smolnar82: At this point, an instance of `AzureADClient` has already been built by `org.pac4j.config.builder.OidcClientBuilder` using the same properties object. While decorating the callback URL resolver we do not need to check those properties because setting it to an instance of `PathParameterCallbackUrlResolver` has no any dependency on any property.
P.S: this piece of code was just re-factored out from `org.apache.knox.gateway.pac4j.filter.Pac4jDispatcherFilter` (as it fits into the decorator pattern) I did not change the "logic".

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/config/Pac4jClientConfigurationDecorator.java
```diff
@@ -0,0 +1,45 @@
...
+public class Pac4jClientConfigurationDecorator implements ClientConfigurationDecorator {
+
+  private static final List<ClientConfigurationDecorator> DEFAULT_DECORATORS = Arrays.asList(new SAML2ClientConfigurationDecorator(), new AzureADClientConfigurationDecorator());
+  private final List<ClientConfigurationDecorator> decorators;
+
+  public Pac4jClientConfigurationDecorator() {
+    this(DEFAULT_DECORATORS);
+  }
+
+  // package protected so that it's visible in unit tests
```
pzampino, line 35: My personal preference is to assign the appropriate visibility based on the functionality, and use reflection if necessary from unit tests.

PR #349, opened 2020-06-18
KNOX-2393 - Add a configurable list of paths  that SSOCookieProvider can ignore

moresandeep:
This patch was tested on local cluster
This change proposes adding a configurable parameter that adds a list of path elements to be ignored by SSOCookieProvider. Default set includes favicon.ico.
This is because Browsers will send unsolicited requests to get favicon.ico which can break SSO experience especially when default topology feature is used when content is served from the root.

## How was this patch tested?

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
@@ -57,5 +57,5 @@ public interface JWTMessages {
   void jwtAudienceValidated();
 
   @Message( level = MessageLevel.INFO, text = "Path {0} is configured as unauthenticated path, letting the request {1} through" )
-  void unAuthenticatePathBypass(String path, String uri);
+  void unauthenticatedPathBypass(String path, String uri);
 }
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
@@ -55,4 +55,7 @@ public interface JWTMessages {
 
   @Message( level = MessageLevel.DEBUG, text = "Audience claim has been validated." )
   void jwtAudienceValidated();
+
+  @Message( level = MessageLevel.INFO, text = "Path {0} is configured as unauthenticated path, letting the request {1} through" )
+  void unAuthenticatePathBypass(String path, String uri);
 }
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
@@ -107,7 +107,7 @@ public class SSOCookieFederationFilter extends AbstractJWTFilter {
       unAuthPathString = DEFAULT_SSO_UNAUTHENTICATED_PATHS_PARAM;
     }
 
-    final StringTokenizer st = new StringTokenizer(unAuthPathString, ";");
+    final StringTokenizer st = new StringTokenizer(unAuthPathString, ";,");
     while (st.hasMoreTokens()) {
       unAuthenticatedPaths.add(st.nextToken());
     }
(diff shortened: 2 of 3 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
```diff
@@ -92,6 +100,18 @@ public void init( FilterConfig filterConfig ) throws ServletException {
...
     }
 
+    /* get unauthenticated paths list */
+    String unAuthPathString = filterConfig.getInitParameter(SSO_UNAUTHENTICATED_PATHS_PARAM);
+    /* if no list specified use default value */
+    if (StringUtils.isBlank(unAuthPathString)) {
+      unAuthPathString = DEFAULT_SSO_UNAUTHENTICATED_PATHS_PARAM;
+    }
+
+    final StringTokenizer st = new StringTokenizer(unAuthPathString, ";");
```
pzampino, line 110: Could we support ',' in addition to ';' as a delimiter here? I've personally been burned by using the incorrect delimiter in some provide config property values.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
```diff
@@ -121,13 +141,25 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+
       if (req.getMethod().equals("OPTIONS")) {
         // CORS preflight requests to determine allowed origins and related config
         // must be able to continue without being redirected
         Subject sub = new Subject();
         sub.getPrincipals().add(new PrimaryPrincipal("anonymous"));
         continueWithEstablishedSecurityContext(sub, req, res, chain);
-      } else {
+      }
+      else {
```
pzampino, line 162: Any good reason to have bumped this else onto its own line?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
```diff
@@ -121,13 +141,25 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+        if (req.getRequestURI().contains(path)) {
+          /* This path is configured as an unauthenticated path let the request through */
+          final Subject sub = new Subject();
+          sub.getPrincipals().add(new PrimaryPrincipal("anonymous"));
+          LOGGER.unAuthenticatePathBypass(path, req.getRequestURI());
+          continueWithEstablishedSecurityContext(sub, req, res, chain);
+        }
+      }
+
       if (req.getMethod().equals("OPTIONS")) {
```
pzampino, line 155: Safer to reverse this to "OPTIONS".equals(req.getMethod()) since it will handle null from getMethod().

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
smolnar82: +1 to Phil's suggestion

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
smolnar82: +1

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
```diff
@@ -55,4 +55,7 @@
 
   @Message( level = MessageLevel.DEBUG, text = "Audience claim has been validated." )
   void jwtAudienceValidated();
+
+  @Message( level = MessageLevel.INFO, text = "Path {0} is configured as unauthenticated path, letting the request {1} through" )
+  void unAuthenticatePathBypass(String path, String uri);
```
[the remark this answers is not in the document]
moresandeep: +1 Ack.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
moresandeep: Yeah that is a good idea, i was conflicted whether to support "," decided not to include to be in line with other config props. 
I will update it to support ",".

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
moresandeep: This was formatting error. I'll fix this.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
moresandeep: Good catch!

PR #350, opened 2020-06-18
KNOX-2392 - Simple file-based TokenStateService implementation

pzampino:

I've created a TokenStateService implementation that avoids the unnecessary overhead associated with the AliasBasedTokenStateService (size of keystore and associated performance of access).

Since KNOX-2377 added the TokenStateJournal for addressing the potential loss of token state due to changes necessary to improve the performance of the AliasBasedTokenStateService, this new implementation leverages that journaling facility alone to manage token state. It is effectively, the AliasBasedTokenStateService without the keystore interactions.

I've not yet made this implementation the default, but I have tested it locally, and I foresee this becoming the default for the near future since the use of the keystore presents an unnecessary burden now that secrets are no longer persisted with token state.

## How was this patch tested?

- _mvn -T1.5C -Ppackage,release clean install_
- Added unit tests (JournalBasedTokenStateServiceTest), and augmented existing tests.
- Changed (only locally) DefaultGatewayServices to use the new implementation, and tested manually, including concurrent load testing. Even with multiple concurrent clients making frequent token requests, the size of the persisted content grows at a small fraction of the rate that the keystore does under similar load.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JournalBasedTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JournalBasedTokenStateService.java
@@ -1,0 +1,173 @@
+/*
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ *
+ */
+package org.apache.knox.gateway.services.token.impl;
+
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.services.ServiceLifecycleException;
+import org.apache.knox.gateway.services.security.token.UnknownTokenException;
+import org.apache.knox.gateway.services.token.impl.state.TokenStateJournalFactory;
+import org.apache.knox.gateway.services.token.state.JournalEntry;
+import org.apache.knox.gateway.services.token.state.TokenStateJournal;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class JournalBasedTokenStateService extends DefaultTokenStateService {
+
+    private TokenStateJournal journal;
+
+    @Override
+    public void init(final GatewayConfig config, final Map<String, String> options) throws ServiceLifecycleException {
+        super.init(config, options);
+
+        try {
+            // Initialize the token state journal
+            journal = TokenStateJournalFactory.create(config);
+
+            // Load any persisted journal entries, and add them to the unpersisted state collection
+            List<JournalEntry> entries = journal.get();
+            for (JournalEntry entry : entries) {
+                String id = entry.getTokenId();
+                try {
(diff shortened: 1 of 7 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JournalBasedTokenStateService.java
```diff
@@ -0,0 +1,190 @@
...
+        try {
+            journal.add(tokenId, issueTime, expiration, maxLifetimeDuration);
+        } catch (IOException e) {
+            log.failedToAddJournalEntry(tokenId, e);
+        }
+    }
+
+    @Override
+    public long getTokenExpiration(final String tokenId, boolean validate) throws UnknownTokenException {
+        // Check the in-memory collection first, to avoid costly keystore access when possible
```
smolnar82, line 80: nit: the comment is misleading, there is no keystore access involve here if I understood correctly.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JournalBasedTokenStateService.java
```diff
@@ -0,0 +1,190 @@
...
+        // If there is no associated state in the in-memory cache, proceed to check the journal
+        long expiration = 0;
+        try {
+            JournalEntry entry = journal.get(tokenId);
+            if (entry == null) {
+                throw new UnknownTokenException(tokenId);
+            }
+
+            expiration = Long.parseLong(entry.getExpiration());
+            super.addToken(tokenId,
```
smolnar82, line 102: Would not be enough to invoke ` super.updateExpiration(tokenId, expiration)` here instead of re-adding the token? I see `super. addToken(...)` sets the max-lifetime too which seems correct.
However, in the `AliasBasedTokenStateService` we invoke `super.updateExpiration`. I think they should be in synch.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JournalBasedTokenStateService.java
pzampino: +1

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JournalBasedTokenStateService.java
pzampino: The difference is due to the difference in the nature of the implementations.

For AliasBasedTokenStateService, invoking _super.addToken_ would require an **additional keystore access** to get the max lifetime, and even then, I don't think the issue time is available; I think it's better to defer this additional keystore access for when the max lifetime is needed, and cache it then. Therefore, only the cached expiration is updated here.

For JournalBasedTokenStateService, expiration and max lifetime are not managed separately (notice, it does not override _setMaxLifetime()_), and its _addToken_ implementation is effectively an update when the journal entry already exists. Since the max lifetime and issue time are available in this case, it seems to me more efficient to update the in-memory cache of the max lifetime here via _addToken_, rather than requiring an additional file access later.

While the code differs, I believe the behavior is ultimately equivalent, and the implementations take advantage of their respective differences for a performance benefit.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JournalBasedTokenStateService.java
```diff
@@ -0,0 +1,190 @@
...
+                log.failedToLoadJournalEntry(e);
+            }
+        }
+        return result;
+    }
+
+    @Override
+    protected List<String> getTokens() {
+        List<String> tokenIds = new ArrayList<>();
+
```
moresandeep, line 136: Should we be checking in-memory implementation first?

pzampino: Hmm, I think we _could_ since this implementation loads all the persisted token state on init(), and any subsequent addition/removal of state should be reflected in the cache.

pzampino: I think that would mean this override simply goes away though, rather than checking the superclass implementation _first_.

PR #353, opened 2020-06-23
KNOX-2344 - Rename gateway-cm-integration module and related resources

smolnar82:

Renamed the `gateway-cm-integration` Maven module to `gateway-topology-hadoopxml` since the feature we implemented there is not CM specific: anyone can generate a Hadoop style XML file have Knox process it.

Additional changes:
- Knox will monitor `.hxr` (**H**adoop **X**ml **R**esources) files within `KNOX_DESCRIPTOR_DIR` (instead of `.cm`)
- logging parser result of Knox provider configurations too (in addition to Knox descriptors)

## How was this patch tested?

Ran a full build:
```
...
[INFO] BUILD SUCCESS
[INFO] Total time: 32:03 min (Wall Clock)
[INFO] Finished at: 2020-06-23T10:43:35+02:00
[INFO] Final Memory: 436M/2415M
```


gateway-release/pom.xml
```diff
@@ -439,7 +439,7 @@
         </dependency>
         <dependency>
             <groupId>org.apache.knox</groupId>
-            <artifactId>gateway-cm-integration</artifactId>
+            <artifactId>gateway-topology-hadoopxml</artifactId>
```
moresandeep, line 442: nit: since hadoop and xml are two different words do you think the name `gateway-topology-hadoop-xml` is more appropriate? it does look a bit long and awkward though.

gateway-topology-hadoopxml/pom.xml
```diff
@@ -25,9 +25,9 @@
         <version>1.5.0-SNAPSHOT</version>
     </parent>
 
-    <artifactId>gateway-cm-integration</artifactId>
-    <name>gateway-cm-integration</name>
-    <description>Cloudera Manager integration related resources</description>
+    <artifactId>gateway-topology-hadoopxml</artifactId>
+    <name>gateway-topology-hadoopxml</name>
```
moresandeep, line 29: Same nit as above.

smolnar82: Done.

gateway-release/pom.xml
smolnar82: Fine by me.

smolnar82:
> Need we be concerned with backward-compatibility at all here? Was this part of the 1.4 release?

Yes, this was released in 1.4. In fact, the JIRA was an outcome of the release e-mail thread where we agreed this should be renamed but it's not a release blocker.
The module I renamed was a new feature in 1.4 and AFAIK it's only been used by someone who integrated with Cloudera Manager. This means that whoever upgrades to this version needs to make sure to rename the CM generated file(s) to end with `.hxr` instead of `.cm`.

pzampino:
LGTM

PR #358, opened 2020-07-01
KNOX-2395 - Make Gateway Services Pluggable

smolnar82:

From now on end-users can select different implementation for certain Knox services (where there are more than one implementations) through a newly introduced `gateway-site.xml` configuration called `gateway.service.$SERVICE.impl`.
Where `$SERVICE` is the short name of the service you want to configure and the value is the fully qualified class name of the desired implementation.  This works as an `override` option for the default implementation (which is hard-coded now in the Java code).
Currently, the following service/implementation pairs are configurable:
- alias
   - `org.apache.knox.gateway.services.security.impl.DefaultAliasService`
   - `org.apache.knox.gateway.backend.hashicorp.vault.HashicorpVaultAliasService`
   - `org.apache.knox.gateway.services.security.impl.RemoteAliasService`
   - `org.apache.knox.gateway.services.security.impl.ZookeeperRemoteAliasService`
- master
   - `org.apache.knox.gateway.services.security.impl.DefaultMasterService`
   - `org.apache.knox.gateway.services.security.impl.CLIMasterService`
- tokenstate
   - `org.apache.knox.gateway.services.token.impl.DefaultTokenStateService`
   - `org.apache.knox.gateway.services.token.impl.AliasBasedTokenStateService`
   - `org.apache.knox.gateway.services.token.impl.JournalBasedTokenStateService`

`DefaultGatewayServices` and `CLIGatewayServices` have been modified to use the new service factories. Different service factories are loaded into the memory once using Java's `ServiceLoader` feature.

## How was this patch tested?

Edited existing JUnit tests, added a bunch of new tests to cover new classes, and executed them locally. As well as running some manual tests:

Ran the `knoxcli` tool:
- created the master secret
- added/listed some aliases

Started the Knox Gateway:
- checked the home page
- logged into the Admin UI
   - modified the 'sandbox' topology
   - modified a service definition (one that was listed in `sandbox`, confirmed that it got redeployed)
- set different values for `gateway.service.[alias|tokenstate].impl` in `gateway-site.xml` and confirmed the factories created the proper instances based on configuration

### DIFF
--- a/gateway-release/home/conf/gateway-site.xml
+++ b/gateway-release/home/conf/gateway-site.xml
@@ -19,6 +19,11 @@ limitations under the License.
 <configuration>
 
     <property>
+        <name>gateway.services</name>
+        <value>alias:remote</value>
+        <description>Overriding some of the gateway services so that Knox will use the selected implementation</description>
+    </property>
+    <property>
         <name>gateway.port</name>
         <value>8443</value>
         <description>The HTTP port for the Gateway.</description>
--- a/gateway-release/home/conf/gateway-site.xml
+++ b/gateway-release/home/conf/gateway-site.xml
@@ -19,9 +19,8 @@ limitations under the License.
 <configuration>
 
     <property>
-        <name>gateway.services</name>
-        <value>alias:remote</value>
-        <description>Overriding some of the gateway services so that Knox will use the selected implementation</description>
+        <name>gateway.service.alias.impl</name>
+        <value>org.apache.knox.gateway.services.security.impl.RemoteAliasService</value>
     </property>
     <property>
         <name>gateway.port</name>
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AliasServiceFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AliasServiceFactory.java
@@ -28,13 +28,10 @@ import org.apache.knox.gateway.services.ServiceType;
 import org.apache.knox.gateway.services.security.AliasService;
 import org.apache.knox.gateway.services.security.impl.DefaultAliasService;
 import org.apache.knox.gateway.services.security.impl.RemoteAliasService;
+import org.apache.knox.gateway.services.security.impl.ZookeeperRemoteAliasService;
 import org.apache.knox.gateway.services.security.impl.ZookeeperRemoteAliasServiceProvider;
 
 public class AliasServiceFactory extends AbstractServiceFactory {
-  private static final String DEFAULT_IMPLEMENTATION_NAME = "default";
-  private static final String HASHICORP_IMPLEMENTATION_NAME = "hashicorp";
-  private static final String REMOTE_IMPLEMENTATION_NAME = "remote";
-  private static final String ZK_IMPLEMENTATION_NAME = "zookeeper";
 
   @Override
   public Service create(GatewayServices gatewayServices, ServiceType serviceType, GatewayConfig gatewayConfig, Map<String, String> options, String implementation)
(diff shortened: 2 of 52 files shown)


gateway-release/home/conf/gateway-site.xml
```diff
@@ -18,6 +18,11 @@ limitations under the License.
 -->
 <configuration>
 
+    <property>
+        <name>gateway.services</name>
```
pzampino, line 22: I had envisioned the use of distinct properties, rather than a single "list" property.
Something more explicit, like:
`<property><name>gateway.service.alias</name><value>org.apache.knox.gateway.services.security.impl.RemoteAliasService</value></property>
<property><name>gateway.service.tokenstate</name><value>org.apache.knox.gateway.services.token.impl.JournalBasedTokenStateService</value></property>`

I think explicit class name values will make it easier for developers to create and deploy Knox customizations/extensions. Given this, I think a _list_ of such things could quickly become difficult to read.

pzampino: I do like the use of the AbstractFactory pattern here.

smolnar82: > I had envisioned the use of distinct properties, rather than a single "list" property.
> Something more explicit, like:
> `<property><name>gateway.service.alias</name><value>org.apache.knox.gateway.services.security.impl.RemoteAliasService</value></property> <property><name>gateway.service.tokenstate</name><value>org.apache.knox.gateway.services.token.impl.JournalBasedTokenStateService</value></property>`
> 
> I think explicit class name values will make it easier for developers to create and deploy Knox customizations/extensions. Given this, I think a _list_ of such things could quickly become difficult to read.

Thanks, @pzampino for your review comment. It makes sense and I'm in favor of your recommended approach with a small alteration: I'd build the property names as `gateway.service.$SERVICE.$PARAM` where service references the service name and param is an arbitrary parameter name that Knox can interpret later on. In this round, I'd add `impl` only but in upcoming changes, there will be others. For instance:
```
<property>
   <value>org.apache.knox.gateway.services.token.impl.ZookeeperTokenStateService</value>
</property>
<property>
   <name>gateway.service.tokenstate.useLocalAliasService</name>
   <value>false</value>
   <description>Indicates whether ZookeeperTokenStateService should use a local alias service when storing tokens or purely do the job in Zookeeper</description>
</property>

Any objection?

Comment (pzampino) [reply gateway-release/home/conf/gateway-site.xml:22]: I am strongly in favor of your suggestion. I believe it's a more flexible pattern than list-value properties. My examples had originally included `gateway.service.$SERVICE.impl` as property names, but I modified them for simplicity, since it was only an example, and not the primary point of my comment.

Comment (smolnar82) [reply gateway-release/home/conf/gateway-site.xml]: Fixed them all. Thanks again for the review, @pzampino !

File: gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AliasServiceFactory.java [hunk 2]
```diff
@@ -44,22 +41,17 @@ public Service create(GatewayServices gatewayServices, ServiceType serviceType,
       final AliasService defaultAliasService = new DefaultAliasService();
       ((DefaultAliasService) defaultAliasService).setMasterService(getMasterService(gatewayServices));
       ((DefaultAliasService) defaultAliasService).setKeystoreService(getKeystoreService(gatewayServices));
-      defaultAliasService.init(gatewayConfig, options); //invoking init on DefaultAliasService twice is ok (in case implementation is set to 'default')
-      switch (implementation) {
-      case DEFAULT_IMPLEMENTATION_NAME:
-      case "":
+      defaultAliasService.init(gatewayConfig, options); // invoking init on DefaultAliasService twice is ok (in case implementation is set to 'default')
+
+      if (matchesImplementation(implementation, DefaultAliasService.class, true)) {
```
pzampino, line 46: Why is the DefaultAliasService always created and initialized?
Could this not be done only when there is no implementation, or if the specified implementation is the default one?

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/RemoteRegistryClientServiceFactory.java
```diff
@@ -26,20 +26,26 @@
...
 // There are two implementations of 'RemoteConfigurationRegistryClientService':
 // - LocalFileSystemRemoteConfigurationRegistryClientService
 // - CuratorClientService
 // However, the first one - local - is for test-only purposes
 public class RemoteRegistryClientServiceFactory extends AbstractServiceFactory {
 
+  private final AliasServiceFactory aliasServiceFactory = new AliasServiceFactory();
+
   @Override
   public Service create(GatewayServices gatewayServices, ServiceType serviceType, GatewayConfig gatewayConfig, Map<String, String> options, String implementation)
```
pzampino, line 40: Can this implementation not be overridden?

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AliasServiceFactory.java
smolnar82: If you check out the rest of the implementations, they all use the `defaultAliasService` instance for their purpose. It was the same way (as a local alias service)

smolnar82:
> It still seems that the possible implementations are constrained to some set already known by Knox. Why can't a developer "plug-in" an arbitrary implementation of any gateway service to suit their individual need?

Hm..this is a good idea. I haven't thought about it until now. It'd also mean that the developer will need to make sure the implementation is on the classpath.
Let me ad it to this PR.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/RemoteRegistryClientServiceFactory.java
smolnar82: I'm not sure if I understood your comment here. Could you please elaborate?

pzampino: This is related to my most recent comment, I guess. Since there isn't another known implementation, it currently cannot be overridden. So, the heart of the matter is the ability to plug-in an arbitrary gateway service implementation.

pzampino:
> > It still seems that the possible implementations are constrained to some set already known by Knox. Why can't a developer "plug-in" an arbitrary implementation of any gateway service to suit their individual need?
> 
> Hm..this is a good idea. I haven't thought about it until now. It'd also mean that the developer will need to make sure the implementation is on the classpath.
> Let me ad it to this PR.

Yes, the implementation being plugged-in would obviously have to be on the classpath. The **ext** dir in a Knox installation is intended for such a purpose.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/RemoteRegistryClientServiceFactory.java
smolnar82: Fixed.

smolnar82:
> It still seems that the possible implementations are constrained to some set already known by Knox. Why can't a developer "plug-in" an arbitrary implementation of any gateway service to suit their individual need?

Thanks for this very useful comment @pzampino. I've just submitted a new commit making it possible to plug-in an arbitrary gateway service implementation.

PR #361, opened 2020-07-14
KNOX-2399 - Implemented ZookeeperTokenStateService

smolnar82:

Added a new token state service implementation that uses Zookeeper as token storage.

## How was this patch tested?

Updated/ran JUnit tests and executed the following manual test steps:

1. Enabled token state service in sandbox:
```
      <param>
          <name>knox.token.exp.server-managed</name>
          <value>true</value>
      </param>
```
2. Configured Knox to use the new ZK Token State Service in `gateway-site.xml`:
```
    <property>
        <value>org.apache.knox.gateway.services.token.impl.ZookeeperTokenStateService</value>
    </property>
    <property>
        <name>gateway.remote.config.registry.zookeeper-client</name>
        <value>type=ZooKeeper;address=$ZK_HOST:2181</value>
        <description>ZooKeeper configuration registry client details.</description>
    </property>
    <property>
        <name>gateway.remote.config.monitor.client</name>
        <value>zookeeper-client</value>
        <description>Remote configuration monitor client name.</description>
    </property>
```
3. Started Knox
4. Acquired a Knox delegation token:
```
$ curl -ivku guest:guest-password https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token
...
{"access_token":"eyJhbGciOiJSUzI1NiJ9.eyJzdWIiOiJndWVzdCIsImF1...,"target_url":"https://localhost:8443/gateway/tokenbased",...,"token_type":"Bearer","expires_in":1594764588683}

gateway.log:
2020-07-14 14:09:48,681 INFO  service.knoxtoken (TokenResource.java:init(169)) - Server management of token state is enabled for the "sandbox" topology.
2020-07-14 14:09:48,682 WARN  service.knoxtoken (TokenResource.java:init(199)) - There are no token renewers white-listed in the "sandbox" topology.
2020-07-14 14:09:48,690 INFO  service.knoxtoken (TokenResource.java:getAuthenticationToken(400)) - Knox Token service (sandbox) issued token eyJhbG...WYtr9Q (483b5976-f5ca-4761-b2d3-65041aa40d09)
```
5. Confirmed that the token became persisted in Zookeeper:
```
$ zookeeper-client -server $ZK_HOST:2181
Connecting to $ZK_HOST:2181
Welcome to ZooKeeper!
JLine support is enabled

WATCHER::

WatchedEvent state:SyncConnected type:None path:null

[zk: $ZK_HOST:2181(CONNECTED) 0] ls /knox/security/topology/__gateway
[483b5976-f5ca-4761-b2d3-65041aa40d09, 483b5976-f5ca-4761-b2d3-65041aa40d09--max]
```

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
@@ -53,10 +53,10 @@ public abstract class AbstractServiceFactory implements ServiceFactory {
       if (service == null && StringUtils.isNotBlank(implementation)) {
         // no known service implementation created, try to create the custom one
         try {
-          service = Service.class.cast(Class.forName(implementation).newInstance());
+          service = (Service) Class.forName(implementation).newInstance();
           logServiceUsage(implementation, serviceType);
         } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
-          throw new ServiceLifecycleException("Errror while instantiating " + serviceType.getShortName() + " service implementation " + implementation, e);
+          throw new ServiceLifecycleException("Error while instantiating " + serviceType.getShortName() + " service implementation " + implementation, e);
         }
       }
     }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AliasServiceFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AliasServiceFactory.java
@@ -54,7 +54,7 @@ public class AliasServiceFactory extends AbstractServiceFactory {
       } else if (matchesImplementation(implementation, RemoteAliasService.class)) {
         service = new RemoteAliasService(defaultAliasService, getMasterService(gatewayServices));
       } else if (matchesImplementation(implementation, ZookeeperRemoteAliasService.class)) {
-        service = new ZookeeperRemoteAliasServiceProvider().newInstance(defaultAliasService, getMasterService(gatewayServices));
+        service = new ZookeeperRemoteAliasServiceProvider().newInstance(gatewayServices, defaultAliasService, getMasterService(gatewayServices));
       }
 
       logServiceUsage(implementation, serviceType);
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/TokenStateServiceFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/TokenStateServiceFactory.java
@@ -31,6 +31,7 @@ import org.apache.knox.gateway.services.ServiceType;
 import org.apache.knox.gateway.services.token.impl.AliasBasedTokenStateService;
 import org.apache.knox.gateway.services.token.impl.DefaultTokenStateService;
 import org.apache.knox.gateway.services.token.impl.JournalBasedTokenStateService;
+import org.apache.knox.gateway.services.token.impl.ZookeeperTokenStateService;
 
 public class TokenStateServiceFactory extends AbstractServiceFactory {
 
@@ -46,6 +47,8 @@ public class TokenStateServiceFactory extends AbstractServiceFactory {
         ((AliasBasedTokenStateService) service).setAliasService(getAliasService(gatewayServices));
       } else if (matchesImplementation(implementation, JournalBasedTokenStateService.class)) {
         service = new JournalBasedTokenStateService();
+      } else if (matchesImplementation(implementation, ZookeeperTokenStateService.class)) {
+        service = new ZookeeperTokenStateService(gatewayServices);
       }
 
       logServiceUsage(implementation, serviceType);
@@ -61,6 +64,7 @@ public class TokenStateServiceFactory extends AbstractServiceFactory {
 
   @Override
   protected Collection<String> getKnownImplementations() {
-    return unmodifiableList(asList(DefaultTokenStateService.class.getName(), AliasBasedTokenStateService.class.getName(), JournalBasedTokenStateService.class.getName()));
+    return unmodifiableList(asList(DefaultTokenStateService.class.getName(), AliasBasedTokenStateService.class.getName(), JournalBasedTokenStateService.class.getName(),
+        ZookeeperTokenStateService.class.getName()));
   }
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
@@ -190,13 +190,18 @@ public class ZookeeperRemoteAliasService implements AliasService {
      *
      * @param clusterName
      *            cluster name
-     * @return List of all the aliases
+     * @return List of all the aliases (an empty list, in case there is no alias for the given cluster)
      */
     @Override
     public List<String> getAliasesForCluster(final String clusterName) throws AliasServiceException {
         final List<String> localAliases = localAliasService.getAliasesForCluster(clusterName);
         if (localAliases == null || localAliases.isEmpty()) {
-          return remoteClient == null ? new ArrayList<>() : remoteClient.listChildEntries(buildClusterEntryName(clusterName));
+          if (remoteClient != null) {
+            final List<String> remoteAliases = remoteClient.listChildEntries(buildClusterEntryName(clusterName));
+            return remoteAliases == null ? new ArrayList<>() : remoteAliases;
+          } else {
+            return new ArrayList<>();
+          }
         } else {
           return localAliases;
         }
@@ -435,7 +440,6 @@ public class ZookeeperRemoteAliasService implements AliasService {
 
         @Override
         public void childEvent(final RemoteConfigurationRegistryClient client, final Type type, final String path) {
-
             final String subPath = StringUtils.substringAfter(path, PATH_KNOX_ALIAS_STORE_TOPOLOGY + PATH_SEPARATOR);
             final String[] paths = StringUtils.split(subPath, '/');
 
@@ -494,11 +498,11 @@ public class ZookeeperRemoteAliasService implements AliasService {
         @Override
         public void entryChanged(final RemoteConfigurationRegistryClient client, final String path, final byte[] data) {
             try {
-                localAliasService.addAliasForCluster(cluster, alias, decrypt(new String(data, StandardCharsets.UTF_8)));
+              localAliasService.addAliasForCluster(cluster, alias, decrypt(new String(data, StandardCharsets.UTF_8)));
             } catch (final Exception e) {
-                /* log and move on */
-                LOG.errorAddingAliasLocally(cluster, alias, e.toString());
-            }
+              /* log and move on */
+              LOG.errorAddingAliasLocally(cluster, alias, e.toString());
+          }
         }
     }
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/ZookeeperTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/ZookeeperTokenStateService.java
@@ -34,11 +34,16 @@ import org.apache.knox.gateway.services.security.impl.ZookeeperRemoteAliasServic
  */
 public class ZookeeperTokenStateService extends AliasBasedTokenStateService {
 
-  private final AliasServiceFactory aliasServiceFactory = new AliasServiceFactory();
   private final GatewayServices gatewayServices;
+  private final AliasServiceFactory aliasServiceFactory;
 
   public ZookeeperTokenStateService(GatewayServices gatewayServices) {
+    this(gatewayServices, new AliasServiceFactory());
+  }
+
+  public ZookeeperTokenStateService(GatewayServices gatewayServices, AliasServiceFactory aliasServiceFactory) {
     this.gatewayServices = gatewayServices;
+    this.aliasServiceFactory = aliasServiceFactory;
   }
 
   @Override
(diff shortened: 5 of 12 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
```diff
@@ -190,13 +190,18 @@ private static void checkPathsExist(final RemoteConfigurationRegistryClient remo
...
      */
     @Override
     public List<String> getAliasesForCluster(final String clusterName) throws AliasServiceException {
-        final List<String> localAliases = localAliasService.getAliasesForCluster(clusterName);
+        final List<String> localAliases = localAliasService == null ? Collections.emptyList() : localAliasService.getAliasesForCluster(clusterName);
         if (localAliases == null || localAliases.isEmpty()) {
-          return remoteClient == null ? new ArrayList<>() : remoteClient.listChildEntries(buildClusterEntryName(clusterName));
+          if (remoteClient != null) {
+            final List<String> remoteAliases = remoteClient.listChildEntries(buildClusterEntryName(clusterName));
+            return remoteAliases == null ? new ArrayList<>() : remoteAliases;
```
moresandeep, line 201: `Collections.emptyList() ` might be better here too like you did above.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
```diff
@@ -190,13 +190,18 @@ private static void checkPathsExist(final RemoteConfigurationRegistryClient remo
...
+          } else {
+            return new ArrayList<>();
```
moresandeep, line 203: `Collections.emptyList() ` might be better here too like you did above. Unless this is mutable list.

gateway-server/src/test/java/org/apache/knox/gateway/services/factory/ServiceFactoryTest.java
```diff
@@ -52,7 +52,7 @@
 
   protected final GatewayServices gatewayServices = EasyMock.createNiceMock(GatewayServices.class);
   protected final GatewayConfig gatewayConfig = EasyMock.createNiceMock(GatewayConfig.class);
-  protected final Map<String, String> options = Collections.emptyMap();
+  protected final Map<String, String> options = new HashMap<>();
```
moresandeep, line 55: Any reason why we need a mutable map?

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
```diff
@@ -47,22 +52,38 @@ public Service create(GatewayServices gatewayServices, ServiceType serviceType,
...
-        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
-          throw new ServiceLifecycleException("Errror while instantiating " + serviceType.getShortName() + " service implementation " + implementation, e);
+    Service service = fetchFromCache(serviceType, implementation);
+    if (service == null) {
+      if (getServiceType() == serviceType) {
+        service = createService(gatewayServices, serviceType, gatewayConfig, options, implementation);
+        if (service == null && StringUtils.isNotBlank(implementation)) {
+          // no known service implementation created, try to create the custom one
+          try {
+            service = Service.class.cast(Class.forName(implementation).newInstance());
```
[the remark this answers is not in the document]
smolnar82: I believe this depends on personal tasting. I like the one I added before too.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
```diff
@@ -47,22 +52,38 @@ public Service create(GatewayServices gatewayServices, ServiceType serviceType,
...
+            logServiceUsage(implementation, serviceType);
+          } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+            throw new ServiceLifecycleException("Errror while instantiating " + serviceType.getShortName() + " service implementation " + implementation, e);
```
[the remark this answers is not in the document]
smolnar82: Because it is an ERRRRRRRRROR! :) I'll fix it.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
```diff
@@ -37,7 +41,8 @@
 
   private static final GatewayMessages LOG = MessagesFactory.get(GatewayMessages.class);
   private static final String IMPLEMENTATION_PARAM_NAME = "impl";
-  private static final String EMPTY_DEFAULT_IMPLEMENTATION = "";
+  protected static final String EMPTY_DEFAULT_IMPLEMENTATION = "";
+  private final Map<ServiceCacheKey, Service> serviceCache = new HashMap<>(); //to avoid creating the same service with the same implementation more than once
```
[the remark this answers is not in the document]
smolnar82: Removed it from the latest patchset.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AliasServiceFactory.java
```diff
@@ -54,7 +57,8 @@ protected Service createService(GatewayServices gatewayServices, ServiceType ser
       } else if (matchesImplementation(implementation, RemoteAliasService.class)) {
         service = new RemoteAliasService(defaultAliasService, getMasterService(gatewayServices));
       } else if (matchesImplementation(implementation, ZookeeperRemoteAliasService.class)) {
-        service = new ZookeeperRemoteAliasServiceProvider().newInstance(defaultAliasService, getMasterService(gatewayServices));
+        final boolean useLocalAliasService = options == null ? true : Boolean.valueOf(options.getOrDefault(USE_LOCAL_ALIAS_IN_ZK_IMPL_PARAM_NAME, "true"));
+        service = new ZookeeperRemoteAliasServiceProvider().newInstance(useLocalAliasService ? defaultAliasService : null, getMasterService(gatewayServices));
```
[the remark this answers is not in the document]
smolnar82: Nope. It means if you set the implementation to ZK in gateway-site.xml, then tokens will be stored in ZK. The default is `DefaultAliasService`

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
smolnar82: I just wanted to be consistent with the `null` case.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/ZookeeperRemoteAliasService.java
smolnar82: I've no idea if it should be mutable and I did not want to touch that part -> this is why I use the same way to return in the other case too.

gateway-server/src/test/java/org/apache/knox/gateway/services/factory/ServiceFactoryTest.java
smolnar82: We _might_ need. In one of our tests, I modified `options` (removed that since then) and the test failed since the map was immutable.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/TokenStateServiceFactory.java
```diff
@@ -61,6 +64,7 @@ protected ServiceType getServiceType() {
 
   @Override
   protected Collection<String> getKnownImplementations() {
-    return unmodifiableList(asList(DefaultTokenStateService.class.getName(), AliasBasedTokenStateService.class.getName(), JournalBasedTokenStateService.class.getName()));
+    return unmodifiableList(asList(DefaultTokenStateService.class.getName(), AliasBasedTokenStateService.class.getName(), JournalBasedTokenStateService.class.getName(),
```
pzampino, line 67: This is not necessarily directly related to this PR, but why isn't the ServiceLoader mechanism used for identifying service implementations available on the classpath?

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
```diff
@@ -47,22 +52,38 @@ public Service create(GatewayServices gatewayServices, ServiceType serviceType,
...
-        try {
-          service = Service.class.cast(Class.forName(implementation).newInstance());
-          logServiceUsage(implementation, serviceType);
-        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
-          throw new ServiceLifecycleException("Errror while instantiating " + serviceType.getShortName() + " service implementation " + implementation, e);
+    Service service = fetchFromCache(serviceType, implementation);
+    if (service == null) {
+      if (getServiceType() == serviceType) {
+        service = createService(gatewayServices, serviceType, gatewayConfig, options, implementation);
+        if (service == null && StringUtils.isNotBlank(implementation)) {
```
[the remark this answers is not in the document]
smolnar82: We **do** have a fallback. There are 2 types of service factories:

1. there is only one known implementation - in this case, that known implementation will be created if it's not overwritten in `gatewy-site.xml`
2. there are more than one known implementations - in this case, one of the known implementations (typically the first one in the `if-else` branches) will be the default one (i.e. none is defined in `gateway-site.xml`)

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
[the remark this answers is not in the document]
smolnar82: Removed that code since.

gateway-server/src/test/java/org/apache/knox/gateway/services/factory/TokenStateServiceFactoryTest.java
```diff
@@ -43,22 +44,31 @@ public void testBasics() throws Exception {
 
   @Test
   public void shouldReturnDefaultAliasService() throws Exception {
```
pzampino, line 46: Method name indicates an alias service test, but the implementation is a TokenStateService test.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/ZookeeperTokenStateService.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.knox.gateway.services.ServiceLifecycleException;
+import org.apache.knox.gateway.services.factory.AliasServiceFactory;
+import org.apache.knox.gateway.services.security.impl.ZookeeperRemoteAliasService;
+
+/**
+ * A Zookeeper Token State Service is actually an Alias based TSS where the 'alias service' happens to be the 'zookeeper' implementation.
+ * This means the only important thing that should be overridden here is the init method where the underlying alias service is configured
+ * properly.
+ */
+public class ZookeeperTokenStateService extends AliasBasedTokenStateService {
```
pzampino, line 35: I know it's a bit more work, but can you extend AliasBasedTokenStateServiceTest to exercise this new implementation?

gateway-server/src/test/java/org/apache/knox/gateway/services/factory/ServiceFactoryTest.java
pzampino: Does the mutability of the map matter for a _test_?

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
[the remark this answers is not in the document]
pzampino: I don't believe there is any functional difference between the two. The _cast_ method didn't exist until generics were added to Java.

pzampino:
> I have few concerns
> 
> 3. ZK is not not build to store data in this case we are storing sensitive data. In case of loss of quorum TSS will be significantly impacted (unless we store the tokens locally and then sync to ZK). Also, if we push a whole bunch of tokens in ZK we might impact other services negatively.

Since this implementation is based on the AliasBasedTokenStateService, it also employs a local cache and the token state journaling facility to mitigate potential token state loss. Further, I think the ZookeeperRemoteAliasService (used by this implementation) keeps the local keystore in sync, correct?

> 4. I foresee performance issues in case token size increases.

There is certainly performance testing yet to be done, at least to understand the limitations of this implementation.

gateway-server/src/test/java/org/apache/knox/gateway/services/factory/ServiceFactoryTest.java
smolnar82: Yes. Since this is a base class for other tests this `options` instance is used within test classes.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/TokenStateServiceFactory.java
smolnar82: We may implement this enhancement in a different JIRA, but I think it's out-of-scope of this one.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/ZookeeperTokenStateService.java
smolnar82: Yes, this is a good idea.

smolnar82: Added a brand new test class that actually starts-up a ZK cluster and the new ZK Token State Service stores the token aliases there: `org.apache.knox.gateway.services.token.impl.ZookeeperTokenStateServiceTest`

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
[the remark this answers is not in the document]
smolnar82: Changed anyway.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/AbstractServiceFactory.java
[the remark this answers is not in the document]
smolnar82: Fixed.

PR #365, opened 2020-07-24
KNOX-2402 - Adding Gateway performance testing

smolnar82:

All is described [here](https://cwiki.apache.org/confluence/display/KNOX/KNOX-2402+-+Develop+a+performance-test+framework+to+test+token+state+service).

## How was this patch tested?

Manual testing: configured a Knox Gateway and executed the performance test tool multiple times. I checked the generated test results as well as the log files. See the above-referenced link for more information.
I even revealed some performance test issues in `AliasBasedTokenStateService` (see [KNOX-2408](https://issues.apache.org/jira/browse/KNOX-2408))

### DIFF
--- a/gateway-performance-test/pom.xml
+++ b/gateway-performance-test/pom.xml
@@ -1,0 +1,126 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.knox</groupId>
+        <artifactId>gateway</artifactId>
+        <version>1.5.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>gateway-performance-test</artifactId>
+    <name>gateway-performance-test</name>
+    <description>A test framework to measure the Knox Gateway's performance</description>
+    
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-i18n</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-i18n-logging-log4j</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-shell</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-spi</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-util-common</artifactId>
+        </dependency>
+
+        <dependency>
+           <groupId>com.fasterxml.jackson.core</groupId>
+           <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+           <groupId>com.fasterxml.jackson.dataformat</groupId>
+           <artifactId>jackson-dataformat-yaml</artifactId>
+        </dependency>
+        <dependency>
+           <groupId>commons-io</groupId>
+           <artifactId>commons-io</artifactId>
+        </dependency>
+        <dependency>
+           <groupId>javax.servlet</groupId>
+           <artifactId>javax.servlet-api</artifactId>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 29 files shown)


smolnar82:
I still need to produce documentation about the tool (and update the PR's description once it's done). I just wanted you guys to review the code as soon as possible.

smolnar82:
Thanks, @risdenk for your comment! It totally makes sense; there is no reason to overcomplicate it. I added a new profile to run performance tests using Maven instead of having a separate release artifact.

gateway-performance-test/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+					                    <systemProperty>
+					                        <key>perf.test.report.generation.target.folder</key>
+					                        <value>${basedir}/target/testResults</value>
+					                    </systemProperty>
+					                    <systemProperty>
+					                        <key>log4j.configuration</key>
+					                        <value>file:${basedir}/src/test/resources/performanceTest-log4j.properties</value>
+					                    </systemProperty>
+					                </systemProperties>
+					            </configuration>
```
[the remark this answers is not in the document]
smolnar82: Ohh...copy-pasted that section from the original Maven page which comes with tabs. Fixed now.

gateway-performance-test/src/main/java/org/apache/knox/gateway/performance/test/PerformanceTestRunner.java
```diff
@@ -0,0 +1,40 @@
...
+
+public class PerformanceTestRunner {
+
+  public static void main(String[] args) throws Exception {
+    final PerformanceTestConfiguration configuration = new PerformanceTestConfiguration(args[0]);
+    final ResponseTimeCache responseTimeCache = new ResponseTimeCache();
+    final GatewayMetricsReporter metricsReporter = new GatewayMetricsReporter(configuration, responseTimeCache);
+    metricsReporter.start();
+    final List<PerformanceTestLifeCyleListener> lifeCyleListeners = Arrays.asList(metricsReporter);
+    for (PerformanceTestUseCase useCase : PerformanceTestUseCase.values()) {
```
pzampino, line 33: Could use ServiceLoader instead of a hard-coded enum to find PerformanceTestUseCase implementations.

gateway-performance-test/src/main/java/org/apache/knox/gateway/performance/test/knoxtoken/KnoxTokenCache.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+class KnoxTokenCache {
+  private static final int CAPACITY = 500;
+  private final Queue<String> knoxTokens = new ConcurrentLinkedQueue<>();
+
+  void saveKnoxToken(String knoxToken) {
+    cleanIfNecessary();
+    this.knoxTokens.offer(knoxToken);
```
pzampino, line 29: Is it acceptable for a token to NOT be saved here? Just wondering why you're using _offer_ instead of _add_.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -154,17 +177,16 @@ public long getTokenExpiration(final String tokenId) throws UnknownTokenExceptio
...
 
+    long expiration = -1;
     synchronized (tokenExpirations) {
-      if (!tokenExpirations.containsKey(tokenId)) {
-        throw new UnknownTokenException(tokenId);
-      }
-      expiration = tokenExpirations.get(tokenId);
+      expiration = tokenExpirations.getOrDefault(tokenId, -1L);
+    }
+    if (expiration == -1) {
```
pzampino, line 188: This doesn't appear to be any different than the previous implementation. If this were a check for _expiration < 0_, I could see a difference. What is the reason for the change?

gateway-shell/src/main/java/org/apache/knox/gateway/shell/hdfs/Ls.java
```diff
@@ -41,6 +42,11 @@ public Request dir( String dir ) {
       return this;
     }
 
+    public Request knoxToken(String knoxToken) {
```
pzampino, line 45: Why not add Knox token support to KnoxSession?

gateway-performance-test/src/main/java/org/apache/knox/gateway/performance/test/knoxtoken/KnoxTokenCache.java
smolnar82: As discussed offline, it is acceptable: this cache works only as a storage to make sure there is an already acquired Knox DT to be renewed/used. It really does not matter if we miss some.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
smolnar82: There is a slight difference: we can save one iteration when checking `tokenExpirations`
`containsKey + get` vs. `getOrDefault`.
But you are right, there is no more difference between the two versions.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/hdfs/Ls.java
smolnar82: IMO, the token handling should happen on request level. But you are right, this should not be limited to the `ls` command -> moving it to the abstract parent.

gateway-performance-test/src/main/java/org/apache/knox/gateway/performance/test/PerformanceTestRunner.java
smolnar82: Thanks for your comment; I made the requested change.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
smolnar82: I left it as-is based on my previous comment.

gateway-shell/src/main/java/org/apache/knox/gateway/shell/hdfs/Ls.java
smolnar82: Fixed.

PR #366, opened 2020-07-30
KNOX-2434 - Knox should fallback to JDK default keystore/truststore type instead of hardcoding JKS

risdenk:

Replace hardcoded `JKS` with `KeyStore.getDefaultType()`

## How was this patch tested?

* `mvn -U -T.75C clean verify -Ppackage,release -Dshellcheck`
* Check that this fixes FIPS crypto when default keystore configured at JDK level


risdenk:
I'm still working on this - just want to make sure it gets a few CI runs through different JDK versions.

PR #369, opened 2020-08-26
KNOX-2436 Add new service for replacing resourceManager property duri…

arvaiadam:

In Oozie 5.x there is a new parameter called resourceManager. This is used instead of jobTracker. However, Oozie can fall back to jobTracker value if no resourceManager is defined. resourceManager shall have the same value as jobTracker has now.

This ticket is intend to track the work on adding a new rewrite rule and a new service returning the resource-manager's address.

Currently the JobTrackerServiceModelGenerator despite its name is actually returning the resource-manager's address. This is not enabled by default because its misleading name since we no longer have job-trackers but resource-managers.

By having a new service which does exactly the same thing as JobTrackerServiceModelGenerator but is enabled by default we could have Oozie work with Knox out of the box.

We should be careful as the resource manager service name is already taken by ResourceManagerUIServiceModelGenerator.

## How was this patch tested?
Updated/ran unit tests.

### DIFF
--- a/gateway-admin-ui/admin-ui/app/new-desc-wizard/new-desc-wizard.component.ts
+++ b/gateway-admin-ui/admin-ui/app/new-desc-wizard/new-desc-wizard.component.ts
@@ -62,6 +62,7 @@ export class NewDescWizardComponent implements OnInit {
         'RANGER',
         'RANGERUI',
         'RESOURCEMANAGER',
+        'RESOURCEMANAGERAPI',
         'SOLR',
         'SPARKHISTORYUI',
         'SPARK3HISTORYUI',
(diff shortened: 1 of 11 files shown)


smolnar82:
@arvaiadam , in the PR description you say this is needed for Oozie 5.x. Would not be better to create a new version within `gateway-service-definitions` (e.g. `5.0.0`) and leave the previous version as-is?

smolnar82 (commented):
@arvaiadam , in the PR description you say this is needed for Oozie 5.x. Would not be better to create a new version within gateway-service-definitions (e.g. 5.0.0) and leave the previous version as-is?

smolnar82 (approved):
LGTM

smolnar82:
@moresandeep, please review this PR when you have time. Thanks!

smolnar82 (approved):
LGTM

arvaiadam:
> @arvaiadam , in the PR description you say this is needed for Oozie 5.x. Would not be better to create a new version within gateway-service-definitions (e.g. 5.0.0) and leave the previous version as-is?

Thanks for your feedback, i changed accordingly.

smolnar82:
> > @arvaiadam , in the PR description you say this is needed for Oozie 5.x. Would not be better to create a new version within gateway-service-definitions (e.g. 5.0.0) and leave the previous version as-is?
> 
> Thanks for your feedback, i changed accordingly.

Thanks, @arvaiadam! I really appreciate your fast feedback and action you've taken here!

moresandeep:
Java8 job failed, re-triggered it. Once it passes we should be good to go.

smolnar82:
There is an ongoing issue with Github Actions for a while (for some PRs it's intermittent, for some it's more severe). Luckily, we have Travis in place which proves this PR is OK to merge so I'm going to do it now.

PR #370, opened 2020-08-28
KNOX-2437 The request url does not have to be coded

apiloqbc:

URI.getPath() returns is an encoded URL
The request does not have to be coded, they should be rewritten but left as they are.
Backend applications expect that they are not.

## How was this patch tested?

all unit tests are working after the fix


moresandeep (approved):
Looks good, sorry for the delay!

PR #371, opened 2020-08-31
KNOX-2408 - Improved AliasBasedTokenState service and house-keeping

smolnar82:

- Let end-users customize the size/TTL attributes of the cache in `DefaultKeystoreService` I introduced while working on [KNOX-2136](https://issues.apache.org/jira/browse/KNOX-2136)
- `AliasBasedTokenStateService.updateExpiration` updates the data in-memory and let the background thread does its job
- if `AliasBasedTokenStateService` is selected as token state service, a new background thread is run when the service is started which reads previously persisted relevant aliases into the in-memory collections (in `DefaultTokenStateService`). With this change, it's not needed to override `getTokens` in `AliasBasedTokenStateService` which is only used by the reaper thread when evicting expired tokens. Instead, the token eviction logic will wait until everything is loaded.
- `DefaultTokenStateService.tokenExpirations` and `DefaultTokenStateService.maxTokenLifetimes` became `ConcurrentHashMap`s -> get rid of many of the synchronization blocks -> easier to read and maintain the code and `CHM` performs better than synchronization + `HashMap`

## How was this patch tested?

Updated and ran JUnit tests as well as executed the performance test tool multiple times to make sure I achieved the performance goals I was looking for with these changes.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultAliasService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultAliasService.java
@@ -197,7 +197,7 @@ public class DefaultAliasService extends AbstractAliasService {
 
   //Overriding the default behavior as we want to avoid loading the keystore N-times from the file system
   @Override
-  public Map<String, char[]> getPasswordAliasMapForGateway() throws AliasServiceException {
+  public Map<String, char[]> getPasswordsForGateway() throws AliasServiceException {
     final Map<String, char[]> passwordAliasMap = new HashMap<>();
     try {
       final KeyStore gatewayCredentialStore = keystoreService.getCredentialStoreForCluster(NO_CLUSTER_NAME);
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultAliasService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultAliasService.java
@@ -22,6 +22,7 @@ import java.security.KeyStoreException;
 import java.security.cert.Certificate;
 import java.util.ArrayList;
 import java.util.Enumeration;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -30,14 +31,14 @@ import org.apache.knox.gateway.GatewayMessages;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.services.ServiceLifecycleException;
-import org.apache.knox.gateway.services.security.AliasService;
+import org.apache.knox.gateway.services.security.AbstractAliasService;
 import org.apache.knox.gateway.services.security.AliasServiceException;
 import org.apache.knox.gateway.services.security.KeystoreService;
 import org.apache.knox.gateway.services.security.KeystoreServiceException;
 import org.apache.knox.gateway.services.security.MasterService;
 import org.apache.knox.gateway.util.PasswordUtils;
 
-public class DefaultAliasService implements AliasService {
+public class DefaultAliasService extends AbstractAliasService {
   private static final GatewayMessages LOG = MessagesFactory.get( GatewayMessages.class );
 
   private KeystoreService keystoreService;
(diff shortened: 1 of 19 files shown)


gateway-server/src/test/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateServiceTest.java
```diff
@@ -195,7 +204,8 @@ public void testAddAndRemoveTokenIncludesCache() throws Exception {
    * Verify that the token state reaper includes token state which has not been cached, so it's not left in the keystore
    * forever.
    */
-  @Test
+  @Ignore("I'm not sure if this is a valid use case since we have everything in the cache when eviction takes place")
```
smolnar82, line 207: @pzampino As you can see I only ignored this test case for now because I'm not sure if it makes sense going forward. Since the eviction thread checks everything in the cache we may remove this test case. What do you think?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultAliasService.java
```diff
@@ -194,6 +195,24 @@ public void removeAliasesForCluster(String clusterName, Set<String> aliases) thr
     return getPasswordFromAliasForCluster(NO_CLUSTER_NAME, alias);
   }
 
+  //Overriding the default behavior as we want to avoid loading the keystore N-times from the file system
+  @Override
```
pzampino, line 199: Does something similar need to be done for cluster-specific aliases?

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/AliasService.java
```diff
@@ -54,6 +54,8 @@ void generateAliasForCluster(String clusterName, String alias)
   char[] getPasswordFromAliasForGateway(String alias)
       throws AliasServiceException;
 
+  Map<String, char[]> getPasswordAliasMapForGateway() throws AliasServiceException;
```
pzampino, line 57: Why not simply getGatewayAliases(), more similar to other signatures in this interface? Why the need to cite the return type in the method signature?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -122,6 +129,44 @@ public void start() throws ServiceLifecycleException {
...
+    try {
+      log.loadingGatewayCredentialsOnStartup();
+      final long start = System.currentTimeMillis();
+      final Map<String, char[]> passwordAliasMap = aliasService.getPasswordAliasMapForGateway();
+      String alias, tokenId;
+      long expiration, maxLifeTime;
+      int count = 0;
+      for (Map.Entry<String, char[]> passwordAliasMapEntry : passwordAliasMap.entrySet()) {
+        alias = passwordAliasMapEntry.getKey();
+        if (alias.endsWith(TOKEN_MAX_LIFETIME_POSTFIX)) {
```
pzampino, line 150: Some comments here might be helpful for future viewers. I understand what is being done here, but others may not.

pzampino (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -344,36 +320,44 @@ protected void validateToken(final String tokenId) throws IllegalArgumentExcepti
...
+          log.failedExpiredTokenEviction(e);
+        }
       }
+    } else {
+      log.skipEviction();
     }
   }
 
+  protected boolean readyForEviction() {
+    return true;
```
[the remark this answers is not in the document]
smolnar82: It's overridden in `AliasBasedTokenStateService` where the returned value depends on the state of loading all aliases from `__gateway-credentials.jckes`

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultAliasService.java
smolnar82: I'm going to re-check and change other methods too if needed.

smolnar82: So, in this case, the cluster we use is `__gateway` and the only reason this method was added is to avoid exhausting I/O operations in the new `AliasBasedTokenStateService.loadGatewayCredentialsOnStartup()` method.
Therefore I don't see if other methods should be changes within the scope of this task.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
smolnar82: Done.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/AliasService.java
smolnar82: Done.

smolnar82:
Thanks, @pzampino for your valuable review comments. I addressed them and submitted a new commit. Please check it out and let me know if there is anything else that should be done or I can merge my changes.

pzampino (approved):
LGTM

PR #372, opened 2020-09-04
KNOX-2412 - Add Logout Link to Home Page for Select Authentication Pr…

lmccay:

* Logout Button on Homepage
* Homepage Logout Configurable from gateway-site.xml
* KNOXSSOUT service changed to be able to be hosted in arbitrary topologies - may break existing use of this service in a dedicated knoxssout.xml topology which was the original intent but it was never really documented as supported.
* KNOX-SESSION API added to return authenticated user and optionally logoutUrl
* A SessionInformation section added to the Homepage with a Welcome {username} and logout button
* KNOXSSOUT and KNOX-SESSION added to homepage.xml topology by default
* homePageLogoutEnabled defaults to true since the default KnoxSSO with form based IDP is supported.

### Planned for Future

* Move session-information module into the header
* Integrate KNOXSSO with TokenStateServer feature for token revocation for cookie tokens
* Extend KNOXSSOUT to revoke tokens on logout
* Pluggable logout providers rather than a single logoutUrl mechanism so other IDP logout mechanisms can be integrated

## How was this patch tested?
* Existing unit tests run
* SessionResourceTest unit test added

image

The following has homePageLogoutEnabled=false

image

### DIFF
--- a/gateway-release/pom.xml
+++ b/gateway-release/pom.xml
@@ -451,7 +451,7 @@
         </dependency>
         <dependency>
             <groupId>org.apache.knox</groupId>
-            <artifactId>gateway-service-metadata</artifactId>
+            <artifactId>gateway-service-session</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/gateway-service-knoxssout/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOutResource.java
+++ b/gateway-service-knoxssout/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOutResource.java
@@ -44,7 +44,7 @@ public class WebSSOutResource {
   private static final String SSO_COOKIE_NAME = "knoxsso.cookie.name";
   private static final String DEFAULT_SSO_COOKIE_NAME = "hadoop-jwt";
 
-  static final String RESOURCE_PATH = "/api/v1/webssout";
+  static final String RESOURCE_PATH = "knoxssout/api/v1/webssout";
 
   private String cookieName;
 
--- a/gateway-service-session/src/main/java/org/apache/knox/gateway/service/session/SessionResource.java
+++ b/gateway-service-session/src/main/java/org/apache/knox/gateway/service/session/SessionResource.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.knox.gateway.service.session;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.MediaType.APPLICATION_XML;
+
+import javax.inject.Singleton;
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.i18n.messages.MessagesFactory;
+import org.apache.knox.gateway.security.SubjectUtils;
+
+@Singleton
+@Path("session/api/v1/")
+public class SessionResource {
+  private static final SessionServiceMessages LOG = MessagesFactory.get(SessionServiceMessages.class);
+
+  @Context
+  HttpServletRequest request;
+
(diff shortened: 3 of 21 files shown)


gateway-service-session/src/main/java/org/apache/knox/gateway/service/session/SessionResource.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.i18n.messages.MessagesFactory;
+import org.apache.knox.gateway.security.SubjectUtils;
+
+@Singleton
+@Path("session/api/v1/")
```
smolnar82, line 36: Perhaps changing this to `api/v1/session` to be consistent with another API endpoints?

knox-homepage-ui/home/app/sessionInformation/session.information.component.ts
```diff
@@ -0,0 +1,80 @@
...
+        this['showSessionInformation'] = true;
+    }
+
+    getUser() {
+        if (this.sessionInformation) {
+            console.debug('SessionInformationComponent --> getUser() --> ' + this.sessionInformation.user);
+            return this.sessionInformation.user;
+        }
+        console.debug('SessionInformationComponent --> getUser() --> dr.who');
+        return 'dr.who';
```
smolnar82, line 42: Why not returning `null`?

knox-homepage-ui/home/app/sessionInformation/session.information.component.ts
```diff
@@ -0,0 +1,80 @@
...
+            return this.sessionInformation.logoutUrl;
+        }
+        return null;
+    }
+
+    logout() {
+        // window.alert('Are you sure???');
+        console.debug('SessionInformationComponent --> attempting logout() --> ');
+        if (this.sessionInformation) {
+            if (this.getLogoutUrl() == null) {
```
smolnar82, line 57: I'd use `this.logoutSupported` so that the logic if logout is supported happens in one place only (in `setSessionInformation`).

knox-homepage-ui/home/app/sessionInformation/session.information.component.ts
```diff
@@ -0,0 +1,80 @@
...
+                window.alert('Logout for the configured is IDP not supported.\nPlease close all browser windows to logout.');
```
smolnar82, line 58: There is a more elegant way of notifying the end-users: using a module called `sweetalert` (https://www.npmjs.com/package/sweetalert). It's been used both on the Admin UI and on the Home Page. See this usage for instance:
https://github.com/apache/knox/blob/master/knox-homepage-ui/home/app/homepage.service.ts#L81

smolnar82 (changes requested):
It looks really good; I only have some minor comments and questions to address.

knox-homepage-ui/home/app/sessionInformation/session.information.component.ts
lmccay: This isn't being used now since it doesn't render the button at all for now. I'm not sure which is better which is why I left it here for now. A missing logout button may seem like a bug but a logout button that doesn't log you out may be bad as well. At least it is known to be intentional and provides some insights though. Planned to follow up with discussion on this. If we go that route we can use sweetalert instead for sure.

knox-homepage-ui/home/app/sessionInformation/session.information.component.ts
lmccay: dr.who is common in Hadoop world. While I don't like it - it is better than the UI saying "Welcome null". :)

gateway-service-knoxssout/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOutResource.java
```diff
@@ -44,7 +44,7 @@
   private static final String SSO_COOKIE_NAME = "knoxsso.cookie.name";
   private static final String DEFAULT_SSO_COOKIE_NAME = "hadoop-jwt";
 
-  static final String RESOURCE_PATH = "/api/v1/webssout";
+  static final String RESOURCE_PATH = "knoxssout/api/v1/webssout";
```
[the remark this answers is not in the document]
lmccay: Not all of the other APIs do that. It was done that way for APIs that were initially intended to be deployed to a single dedicated topology. Like the admin.xml.

gateway-release/pom.xml
```diff
@@ -451,7 +451,7 @@
         </dependency>
         <dependency>
             <groupId>org.apache.knox</groupId>
-            <artifactId>gateway-service-metadata</artifactId>
+            <artifactId>gateway-service-session</artifactId>
```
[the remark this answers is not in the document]
lmccay: Don't know what you mean here.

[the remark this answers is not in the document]
lmccay: Not sure why the patch only shows up like this but the original gateway-service-metadata is still in there. I had originally copy and pasted it but forgot to change the name then went back and changed it. The merged file looks fine.

gateway-service-session/src/main/java/org/apache/knox/gateway/service/session/SessionResource.java
lmccay: I don't believe so - this is the session API, for now there is a sessioninfo but we could potentially add other things to it. It can be added to any topology and addressed via the session context path.

knox-homepage-ui/home/app/sessionInformation/session.information.component.ts
lmccay: Good point. I'll make that change - thanks!

PR #381, opened 2020-10-13
KNOX-2462 - Make credential store type configurable

smolnar82:
KNOX-2463 - Let end-users customize security algorithm for internal credential stores
KNOX-2464 - KnoxCLI should pass GatewayConfig when setting up master secret so that encryptor becomes initialized

This PR contains resolution for 3 JIRAs at once (using different commits, of course) as they are related to each other. In these commits, I made sure that:
- credential store type is configurable via `gateway-site.xml` (defaults to `JCEKS`)
- the algorithm Knox uses when creating an entry in a credential is configurable via `gateway-site.xml` (defaults to `AES`)
- fixed the bug in KnoxCLI of not considering inputs from `gateway-site.xml`

## How was this patch tested?

Tested manually in a CM managed cluster where I set:
```
gateway.credential.store.type = BCFKS
gateway.credential.store.alg = HMACSHA512
gateway.crypto.salt.size = 16
```
All changes were picked up and Knox functioned as expected.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayMessages.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayMessages.java
@@ -187,6 +187,9 @@ public interface GatewayMessages {
   @Message( level = MessageLevel.INFO, text = "Credential store found for the cluster: {0} - no need to create one." )
   void credentialStoreForClusterFoundNotCreating(String clusterName);
 
+  @Message(level = MessageLevel.WARN, text = "An existing credential store found for the cluster {0} with a different type of {1}")
+  void credentialStoreForClusterFoundWithDifferentType(String clusterName, String existingCredentialStoreType);
+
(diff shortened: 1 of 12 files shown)


risdenk (commented):
What happens if the credential store type is changed after Knox has already been started? Is there a helpful error message? Does Knox just blindly create a new one and leave the old one there?

risdenk:
> Do you mean if credential store type changes AND Knox is restarted?

Yup. Like you start with the default, and then go oops changed my mind. Or you change it and then want to go back to the default

smolnar82:
This is a valid point. Let me think and come up with a proposal.

lmccay:
I think logging + creating a new one and leaving the old one will be sufficient.
Let's not try and be smart about it and delete something that may be needed.
It seems like an unusual or at least infrequent thing to do anyway.

smolnar82:
Added a check to log a WARN message if a credential store existed before with a different extension.

risdenk (approved):

PR #383, opened 2020-11-15
KNOX-2469 - Fixing Knox keystore path directory creation for symlinks

jameschen1519:

This patch fixes a potential issue regarding the creation of Knox's keystores. The current logic checks to see if the keystore path exists--if it doesn't, it tries to create the parent folder of the keystore path. However, there is an edge case, as described in JDK-8130464, where the directory creation fails if the final, parent directory of the keystore path is a symlink. This causes a failure during startup. This PR remedies this by checking if the keystore parent directory exists instead of checking the keystore itself, as checking directories is symlink-safe. There is also no extra logic after the keystore creation, so if the keystore does exist, this turns into a no-op.

## How was this patch tested?

(If this patch involves UI changes, please attach a screen-shot; otherwise, remove this)

This PR was tested against an environment in which the keystore path's parent directory is a symlink on Knox 1.4.0. The tests that pass before this change pass after this change, locally.

One issue that may be worth noting is that this does not fix the case where some ancestor directory within the keystore path directory chain is an invalid symlink. If C:\a is a symlink to C:\b but C:\b does not exist, then the attempt to create C:\a\z will fail. However, depending on how we would like to do this, this might be a task best assigned to the users.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
@@ -516,7 +516,7 @@ public class DefaultKeystoreService implements KeystoreService {
     // Ensure the parent directory exists...
     // This is symlink safe.
     Path parentPath = keystoreFilePath.getParent();
-    if (parentPath == null || !Files.isDirectory(parentPath)) {
+    if (parentPath != null && !Files.isDirectory(parentPath)) {
       try{
         // This will attempt to create all missing directories.  No failures will occur if the
         // directories already exist.
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
@@ -513,11 +513,11 @@ public class DefaultKeystoreService implements KeystoreService {
   // Package private for unit test access
   // We need this to be synchronized to prevent multiple threads from using at once
   synchronized KeyStore createKeyStore(Path keystoreFilePath, String keystoreType, char[] password) throws KeystoreServiceException {
-    if (Files.notExists(keystoreFilePath)) {
     // Ensure the parent directory exists...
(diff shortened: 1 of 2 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
```diff
@@ -513,12 +513,14 @@ private synchronized boolean isKeyStoreAvailable(final Path keyStoreFilePath, St
   // Package private for unit test access
   // We need this to be synchronized to prevent multiple threads from using at once
   synchronized KeyStore createKeyStore(Path keystoreFilePath, String keystoreType, char[] password) throws KeystoreServiceException {
-    if (Files.notExists(keystoreFilePath)) {
-      // Ensure the parent directory exists...
-      try {
+    // Ensure the parent directory exists...
+    // This is symlink safe.
+    Path parentPath = keystoreFilePath.getParent();
```
lmccay, line 518: @jameschen1519 - it is possible for this method to return null, I believe.
I am not sure what the implications on the next line would be in that case,
I think that you should extend the existing test cases in DefaultKeystoreServiceTest to provide proper assurances here.

lmccay (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultKeystoreService.java
jameschen1519: Thanks for the catch. I think it would be sufficient to just add a null check to the if statement, since we want the call to getParent to be successful when the keystore directory happens to be the root directory.

I'll try adding in a test case to check if this fix works if the last directory is a symlink.

jameschen1519: Test has been added.

lmccay (approved):
LGTM
+1

lmccay:
@jameschen1519 - thank you for this contribution!
It will be included in the upcoming 1.5.0 release.

PR #391, opened 2020-12-08
KNOX-2479 - Fix an issue where Knox munges set-cookie header

moresandeep:

Fix an issue where set-cookie responses from are broken when backend services send set-cookies with no spaces.

## How was this patch tested?
Locally tested

### DIFF
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultDispatch.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultDispatch.java
@@ -382,7 +382,14 @@ public class DefaultDispatch extends AbstractGatewayDispatch {
           return ""; // we should exclude all -> there should not be any value added with this header
         } else {
           final String separator = SET_COOKIE.equalsIgnoreCase(headerNameToCheck) ? "; " : " ";
-          Set<String> headerValuesToCheck = new HashSet<>(Arrays.asList(headerToCheck.getValue().trim().split("\\s+")));
+          Set<String> headerValuesToCheck;
+          if(headerToCheck.getName().equalsIgnoreCase(SET_COOKIE)) {
+              headerValuesToCheck = new HashSet<>(Arrays.asList(headerToCheck.getValue().trim().split(";")));
+              /* trim */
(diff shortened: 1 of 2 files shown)


smolnar82 (approved):
LGTM

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultDispatch.java
```diff
@@ -382,7 +382,14 @@ private String calculateResponseHeaderValue(Header headerToCheck, Map<String, Se
...
-          Set<String> headerValuesToCheck = new HashSet<>(Arrays.asList(headerToCheck.getValue().trim().split("\\s+")));
+          Set<String> headerValuesToCheck;
+          if(headerToCheck.getName().equalsIgnoreCase(SET_COOKIE)) {
+              headerValuesToCheck = new HashSet<>(Arrays.asList(headerToCheck.getValue().trim().split(";")));
+              /* trim */
+              headerValuesToCheck = headerValuesToCheck.stream().map(String::trim).collect(Collectors.toSet());
+          } else {
+              headerValuesToCheck = new HashSet<>(Arrays.asList(headerToCheck.getValue().trim().split("\\s+")));
+          }
           headerValuesToCheck = headerValuesToCheck.stream().map(h -> h.replaceAll(separator.trim(), "")).collect(Collectors.toSet());
```
pzampino, line 393: Why the need to trim the separator here? Why not set the value to the trimmed form in the first place?

moresandeep: Sorry i did not follow, in case of line 389 we need trimming for the case where cookies use spaces and ; together.

moresandeep: I see it now, line #393 can be eliminated by trimming #384 
I'll test the changes and update the patch, hopefully nothing breaks :)

pzampino: Ok, I see separator gets used for the output as well. LGTM.

pzampino (approved):

PR #398, opened 2021-01-20
KNOX-2530 - Support qualifying service params for CM discovery control

pzampino:

In the case of some services, or specific deployments of those services, there exists the need to direct the discovery process for topology generation. One example is HA HDFS deployments, where a particular nameservice can be desired to be proxied. Another example is multiple Solr instances within the same cluster, where one should be proxied but not the other(s).

This change introduces for Cloudera Manager discovery the concept of discovery-time-only service parameters in simple descriptors. This is already supported by Amabari service discovery for the aforementioned HDFS HA scenario, and the same pattern is being applied and extended for Cloudera Manager.

Service parameters beginning with "discovery-" can be declared for supported services in simple descriptors. The actual names of the parameters are service-specific.

This is an example for the HDFS nameservice selection:
`"services": [
  {
    "name": "WEBHDFS",
    "params": {
      "discovery-nameservice": "ns2"
    }
  },`

These discovery-time-only service params do not manifest in the resulting generated topology.

## How was this patch tested?

Manual testing, augmented/modified unit tests for the affected ServiceModelGenerator implementations.

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ServiceModel.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ServiceModel.java
@@ -37,6 +37,9 @@ public class ServiceModel {
   private final String roleType;
   private final String serviceUrl;
 
+  // Metadata for the model object, which is not directly from the service or role configuration
+  private final Map<String, String> qualifyingServiceParams = new ConcurrentHashMap<>();
+
   // The service configuration properties used to created the model
   private final Map<String, String> serviceConfigProperties = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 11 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ServiceModel.java
```diff
@@ -37,6 +37,9 @@
   private final String roleType;
   private final String serviceUrl;
 
+  // Metadata for the model object, which is not directly from the service or role configuration
+  private final Map<String, String> modelProperties = new ConcurrentHashMap<>();
```
smolnar82, line 41: I'm wondering if `qualifyingServiceParameters` would be a better name for this field?

pzampino: You're probably correct. I went back and forth on that naming choice.

moresandeep (approved):
Looks good

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ServiceModel.java
smolnar82: Other than that it looks good to me. Will you submit a new PS?

pzampino: New PS in progress...

smolnar82 (approved):
LGTM

PR #399, opened 2021-01-21
KNOX-2531 Kill Application button in YARN does not work through KNOX

iain-buclaw-sociomantic:

The Kill Application button is broken, KNOX matches and rewrites the URL twice into an invalid URL.

This change removes the parentheses around the first pattern for `/ws/v1/`, so that relative paths are not rewritten twice.

It seems to work, though I'm not sure if it silently breaks other URLs just yet.

## How was this patch tested?

Tested in local production cluster. Spawned up a new spark job and killed it via the Kill Application button in YARN.

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
@@ -207,7 +207,7 @@
 </filter>
 <filter name="YARNUI/yarn/outbound/filter/cluster">
     <content type="*/html">
-        <apply path="(https?://[^/':,]+:[\d]+)?/ws/v1/cluster/apps/application" rule="YARNUI/yarn/outbound/apps/cluster1"/>
+        <apply path="https?://[^/':,]+:[\d]+/ws/v1/cluster/apps/application" rule="YARNUI/yarn/outbound/apps/cluster1"/>
         <apply path="/ws/v1/.*" rule="YARNUI/yarn/outbound/ws1"/>
         <apply path="https?://[^/':,]+:[\d]+/cluster/scheduler.*" rule="YARNUI/yarn/outbound/scheduler1"/>
         <apply path="/cluster/scheduler.*" rule="YARNUI/yarn/outbound/scheduler2"/>


gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
```diff
@@ -219,6 +218,11 @@
         <apply path="https?://[^/':,]+:[\d][^']*" rule="YARNUI/yarn/outbound/node2"/>
     </content>
 </filter>
+<filter name="YARNUI/yarn/outbound/filter/cluster1">
```
moresandeep, line 221: So, how is this filter called? I don't see any matching route for `YARNUI/yarn/outbound/filter/cluster1` in service.xml

moresandeep (changes requested):

gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
iain-buclaw-sociomantic: That's probably why it works for me then. :-)

iain-buclaw-sociomantic: @moresandeep going to try altering the path for the `YARNUI/yarn/outbound/apps/cluster1` then.  Currently rebuilding and deploying locally.

iain-buclaw-sociomantic: > @moresandeep going to try altering the path for the `YARNUI/yarn/outbound/apps/cluster1` then. Currently rebuilding and deploying locally.

So I see no problems yet running with this patch.  The Kill button works, and everything else seems functional (except for the history url noted in KNOX-2528/KNOX-1032).

If I understand the rules and patterns correctly, I can't imagine anything should really break as a result of this, because `/ws/v1/.*` catches and handles `/ws/v1/cluster/apps/application` correctly.

gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
```diff
@@ -207,7 +207,7 @@
 </filter>
 <filter name="YARNUI/yarn/outbound/filter/cluster">
     <content type="*/html">
-        <apply path="(https?://[^/':,]+:[\d]+)?/ws/v1/cluster/apps/application" rule="YARNUI/yarn/outbound/apps/cluster1"/>
+        <apply path="https?://[^/':,]+:[\d]+/ws/v1/cluster/apps/application" rule="YARNUI/yarn/outbound/apps/cluster1"/>
```
moresandeep, line 210: My regex skills are rusty :( can you explain briefly what changed here.

moresandeep (commented):
Thanks @iain-buclaw-sociomantic and sorry for the delay in responding. Can you just explain what changed here? I'll spin up a cluster and test the patch out.

gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
iain-buclaw-sociomantic: So, the following two regexes:
```
(https?://[^/':,]+:[\d]+)?/ws/v1/cluster/apps/application
/ws/v1/.*
```
Both match the following link of the kill button in YARN.
```
/ws/v1/cluster/apps/application_1611223368889_0003/state
```
You can confirm this here https://regex101.com/r/QkGNeb/1

Because of this, both the `outbound/apps/cluster1` and `outbound/ws1` rules got applied on the link, resulting in a broken URL.

What this change does is make `https?://[^/':,]+:[\d]+` an _explicit_ match, rather than an optional one, so the `outbound/apps/cluster1` rule will now only ever be applied to links in the form of:
```
http://some-internal-domain:1234/ws/v1/cluster/apps/application
```

pzampino (approved):
LGTM

gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
smolnar82: @moresandeep @iain-buclaw-sociomantic 's explanation seems reasonable to me. Shall we merge this PR?

PR #405, opened 2021-03-01
KNOX-2540: fix dependency conflict issue

Bing-ok:
[KNOX-2540](https://issues.apache.org/jira/browse/KNOX-2540): fix dependency conflict issue for org.apache.curator:curator-client:jar

## How was this patch tested?
 running automated unit/integration tests


risdenk (changes requested):
This should not be done in `hadoop-examples/pom.xml`. There is a Hadoop version in the top level pom. In general, no versions should be outside the top level pom.xml.

PR #410, opened 2021-03-12
KNOX-2547 - Token-based providers should perform signature verificati…

pzampino:
…on last

Moved token signature verification to be performed after all other token validations. I've also increased the signature verification cache maximum to 250 and added the explicit removal of signature verification records for expired tokens.

## How was this patch tested?

Added AbstractJWTFilterTest#testExpiredTokensEvictedFromSignatureVerificationCache() to validate the explicit eviction of expired tokens' signature verification. Ran all other existing tests.


lmccay (commented):
+1

PR #416, opened 2021-03-16
KNOX-2553 - Adding the 'managed_token' flag in the issued JWT as well as the JSON response of our KnoxToken service

smolnar82:

In order to avoid adding new and confusing methods into `JWTokenAuthority`, I introduced a model object to store possible JWT attributes and kept only one method to issue a token.
So, from now on, the `managed` flag is available in:

- the JSON response as a new element
- the issued JWT token within a claim called `managed.token`

## How was this patch tested?

Updated/added/ran unit tests and executed manual testing:

1. added the `KNOXTOKEN` service into `sandbox`. Set the `knox.token.exp.server-managed` to `false` and then changed it to `true` (see results below)
2. added a new topology called `tokenbased`
3. I could login to the Knox Home page and to Admin UI
4. Issued the following `curl` commands

4.1 `knox.token.exp.server-managed = false`
```
$ curl -iku admin:admin-password https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token
HTTP/1.1 200 OK
...

{"access_token":"eyJhbGciO...4kjahYA","token_id":"49fd0e9e-a5d9-4a14-86cf-c11e5bac151d","managed":"false","target_url":"https://localhost:8443/gateway/tokenbased","endpoint_public_cert":"MIIDeTCCAmGgAwI...M0XFRx+jhA=","token_type":"Bearer","expires_in":1615886851180}
```
4.2 `knox.token.exp.server-managed = true`
```
$ curl -iku admin:admin-password https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token
HTTP/1.1 200 OK
...

{"access_token":"eyJhbGciO...F0rC4A0g","token_id":"efbed646-2210-4731-9dab-71df641943b0","managed":"true","target_url":"https://localhost:8443/gateway/tokenbased","endpoint_public_cert":"MIIDeTCCAmGgAwI...M0XFRx+jhA=","token_type":"Bearer","expires_in":1615887095677}
```
5. Verified both tokens using the `tokenbased` topology.

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
@@ -38,6 +38,8 @@ import org.apache.knox.gateway.provider.federation.jwt.JWTMessages;
 import org.apache.knox.gateway.services.ServiceType;
 import org.apache.knox.gateway.services.GatewayServices;
 import org.apache.knox.gateway.services.registry.ServiceRegistry;
+import org.apache.knox.gateway.services.security.token.JWTokenAttributes;
+import org.apache.knox.gateway.services.security.token.JWTokenAttributesBuilder;
 import org.apache.knox.gateway.services.security.token.JWTokenAuthority;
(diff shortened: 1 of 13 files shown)


pzampino (approved):
I'm sure I've missed some discussion, so the new "managed" claim feels odd to me (I think since it reflects server-side config if I'm understanding correctly). That point aside, these changes LGTM.

PR #423, opened 2021-03-24
KNOX-2559: Added functionality to append headers in request in Config…

mudit-97:
JIRA: https://issues.apache.org/jira/browse/KNOX-2559

Added Support of requestAppendHeaders in ConfigurableDispatch so if any service.xml provides this param like this:
<param>
<name>requestAppendHeaders</name>
<value>{"a":"b"}</value>
</param>

Requests getting dispatched from this service will have an extra header a having value b along with the other headers

## How was this patch tested?

Added 3 Testcases for ConfigurableDispatch Class covering 3 cases:
1. When some new headers were appended and no headers were excluded --> Default Exclusion of header took place and new headers were appended
2. When some new headers were appended and some headers excluded --> Both exclusion and append worked as expected
3. When a header from request was excluded but same was also appended but with new value --> Header came with new value as per append logic

Apart from this, tested this scenario on local Knox build. Headers were getting appended

### DIFF
--- a/gateway-spi/pom.xml
+++ b/gateway-spi/pom.xml
@@ -160,6 +160,15 @@
         </dependency>
 
         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.knox</groupId>
             <artifactId>gateway-test-utils</artifactId>
             <scope>test</scope>
@@ -169,15 +178,6 @@
             <groupId>org.apache.velocity</groupId>
             <artifactId>velocity</artifactId>
             <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-core</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/gateway-spi/pom.xml
+++ b/gateway-spi/pom.xml
@@ -160,15 +160,6 @@
         </dependency>
 
         <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-core</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>org.apache.knox</groupId>
             <artifactId>gateway-test-utils</artifactId>
             <scope>test</scope>
--- a/gateway-spi/pom.xml
+++ b/gateway-spi/pom.xml
@@ -170,5 +170,14 @@
             <artifactId>velocity</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/AbstractGatewayDispatch.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/AbstractGatewayDispatch.java
@@ -30,8 +30,10 @@ import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 public abstract class AbstractGatewayDispatch implements Dispatch {
@@ -132,10 +134,20 @@ public abstract class AbstractGatewayDispatch implements Dispatch {
         outboundRequest.addHeader( name, value );
       }
     }
+
+    //If there are some headers to be appended, append them
+    Map<String, String> extraHeaders = getOutboundRequestAppendHeaders();
+    if(null != extraHeaders){
+      extraHeaders.keySet().forEach(header -> outboundRequest.addHeader(header, extraHeaders.get(header)));
+    }
   }
 
   public Set<String> getOutboundRequestExcludeHeaders() {
     return REQUEST_EXCLUDE_HEADERS;
+  }
+
+  public Map<String, String> getOutboundRequestAppendHeaders() {
+    return Collections.emptyMap();
   }
 
   protected void encodeUnwiseCharacters(StringBuffer str) {
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/AbstractGatewayDispatch.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/AbstractGatewayDispatch.java
@@ -30,10 +30,8 @@ import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.Enumeration;
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Set;
 
 public abstract class AbstractGatewayDispatch implements Dispatch {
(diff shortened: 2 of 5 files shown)


gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/ConfigurableDispatch.java
```diff
@@ -41,13 +47,26 @@
...
   private Set<String> responseExcludeHeaders = super.getOutboundResponseExcludeHeaders();
+  private Map<String, String> requestAppendHeaders = new HashMap<>();
   private Set<String> responseExcludeSetCookieHeaderDirectives = super.getOutboundResponseExcludedSetCookieHeaderDirectives();
   private Boolean removeUrlEncoding = false;
 
   private Set<String> convertCommaDelimitedHeadersToSet(String headers) {
     return headers == null ?  Collections.emptySet(): new HashSet<>(Arrays.asList(headers.split("\\s*,\\s*")));
   }
 
+  private Map<String, String> convertJSONHeadersToMap(String headers) {
```
moresandeep, line 58: Putting JSON in XML looks wrong, we should follow the same pattern here as we follow elsewhere to extract parameters from service.xml file. I think NiFi dispatch does something like that with twoWaySSL.

moresandeep (changes requested):
Thanks for the patch, needs few changes otherwise looks good.

moresandeep:
@lmccay @pzampino @smolnar82 thougts?

gateway-spi/pom.xml
```diff
@@ -170,5 +170,14 @@
             <artifactId>velocity</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
```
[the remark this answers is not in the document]
mudit-97: ack

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/AbstractGatewayDispatch.java
```diff
@@ -132,12 +134,22 @@ public void copyRequestHeaderFields(HttpUriRequest outboundRequest,
         outboundRequest.addHeader( name, value );
       }
     }
+
+    //If there are some headers to be appended, append them
+    Map<String, String> extraHeaders = getOutboundRequestAppendHeaders();
+    if(null != extraHeaders){
```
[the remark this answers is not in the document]
mudit-97: ack, used MapUtils.isNotEmpty

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/AbstractGatewayDispatch.java
```diff
@@ -132,12 +134,22 @@ public void copyRequestHeaderFields(HttpUriRequest outboundRequest,
...
+      extraHeaders.keySet().forEach(header -> outboundRequest.addHeader(header, extraHeaders.get(header)));
```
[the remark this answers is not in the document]
mudit-97: ack

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/AbstractGatewayDispatch.java
```diff
@@ -132,12 +134,22 @@ public void copyRequestHeaderFields(HttpUriRequest outboundRequest,
         outboundRequest.addHeader( name, value );
       }
     }
+
+    //If there are some headers to be appended, append them
+    Map<String, String> extraHeaders = getOutboundRequestAppendHeaders();
```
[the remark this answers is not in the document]
mudit-97: ack

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/ConfigurableDispatch.java
```diff
@@ -41,13 +47,26 @@
 public class ConfigurableDispatch extends DefaultDispatch {
   private Set<String> requestExcludeHeaders = super.getOutboundRequestExcludeHeaders();
   private Set<String> responseExcludeHeaders = super.getOutboundResponseExcludeHeaders();
+  private Map<String, String> requestAppendHeaders = new HashMap<>();
```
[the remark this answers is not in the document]
mudit-97: ack

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/ConfigurableDispatch.java
mudit-97: I had some doubts around this. I checked that in CustomDispatch, there is one xml param support:

@XmlAttribute(name = "use-two-way-ssl")
  public boolean getUseTwoWaySsl() {
    return useTwoWaySsl;
  }
and xml will look like this:
<requestAppendHeaders>
    <header>
      <name>a</name>
      <value>b</value>
    </header>
    <header>
      <name>c</name>
      <value>d</value>
    </header>
  </requestAppendHeaders>

Are you suggesting that we should have a new parameter requestAppendHeaders

or are you suggesting a layout like this:

<param>
  <requestAppendHeaders>
    <header>
      <name>a</name>
      <value>b</value>
    </header>
    <header>
      <name>c</name>
      <value>d</value>
    </header>
  </requestAppendHeaders>
</param>

mudit-97: Analysed both the approaches which I posted above, first one looks like we are allowing a new special variable which does not come under param scope, that looked a little bad to me
Second one however looks functionally more correct but will require many changes in param structure in  CustomDispatch class, which currently allows only name, value pairs inside it

Not sure what should be done here

moresandeep: The twoWaySSL is implemented as attribute :( 
I was thinking using <params> to keep the XML generic, something like 
` <param>
            <name>requestAppendHeaders</name>
            <value>a:b,c:d</value>
        </param>`

Now, that brings up an interesting point, what happens if the header contains `:` how do we handle it?

moresandeep: Or we could have a attribute `requestAppendHeaders` on `param` and use name value.

mudit-97: Sorry for the delay in response here, I have pushed a new commit. What I am trying to do is:

Support for a values inside DispatchParam. So either a param can be a simple param like:
<param>
<name>a</name>
<value>b</value>
</param>

Or it can be a complex param like this:

<param>
<name>a</name>
<values>
  <param>
    <name>a1</name>
    <value>b1</value>
  </param>
  <param>
    <name>a2</name>
    <value>b2</value>
  </param>
</values>
</param>

mudit-97: I have not added test cases for these yet, please let me know if this sounds fine to you

moresandeep: Apologies for the delay, the simple param example looks good to me, I cannot think of a case where we need complex param, do you have one in mind? AFA headers are concerned they are simple <key><value> so I do not think we need to complicate this, now depending on header the separator might vary (e.g. SET-COOKIE header) but that should not be a problem here. We could do a simple validation to make sure that the header values don't break the request. 

Test would be great to have we don't this to break accidentally due to any later change.
Thanks.

mudit-97: yeah I also think the same, I felt that I am complicating the code but was not sure about your view. So, now can I implement something like this: <param> <name>requestAppendHeaders</name> <value>a:b,c:d</value> </param>

mudit-97: I will add relevant checks and guardrails with logs

moresandeep: Sounds good, just to clarify what I mean following are few examples
```
<param> 
     <name>requestAppendHeaders</name> 
     <value>Transfer-Encoding:gzip, chunked; Keep-Alive: timeout=5, max=1000</value> 
</param>
```
Here `;` is used as separator BUT `;` is used in SET-COOKIE header e.g.
`Set-Cookie: <cookie-name>=<cookie-value>; Domain=<domain-value>; Secure; HttpOnly`

The question here is should we be allowing to set headers like SET-COOKIE? if yes how do we handle `;` 
one way would be to use an escape `\;` 

Also, I think your patch is for inbound requests only right? i.e. requests coming in from client to Knox and knox setting headers on requests it dispatches, not the other way around. If it is both ways then we might have some issues. 
See, now this is starting to look complicated :)

mudit-97: @moresandeep  I have done all the changes from my side, please check once. Added testcases also

moresandeep: Thank you @mudit-97 I will build the patch and check it out, thanks for quick turnaround.

mudit-97: @moresandeep  let me know if any more changes are required

mudit-97: @moresandeep  did you get time to check this?, please let me know if further changes are required

moresandeep: Sorry @mudit-97 got buried in other stuff, this is still on top of my todo list. Last week I tried to build a docker container for your patch (https://github.com/moresandeep/knox-dev-docker) but it failed, so i'll have to manually download the patch and check out.

mudit-97: Thanks for the update

mudit-97: @moresandeep , I tried testing this on my local, that was working fine. I tested one positive and one negative case, let me know if you face any issues

moresandeep (approved):
Looks good

PR #426, opened 2021-03-30
KNOX-2562 - TokenStateService getTokenMetadata method should throw Un…

pzampino:
…knownTokenException

To align with other methods in the TokenStateService interface, the getTokenMetadata(String tokenId) method has modified to throw UnknownTokenException when there is no metadata for the specified token identifier.

## How was this patch tested?

Existing tests have been modified/executed, and I've added org.apache.knox.gateway.services.token.impl.DefaultTokenStateServiceTest#testGetMetadata_InvalidToken(), which also gets invoked for the Alias- and Journal-based TokenStateService implementations.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
@@ -446,13 +446,22 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService implem
   }
 
   @Override
-  public TokenMetadata getTokenMetadata(String tokenId) {
-    TokenMetadata tokenMetadata = super.getTokenMetadata(tokenId);
+  public TokenMetadata getTokenMetadata(String tokenId) throws UnknownTokenException {
+    TokenMetadata tokenMetadata = null;
+    try {
+      tokenMetadata = super.getTokenMetadata(tokenId);
+    } catch (UnknownTokenException e) {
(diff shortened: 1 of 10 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
```diff
@@ -446,13 +446,15 @@ protected void addMetadataInMemory(String tokenId, TokenMetadata metadata) {
   }
 
   @Override
-  public TokenMetadata getTokenMetadata(String tokenId) {
+  public TokenMetadata getTokenMetadata(String tokenId) throws UnknownTokenException {
     TokenMetadata tokenMetadata = super.getTokenMetadata(tokenId);
```
smolnar82, line 450: UnknownTokenException should be handled. In HA case it may happen that the in-memory collection will not have the given token metadata -> super will throw UTE. We should catch it and continue with the alias service action.

smolnar82 (changes requested):

moresandeep (approved):

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
pzampino: Do you think we need to log the fact that it's missing from the in-memory record? I think not, but I'm interested in your opinion.

smolnar82 (approved):
LGTM

PR #427, opened 2021-03-31
KNOX-2566 - JWT Token Signature Verification Caching NPE

pzampino:

Added some checks for missing Knox token UUID claim around the signature verification caching (which was added as part of KNOX-2544) to avoid NullPointerException when JWTs which were not issued by Knox (but which Knox can verify) are received.

## How was this patch tested?

Added org.apache.knox.gateway.provider.federation.AbstractJWTFilterTest#testJWTWithoutKnoxUUIDClaim() and org.apache.knox.gateway.provider.federation.AbstractJWTFilterTest#testExpiredJWTWithoutKnoxUUIDClaim()to reproduce the NPE condition and then to verify the fix. Ran all other existing tests as part of the build.

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
@@ -69,4 +69,8 @@ public interface JWTMessages {
             text = "The configuration value ({0}) for maximum token verification cache is invalid; Using the default value." )
   void invalidVerificationCacheMaxConfiguration(String value);
 
+  @Message( level = MessageLevel.ERROR,
+            text = "Missing token passcode." )
+  void missingTokenPasscode();
+
 }
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
@@ -357,10 +357,12 @@ public abstract class AbstractJWTFilter implements Filter {
         } else {
           log.tokenHasExpired(displayableToken, displayableTokenId);
 
-          // Explicitly evict the record of this token's signature verification (if present).
-          // There is no value in keeping this record for expired tokens, and explicitly removing them may prevent
-          // records for other valid tokens from being prematurely evicted from the cache.
-          removeSignatureVerificationRecord(tokenId);
+          if (tokenId != null) {
+            // Explicitly evict the record of this token's signature verification (if present).
+            // There is no value in keeping this record for expired tokens, and explicitly removing them may prevent
+            // records for other valid tokens from being prematurely evicted from the cache.
+            removeSignatureVerificationRecord(tokenId);
+          }
 
           handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
                                 "Bad request: token has expired");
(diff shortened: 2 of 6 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -385,12 +387,18 @@ protected boolean validateToken(final HttpServletRequest request,
...
+          } else {
+            log.tokenHasExpired(Tokens.getTokenIDDisplayText(tokenId));
+            handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
+                    "Bad request: token has expired");
+          }
         } else {
-          log.tokenHasExpired(Tokens.getTokenIDDisplayText(tokenId));
+          log.missingTokenPasscode();
           handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-                                "Bad request: token has expired");
```
lmccay, line 393: It seems has though we are assuming that if token state server is enabled for a given KnoxToken service that it cannot also accept 3rd party tokens. Is this a necessary requirement? Would falling back to only using the token expiry be acceptable? It is sort of unclear how we would verify the signature given today's approach to signing keys but if we were to add support for jwks for getting the public key for verification, it seems that we could fully support 3rd party tokens as well as KnoxTokens in the same provider deployment. If we waited for the jwks support to make that work it may be harder to find and fix these.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -407,7 +415,7 @@ protected boolean verifyTokenSignature(final JWT token) {
     String tokenId = TokenUtils.getTokenId(token);
 
     // Check if the token has already been verified
-    verified = hasSignatureBeenVerified(tokenId);
+    verified = (tokenId != null) && hasSignatureBeenVerified(tokenId);
```
lmccay, line 418: if we do decide to support 3rd party tokens, we would likely want to make sure that such tokens also benefit from the caching. Which seems to imply that we would either need to derive a tokenId from the 3rd party token somehow or use the actual token as the id. The latter would have uniqueness implications across jobs but may be okay for this optimization. If the token is exactly the same then it is the same user and created at exactly the same time with the exact same expiry?

lmccay (changes requested):

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
```diff
@@ -69,4 +69,8 @@
             text = "The configuration value ({0}) for maximum token verification cache is invalid; Using the default value." )
   void invalidVerificationCacheMaxConfiguration(String value);
 
+  @Message( level = MessageLevel.ERROR,
```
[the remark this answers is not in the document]
pzampino: The context from which this is logged is not applicable to the JWT scenario. It is only ever logged when the token ID is used as a passcode, which requires server-managed token state to be enabled because we can't authenticate based on the ID alone (we need the expiry/metadata). In this context, I believe this is an error.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
pzampino: Again, this code is not applicable to the cases where a JWT is provided for authentication; It is only for those cases where ONLY the token ID is provided, in which case the server-managed token state is absolutely necessary since there is nothing about the ID which can be verified by itself.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
pzampino: This is a valid point, and one that I would argue can be addressed as part of revisiting the the signature verification caching implementation altogether.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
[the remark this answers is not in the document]
lmccay: okay.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
lmccay: okay.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
lmccay: I can buy that.

lmccay:
+1

PR #428, opened 2021-04-06
KNOX-2565 - Fix login sso with oidc error

cdmikechen:

Fix JIRA issue https://issues.apache.org/jira/browse/KNOX-2565

## How was this patch tested?

Have tested login with keycloak (10.0.2) oidc.


lmccay:
Based on new comments in https://issues.apache.org/jira/browse/KNOX-2565, it seems that perhaps this is not needed. Is there any functional value to the changes that you have made here that we would not gain with the upgrade to pac4j 4.3.0?
If so, we would need these described and additional unit tests added.

PR #429, opened 2021-04-06
KNOX-2570 - Add support for JWKS endpoint

moresandeep:
This PR adds support for JWKS endpoint that exposes the public key used to sign the JWT tokens. The format used is the standard JWKS format. 

This is an example of rest request
` curl -ku admin:admin-password 'https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/.well-known/jwks.json`

and the response
```
{
   "keys":[
      {
         "kty":"RSA",
         "e":"AQAB",
         "use":"sig",
         "kid":"7T805L6i36S0U0mVcO\/WjuPD1TtNehUXjQa7uBdet34=",
         "alg":"RS256",
         "n":"t2muUKwV9Zx8Q3NWy5vhu_EszpBkfa8cZjhJpF8IHNkx2kbf8aeRTr3KwGInPYx6bVVD1eNzjDX_m5afyHfQr4O_PHSiaxtJPN16kmoVnxQW8ZxnT2GlMUY6TL2t79Ovrhnkig-4_rP5qH0ruZZJpNcly-FLKoMd23apy8BCe9lSaSIbWUVU5YrVzWzuO3TWILSQCoIS8-4j6G0JmzaOu4t3f0yATE7WaJbqW9UmEAyTDIlft0Xnh1Oz_UHWAe5LLVYyVh6Xr8lMUOBj1oAVAKLJrd0ZeKb-UerTKuCzuRM2yI-idXCZFaJQ3x4KklYihL6Qc7FPjRestPxIeBJl1w"
      }
   ]
}
```

## How was this patch tested?
The patch was tested locally.

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.service.knoxtoken;
+
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.jwk.JWKSet;
+import com.nimbusds.jose.jwk.KeyUse;
+import com.nimbusds.jose.jwk.RSAKey;
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.services.GatewayServices;
+import org.apache.knox.gateway.services.ServiceType;
+import org.apache.knox.gateway.services.security.KeystoreService;
+import org.apache.knox.gateway.services.security.KeystoreServiceException;
+
+import javax.annotation.PostConstruct;
+import javax.inject.Singleton;
+import javax.servlet.ServletContext;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.QueryParam;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import java.security.KeyStoreException;
+import java.security.interfaces.RSAPublicKey;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+@Singleton
+@Path(JWKSResource.RESOURCE_PATH)
+public class JWKSResource {
+
+  static final String RESOURCE_PATH = "knoxtoken/api/v1/.well-known";
+  static final String JWKS_PATH = "/jwks.json";
+  @Context
+  HttpServletRequest request;
+  @Context
+  ServletContext context;
+  private KeystoreService keystoreService;
+
+  @PostConstruct
+  public void init() {
+    final GatewayServices services = (GatewayServices) context
+        .getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
+    keystoreService = services.getService(ServiceType.KEYSTORE_SERVICE);
+  }
+
+  @GET
+  @Path(JWKS_PATH)
+  @Produces({ APPLICATION_JSON })
+  public Response getJwksResponse() {
+    return getJwks(null);
+  }
+
+  private Response getJwks(final String keystore) {
+    JWKSet jwks;
+    try {
+      final RSAPublicKey rsa = getPublicKey(keystore);
+      final RSAKey.Builder builder = new RSAKey.Builder(rsa)
+          .keyUse(KeyUse.SIGNATURE)
+          .algorithm(new JWSAlgorithm(rsa.getAlgorithm()))
+          .keyIDFromThumbprint();
+
+      jwks = new JWKSet(builder.build());
+
+    } catch (KeyStoreException | JOSEException e) {
+      return Response.status(500)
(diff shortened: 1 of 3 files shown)


gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
```diff
@@ -0,0 +1,108 @@
...
+import java.security.KeyStoreException;
+import java.security.interfaces.RSAPublicKey;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+@Singleton
+@Path(JWKSResource.RESOURCE_PATH)
+public class JWKSResource {
+
+  static final String RESOURCE_PATH = "knoxtoken/api/v1/.well-known";
```
lmccay, line 50: I'd like to pick at the use of .well-known here - perhaps based on ignorance though.
My assertion would be that if you aren't providing this in a central location host:port/gateway/.well-known sort of thing then it is not actually well known.
I don't know that it is adding any value here.
Perhaps, however, it is referring to the actualy "/jwks.json" as being a well known resource?
Maybe it is just part of the standard that I am unaware of.

I'd like to hear the rationale for it and what value it proposes.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
```diff
@@ -0,0 +1,108 @@
...
+              + "\"\n}\n").build();
+    }
+    return Response.ok()
+        .entity(jwks.toJSONObject().toString()).build();
+  }
+
+  protected RSAPublicKey getPublicKey(final String keystore)
+      throws KeystoreServiceException, KeyStoreException {
+    return (RSAPublicKey) keystoreService.getSigningKeystore(keystore)
+        .getCertificate(getSigningKeyAlias()).getPublicKey();
```
lmccay, line 98: Need to insure that there is no NPE here. We have added HMAC/shared secret support as well. So it is possible that in this case there will be a null signingKeystore and if not that will be a possible bug in those deployments that needs to be addressed.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private String getSigningKeyAlias() {
+    final GatewayConfig config = (GatewayConfig) request.getServletContext()
+        .getAttribute(GatewayConfig.GATEWAY_CONFIG_ATTRIBUTE);
+    final String alias = config.getSigningKeyAlias();
```
lmccay, line 104: may need a null check.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
```diff
@@ -0,0 +1,108 @@
...
+    return (alias == null) ? GatewayConfig.DEFAULT_SIGNING_KEY_ALIAS : alias;
```
lmccay, line 105: should getSigningKeyAlias be changed to return the default if it isn't explicitly set? Is that all this is doing?

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/JWKSResourceTest.java
```diff
@@ -0,0 +1,144 @@
...
+    claimArray[3] = null;
+
+    final String algorithm = "RS256";
+    final JWT token = new JWTToken(algorithm, claimArray,
+        Collections.singletonList("aud"), false);
+    final JWSSigner signer = new RSASSASigner(privateKey, true);
+    token.sign(signer);
+    return token;
+  }
+
```
lmccay, line 143: Can you add a test for when shared secret is being used to insure that we do not NPE or the like somewhere along the line? Just to tease out assumptions.

lmccay (changes requested):
Thanks for this PR, Sandeep!
This is a great improvement and will lead to easier provisioning, deployment and management.

We may need to consider how cert rotation, rolling may affect this. I think you may end up with more than one. There is another attribute called kid that I think you are supposed to match with the cert or map internally within a cache in the client for tokens that are signed by different key material. This could possibly be added in a follow up change as well.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
moresandeep: There is no official standard as such. I was just going by the pattern used by services such as [auth0](https://auth0.com/docs/tokens/json-web-tokens/json-web-key-sets/locate-json-web-key-sets) This is an example
`https://YOUR_DOMAIN/.well-known/openid-configuration`
I do agree that this does not add any value given it not a well known endpoint (since it is part of a topology) there is no point in adding `.well-known` 
I will update it to `knoxtoken/api/v1/jwks.json`

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
moresandeep: I see, yeah for HMAC this might not work. prolly not blow up `keystoreService.getSigningKeystore(keystore)` is null safe the other ones can. I will update it.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
moresandeep: Ack

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
moresandeep: I do not fully follow the first question, do you mean change the `config.getSigningKeyAlias()` method?
But this method is just returning default signing key alias in case none is found.

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/JWKSResourceTest.java
moresandeep: Right, yes I will.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/JWKSResource.java
moresandeep: On second pass, there is a null check that checks for `(alias == null)` in the next line.

PR #432, opened 2021-04-07
KNOX-2572 - Unique token identifiers still being logged in entirety

pzampino:

Apparently, there are a number of places where the UUIDs associated with Knox JWTs are still being logged in their entirety.
This change is to mask all of those remaining messages.

## How was this patch tested?

Added ZookeeperTokenStateServiceTest#testTokenIDDisplayText() to test the masking of token ID in aliases.
Added MultiFileTokenStateJournalTest#testGetDisplayableJournalFilepathWithID() and MultiFileTokenStateJournalTest#testGetDisplayableJournalFilepathWithoutID() to test the masking of journal entry file paths that contain UUIDs.
The rest of the changes involving plain tokens uses the previously tested Tokens.getTokenIDDisplayText(String id) facility.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
@@ -45,6 +45,7 @@ import org.apache.knox.gateway.services.security.token.UnknownTokenException;
 import org.apache.knox.gateway.services.security.token.impl.JWT;
 import org.apache.knox.gateway.services.security.token.impl.JWTToken;
 import org.apache.knox.gateway.services.token.TokenStateServiceStatistics;
+import org.apache.knox.gateway.util.Tokens;
 
 /**
  * In-Memory authentication token state management implementation.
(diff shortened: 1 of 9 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
```diff
@@ -286,7 +287,15 @@ private void removeTokenState(final Set<String> tokenIds) {
     tokenExpirations.keySet().removeAll(tokenIds);
     maxTokenLifetimes.keySet().removeAll(tokenIds);
     metadataMap.keySet().removeAll(tokenIds);
-    log.removedTokenState(String.join(", ", tokenIds));
+    log.removedTokenState(String.join(", ", getDisplayableTokenIDsText(tokenIds)));
+  }
+
+  protected Set<String> getDisplayableTokenIDsText(final Set<String> tokenIds) {
```
smolnar82, line 293: It feels better to have this method in `Tokens`.

smolnar82 (approved):
LGTM; I only had a minor comment.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateService.java
pzampino: I agree.

smolnar82 (approved):

PR #433, opened 2021-04-13
KNOX-2554 - Implemented JDBC Token State Service

smolnar82:

This change introduces a new `TokenStateService` implementation that operates using pure JDBC connections to relational databases. In the first version, only `Postgres` and `Derby` database types are supported.

## How was this patch tested?

Added new JUnit test classes and updated existing test cases as needed.

I also executed manual testing:

1. Used our new `tokengen` application to fetch a token
```
2021-04-19 23:24:38,972 DEBUG knox.gateway (GatewayFilter.java:doFilter(116)) - Received request: GET /knoxtoken/api/v1/token
2021-04-19 23:24:38,982 INFO  service.knoxtoken (TokenResource.java:getAuthenticationToken(419)) - Knox Token service (homepage) issued token eyJhbG...WdNDTw (8c419cf4...675b1dc14cce)
2021-04-19 23:24:39,001 DEBUG token.state (JDBCTokenStateService.java:addToken(59)) - Token 8c419cf4...675b1dc14cce has been saved in the database
2021-04-19 23:24:39,002 DEBUG token.state (DefaultTokenStateService.java:addToken(144)) - Added token 8c419cf4...675b1dc14cce, expiration 2021-05-19T21:24:38.978Z
2021-04-19 23:24:39,011 DEBUG token.state (JDBCTokenStateService.java:addMetadata(186)) - Updated metadata for 8c419cf4...675b1dc14cce in the database
2021-04-19 23:24:39,012 DEBUG service.knoxtoken (TokenResource.java:getAuthenticationToken(446)) - Knox Token service (homepage) stored state for token eyJhbG...WdNDTw (8c419cf4...675b1dc14cce)

postgres=# select * FROM KNOX_TOKENS where token_id LIKE '8c419cf4%';
               token_id               |  issue_time   |  expiration   | max_lifetime  | username | comment 
--------------------------------------+---------------+---------------+---------------+----------+---------
 8c419cf4-ff57-4412-93dd-675b1dc14cce | 1618867478983 | 1621459478978 | 1619472278983 | admin    | 
(1 row)
```

2. Renewed the previously acquired token:
```
$ curl -ku admin:admin-password -d "@token.txt" -X POST https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token/renew
{
  "renewed": "true",
  "expires": "1618954035727"
}

2021-04-19 23:27:15,711 DEBUG knox.gateway (GatewayFilter.java:doFilter(116)) - Received request: POST /knoxtoken/api/v1/token/renew
2021-04-19 23:27:15,713 INFO  knox.gateway (KnoxLdapRealm.java:getUserDn(688)) - Computed userDn: uid=admin,ou=people,dc=hadoop,dc=apache,dc=org using dnTemplate for principal: admin
2021-04-19 23:27:15,740 DEBUG token.state (JDBCTokenStateService.java:updateExpiration(108)) - Updated expiration for 8c419cf4...675b1dc14cce in the database to 1,618,954,035,727
2021-04-19 23:27:15,740 DEBUG token.state (DefaultTokenStateService.java:renewToken(219)) - Renewed token 8c419cf4...675b1dc14cce, expiration 2021-04-20T21:27:15.727Z
2021-04-19 23:27:15,741 INFO  service.knoxtoken (TokenResource.java:renew(280)) - Knox Token service (sandbox) renewed the expiration for token eyJhbG...WdNDTw (8c419cf4...675b1dc14cce) (renewer=admin)

postgres=# select * FROM KNOX_TOKENS where token_id LIKE '8c419cf4%';
               token_id               |  issue_time   |  expiration   | max_lifetime  | username | comment 
--------------------------------------+---------------+---------------+---------------+----------+---------
 8c419cf4-ff57-4412-93dd-675b1dc14cce | 1618867478983 | 1618954035727 | 1619472278983 | admin    | 
(1 row)
```
3. Used that token to list `WEBHDFS` status. This time the configured `WEBHDFS` backend pointed to a non-existing cluster, but the point is that the supplied `Passcode` token could be validated. I also covered this case by restarting Knox and repeated the same step, so that a DB lookup was necessary as the token was not available in the in-memory collections:
```
$ curl -ku Passcode:8c419cf4-ff57-4412-93dd-675b1dc14cce https://localhost:8443/gateway/tokenbased/webhdfs/v1?op=LISTSTATUS
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 500 java.io.IOException: Service connectivity error.</title>
</head>
...
</html>


2021-04-19 23:32:49,216 DEBUG knox.gateway (GatewayFilter.java:doFilter(116)) - Received request: GET /webhdfs/v1
2021-04-19 23:32:49,336 DEBUG token.state (JDBCTokenStateService.java:getTokenExpiration(90)) - Fetched expiration for 8c419cf4...675b1dc14cce from the database : 1,618,954,035,727
2021-04-19 23:32:49,343 DEBUG token.state (JDBCTokenStateService.java:getTokenMetadata(215)) - Fetched metadata for 8c419cf4...675b1dc14cce from the database

postgres=# select * FROM KNOX_TOKENS where token_id LIKE '8c419cf4%';
               token_id               |  issue_time   |  expiration   | max_lifetime  | username | comment 
--------------------------------------+---------------+---------------+---------------+----------+---------
 8c419cf4-ff57-4412-93dd-675b1dc14cce | 1618867478983 | 1618954035727 | 1619472278983 | admin    | 
(1 row)
```
4. Revoked that token:
```
$ curl -ku admin:admin-password -d "@token.txt" -X POST https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token/revoke
{
  "revoked": "true"
}


2021-04-19 23:34:36,081 DEBUG knox.gateway (GatewayFilter.java:doFilter(116)) - Received request: POST /knoxtoken/api/v1/token/revoke
2021-04-19 23:34:36,124 INFO  knox.gateway (KnoxLdapRealm.java:getUserDn(688)) - Computed userDn: uid=admin,ou=people,dc=hadoop,dc=apache,dc=org using dnTemplate for principal: admin
2021-04-19 23:34:36,808 DEBUG token.state (DefaultTokenStateService.java:removeTokenState(290)) - Removed state for tokens 8c419cf4...675b1dc14cce
2021-04-19 23:34:36,808 DEBUG token.state (JDBCTokenStateService.java:removeToken(159)) - Token 8c419cf4...675b1dc14cce has been removed from the database
2021-04-19 23:34:36,809 DEBUG token.state (DefaultTokenStateService.java:revokeToken(244)) - Revoked token 8c419cf4...675b1dc14cce
2021-04-19 23:34:36,809 INFO  service.knoxtoken (TokenResource.java:revoke(327)) - Knox Token service (sandbox) revoked token eyJhbG...WdNDTw (8c419cf4...675b1dc14cce) (renewer=admin)

postgres=# select * FROM KNOX_TOKENS where token_id LIKE '8c419cf4%';
 token_id | issue_time | expiration | max_lifetime | username | comment 
----------+------------+------------+--------------+----------+---------
(0 rows)
```

I also let the Knox Gateway running for the long weekend and confirmed that expired tokens got removed by the reaper thread.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
@@ -17,19 +17,6 @@
  */
 package org.apache.knox.gateway.config.impl;
 
-import org.apache.commons.io.FilenameUtils;
-import org.apache.commons.lang3.StringUtils;
-import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.Path;
-import org.apache.knox.gateway.GatewayMessages;
-import org.apache.knox.gateway.config.GatewayConfig;
-import org.apache.knox.gateway.dto.HomePageProfile;
-import org.apache.knox.gateway.i18n.messages.MessagesFactory;
-import org.apache.knox.gateway.services.security.impl.ZookeeperRemoteAliasService;
-import org.joda.time.Period;
-import org.joda.time.format.PeriodFormatter;
-import org.joda.time.format.PeriodFormatterBuilder;
-
 import static org.apache.knox.gateway.services.security.impl.RemoteAliasService.REMOTE_ALIAS_SERVICE_TYPE;
 
 import java.io.File;
@@ -51,6 +38,20 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.Path;
+import org.apache.knox.gateway.GatewayMessages;
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.dto.HomePageProfile;
+import org.apache.knox.gateway.i18n.messages.MessagesFactory;
+import org.apache.knox.gateway.services.security.impl.ZookeeperRemoteAliasService;
+import org.joda.time.Period;
+import org.joda.time.format.PeriodFormatter;
+import org.joda.time.format.PeriodFormatterBuilder;
+
 
 /**
  * The configuration for the Gateway.
@@ -270,6 +271,12 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   private static final Set<String> KNOX_HOMEPAGE_HIDDEN_TOPOLOGIES_DEFAULT = new HashSet<>(Arrays.asList("admin", "manager", "knoxsso", "metadata", "homepage"));
   private static final String KNOX_HOMEPAGE_LOGOUT_ENABLED =  "knox.homepage.logout.enabled";
   private static final String KNOX_INCOMING_XFORWARDED_ENABLED = "gateway.incoming.xforwarded.enabled";
+
+  //Gateway Database related properties
+  private static final String GATEWAY_DATABASE_TYPE = GATEWAY_CONFIG_FILE_PREFIX + ".database.type";
+  private static final String GATEWAY_DATABASE_HOST =  GATEWAY_CONFIG_FILE_PREFIX + ".database.host";
+  private static final String GATEWAY_DATABASE_PORT =  GATEWAY_CONFIG_FILE_PREFIX + ".database.port";
+  private static final String GATEWAY_DATABASE_NAME =  GATEWAY_CONFIG_FILE_PREFIX + ".database.name";
 
   public GatewayConfigImpl() {
     init();
@@ -1230,5 +1237,25 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
     profiles.put("thin", HomePageProfile.getThinProfileElemens());
     profiles.put("token", HomePageProfile.getTokenProfileElements());
     return profiles;
+  }
+
+  @Override
+  public String getDatabaseType() {
+    return get(GATEWAY_DATABASE_TYPE, "none");
+  }
+
+  @Override
+  public String getDatabaseHost() {
+    return get(GATEWAY_DATABASE_HOST);
+  }
+
+  @Override
+  public int getDatabasePort() {
+    return getInt(GATEWAY_DATABASE_PORT, 0);
+  }
+
+  @Override
+  public String getDatabaseName() {
+    return get(GATEWAY_DATABASE_NAME, "GATEWAY_DATABASE");
   }
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
@@ -153,7 +153,7 @@ public class JDBCTokenStateService extends DefaultTokenStateService {
   @Override
   protected void removeToken(String tokenId) throws UnknownTokenException {
     try {
-      final boolean removed = tokenDatabase.removeToken(Tokens.getTokenIDDisplayText(tokenId));
+      final boolean removed = tokenDatabase.removeToken(tokenId);
       if (removed) {
         super.removeToken(tokenId);
         log.removedTokenFromDatabase(Tokens.getTokenIDDisplayText(tokenId));
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
@@ -1,0 +1,180 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.services.token.impl;
+
+import java.sql.SQLException;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.services.ServiceLifecycleException;
+import org.apache.knox.gateway.services.security.AliasService;
+import org.apache.knox.gateway.services.security.token.TokenMetadata;
+import org.apache.knox.gateway.services.security.token.UnknownTokenException;
+import org.apache.knox.gateway.util.JDBCUtils;
+import org.apache.knox.gateway.util.Tokens;
+
+public class JDBCTokenStateService extends DefaultTokenStateService {
+  private AliasService aliasService; // connection username/pw is stored here
+  private TokenStateDatabase tokenDatabase;
+
+  public void setAliasService(AliasService aliasService) {
+    this.aliasService = aliasService;
+  }
+
+  @Override
+  public void init(GatewayConfig config, Map<String, String> options) throws ServiceLifecycleException {
+    super.init(config, options);
+    if (aliasService == null) {
+      throw new ServiceLifecycleException("The required AliasService reference has not been set.");
+    }
+    try {
+      this.tokenDatabase = new TokenStateDatabase(JDBCUtils.getDataSource(config, aliasService));
+    } catch (Exception e) {
+      throw new ServiceLifecycleException("Error while initiating JDBCTokenStateService: " + e, e);
+    }
+  }
+
+  @Override
+  public void addToken(String tokenId, long issueTime, long expiration, long maxLifetimeDuration) {
+    super.addToken(tokenId, issueTime, expiration, maxLifetimeDuration);
+    try {
+      final boolean added = tokenDatabase.addToken(tokenId, issueTime, expiration, maxLifetimeDuration);
+      if (added) {
+        log.savedTokenInDatabase(Tokens.getTokenIDDisplayText(tokenId));
+      } else {
+        log.failedToSaveTokenInDatabase(Tokens.getTokenIDDisplayText(tokenId));
+      }
+    } catch (SQLException e) {
+      log.errorSavingTokenInDatabase(Tokens.getTokenIDDisplayText(tokenId), e.getMessage(), e);
+    }
+  }
+
+  @Override
+  protected void removeTokens(Set<String> tokenIds) {
+    try {
+      boolean removed = tokenDatabase.removeTokens(tokenIds);
+      if (removed) {
+        log.removedTokensFromDatabase(tokenIds.size());
+      } else {
+        log.failedToRemoveTokensFromDatabase(tokenIds.size());
+      }
+    } catch (SQLException e) {
+      log.errorRemovingTokensFromDatabase(tokenIds.size(), e.getMessage(), e);
+    }
+    super.removeTokens(tokenIds);
+  }
+
+  @Override
+  public long getTokenExpiration(String tokenId, boolean validate) throws UnknownTokenException {
+    try {
+      // check the in-memory cache, then
+      return super.getTokenExpiration(tokenId, validate);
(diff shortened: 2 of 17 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
```diff
@@ -0,0 +1,180 @@
...
+      }
+    }
+    return isUnknown;
+  }
+
+  @Override
+  protected List<String> getTokenIds() {
+    List<String> tokenIds = new LinkedList<>();
+    try {
+      tokenIds = tokenDatabase.getTokenIds();
```
moresandeep, line 156: This seems like an expensive operation and can potentially block a lot of CPU

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
```diff
@@ -0,0 +1,145 @@
...
+      addTokenStatement.setLong(2, issueTime);
+      addTokenStatement.setLong(3, expiration);
+      addTokenStatement.setLong(4, issueTime + maxLifetimeDuration);
+      return addTokenStatement.executeUpdate() == 1;
+    }
+  }
+
+  // This needs to be done in batches as many DB vendors have limitations on the number of items within the 'IN' clause
+  boolean removeTokens(Set<String> tokenIds) throws SQLException {
+    int removed = 0;
```
moresandeep, line 62: Looks unsafe should be synchronized

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
```diff
@@ -0,0 +1,145 @@
...
+  long getMaxLifetime(String tokenId) throws SQLException {
+    try (Connection connection = dataSource.getConnection(); PreparedStatement getMaxLifetimeStatement = connection.prepareStatement(GET_MAX_LIFETIME_SQL)) {
+      getMaxLifetimeStatement.setString(1, tokenId);
+      try (ResultSet rs = getMaxLifetimeStatement.executeQuery()) {
+        return rs.next() ? rs.getLong(1) : -1;
+      }
+    }
+  }
+
+  List<String> getTokenIds() throws SQLException {
```
moresandeep, line 124: This looks risky :) we might overwhelm Knox depending on number of tokens, could end up DOSing

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.Set;
+
+import javax.sql.DataSource;
+
+import org.apache.knox.gateway.services.security.token.TokenMetadata;
+
+public class TokenStateDatabase {
+  private static final String TOKENS_TABLE_NAME = "KNOX_TOKENS";
+  private static final String ADD_TOKEN_SQL = "INSERT INTO " + TOKENS_TABLE_NAME + "(token_id, issue_time, expiration, max_lifetime) VALUES(?, ?, ?, ?)";
+  private static final String REMOVE_TOKENS_SQL_PREFIX = "DELETE FROM " + TOKENS_TABLE_NAME + " WHERE token_id IN (";
```
moresandeep, line 36: Just a thought, why not list them under resources as a *.sql statements so all of them are in one place and easy to manage and look at.

moresandeep (commented):
Looks great Sandor!!
I did a quick review and noted few observations. Also, synchronization can be issue here, might want to consider using CompletableFutures for optimization, failures and thread safety.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
smolnar82: This is a simple `SELECT * FROM` query, nothing special happens here. Even if this was the case we need to get them from the Database to support HA.
My two cents:
- in case of `AliasBasedTokenStateService` there is a separate thread that reads all relevant aliases from the keystore into the memory -> this method does not need to be overridden. This is strictly non-HA.
- in case of `ZKTokenStateService` the listener mechanism makes sure the given Knox node collects all aliases into the in-memory -> this method does not need to be overridden.

Please note this method invoked by the reaper thread to evict expired tokens. This means we could do a better job here by adding a  `WHERE` clause to filter records that are expired already (or close enough to be expired).

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
smolnar82: I'm not sure if I follow. Do you think that having a `.sql` file is better than having all token-related SQL statements in one place (in this DAO class)?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
smolnar82: Two notes here:
- It's not part of any of the public API invocations so there is no risk here in terms of DOS
- Even if we are talking about a large number of tokens, getting them from the database should not be a problem

I'm going to run our own performance testing tool with this implementation to see actual numbers (e.g. having 1 million tokens in the DB) but I doubt that querying even millions of tokens from a DB causes a problem.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
smolnar82: Nice catch :)

smolnar82 (commented):
Thanks, @moresandeep !
I'm going to submit a new patch soon with the `getTokenIds` optimization and the removal locking the rest should be ok IMO.

smolnar82:
@moresandeep - thanks again for your review. In fact, I made it even more clean and effective by overriding the `evictExpiredTokens` from the parent class. This way a simple `DELETE` statement is enough.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
moresandeep: My concern is the SELECT clause blowing up, atleast we should use LIMIT clause 
ie
`SELECT * FROM tbl LIMIT 5,10;  # Retrieve rows 6-15`

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
moresandeep: cool! future devs might use this method accidentally and could cause issues but if it not public should be ok.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
moresandeep: I have seen this implemented in other projects where SQL was used heavily. The idea being all SQL stays in one resource place and if you need to tweak the SQL statements you don't have to worry about changing the java files. Even better if you can have them as external resources you could change them at runtime without changing class files (but that is not what I am expecting just something that came to me while writing :) )
I am not opposed to having it in DAO but my thinking is SQLs statements are more like resources.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -271,6 +272,12 @@
   private static final String KNOX_HOMEPAGE_LOGOUT_ENABLED =  "knox.homepage.logout.enabled";
   private static final String KNOX_INCOMING_XFORWARDED_ENABLED = "gateway.incoming.xforwarded.enabled";
 
+  //Gateway Database related properties
+  private static final String GATEWAY_DATABASE_TYPE = GATEWAY_CONFIG_FILE_PREFIX + ".database.type";
+  private static final String GATEWAY_DATABASE_HOST =  GATEWAY_CONFIG_FILE_PREFIX + ".database.host";
+  private static final String GATEWAY_DATABASE_PORT =  GATEWAY_CONFIG_FILE_PREFIX + ".database.port";
+  private static final String GATEWAY_DATABASE_NAME =  GATEWAY_CONFIG_FILE_PREFIX + ".database.name";
```
moresandeep, line 279: What if we need to add another database for other functions? 
e.g. this is for token, I am thinking of leveraging this to store public key and kids for JWKS service so we might use a different DB name for that.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
smolnar82: With the new implementation this is not a problem anymore (I did not even overwrite this method)

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
smolnar82: With the new implementation, this is not a problem anymore (I did not even overwrite this method)

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
smolnar82: With the new implementation, this is not a problem anymore (I did not even overwrite this method)

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
smolnar82: Then a new table has to be created in the same DB.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
smolnar82: I found [this article](https://stackoverflow.com/questions/1661921/java-programming-where-should-sql-statements-be-stored) on SO. Let me quote the pros/cons of my approach and I do agree with them:
```
Hardcoded/encapsulated in DAO layer

Pros:

SQL is kept in the objects that access data (encapsulation)
SQL is easy to write (speed of development)
SQL is easy to track down when changes are required
A simple solution (no messy architecture)

Cons:

SQL cannot be reviewed/changed by DBA
SQL is likely to become DB-specific
SQL can become hard to maintain
```
Given the fact we are going to keep authentication tokens in this DB table, I have the impression that these SQL statements should reside in the Java class so that they cannot be changed/manipulated/removed easily.

At my previous workplaces, where pure JDBC was used, no ORM (Hibernate, EclipseLink, etc) frameworks, we followed this approach because of the above-listed `pros`.
There was an exception where lots of business logic was written in PL/SQL but - to be honest - I really hated that part.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
moresandeep: Right, I was thinking of a table and not database.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
```diff
@@ -0,0 +1,180 @@
...
+      log.fetchedAllTokenIdsFromDatabase(tokenIds.size());
+    } catch (SQLException e) {
+      log.errorFetchingAllTokenIdsFromDatabase(e.getMessage(), e);
+    }
+    return tokenIds;
+  }
+
+  @Override
+  public void addMetadata(String tokenId, TokenMetadata metadata) {
```
lmccay, line 182: It seems that the addMetadata is decoupled from the addToken. I can imagine this came about due to the alias service based approach but it seems sort of artificial for JDBC approach. Perhaps this can be refactored in the future to deprecate this method and include an addToken that accepts a TokenMetadata object as well? Should we actually do that now and leave the separate method for where it is currently used but deprecate it?

lmccay (changes requested):
Looks good, added one review comment.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateService.java
smolnar82: Good idea; let me submit a new PS with that change.

smolnar82: @lmccay I went through the code and checked the current usage of these methods and analyzed the amount of changes to be made and I decided to handle this as a separate JIRA: https://issues.apache.org/jira/browse/KNOX-2580

smolnar82:
Merging based on @lmccay 's comment: this change looks good except for the suggested refactoring which I created a JIRA for.

PR #435, opened 2021-04-15
KNOX-2577 - In Livy UI, make application and container log links point to YARN UI v2

wypoon:

In the Livy UI, there are links to the Spark application and the driver log. They point to the old YARN UI.
Make the links point to the YARN UI v2.

## How was this patch tested?

Manually modified the rewrite.xml in a Knox gateway and tested the links.

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/livy/0.4.3/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/livy/0.4.3/rewrite.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<!--
+   Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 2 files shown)


moresandeep (changes requested):
@wypoon thank you for the patch, looks good, the only thing I would like is to create another service def e.g. 0.4.3 for this new change. We don't want to break the compatibility with older Livy releases.

wypoon:
> @wypoon thank you for the patch, looks good, the only thing I would like is to create another service def e.g. 0.4.3 for this new change. We don't want to break the compatibility with older Livy releases.

Ok, I can do that.

wypoon:
@moresandeep I added a new service version, 0.4.3, as you suggested. The service definition is the same as the existing one (0.4.0), while the rewrite rules have the proposed changes.

moresandeep (approved):
Great, thank you!

PR #436, opened 2021-04-15
KNOX-2575 - Add  `kid` and `jku`  claims to JWT tokens issues by Knox

moresandeep:
This change adds `kid` and `jku` claims to JWT tokens issues by Knox

## How was this patch tested?
The patch was tested locally on a local cluster.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
@@ -33,10 +33,12 @@ import java.text.ParseException;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import org.apache.knox.gateway.GatewayResources;
 import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.i18n.resources.ResourcesFactory;
 import org.apache.knox.gateway.services.Service;
 import org.apache.knox.gateway.services.ServiceLifecycleException;
@@ -73,6 +75,7 @@ import com.nimbusds.jwt.proc.JWTClaimsSetVerifier;
 
 public class DefaultTokenAuthorityService implements JWTokenAuthority, Service {
   private static final GatewayResources RESOURCES = ResourcesFactory.get(GatewayResources.class);
+  private static final TokenAuthorityServiceMessages LOG = MessagesFactory.get(TokenAuthorityServiceMessages.class);
 
   // Only standard RSA and HMAC signature algorithms are accepted
   // https://tools.ietf.org/html/rfc7518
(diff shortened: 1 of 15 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -311,4 +327,8 @@ else if (! (key instanceof RSAPrivateKey)) {
   @Override
   public void stop() throws ServiceLifecycleException {
   }
+
+  protected Optional<String> getCachedSigningKeyKID() {
```
pzampino, line 331: "KID" is an acronym for "KeyID", correct? If so, then I think this should be named `getCachedSigningKeyID()`

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -86,6 +89,8 @@
   private byte[] cachedSigningHmacSecret;
   private RSAPrivateKey signingKey;
 
+  private Optional<String> cachedSigningKeyKID = Optional.empty();
```
pzampino, line 92: `cachedSigningKeyID` ?

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/impl/JWTToken.java
```diff
@@ -80,12 +82,17 @@ public JWTToken(String alg, String[] claimsArray, List<String> audiences, boolea
     if(claimsArray[3] != null) {
       builder = builder.expirationTime(new Date(Long.parseLong(claimsArray[3])));
     }
+    if(claimsArray[4] != null) {
```
pzampino, line 85: Are there any callers of this which may not have a claims array of this length? In other words, ought this check the length of the claimsArray before checking whether the value is null? I think it only gets used by Knox issuers.

pzampino (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
moresandeep: Yup, I'll change it!

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
moresandeep: ah, the silly KeyKid will change it, thanks!

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/impl/JWTToken.java
moresandeep: It does but that is a good test before we stumble into ArrayIndexOutOfBounds exception

moresandeep: Here I just log and move on, since we are dealing with arrays of fix size initialized earlier in a different class there is no easy way to recover.

pzampino (approved):
LGTM

PR #438, opened 2021-04-26
KNOX-2595 - KNOX_TOKENS table is created automatically if not exists

smolnar82:

Instead of relying on that `KNOX_TOKENS` is being created externally, we do create that table as part of `JDBCTokenStateService` init if the table does not exist.

## How was this patch tested?

Updated JUnit tests (removed table creation) so that the Derby path is tested.
I ran manual testing against a PostgresSQL instance and started/stopped the server several times. Between these operations, I dropped the table + created tokens and confirmed the outcome matches my expectations.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
@@ -17,16 +17,23 @@
  */
 package org.apache.knox.gateway.services.token.impl;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+import java.io.InputStream;
 import java.sql.Connection;
+import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Locale;
 
 import javax.sql.DataSource;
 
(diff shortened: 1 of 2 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
```diff
@@ -41,6 +48,33 @@
 
   TokenStateDatabase(DataSource dataSource) throws Exception {
     this.dataSource = dataSource;
+    createKnoxTokensTableIfNotExists();
+  }
+
+  private void createKnoxTokensTableIfNotExists() throws Exception {
```
pzampino, line 54: Just curious what happens if the table creation throws an Exception.

pzampino (approved):
LGTM

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
smolnar82: The Knox Gateway won't start. IMO, this is correct since table creation takes places only if token state service is explicitly set to JDBC and in this case, it's a must that the table is created in the configured DB.

pzampino: +1

PR #443, opened 2021-04-29
KNOX-2599 - Improve tokengen UI

smolnar82:

- let end-users declare a more refined lifespan
- let end-users add a comment to the generated

## How was this patch tested?

Updated and ran JUnit tests. Additionally, I did manual testing.

1. Tested a wrong lifetime:
Screen Shot 2021-04-30 at 1 31 30 AM

2. Tested custom lifespan (token generated at 30 Apr 2021 01:32:14 AM)
Screen Shot 2021-04-30 at 1 32 35 AM

3. Saved comment with tricky characters:
Screen Shot 2021-04-30 at 1 34 17 AM

```
postgres=# select * from knox_tokens where token_id = '0dd15ada-4986-4de8-a261-001668af93f5';
               token_id               |  issue_time   |  expiration   | max_lifetime  | username |                                  comment                                  
--------------------------------------+---------------+---------------+---------------+----------+---------------------------------------------------------------------------
 0dd15ada-4986-4de8-a261-001668af93f5 | 1619739244250 | 1619825644246 | 1620344044250 | admin    | comment with a question mark ? and my favorite cartoon tom & jerry !!! :)
(1 row)
```

### DIFF
--- a/gateway-applications/src/main/resources/applications/tokengen/app/js/tokengen.js
+++ b/gateway-applications/src/main/resources/applications/tokengen/app/js/tokengen.js
@@ -125,6 +125,9 @@ var gen = function() {
         var apiUrl = tokenURL;
         //Instantiate HTTP Request
         var params = '?lifespan=P' + lt_days + "DT" + lt_hours + "H" + lt_mins + "M";  //we need to support Java's Duration pattern
+        if (form.comment.value != '') {
+            params = params + '&comment=' + encodeURIComponent(form.comment.value);
+        }
         var request = ((window.XMLHttpRequest) ? new XMLHttpRequest() : new ActiveXObject("Microsoft.XMLHTTP"));
         request.open("GET", apiUrl + params, true);
         request.send(null);
--- a/gateway-applications/src/main/resources/applications/tokengen/app/js/tokengen.js
+++ b/gateway-applications/src/main/resources/applications/tokengen/app/js/tokengen.js
@@ -110,6 +110,7 @@ function setTokenStateServiceStatus() {
 }
 
 var gen = function() {
+	$('#invalidLifetimeText').hide();
     var pathname = window.location.pathname;
     var topologyContext = pathname.replace(loginPageSuffix, "");;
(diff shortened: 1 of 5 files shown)


gateway-applications/src/main/resources/applications/tokengen/app/js/tokengen.js
```diff
@@ -53,20 +53,23 @@ function b64DecodeUnicode(str) {
...
+    var lt_days = form.lt_days.value;
+    var lt_hours = form.lt_hours.value;
+    var lt_mins = form.lt_mins.value;
     var _gen = function() {
         var apiUrl = tokenURL;
         //Instantiate HTTP Request
-        var params = '';
-        if (lifespan != '') {
-        	params = '?lifespan=' + lifespan;
+        var params = '?lifespan=P' + lt_days + "DT" + lt_hours + "H" + lt_mins + "M";  //we need to support Java's Duration pattern
```
lmccay, line 70: So, this would end up with a query param like: "1DT2H3M"?
Why are we using DT instead of D?
We are also going to be exposing this as a general API.
Wondering whether we would like to be able to just say 8H for 8 hours instead of "0DT8H0M".

lmccay (changes requested):
Looks good.
Made a comment about the lifespan query param.

gateway-applications/src/main/resources/applications/tokengen/app/js/tokengen.js
smolnar82: Thanks, @lmccay for your review.
The required pattern by Java's [Duration.parse](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-) is :`PnDTnHnMn.nS`.
Since I did not want to reinvent the wheel, we could re-use Java's own public API. Any objection?

smolnar82: @lmccay - I'm going to merge this change as-is, if you want we can open a follow-up JIRA with a different `queryParam` language later.

PR #444, opened 2021-05-02
KNOX-2600 - It's now possible to setup a PostgreSQL connection using a JDBC URL

smolnar82:

Instead of declaring connection properties one-by-one in `gateway-site.xml` a simple `connection URL` is enough to create connections to PostgreSQL DB server.

## How was this patch tested?

Updated and ran JUnit tests as well as executing the following manual test steps:
1. Configured the Knox Gateway as follows and started it successfully
```
    <property>
        <value>org.apache.knox.gateway.services.token.impl.JDBCTokenStateService</value>
    </property>
     <property>
        <value>postgresql</value>
    </property>
    <property>
        <value>jdbc:postgresql://smolnar-MBP15.local:5432/postgres?user=postgres&amp;ssl=true&amp;sslmode=verify-full&amp;sslrootcert=/usr/local/var/postgresql@10/data/root.crt</value>
    </property>
```
2. Generated some tokens with our `tokengen` application
3. Confirmed that the tokens were successfully saved in my local PostgreSQL DB

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
@@ -274,6 +274,7 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
 
   //Gateway Database related properties
   private static final String GATEWAY_DATABASE_TYPE = GATEWAY_CONFIG_FILE_PREFIX + ".database.type";
+  private static final String GATEWAY_DATABASE_CONN_URL = GATEWAY_CONFIG_FILE_PREFIX + ".database.connection.url";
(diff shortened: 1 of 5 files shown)


moresandeep (changes requested):
It would be nice if we can validate JDBC url syntax and let users know that the configured URL was wrong in the logs.

smolnar82:
> It would be nice if we can validate JDBC url syntax and let users know that the configured URL was wrong in the logs.

Thanks, @moresandeep for your review.
This is being done by `org.postgresql.Driver.parseURL(String, Properties)`. This method logs messages such as:
`JDBC URL must start with \"jdbc:postgresql:\"`
`JDBC URL must contain a / at the end of the host or port:`

pzampino (approved):
LGTM

PR #445, opened 2021-05-03
Knox 2582

moresandeep:
This change adds unauthenticated path list support for the following two authentication providers (along with SSOCookieProvider)

- JWTProvider
- HadoopAuthProvider

By default the path `/knoxtoken/api/v1/jwks.json` is added by default. Note that the path has to be exact, i.e. the request path after {gateway}/{topology} should EXACTLY match the configured path. 

Following properties can be used to add more paths to these providers

- hadoop.auth.unauthenticated.path.list
- jwt.unauthenticated.path.list

## How was this patch tested?

### DIFF
--- a/gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
+++ b/gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
@@ -17,7 +17,6 @@
  */
 package org.apache.knox.gateway.hadoopauth.filter;
 
-import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.authorize.AuthorizationException;
@@ -64,7 +63,6 @@ import java.util.HashSet;
 import java.util.Locale;
 import java.util.Properties;
 import java.util.Set;
-import java.util.StringTokenizer;
 
 import static org.apache.knox.gateway.util.AuthFilterUtils.DEFAULT_AUTH_UNAUTHENTICATED_PATHS_PARAM;
 
@@ -155,16 +153,10 @@ public class HadoopAuthFilter extends
       LOG.initializedJwtFilter();
     }
 
-    /* add default unauthenticated paths list */
-    AuthFilterUtils.parseAndAddUnauthPathList(unAuthenticatedPaths, DEFAULT_AUTH_UNAUTHENTICATED_PATHS_PARAM);
-
-    /* add provided unauthenticated paths list if specified */
     final String unAuthPathString = filterConfig
         .getInitParameter(HADOOP_AUTH_UNAUTHENTICATED_PATHS_PARAM);
-    /* if list specified add it */
-    if (!StringUtils.isBlank(unAuthPathString)) {
-      AuthFilterUtils.parseAndAddUnauthPathList(unAuthenticatedPaths, unAuthPathString);
-    }
+    /* prepare a list of allowed unauthenticated paths */
+    AuthFilterUtils.addUnauthPaths(unAuthenticatedPaths, unAuthPathString, DEFAULT_AUTH_UNAUTHENTICATED_PATHS_PARAM);
   }
 
   @Override
--- a/gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
+++ b/gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
@@ -170,7 +170,7 @@ public class HadoopAuthFilter extends
   public void doFilter(ServletRequest request, ServletResponse response, FilterChain filterChain) throws IOException, ServletException {
     /* check for unauthenticated paths to bypass */
     if(doesRequestContainUnauthPath(request)) {
-      checkForUnauthenticatedPaths(request, response, filterChain);
+      continueWithAnonymousSubject(request, response, filterChain);
       return;
     }
     if (shouldUseJwtFilter(jwtFilter, (HttpServletRequest) request)) {
@@ -185,7 +185,7 @@ public class HadoopAuthFilter extends
   protected void doFilter(FilterChain filterChain, HttpServletRequest request, HttpServletResponse response) throws IOException, ServletException {
     /* check for unauthenticated paths to bypass */
     if(doesRequestContainUnauthPath(request)) {
-      checkForUnauthenticatedPaths(request, response, filterChain);
+      continueWithAnonymousSubject(request, response, filterChain);
       return;
     }
     if (shouldUseJwtFilter(jwtFilter, request)) {
@@ -272,8 +272,9 @@ public class HadoopAuthFilter extends
    * @param response
    * @param chain
    */
-  private void checkForUnauthenticatedPaths(final ServletRequest request,
-      final ServletResponse response, final FilterChain chain) {
+  private void continueWithAnonymousSubject(final ServletRequest request,
+      final ServletResponse response, final FilterChain chain)
+      throws ServletException, IOException {
     try {
       /* This path is configured as an unauthenticated path let the request through */
       final Subject sub = new Subject();
@@ -283,9 +284,9 @@ public class HadoopAuthFilter extends
           (HttpServletResponse) response, chain);
 
     } catch (final Exception e) {
-      /* in case anything fails here log and move on */
       LOG.unauthenticatedPathError(
           ((HttpServletRequest) request).getRequestURI(), e.toString());
+      throw e;
     }
   }
 
--- a/gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
+++ b/gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
@@ -17,31 +17,33 @@
  */
 package org.apache.knox.gateway.hadoopauth.filter;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.security.UserGroupInformation;
 import org.apache.hadoop.security.authorize.AuthorizationException;
 import org.apache.hadoop.security.authorize.ProxyUsers;
 import org.apache.hadoop.util.HttpExceptionUtils;
 import org.apache.knox.gateway.GatewayServer;
+import org.apache.knox.gateway.audit.api.Action;
+import org.apache.knox.gateway.audit.api.ActionOutcome;
+import org.apache.knox.gateway.audit.api.AuditContext;
+import org.apache.knox.gateway.audit.api.AuditService;
+import org.apache.knox.gateway.audit.api.AuditServiceFactory;
+import org.apache.knox.gateway.audit.api.Auditor;
+import org.apache.knox.gateway.audit.api.ResourceType;
+import org.apache.knox.gateway.audit.log4j.audit.AuditConstants;
 import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.filter.AbstractGatewayFilter;
 import org.apache.knox.gateway.hadoopauth.HadoopAuthMessages;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.provider.federation.jwt.filter.JWTFederationFilter;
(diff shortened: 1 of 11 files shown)


gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
```diff
@@ -201,6 +246,86 @@ public Principal getUserPrincipal() {
...
   }
 
+  /**
+   * A helper method that checks whether request contains
+   * unauthenticated path
+   * @param request
+   * @return
+   */
+  private boolean doesRequestContainUnauthPath(final ServletRequest request) {
+    for (final String path : unAuthenticatedPaths) {
```
pzampino, line 256: Is there a reason to not use `return unAuthenticatedPaths.contains(((HttpServletRequest) request).getPathInfo())` ?

gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
```diff
@@ -201,6 +246,86 @@ public Principal getUserPrincipal() {
...
+   * A function that let's configured unauthenticated path requests to
+   * pass through without requiring authentication.
+   * An anonymous subject is created and the request is audited.
+   *
+   * Fail gracefully by logging error message.
+   * @param request
+   * @param response
+   * @param chain
+   */
+  private void checkForUnauthenticatedPaths(final ServletRequest request,
```
pzampino: This method does not appear to be "checking" for unauthenticated paths. I think it could be more accurately named something like `handleUnauthenticatedPath` or `bypassAuthForPath`.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
```diff
@@ -59,8 +59,11 @@
   @Message( level = MessageLevel.DEBUG, text = "Audience claim has been validated." )
   void jwtAudienceValidated();
 
-  @Message( level = MessageLevel.INFO, text = "Path {0} is configured as unauthenticated path, letting the request {1} through" )
-  void unauthenticatedPathBypass(String path, String uri);
+  @Message( level = MessageLevel.INFO, text = "Request {0} matches unauthenticated path list {1}, letting it through" )
+  void unauthenticatedPathBypass(String uri, String unauthPathList);
+
+  @Message( level = MessageLevel.ERROR, text = "Error {1} while checking whether path {0} should be allowed unauthenticated access" )
```
pzampino, line 65: nit: I would prefer to see the error at the end of the message; I think it's clearer that way.
"Error while checking whether path {0} should be allowed unauthenticated access : {1}"

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -31,15 +35,22 @@
...
 import java.util.Locale;
+import java.util.Set;
+import java.util.StringTokenizer;
 
 public class JWTFederationFilter extends AbstractJWTFilter {
 
+  private static final JWTMessages LOGGER = MessagesFactory.get( JWTMessages.class );
+  /* A semicolon separated list of paths that need to bypass authentication */
+  public static final String JWT_UNAUTHENTICATED_PATHS_PARAM = "jwt.unauthenticated.path.list";
+  public static final String DEFAULT_JWT_UNAUTHENTICATED_PATHS_PARAM = "/knoxtoken/api/v1/jwks.json";
```
pzampino, line 52: It would be ideal if these constants could be defined elsewhere and shared between HadoopAuthFilter and here instead of being defined in both places.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -90,6 +102,18 @@ public void init( FilterConfig filterConfig ) throws ServletException {
       publicKey = CertificateUtils.parseRSAPublicKey(verificationPEM);
     }
 
+    /* get unauthenticated paths list */
+    String unAuthPathString = filterConfig.getInitParameter(JWT_UNAUTHENTICATED_PATHS_PARAM);
+    /* if no list specified use default value */
+    if (StringUtils.isBlank(unAuthPathString)) {
+      unAuthPathString = DEFAULT_JWT_UNAUTHENTICATED_PATHS_PARAM;
```
pzampino, line 109: Should this not always be added to the Set?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -186,4 +215,46 @@ protected void handleValidationError(HttpServletRequest request, HttpServletResp
...
+   * A function that let's configured unauthenticated path requests to
+   * pass through without requiring authentication.
+   * An anonymous subject is created and the request is audited.
+   *
+   * Fail gracefully by logging error message.
+   * @param request
+   * @param response
+   * @param chain
+   */
+  private void checkForUnauthenticatedPaths(final ServletRequest request,
```
pzampino: Same as previous comment about this method name in HadoopAuthFilter

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
```diff
@@ -146,11 +146,11 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
     if (ssoCookies.isEmpty()) {
       /* check for unauthenticated paths to bypass */
       for (final String path : unAuthenticatedPaths) {
-        if (req.getRequestURI().contains(path)) {
+        if (req.getPathInfo().equals(path)) {
```
pzampino, line 149: Could unAuthenticatedPaths.contains(req.getPathInfo()) be employed here? Should this logic go into a util method, which could be shared across all these implementations?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
```diff
@@ -146,11 +146,11 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
           /* This path is configured as an unauthenticated path let the request through */
           final Subject sub = new Subject();
           sub.getPrincipals().add(new PrimaryPrincipal("anonymous"));
-          LOGGER.unauthenticatedPathBypass(path, req.getRequestURI());
+          LOGGER.unauthenticatedPathBypass(req.getRequestURI(), unAuthenticatedPaths.toString());
           continueWithEstablishedSecurityContext(sub, req, res, chain);
         }
```
pzampino, line 155: Is a break statement needed here to terminate the for loop?

pzampino (changes requested):

gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
pzampino: Nevermind...I saw the follow-on commit.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
pzampino: Nevermind...I saw the follow-on commit.

gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
```diff
@@ -129,10 +152,27 @@ public void init(FilterConfig filterConfig) throws ServletException {
       jwtFilter.init(filterConfig);
       LOG.initializedJwtFilter();
     }
+
+    /* get unauthenticated paths list */
+    String unAuthPathString = filterConfig.getInitParameter(HADOOP_AUTH_UNAUTHENTICATED_PATHS_PARAM);
+    /* if no list specified use default value */
+    if (StringUtils.isBlank(unAuthPathString)) {
+      unAuthPathString = DEFAULT_HADOOP_AUTH_UNAUTHENTICATED_PATHS_PARAM;
```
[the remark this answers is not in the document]
moresandeep: right, i'll change it.

gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
moresandeep: yup, that will simplify the code.

gateway-provider-security-hadoopauth/src/main/java/org/apache/knox/gateway/hadoopauth/filter/HadoopAuthFilter.java
moresandeep: I believe this was for older commit. This is fixed now.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
moresandeep: +1

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
moresandeep: I thought about that but then these auth filters are part of separate packages and extend different filters (one extends Hadoop's and other servlet filter). I don't think there is a suitable common place to house these constants.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
moresandeep: yup

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
moresandeep: fixed.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
pzampino: The HadoopAuthFilter already references the `org.apache.knox.gateway.provider.federation.jwt.filter.JWTFederationFilter`; Couldn't the constants could be defined there, or in AbstractJWTFilter since the module dependency apparently already exists?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
moresandeep: yes, it should, I will refactor the code.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
moresandeep: Refactored this code so this should not be an issue.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
moresandeep: I moved it to a common util class along with other methods you pointed out.

pzampino (approved):
LGTM!

PR #448, opened 2021-05-14
KNOX-2603 - Caching passcode token verifications

smolnar82:

In order to avoid expensive MAC calculation at request verification time, previous successful passcode token verifications are cached and checked when validating a passcode token.

## How was this patch tested?

Updated existing JUnit tests: I made sure common `testVerificationOptimization` is invoked in `TokenIDAsHTTPBasicCredsFederationFilterTest` (where we test passcode tokens)

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SignatureVerificationCache.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SignatureVerificationCache.java
@@ -30,8 +30,8 @@ import java.util.concurrent.ConcurrentHashMap;
  */
 public class SignatureVerificationCache {
 
-    public static final String JWT_VERIFIED_CACHE_MAX = "jwt.verified.cache.max";
-    public static final int    JWT_VERIFIED_CACHE_MAX_DEFAULT = 250;
+    public static final String TOKENS_VERIFIED_CACHE_MAX = "tokens.verified.cache.max";
+    private static final int   TOKENS_VERIFIED_CACHE_MAX_DEFAULT = 250;
 
     static final String DEFAULT_CACHE_ID = "default-cache";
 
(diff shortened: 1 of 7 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SignatureVerificationCache.java
```diff
@@ -30,8 +30,8 @@
  */
 public class SignatureVerificationCache {
 
-    public static final String JWT_VERIFIED_CACHE_MAX = "jwt.verified.cache.max";
-    public static final int    JWT_VERIFIED_CACHE_MAX_DEFAULT = 250;
+    public static final String VERIFIED_TOKENS_CACHE_MAX = "verified.tokens.cache.max";
```
pzampino, line 33: nit: `tokens.verified.cache.max`

pzampino (approved):
One minor naming comment; Otherwise, LGTM

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SignatureVerificationCache.java
smolnar82: I'm not sure I understand the nuance between the one I used and your recommendation. Do you think I can ship this way or you insist on changing it?

moresandeep (approved):

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SignatureVerificationCache.java
pzampino: Other similar properties begin with the context (e.g., jwt, sso, etc...), so it seems appropriate to begin with the context (i.e., tokens) with this one also.

pzampino: I've approved the change, so not insistent on changing it, but consistency is good IMO.

smolnar82: Ack; I've submitted a new PS.

PR #452, opened 2021-06-02
KNOX-2616 - Removed trailing slashes service context generation

smolnar82:

Extra trailing slash in the generated service context caused problems.

Additionally, to avoid regression, all UI and API_AND_UI type service definitions were modified to contain the trailing slash except the ones with query parameters. Thanks, @stoty for the deliberate testing.

## How was this patch tested?

Updated JUnit tests and tested many of the supported service UIs in my own cluster.

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/ambariui/2.2.0/service.xml
+++ b/gateway-service-definitions/src/main/resources/services/ambariui/2.2.0/service.xml
@@ -17,7 +17,7 @@
 <service role="AMBARIUI" name="ambariui" version="2.2.0">
     <metadata>
         <type>UI</type>
-        <context>/ambari</context>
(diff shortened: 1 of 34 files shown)


stoty:
When I tested, the UI pages for some services, look the Oozie UI, didn't load without the trailing slash.
As it is kind of hopeless to test every single service, the most straightforward thing to do is to re-add the slash in the service.xml files (except where the <context> tag contains URL parameters)

smolnar82:
Thanks, @stoty for your comment. I did a smoke test in a local cluster with all the service UIs available and found the following services need the trailing slash: Oozie, Ranger, Solr, Zeppelin.

stoty (approved):
+1 LGTM

PR #459, opened 2021-06-18
[KNOX-2622] Support Deflate Encoding for the Inbound Response

xuzikun2003:

Currently, Knox only supports the Content-Encoding "gzip" and the Content-Encoding "deflate" is not supported. When the upstream server (such as Livy 0.7 Server) returns a response which is deflated, Knox does not know how to decode it.

The current code only considers the "gzip" format and the "deflate" format is not considered. We should also handle the case when the format is "deflate" and use InflaterInputStream to read the in-bound response and use DeflaterOutputStream to write the out-bound response.

Here is the JIRA (https://issues.apache.org/jira/browse/KNOX-2622) that describes this issue.

## How was this patch tested?
Existing unit tests and newly added unit tests.

### DIFF
--- a/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/UrlRewriteResponse.java
+++ b/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/UrlRewriteResponse.java
@@ -126,7 +126,7 @@ public class UrlRewriteResponse extends GatewayResponseWrapper implements Params
   }
 
   private void setContentEncoding(String name, String value) {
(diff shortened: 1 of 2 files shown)


lmccay:
@xuzikun2003 - it seems you have some checkstyle issues that need to be resolved.
Please see https://travis-ci.com/github/apache/knox/jobs/515699869 for the issues towards the bottom of the build output.

xuzikun2003:
@lmccay , the style issue is already fixed.

lmccay:
@xuzikun2003  - thank you for your contribution! I will assign the JIRA to you and this will be available in Knox 1.6.0.

PR #461, opened 2021-07-08
KNOX-2624 - Introducing Knox Token Management UI

smolnar82:

A new Angular-based web UI is created to allow end-users to manage Knox tokens created by the logged-in users. In order to make this happen, the following changes were necessary:

**On the back-end side**

1. A new REST API endpoint had to be created to fetch Knox Tokens for the logged-in user. A sample JSON response looks like this:
```
{
	"tokens": [{
		"tokenId": "e0aee0f0-dabb-4cc0-8133-37799c7d17da",
		"issueTime": "2021-07-08T12:30:25.958+0200",
		"expiration": "2021-07-08T13:30:25.955+0200",
		"maxLifetime": "2021-07-15T12:30:25.958+0200",
		"metadata": {
			"enabled": true,
			"userName": "guest",
			"comment": "guest token"
		}
	}]
}
```
2. I'd to modify the `revoke` API to work with token IDs too: if the submitted `token` is not an entire `JWT` but 'only' a token ID (which we have available on the new token management UI) token revocation should work. In fact, the current token state service only extracts the token ID from the supplied JWT, so the token state service remains untouched in this respect; only `TokenResource` had to be changed.
3. From now on, using `disabled` tokens on the verification side will result in HTTP error `UNAUTHORIZED (401)`

**On the UI side**
1. There is a new link in the `General Proxy Information` section on the Knox Home page that points to the new token management UI
2. A new token management UI is created using TypeScript and Angular2. the new project's name is `token-management`.

**Updated configuration**
1. Enabled token state management in the `homepage` topology
2. Added `admin` as a renewer (so that it can revoke its own tokens)
3. Added the new `token-management` application into the `homepage` topology.
3. `gateway.knox.token.eviction.grace.period` is set to `0` so that expired tokens will be evicted w/o waiting for another day

## How was this patch tested?

Updated and executed JUnit tests.

I also tested the new functionality with `JDBCTokenStateSevice` and `AliasBasedTokenStateService`.

Sample UI screenshot:
Screenshot 2021-07-08 at 21 52 50


smolnar82:
Cc. @zeroflag

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -370,6 +375,11 @@ protected boolean validateToken(final HttpServletRequest request, final HttpServ
     return false;
   }
 
+  private boolean isTokenEnabled(String tokenId) throws UnknownTokenException {
+    final TokenMetadata tokenMetadata = tokenStateService == null ? null : tokenStateService.getTokenMetadata(tokenId);
```
zeroflag, line 379: Under what circumstances we don't have metadata for a token? Is it a valid scenario or should it only happen by accident?

smolnar82: There are 2 possible scenarios where there is no token metadata:

1. if the token management is turned off for the given topology (on KNOXTOKEN service)
2. the metadata alias is something that we invented in a later phase of Knox token management (but before adding the JDBC implementation). So in theory it may happen that only the basic information is stored in the underlying token store (keystore, Zookeeper) but the metadata is not yet there.

In both cases, we should consider the given token as an enabled one.

zeroflag: Ok, makes sense.

PR #462, opened 2021-07-09
KNOX-2628 - Metadata and issue time aliases are removed too during token revocation

smolnar82:

Added the missing alias names while removing a token from the keystore (and from memory).

## How was this patch tested?

Repeated the same steps as described in the JIRA and confirmed that all token aliases were removed from the credential store.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
@@ -431,10 +431,12 @@ public class AliasBasedTokenStateService extends DefaultTokenStateService implem
       unpersistedState.removeAll(unpersistedToRemove);
     }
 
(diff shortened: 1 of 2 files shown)


smolnar82:
Cc. @zeroflag

zeroflag:
LGTM. 
What's the relation between these aliases and the issue_time/experiation/max_lifetime columns used by TokenStateDatabase. Are these the same, or used together, or this is an alternative implementation for storing the same info?

pzampino (approved):
LGTM

pzampino:
What about a test?

smolnar82:
> What about a test?

Thanks for your review, @pzampino !
I modified `AliasBasedTokenStateServiceTest.testAddAndRemoveTokenIncludesCache()` to match the exact alias name list at removal time (instead of `anyObject`), so that we can make sure the appropriate alias removals happen.
In addition to this, I added code to make sure that information is moved from the in-memory cache too.

PR #463, opened 2021-07-11
KNOX-2627 - Limiting the number of managed tokens per user

smolnar82:

From now on, users are limited to manage only up to a pre-configured number of tokens (defaults to `10`). It's possible to eliminate this limit by setting the new `gateway.knox.token.limit.per.user` to `-1`. In this case, end-users are allowed to manage as many tokens as they want.
This configuration only applies when server-managed token state is enabled either in gateway-site or at the topology level.

## How was this patch tested?

Added new JUnit tests and executed manual test steps using the token generation UI.


smolnar82:
Cc. @zeroflag

zeroflag:
LGTM

PR #468, opened 2021-07-14
KNOX-2633 - Handling supplied client data with multiple '=' signs when generating a token

smolnar82:

Updated the logic which parses configured `knox.token.client.data` to honour values with multiple `=` characters.

## How was this patch tested?

Updated an existing JUnit test case to check `knox.token.client.data` too.

Modified our `homepage` topology to contain the appropriate profile name and the `sandbox` topology, then redeployed Knox. I confirmed the `Homepage` link on the token generation UI pointed to the configured URL and it actually opened the homepage with the proper filters in place.

### DIFF
--- a/gateway-release/home/conf/topologies/homepage.xml
+++ b/gateway-release/home/conf/topologies/homepage.xml
@@ -87,7 +87,7 @@
       </param>
       <param>
          <name>knox.token.client.data</name>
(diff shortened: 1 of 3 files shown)


smolnar82:
Cc. @zeroflag

pzampino (approved):
LGTM

PR #469, opened 2021-07-14
KNOX-2634 - Disable HA loadbalancing for ODBC driver and make this setting configurable.

moresandeep:
Add a way to turn off HA loadbalancing based on user agent.
This is done because some non-browser clients such as ODBC drivers do not honor multiple SET-COOKIE headers in response body. This config is to work around that issue. 

This PR proposed a configuration in HA Provider called `disableLoadBalancingForUserAgents` which takes a comma separated list of user agents that should disable the HA loadbalancing ifeature.

Following is an example config 
```
<provider>
             <role>ha</role>
             <name>HaProvider</name>
             <enabled>true</enabled>

             <param>
                 <name>HIVE</name>     <value>enableStickySession=true;enableLoadBalancing=true;enabled=true;maxFailoverAttempts=42;failoverSleep=50;maxRetryAttempts=1;disableLoadBalancingForUserAgents=Test User Agent, Test User Agent2,Test User Agent3 ,Test User Agent4 ;retrySleep=1000</value>
             </param>
</provider>
```

Default value of `disableLoadBalancingForUserAgents=ClouderaODBCDriverforApacheHive`

## How was this patch tested?
This patch was tested on a local cluster

### DIFF
--- a/gateway-provider-ha/pom.xml
+++ b/gateway-provider-ha/pom.xml
@@ -106,6 +106,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.curator</groupId>
             <artifactId>curator-test</artifactId>
             <scope>test</scope>
--- a/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
+++ b/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
@@ -48,7 +48,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 /**
@@ -87,9 +87,7 @@ public class ConfigurableHADispatch extends ConfigurableDispatch {
    *  connection which will be an issue.
    *  This variable keeps track of non-LB'ed url and updated upon failover.
    */
-  private String activeURL;
-  private final ReentrantLock activeURLlock = new ReentrantLock();
-
+  private AtomicReference<String> activeURL =  new AtomicReference();
   @Override
   public void init() {
     super.init();
@@ -113,12 +111,7 @@ public class ConfigurableHADispatch extends ConfigurableDispatch {
     }
 
     /* setup the active URL for non-LB case */
-    activeURLlock.lock();
-    try {
-      activeURL = haProvider.getActiveURL(getServiceRole());
-    } finally {
-      activeURLlock.unlock();
-    }
+    activeURL.set(haProvider.getActiveURL(getServiceRole()));
 
     // Suffix the cookie name by the service to make it unique
     // The cookie path is NOT unique since Knox is stripping the service name.
@@ -173,13 +166,10 @@ public class ConfigurableHADispatch extends ConfigurableDispatch {
        * use the activeURL
       */
       if(loadBalancingEnabled && userAgentDisabled) {
-        activeURLlock.lock();
         try {
-          ((HttpRequestBase) outboundRequest).setURI(updateHostURL(outboundRequest.getURI(), activeURL));
+          ((HttpRequestBase) outboundRequest).setURI(updateHostURL(outboundRequest.getURI(), activeURL.get()));
         } catch (final URISyntaxException e) {
           LOG.errorSettingActiveUrl();
-        } finally {
-          activeURLlock.unlock();
         }
       }
 
@@ -318,12 +308,8 @@ public class ConfigurableHADispatch extends ConfigurableDispatch {
       LOG.failingOverRequest(outboundRequest.getURI().toString());
 
       /* in case of failover update the activeURL variable */
-      activeURLlock.lock();
-      try {
-        activeURL = outboundRequest.getURI().toString();
-      } finally {
-        activeURLlock.unlock();
-      }
+      activeURL.set(outboundRequest.getURI().toString());
+
       executeRequest(outboundRequest, inboundRequest, outboundResponse);
     } else {
       LOG.maxFailoverAttemptsReached(maxFailoverAttempts, getServiceRole());
--- a/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
+++ b/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
@@ -18,6 +18,7 @@
 package org.apache.knox.gateway.ha.dispatch;
 
 import org.apache.commons.codec.digest.DigestUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.client.methods.HttpUriRequest;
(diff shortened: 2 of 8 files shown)


gateway-provider-ha/pom.xml
```diff
@@ -105,6 +105,12 @@
             <artifactId>javax.servlet-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <scope>compile</scope>
```
smolnar82, line 111: No need for the `compile` scope; that's the default.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
```diff
@@ -89,6 +91,13 @@ public void init() {
       if(stickySessionsEnabled) {
         stickySessionCookieName = serviceConfig.getStickySessionCookieName();
       }
+      /* if a configurable user agent list is defined to turn off Knox LB used it */
+      if(!StringUtils.isBlank(serviceConfig.getDisableStickySessionForUserAgents())) {
```
smolnar82, line 95: `StringUtils.isNotBlank` is easier to read than `!StringUtils.isBlank`

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
```diff
@@ -89,6 +91,13 @@ public void init() {
...
+        /* configured user agents are comma separate list which *can* contain whitespaces */
+        disableLoadBalancingForUserAgents = Arrays.asList(serviceConfig.getDisableStickySessionForUserAgents()
```
smolnar82, line 97: IMO, `HaServiceConfig.getDisableStickySessionForUserAgents()` 's return type should be `List<String>` and this logic should be implemented there.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/i18n/HaDispatchMessages.java
```diff
@@ -44,4 +44,7 @@
 
   @Message(level = MessageLevel.ERROR, text = "noFallback flag is turned on for sticky session so aborting request without retrying")
   void noFallbackError();
+
+  @Message(level = MessageLevel.INFO, text = "Knox HA loadbalancing disabled, detected user agent user-agent: {0}, configured user-agents to disable loadbalancing: {1}")
```
smolnar82, line 48: `user agent user-agent` seems odd to me.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/provider/HaServiceConfig.java
```diff
@@ -57,4 +57,8 @@
   boolean isNoFallbackEnabled();
 
   void setNoFallbackEnabled(boolean noFallbackEnabled);
+
+  void setDisableStickySessionForUserAgents(String disableStickySessionForUserAgents);
+
+  String getDisableStickySessionForUserAgents();
```
smolnar82, line 63: Should return a `List<String>`

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/provider/impl/HaDescriptorFactory.java
```diff
@@ -31,19 +31,81 @@ public static HaDescriptor createDescriptor() {
...
+      int failoverSleep = DEFAULT_FAILOVER_SLEEP;
+      boolean stickySessionsEnabled = DEFAULT_STICKY_SESSIONS_ENABLED;
+      boolean loadBalancingEnabled = DEFAULT_LOAD_BALANCING_ENABLED;
+      boolean noFallbackEnabled = DEFAULT_NO_FALLBACK_ENABLED;
+      String stickySessionCookieName = DEFAULT_STICKY_SESSION_COOKIE_NAME;
+      String disableStickySessionForUserAgents = DEFAULT_DISABLE_LB_USER_AGENTS;
+
+     String enabledValue = configMap.get(CONFIG_PARAM_ENABLED);
+     String maxFailoverAttemptsValue = configMap.get(CONFIG_PARAM_MAX_FAILOVER_ATTEMPTS);
+     String failoverSleepValue = configMap.get(CONFIG_PARAM_FAILOVER_SLEEP);
```
smolnar82, line 45: You may use `Map.getOrDefault(...)` to save the above default settings.

Even better, if I were you I'd add some private methods like that:
```
getBooleanValue(Map configMap, String configName, boolean default) {
  final configValue = configMap.get(configName);
  return StringUtils.isNotBlank(configValue) ? Boolean.parseBoolean(configValue) : default;
}
```
Then you could initialize your local variables like this:
```
final boolean enabled = getBooleanValue(configMap, CONFIG_PARAM_ENABLED, DEFAULT_ENABLED);
final int maxFailoverAttempts = getIntValue(configMap, CONFIG_PARAM_MAX_FAILOVER_ATTEMPTS, DEFAULT_MAX_FAILOVER_ATTEMPTS);
...
```

smolnar82 (changes requested):

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/provider/HaServiceConfig.java
pzampino: (minor) nit: The method name is misleading to me. 
List<String> getStickySessionDisabledUserAgents() ?

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/provider/HaServiceConfig.java
```diff
@@ -57,4 +57,8 @@
   boolean isNoFallbackEnabled();
 
   void setNoFallbackEnabled(boolean noFallbackEnabled);
+
+  void setDisableStickySessionForUserAgents(String disableStickySessionForUserAgents);
```
pzampino, line 61: (minor) nit: The method name is misleading to me. 
setStickySessionDisabledUserAgents() ?

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/provider/impl/HaServiceConfigConstants.java
```diff
@@ -53,4 +59,6 @@
    boolean DEFAULT_NO_FALLBACK_ENABLED = false;
 
    String DEFAULT_STICKY_SESSION_COOKIE_NAME = "KNOX_BACKEND";
+
+   String DEFAULT_DISABLE_LB_USER_AGENTS = "ClouderaODBCDriverforApacheHive";
```
pzampino, line 63: What if the default includes multiple values in the future? Should this be an array type instead?

moresandeep: Since this is coming from the HA config and will be a string I had it set up to mimic the config but looking at Sandor's and your comment looks like array/list might be appropriate.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
moresandeep: Right, Phil mentioned the same, I'll change it.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/provider/HaServiceConfig.java
moresandeep: true, will change it.

smolnar82 (approved):
LGTM

PR #470, opened 2021-07-15
KNOX-2635 - GatewayAppFuncTest is flaky

zeroflag:

Typical failure look like:

```
[ERROR]   GatewayAppFuncTest.testMultiApps:338 1 expectation failed.
Expected status code <200> but was <404>. 
```

It happens in test which are running right after testDeploymentCleanup or testDefaultTopology. These tests restart the gateway service during the test run.

The subsequent test overwrite the topology file which are used by all tests but file access date granularity is 1 second (msec part is always 000, this is likley FS specific) therefore the topology is never redeployed.

Adding clean up logic in between these restarts seem to solve the issue.

## How was this patch tested?

```bash
mvn test -Dtest=GatewayAppFuncTest   
while [ $? -eq 0 ]; do
    mvn test -Dtest=GatewayAppFuncTest   
done
```

### DIFF
--- a/gateway-test/src/test/java/org/apache/knox/gateway/GatewayAppFuncTest.java
+++ b/gateway-test/src/test/java/org/apache/knox/gateway/GatewayAppFuncTest.java
@@ -18,6 +18,7 @@
 package org.apache.knox.gateway;
 
 import java.io.File;
+import java.io.IOException;
 import java.net.URL;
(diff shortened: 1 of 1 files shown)


smolnar82 (approved):
LGTM!
Thanks, @zeroflag

PR #472, opened 2021-07-21
KNOX-2638 - Shiro propererties are missing after topology deployment

zeroflag:

STR:

1. Have an existing topology (test.xml), see if blockSemicolon/blockBackslash/blockNonAscii are in the get topology output:

```bash
$ curl -u admin:admin-password -k -X GET https://localhost:8443/gateway/admin/api/v1/topologies/test | grep blockSemicolon
<name>main.invalidRequest.blockSemicolon</name>
```

2. Deploy a new topology file (test2.xml)

```bash
cp ./conf/topologies/test.xml ./conf/topologies/test2.xml 
```

3. blockSemicolon/blockBackslash/blockNonAscii are no longer in the orignial test.xml output:

```bash
$ curl -u admin:admin-password  -k -X GET https://localhost:8443/gateway/admin/api/v1/topologies/test | grep  blockSemicolon 
```

The reason why these shiro properties are missing is because we only add these after a topology deployment to the in memory representation of a topology. When we deploy a new topology we also reload all the existing topologies but only redeploy the newly added one.

```java
@Override
public void reloadTopologies() {
  try {
    synchronized (this) {
      Map<File, Topology> oldTopologies = topologies;
      Map<File, Topology> newTopologies = loadTopologies(topologiesDirectory); // <= parse everything from disk
      List<TopologyEvent> events = createChangeEvents(oldTopologies, newTopologies);
      topologies = newTopologies;
      notifyChangeListeners(events); // <= this will redeploy the new topology and add the shiro properties, but those will be missing from all other existing topologies
    }
  } catch (Exception e) {
    // Maybe it makes sense to throw exception
    log.failedToReloadTopologies(e);
  }
} 
```

One open question is whether we wan't to write out these properties to the disk or not.

## How was this patch tested?

This was discovered while investigating the flakiness of an existing test: GatewayAdminTopologyFuncTest.
After the fixed I verified that GatewayAdminTopologyFuncTest is no longer flaky. Existing tests cover this case that's why no new tests were added.

### DIFF
--- a/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/deploy/impl/ShiroConfig.java
+++ b/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/deploy/impl/ShiroConfig.java
@@ -33,10 +33,6 @@ public class ShiroConfig {
     String sectionName;
     String value;
 
-    params.putIfAbsent("main.invalidRequest.blockSemicolon", "false");
-    params.putIfAbsent("main.invalidRequest.blockBackslash", "false");
-    params.putIfAbsent("main.invalidRequest.blockNonAscii", "false");
-
     for(Entry<String, String> entry : params.entrySet()) {
       int sectionDot = entry.getKey().indexOf('.');
       if (sectionDot > 0) {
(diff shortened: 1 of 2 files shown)


zeroflag:
cc: @smolnar82 @moresandeep @pzampino @lmccay4

smolnar82 (approved):
LGTM

IMO, we should write these properties out into the topology file.

pzampino (commented):
I've not been following this issue, but why are provider-specific attributes now being applied in the DefaultTopologyService?

pzampino (approved):
LGTM

PR #473, opened 2021-07-26
KNOX-2534. Allow alias to be used in pac4j topology block

zeroflag:

The patch makes possible to use an alias in pac4j provider config so that `oidc.secret` is no longer need to be hardcoded in the topology as plain text.

## How was this patch tested?

Added an alias for oidc.secret in knoxsso.xml.

```
<param>
  <name>oidc.secret</name>                  
  <value>${ALIAS=myOidcSecret}</value>
</param>
```

Created the alias with knoxcli.

```bash
$ bin/knoxcli.sh create-alias myOidcSecret --value <my-secret> --cluster knoxsso
````

Tested authentication with auth0.com.

### DIFF
--- a/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
+++ b/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
@@ -175,7 +175,7 @@ public class Pac4jDispatcherFilter implements Filter {
       addDefaultConfig(clientNameParameter, properties);
       while (names.hasMoreElements()) {
         final String key = names.nextElement();
-        properties.put(key, filterConfig.getInitParameter(key));
+        properties.put(key, resolveAlias(clusterName, key, filterConfig.getInitParameter(key)));
       }
       final PropertiesConfigFactory propertiesConfigFactory = new PropertiesConfigFactory(pac4jCallbackUrl, properties);
       config = propertiesConfigFactory.build();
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by zeroflag** [comment]
cc: @smolnar82 @moresandeep @pzampino @lmccay

File: gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java [hunk 1]
```diff
@@ -214,6 +214,18 @@ public void init( FilterConfig filterConfig ) throws ServletException {
 
   }
 
+  private String resolveAlias(String clusterName, String key, String value) throws ServletException {
+    if (value.startsWith("${ALIAS=") && value.endsWith("}")) {
+      String alias = value.substring("${ALIAS=".length());
```
pzampino, line 219: Why not remove the end delimiter here instead of performing another substring subsequently?

pzampino (changes requested):

moresandeep (approved):
look good, thanks @zeroflag!

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
zeroflag: @pzampino good idea, fixed it.

smolnar82: `String alias = value.substring(8, alias.lenght()-1);`

pzampino (approved):
LGTM

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
```diff
@@ -214,6 +214,18 @@ public void init( FilterConfig filterConfig ) throws ServletException {
 
   }
 
+  private String resolveAlias(String clusterName, String key, String value) throws ServletException {
+    if (value.startsWith("${ALIAS=") && value.endsWith("}")) {
```
pzampino, line 218: Is it worth making this a constant?

zeroflag: I extracted out to a constant.

smolnar82 (approved):

PR #478, opened 2021-08-06
KNOX-2574 - Missing proper logging when hmac secret is short

zeroflag:

When the hmac secret length is shorter than the required then token generation will fail with no meaningful message.

```json
{ "Unable to acquire token." }
```

The gateway log doesn't contain any information about what went wrong.

I found no simply way to add validation logic to the secret creation (bin/knoxcli.sh create-alias gateway.signing.hmac.secret --value <..>) and prevent creating the secret with invalid length, so I added a check to TokenResorce>>init() instead.

This throws an exception from the init and prevent the initializaiton of TokenResource, result http 500 when calling a method.

One thing to consider: Instead of throwing exception maybe we want to log the error and continue. Not sure which one would be better. 

## How was this patch tested?

1. Created a shorter than expected secret:

```
$ bin/knoxcli.sh create-alias gateway.signing.hmac.secret --value 1234
```

2. Set the signing alg to HS256 (keylength = 32)

```   
<param>
      <name>knox.token.sigalg</name>
      <value>HS256</value>
   </param>
```

3. Tried to create a token

```
$ curl -iku admin:admin-password https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token
```

This returns a http 500.

4. Checked the exception in gateway.logs
```
com.nimbusds.jose.KeyLengthException: Unexpected key length (for HS256 algorithm)
```

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -48,6 +48,10 @@ import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.KeyLengthException;
+import com.nimbusds.jose.crypto.MACSigner;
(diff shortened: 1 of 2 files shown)


zeroflag:
cc: @moresandeep @smolnar82 @pzampino @lmccay

smolnar82 (approved):
LGTM

PR #480, opened 2021-08-11
KNOX-2396 - TokenResource Renew and Revoke Should Respond With Error Codes That Identify Specific Errors

zeroflag:

TokeResource currently responds to failed requests to renew or revoke a token with an error message. Any client behavior based on the nature of the failure must currently depend on the error message content, which is not robust.

This patch extends the response with an error code, which is backward compatible with the existing responses.

## How was this patch tested?

Created a token with API call:

```
$ curl -v -k -u admin:admin-password  https://localhost:8443/gateway/homepage/knoxtoken/api/v1/token
```

Tried renew/revoke/enable/disable endpoints:

```
$ curl -X PUT -d "<tokenId>" -vvv -k -u admin:admin-password https://localhost:8443/knoxtoken/api/v1/token/enable

{
  "setEnabledFlag": "false",
  "error": "Token is already enabled",
  "code": 70
}
```

```
$ curl -X PUT -d "<tokenId>" -vvv -k -u admin:admin-password https://localhost:8443/knoxtoken/api/v1/token/disable
{
  "setEnabledFlag": "false",
  "error": "Token is already disabled",
  "code": 60
}
```

``` 
$ curl -d "invalid" -vvv -k -u admin:admin-password https://localhost:8443/knoxtoken/api/v1/token/revoke
{
  "revoked": "false",
  "error": "Invalid serialized unsecured/JWS/JWE object: Missing part delimiters",
  "code": 40
}
```

```
curl -d "invalid" -vvv -k -u admin:admin-password https://localhost:8443/knoxtoken/api/v1/token/renew
{
  "renewed": "false",
  "error": "Invalid serialized unsecured/JWS/JWE object: Missing part delimiters",
  "code": 40
}
```

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -155,6 +155,27 @@ public class TokenResource {
   @Context
   ServletContext context;
 
+  public enum ErrorCode {
+    UNKNOWN(0),
+    CONFIGURATION_ERROR(10),
+    UNAUTHORIZED(20),
+    INTERNAL_ERROR(30),
+    INVALID_TOKEN(40),
+    UNKNOWN_TOKEN(50),
+    ALREADY_DISABLED(60),
+    ALREADY_ENABLED(70);
+
+    private final int code;
+
+    ErrorCode(int code) {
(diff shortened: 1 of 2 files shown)


zeroflag:
cc: @pzampino @smolnar82 @moresandeep @lmccay

pzampino (approved):
LGTM
We should add these codes to the Knox user guide for them to be really useful.

PR #482, opened 2021-08-16
Shallow cloning on travis

YunLemon:
According to [https://docs.travis-ci.com/user/customizing-the-build#git-clone-depth](url), Travis CI provide a way to shallow clone a repository. This has the obvious benefit of speed, since you only need to download a small number of commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ before_install:
 script:
   - $DOCKERRUN $IMAGE mvn -T.75C clean verify -U -Dshellcheck=true -Dsurefire.useFile=false -Djavax.net.ssl.trustStorePassword=changeit -Dmaven.wagon.http.retryHandler.count=5 -Dmaven.wagon.http.retryHandler.class=default -Dmaven.wagon.http.retryHandler.nonRetryableClasses=java.io.InterruptedIOException -Dhttp.keepAlive=false -B -V
 git:
-  depth: 1000
+  depth: 3
 cache:
   directories:
     - $HOME/.m2


.travis.yml
```diff
@@ -41,7 +41,7 @@ before_install:
 script:
   - $DOCKERRUN $IMAGE mvn -T.75C clean verify -U -Dshellcheck=true -Dsurefire.useFile=false -Djavax.net.ssl.trustStorePassword=changeit -Dmaven.wagon.http.retryHandler.count=5 -Dmaven.wagon.http.retryHandler.class=default -Dmaven.wagon.http.retryHandler.nonRetryableClasses=java.io.InterruptedIOException -Dhttp.keepAlive=false -B -V
 git:
-  depth: 1000
+  depth: 3
```
moresandeep, line 44: Do you think we need to increase this to say 5?

moresandeep (changes requested):
Thanks @YunLemon! There are cases where a PR can have more than 3 commits e.g. https://github.com/apache/knox/pull/469/commits. Will this have any effect on the builds? 
if I understand the `depth` flag it basically squashes all the commits beyond `depth` flag, is that right? in which case this might not be an issue at all.

YunLemon:
`depth` doesn't squash commits, instead of cloning the complete repository (as usually done with git), using clone depth just clones the last clone-depth-number revisions of your repository.
According to [Disadvantages of shallow cloning on Travis and other CI services?](https://stackoverflow.com/questions/31278233/disadvantages-of-shallow-cloning-on-travis-and-other-ci-services), there are two main problems with using shallow clones:
1. Push too fast. Can you really push so fast?
2. Restart an old build. Do you have this need?
For a PR can have more than 3 commits, I don't think `depth` will be an issue.

pzampino:
Is there any data which shows this improves the Knox Travis CI builds? These builds are already a little flaky, and I would prefer to avoid possibly making them even more flaky, especially if there is no noticeable benefit. Admittedly, this is a git feature with which I am not yet very familiar.

YunLemon:
I compare these two builds [https://app.travis-ci.com/github/apache/knox/jobs/531547372](https://app.travis-ci.com/github/apache/knox/jobs/531547372) and [https://app.travis-ci.com/github/apache/knox/jobs/532288500](https://app.travis-ci.com/github/apache/knox/jobs/532288500). 1.47s vs 2.71s. Although the speed has increased twice, the absolute value is very small.....

smolnar82:
I agree with @pzampino that changing this value does not really bring new benefits (proven by the above-referenced jobs). So let me close this PR for now w/o merging it. Feel free to re-open and convince the Knox community if you think otherwise.

PR #484, opened 2021-08-24
KNOX-2646. The tokenLimitPerUser check doesn't always work.

zeroflag:

We have an equality check when checking the max number of tokens per user. But if the user already have N number of tokens, and later the admin changes the gateway.knox.token.limit.per.user to a smaller number then this check will never trigger.

## How was this patch tested?

1. Having no limit and generating N tokens
2. Changing gateway.knox.token.limit.per.user to a lower number than N and restarting the service
3. Trying to generate one more token and getting a limit exception

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -740,8 +740,4 @@ public class TokenResource {
     String message = t.getMessage();
     return message != null ? message : "null";
   }
-
-  void setTokenLimitPerUser(int tokenLimitPerUser) { // visible for testing
-    this.tokenLimitPerUser = tokenLimitPerUser;
-  }
 }
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -605,7 +605,7 @@ public class TokenResource {
 
     if (tokenStateService != null) {
       if (tokenLimitPerUser != -1) { // if -1 => unlimited tokens for all users
-        if (tokenStateService.getTokens(p.getName()).size() == tokenLimitPerUser) {
+        if (tokenStateService.getTokens(p.getName()).size() >= tokenLimitPerUser) {
           log.tokenLimitExceeded(p.getName());
           return Response.status(Response.Status.FORBIDDEN).entity("{ \"Unable to get token - token limit exceeded.\" }").build();
         }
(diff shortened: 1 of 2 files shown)


zeroflag:
cc: @pzampino @moresandeep @lmccay

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
```diff
@@ -986,6 +986,30 @@ public void testUnlimitedTokensPerUser() throws Exception {
...
+            .get("tokens");
+    assertEquals(tokens.size(), numberOfPreExistingTokens);
+    // change the limit and try generate one more
+    tr.setTokenLimitPerUser(numberOfPreExistingTokens -1);
+    Response response = tr.doGet();
+    assertTrue(response.getEntity().toString().contains("Unable to get token - token limit exceeded."));
+  }
+
   @Test
   public void tesTokenLimitPerUserExceeded() throws Exception {
```
pzampino, line 1014: Could you also correct this test name to tes**t**TokenLimitPerUserExceeded ?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -741,4 +741,7 @@ private String safeGetMessage(Throwable t) {
     return message != null ? message : "null";
   }
 
+  void setTokenLimitPerUser(int tokenLimitPerUser) { // visible for testing
```
pzampino, line 744: You could also use reflection in the test to set this, rather than adding a method ONLY for testing.

pzampino (approved):
LGTM

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
zeroflag: Sure.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
zeroflag: I'm not a fan of any of those (adding the accessor / using reflection) but thought this way is a little bit better. With reflection I need to reference the field name as a String which is not refactoring friendly.
I'll see if I can reconfigure the mocks and call the init() on the TokenResource again to change the limit.

smolnar82 (approved):

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
zeroflag: I managed to change the limit without the accessor method or reflection.

moresandeep (approved):

pzampino (approved):
LGTM

PR #486, opened 2021-08-27
KNOX-2650 - Loading token management page can be slow if there are lots of tokens

zeroflag:

Having lots of tokens causes slowness when loading the token management page. The reason is that the token metadata is loaded in a separate query and for each token we open a new connection when running the query.

This patch merges the 2 queries into one.

## How was this patch tested?

1. Creating few hundred tokens
2. Loading the token management page

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
@@ -17,8 +17,12 @@
  */
 package org.apache.knox.gateway.services.token.impl;
 
-import static java.nio.charset.StandardCharsets.UTF_8;
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.io.IOUtils;
+import org.apache.knox.gateway.services.security.token.KnoxToken;
+import org.apache.knox.gateway.services.security.token.TokenMetadata;
 
+import javax.sql.DataSource;
 import java.io.InputStream;
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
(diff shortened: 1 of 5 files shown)


gateway-server/src/test/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateServiceTest.java
```diff
@@ -49,6 +36,25 @@
...
+import java.sql.SQLException;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Locale;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
```
smolnar82, line 53: We need to make sure your IDE is configured properly in terms of imports (i.e. static imports must come first)

smolnar82 (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
```diff
@@ -192,24 +191,34 @@ TokenMetadata getTokenMetadata(String tokenId) throws SQLException {
...
       try (ResultSet rs = getTokenIdsStatement.executeQuery()) {
         while(rs.next()) {
-          tokens.add(new KnoxToken(rs.getString(1), rs.getLong(2), rs.getLong(3), rs.getLong(4)));
+          String tokenId = rs.getString(1);
+          long issueTime = rs.getLong(2);
+          long expiration = rs.getLong(3);
+          long maxLifeTime = rs.getLong(4);
+          String metaName = rs.getString(5);
+          String metaValue = rs.getString(6);
+          KnoxToken token = tokenMap.computeIfAbsent(tokenId, id -> new KnoxToken(tokenId, issueTime, expiration, maxLifeTime));
```
smolnar82, line 217: There is a KnoxToken constructor that takes a TokenMetadata; I'd use that one and remove the constructor being used here (in fact I added the one w/o metadata to support the 'poor performance' getTokens method which you are fixing now).

gateway-server/src/test/java/org/apache/knox/gateway/services/token/impl/JDBCTokenStateServiceTest.java
zeroflag: I reconfigured it so that static comes first, then java.* then org.*. This order matches the original order of the affected files. But I see exceptions, for example in GatewayServer.java, org.* imports are followed by java.* imports.

smolnar82 (approved):
LGTM

pzampino (approved):
LGTM

PR #487, opened 2021-08-30
KNOX-2651 - NPE when token value is missing

zeroflag:

There can be a NPE when there is a space between the username and password.

## How was this patch tested?

```
$ curl -vvv -iku Token: xxxx https://localhost:8443/gateway/sandbox/webhdfs/v1/?op=LISTSTATUS
```

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
@@ -126,7 +126,7 @@ public class JWTFederationFilter extends AbstractJWTFilter {
     }
     final Pair<TokenType, String> wireToken = getWireToken(request);
 
-    if (wireToken != null) {
+    if (wireToken != null && wireToken.getLeft() != null && wireToken.getRight() != null) {
       TokenType tokenType  = wireToken.getLeft();
       String    tokenValue = wireToken.getRight();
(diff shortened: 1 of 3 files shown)


zeroflag:
cc: @pzampino @smolnar82 @moresandeep

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -126,7 +126,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
     }
     final Pair<TokenType, String> wireToken = getWireToken(request);
 
-    if (wireToken != null) {
+    if (wireToken != null && wireToken.getLeft() != null && wireToken.getRight() != null) {
```
smolnar82, line 129: I think we could log the fact that token type/value is missing. It'd be a great help when debugging 401 issues.

smolnar82 (changes requested):

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
zeroflag: Good point.

pzampino (approved):
LGTM

smolnar82 (approved):

PR #489, opened 2021-08-31
KNOX-2653:- Update Atlas Service definition for knox logout/timeout in KNOX-2625

nixonrodrigues:
Updated Atlas Service definition for knox logout/timeout to support knox logout page in Atlas timeout and logout flow.

## How was this patch tested?
This patch was tested  by deploying updated servicedef in knox and test against Atlas changes on CDP cluster.
Manually tested the logout and timeout flow in Atlas redirecting to knox logout.jsp page.


gateway-service-definitions/src/main/resources/services/atlas/2.1.0/rewrite.xml
```diff
@@ -32,14 +32,20 @@
...
-    <rule dir="OUT" name="ATLAS/atlas/outbound/login/headers/location">
-        <match pattern="*://*:*/login.jsp"/>
-        <rewrite template="{$frontend[url]}/atlas/login.jsp"/>
+    <rule dir="OUT" name="ATLAS/atlas/outbound/login/headers/location" flow="OR">
+        <match pattern="*://*:*/login.jsp">
+            <rewrite template="https://{$frontend[addr]}/gateway/knoxsso/knoxauth/logout.jsp?originalUrl={$postfix[url,/atlas/]}" />
+        </match>
+
+        <match pattern="*://*:*/logout.html?{**}">
+            <rewrite template="https://{$frontend[addr]}/gateway/knoxsso/knoxauth/logout.jsp?{**}"/>
```
moresandeep, line 41: Here we are assuming that the cluster name is `gateway` which will not be the case always.

gateway-service-definitions/src/main/resources/services/atlas/2.1.0/rewrite.xml
```diff
@@ -32,14 +32,20 @@
         <rewrite template="{$frontend[path]}/atlas/"/>
     </rule>
 
-    <rule dir="OUT" name="ATLAS/atlas/outbound/login/headers/location">
-        <match pattern="*://*:*/login.jsp"/>
-        <rewrite template="{$frontend[url]}/atlas/login.jsp"/>
+    <rule dir="OUT" name="ATLAS/atlas/outbound/login/headers/location" flow="OR">
+        <match pattern="*://*:*/login.jsp">
+            <rewrite template="https://{$frontend[addr]}/gateway/knoxsso/knoxauth/logout.jsp?originalUrl={$postfix[url,/atlas/]}" />
```
moresandeep, line 37: Here we are assuming that the cluster name is gateway which will not be the case always.

moresandeep (changes requested):

gateway-service-definitions/src/main/resources/services/atlas/2.1.0/rewrite.xml
ashutoshm: @moresandeep Fix would be to replace gateway with {gateway} and knoxsso with {knoxsso}. Correct?

moresandeep: We can leave knoxsso as is since it is not a custom topology name, for gateway I think we can use `{gateway.path}`

PR #493, opened 2021-09-14
KNOX-2662 - Added TLS certs into General Proxy Information for the token profile

smolnar82:

The `General Proxy Information` section is extended with Knox's TLS public certs when the `token` profile is used.

## How was this patch tested?

Added relevant JUnit tests and did a manual test round too:

Screenshot 2021-09-14 at 9 56 53


smolnar82:
Cc. @zeroflag

zeroflag:
LGTM

PR #494, opened 2021-09-14
KNOX-2661 - Consolidated HTTP methods in TokenResource

smolnar82:

There were some public API endpoints in `TokenResource` with wrong HTTP methods. I changed hem and now the resource looks like this:
- GET - `/knoxtoken/api/v1/token`
- GET - `/knoxtoken/api/v1/token/getUserTokens`
- GET - `/knoxtoken/api/v1/token/getTssStatus`
- POST - `/knoxtoken/api/v1/token`
- PUT - `/knoxtoken/api/v1/token/renew`
- PUT - `/knoxtoken/api/v1/token/enable`
- PUT - `/knoxtoken/api/v1/token/disable`
- DELETE - `/knoxtoken/api/v1/token/revoke`

## How was this patch tested?

Ran manual test steps (including token management UI operations and `curl` requests)

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -39,6 +39,7 @@ import javax.annotation.PostConstruct;
 import javax.inject.Singleton;
 import javax.servlet.ServletContext;
 import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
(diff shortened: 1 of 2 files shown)


smolnar82:
Cc. @zeroflag

zeroflag:
Does this change effect existing users or is this not yet used by anyone?

smolnar82:
> Does this change effect existing users or is this not yet used by anyone?

We only use the `revoke` and `enable\disable` API on the token management page. Whoever uses these APIs will have to modify their clients to match the new HTTP method.

zeroflag:
I see, then I'm ok with it. Although the revoke + delete (and put + enable) feels a bit weird to me. There is a verb in both the HTTP method and in the path. But if it's mainly used by the UI then I think it's ok.

PR #495, opened 2021-09-15
KNOX-2664 - Let end-users revoke their own tokens

smolnar82:

Users can revoke tokens created for them even they are not listed in `knox.token.renewer.whitelist`
## How was this patch tested?

Added a new JUnit test case and ran manual test steps:

1. installed the new version of Knox
2. logged into the Knox Home page as `guest` (this user is not added into `knox.token.renewer.whitelist` in to `homepage` topology)
3. generated a token on the token generation page
4. revoked the previously generated token on the token management page successfully

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -489,27 +489,27 @@ public class TokenResource {
       error = "Token revocation support is not configured";
       errorCode = ErrorCode.CONFIGURATION_ERROR;
     } else {
-      String renewer = SubjectUtils.getCurrentEffectivePrincipalName();
-      if (allowedRenewers.contains(renewer)) {
-        try {
-          final String tokenId = getTokenId(token);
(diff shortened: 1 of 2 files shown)


smolnar82:
Cc. @zeroflag

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -527,6 +527,12 @@ public Response revoke(String token) {
     return resp;
   }
 
+  private boolean triesToRevokeOwnToken(String tokenId, String revoker) throws UnknownTokenException {
+    final TokenMetadata metadata = tokenStateService.getTokenMetadata(tokenId);
+    final String tokenUserName = metadata == null ? "" : metadata.getUserName();
+    return revoker != null && revoker.equals(tokenUserName);
```
zeroflag, line 533: Revoker is never an empty string, right?

smolnar82: In theory, it should always be set, but you are right: an additional emptiness checking does not hurt and could avoid weird issues.
I submitted a new PS.

PR #501, opened 2021-09-23
KNOX-2670 - Deleting unnecessary removeToken override from AliasBasedTokenStateService

smolnar82:

See the root cause analysis in https://issues.apache.org/jira/browse/KNOX-2670.
Additional note: the newly implemented `triesToRevokeOwnToken(String, String)` method solves the original issue 1 layer above, but the problem with the `AliasBasedTokenStateService` would still exist w/o removing the `removeToken(String)` method.

## How was this patch tested?

I ran Junit tests and I executed the steps I used for reproducing the original issue (I temporary removed the `triesToRevokeOwnToken(String, String)` call to make sure my change actually solves the issue).

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/AliasBasedTokenStateService.java
@@ -20,7 +20,6 @@ import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.Collections;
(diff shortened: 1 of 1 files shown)


smolnar82:
Cc. @zeroflag

zeroflag:
LGTM. The patch looks good to me, but the inheritance hierarchy is a bit difficult to follow.

smolnar82:
> LGTM. The patch looks good to me, but the inheritance hierarchy is a bit difficult to follow.

It's not that complicated IMO.

Screenshot 2021-09-23 at 19 46 31

pzampino (approved):
LGTM

PR #507, opened 2021-10-07
KNOX-2677 - HADispatch keeps retrying the same URL on failover

zeroflag:
If a host is unreachable Knox marks the URL but keeps trying the same URL upon failover.

The new URL is now chosen from haProvider.

## How was this patch tested?

Add these to sandbox.xml:

```xml
<provider>
     <role>ha</role>
     <name>HaProvider</name>
     <enabled>true</enabled>
     <param>
         <name>HIVE</name>
         <value>maxFailoverAttempts=3;failoverSleep=1000;enabled=true;enableStickySession=true;enableLoadBalancing=true</value>
     </param>
</provider>
```

```xml
<service>
    <role>HIVE</role>
    <url>http://deadhost.site</url>
    <url>http://localhost:1701/alive</url>
    <param>
        <name>replayBufferSize</name>
        <value>8</value>
    </param>
</service>
```

Mini HTTP server:

```smalltalk
Teapot on
	GET: '/alive' -> 'OK';
	start.
```

```bash
$  curl -k -u admin:admin-password  https://localhost:8443/gateway/sandbox/hive
```


zeroflag:
cc: @moresandeep @smolnar82 @pzampino @lmccay

PR #512, opened 2021-10-31
KNOX-2685 - Show/hide (enable/disable) Knox token management links on Home Page if required alias is created

smolnar82:

The Token Integration link on the Home page must be hidden if the required `knox.token.hash.key` Gateway-level alias is missing.

## How was this patch tested?

Manual testing.

Without the alias:
Screenshot 2021-10-31 at 21 09 54

With the alias (running `bin/knoxcli.sh generate-jwk --saveAlias knox.token.hash.key`):

Screenshot 2021-10-31 at 21 29 50


lmccay:
LGTM!

+1

lmccay (approved):
+1

PR #516, opened 2021-11-12
KNOX-2690 - Service URLs are missing from Knox UI after a temporary discovery failure

zeroflag:

When CM is intermittently unavailable, Knox still regenerates topologies with an empty service list. It should skip topology regeneration and log the error instead.

This patch fixes it by throwing an exception and avoiding the topology regeneration in case of a discovery failure.

The patch also removes some unused code around discovery.

## How was this patch tested?

Changed `discovery-address` in `/var/lib/knox/gateway/conf/descriptors/cdp-proxy.json` to an invalid address.

Checked the logs for the error message:

```
2021-11-12 11:14:41,394 INFO  discovery.cm (ClouderaManagerServiceDiscovery.java:discoverCluster(215)) - Performing cluster discovery for "Cluster 1"
2021-11-12 11:14:41,396 ERROR discovery.cm (ClouderaManagerServiceDiscovery.java:getClusterServices(276)) - Failed to access the service configurations for cluster (Cluster 1) discovery: com.cloudera.api.swagger.client.ApiException: java.net.ConnectException: Failed to connect to amagyar-2.amagyar.***/***7180
2021-11-12 11:14:41,396 ERROR knox.gateway (DefaultTopologyService.java:onFileChange(870)) - Unable to complete service discovery for cluster Cluster 1 topology = cdp-proxy.
```

Verified that the URL-s are still present on the Knox UI.

### DIFF
--- a/gateway-discovery-ambari/src/main/java/org/apache/knox/gateway/topology/discovery/ambari/AmbariServiceDiscovery.java
+++ b/gateway-discovery-ambari/src/main/java/org/apache/knox/gateway/topology/discovery/ambari/AmbariServiceDiscovery.java
@@ -194,36 +194,6 @@ class AmbariServiceDiscovery implements ServiceDiscovery {
         return TYPE;
     }
 
-
-    @Override
-    public Map<String, Cluster> discover(GatewayConfig gatewayConfig, ServiceDiscoveryConfig discoveryConfig) {
-        Map<String, Cluster> clusters = new HashMap<>();
-
-        init(gatewayConfig);
-
-        String discoveryAddress = discoveryConfig.getAddress();
-
-        // Invoke Ambari REST API to discover the available clusters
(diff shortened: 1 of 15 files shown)


zeroflag:
cc: @smolnar82 @pzampino @moresandeep @lmccay

smolnar82 (approved):
LGTM

Just to clarify: the removal of "some unused code" was deleting the `public Map<String, Cluster> discover(GatewayConfig gatewayConfig, ServiceDiscoveryConfig discoveryConfig)` method because it was only used in tests, right?

Cc. @pzampino

zeroflag:
> Just to clarify: the removal of "some unused code" was deleting the `public Map<String, Cluster> discover(GatewayConfig gatewayConfig, ServiceDiscoveryConfig discoveryConfig)` method because it was only used in tests, right?

Yes.

PR #517, opened 2021-11-13
KNOX-2679 - Add ability to remove custom attributes from Pac4j cookie

moresandeep:
There are cases where IdP sends back custom attributes that might be too big. For instance groups name attribute could be `https://knox.apache.org/SAML/Attributes/groups` which might contain large number of groups that might prevent setting cookies properly. This patch adds the ability to remove custom attributes from pac4j profile cookie.

`pac4j.session.store.exclude.custom.attributes` is the configuration setting and it takes a comma separated list of values. Default is blank string.

## How was this patch tested?


moresandeep:
@zeroflag fyi

PR #518, opened 2021-11-15
KNOX-2689 - Redeploying a topology only if it was actually changed

smolnar82:

From now on, a topology is redeployed only, and only if, there is a change within the topology itself. Not directly in the topology file, because Knox does not count adding/removing comments in the topology as an update. The change has to be a meaningful change such as a change in any of the services, providers, applications, or topology-level attributes.

With this change in place, the possibility to redeploy a topology, by simply touching it, is removed. Instead, clients should use KnoxCLI like this:
```
knoxcli.sh redeploy --cluster topologyName

E.g. knoxcli.sh redeploy --cluster sandbox
```
## How was this patch tested?

Updated and executed JUnit tests:
```
$ mvn clean -Dshellcheck=true verify -Prelease,package
...
[INFO] BUILD SUCCESS
[INFO] Total time:  26:06 min
[INFO] Finished at: 2021-11-15T15:37:32+01:00
```

In addition to unit testing, I ran the following E2E test steps after redeploying and starting the Knox Gateway:
1. touched `sandbox.xml` -> Knox did not redeploy the topology
1. ran `knoxcli.sh redeploy --cluster sandbox` -> Knox redeployed `sandbox` successfully
1. edited and saved a `WEBHDFS`'s service definition on the Admin UI -> Knox redeployed `sandbox` successfully

Repeated steps 1 and 2 with other topologies too.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
@@ -79,6 +79,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
@@ -88,6 +89,8 @@ public class DefaultTopologyService extends FileAlterationListenerAdaptor implem
     TopologyProvider, FileFilter, FileAlterationListener, ServiceDefinitionChangeListener {
 
   private static final JAXBContext jaxbContext = getJAXBContext();
+  private static final String TOPOLOGY_CLOSING_XML_ELEMENT = "</topology>";
+  private static final String REDEPLOY_TIME_TEMPLATE = "   <redeployTime>%d</redeployTime>\n" + TOPOLOGY_CLOSING_XML_ELEMENT;
 
   private static final Auditor auditor = AuditServiceFactory.getAuditService().getAuditor(
     AuditConstants.DEFAULT_AUDITOR_NAME, AuditConstants.KNOX_SERVICE_NAME,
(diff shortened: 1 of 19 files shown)


smolnar82:
Cc. @zeroflag

gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
```diff
@@ -179,32 +182,21 @@ private void redeployTopology(Topology topology) {
...
-        } catch (InterruptedException e) {
-          auditor.audit(Action.REDEPLOY, topology.getName(), ResourceType.TOPOLOGY, ActionOutcome.FAILURE);
-          log.failedToRedeployTopology(topology.getName(), e);
-          Thread.currentThread().interrupt();
-        }
+      // Since KNOX-2689, updating the topology file's timestamp is not enough.
+      // We need to make an actual change in the topology XML to redeploy it
+      // This change is: updating a new XML element called redeployTime
+      try {
+        final String currentTopologyContent = FileUtils.readFileToString(topologyFile, StandardCharsets.UTF_8);
```
zeroflag, line 189: What if there is no `<redeployTime>` tag in the file? Is this always supposed to be there? 

I wonder if it's possible to use some internal notification mechanism to send some event to the watcher to redeploy the topology without needing to modify the file.
I understand if the change comes from outside (from knoxcli) this is needed, but here we are inside knox so theoretically we should be able to just call the code that does the redeployment.

I also wonder if it would make sense to make redeployTime as an attribute of the root tag for example, making it less apparent.

smolnar82: Thanks, @zeroflag for your review comments. Let me try to reply to them one by one:
> What if there is no <redeployTime> tag in the file? Is this always supposed to be there?

It's not a problem. In that case, the default of `-1` is used.

> I wonder if it's possible to use some internal notification mechanism to send some event to the watcher to redeploy the topology without needing to modify the file.

Unfortunately, currently, modifying the topology file is the _only_ option. Of course, we _could_ implement a framework that listens/receives such events, but it'd very fragile and - IMO - way too heavy for this operation.

> I also wonder if it would make sense to make redeployTime as an attribute of the root tag for example, making it less apparent.

That's a good idea and I also considered that option but I found it more convenient to implement it as a separate XML tag (following the current XML layout we have in topologies). If you, or the rest of the team, convinces me it's a must before we merge this change I'll update my changes.

zeroflag: > Of course, we could implement a framework that listens/receives such events

What's the difference between modifying the file and calling:

```java
notifyChangeListeners(Arrays.asList(new TopologyEvent(TopologyEvent.Type.UPDATED, topology)));
```

All the infrastructure that is required to trigger a redeploy seems to be already available.

moresandeep: Isn't this an expensive operation, reading the entire file? what I am thinking is, relying on file timestamp was quick and dirty solution. Reading and parsing xml file is a bit error prone.

gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
```diff
@@ -179,32 +182,21 @@ private void redeployTopology(Topology topology) {
...
+        String updated = currentTopologyContent.replaceAll("^*<redeployTime>.*", "");
+
+        //add the current timestamp
+        updated = updated.replace(TOPOLOGY_CLOSING_XML_ELEMENT, String.format(Locale.getDefault(), REDEPLOY_TIME_TEMPLATE, System.currentTimeMillis()));
+
+        //save the updated content in the file
+        FileUtils.write(topologyFile, updated, StandardCharsets.UTF_8);
+      } catch (IOException e) {
```
moresandeep, line 197: We should be closing the File object in finally.

moresandeep (commented):
My only concern here is that parsing XML is error prone especially when there are cases where XML is written by another process (changes to descriptor etc.). Reading and parsing file is also an expensive operation.

One suggestion that comes to mind is to use a timestamp file next to topology, so instead of storing timestamps in topology just store them something as `<topology>-ts` and when we write topology just touch this file and when we need to check if something is changed just read the timestamp. Bonus this would help in debugging too. 

I am also worries that a lot of folks that work with Knox know that just touching topology deploys them this might be a a bad experience so we should atleast announce this on the dev and user list before we commit it. 

my two cents :)

gateway-server/src/main/java/org/apache/knox/gateway/services/topology/impl/DefaultTopologyService.java
moresandeep: > 

Had an offline discussion with @smolnar82 and looks like we are already parsing the xml so this change won't add any overhead than we already have.

moresandeep: About file closing `FileUtils` takes care of that.

moresandeep (approved):
+1 looks good!

PR #520, opened 2021-11-29
KNOX-2692 - Topology redeployment can be configured to require changes in topologies

smolnar82:

To support backward compatibility, a new Gateway-level configuration is introduced to allow end-users to configure if they require changes in topologies for redeployment or a simple `touch` is enough (relying on the topology file's timestamp).

## How was this patch tested?

Updated and executed JUnit tests. In addition to unit testing, I ran the following manual test steps:
- redeployed Knox w/ my changes
- confirmed that `touch conf/topologies/sandbox.xml` re-deployed the `sandbox` topology
- edit `gateway-site.xml` (set `gateway.topology.redeploy.requires.changes=true`) and re-started Knox
- confirmed that
  - `touch conf/topologies/sandbox.xml` did not re-deploy the `sandbox` topology
  - `bin/knoxcli.sh redeploy --cluster sandbox` re-deployed the `sandbox` topology


smolnar82:
Cc. @zeroflag

zeroflag:
LGTM

PR #528, opened 2022-01-11
KNOX-2700 - Fix broken unit tests

moresandeep:

Fix unit tests.

## How was this patch tested?
Patch was tested locally.

### DIFF
--- a/gateway-test-release/webhdfs-kerb-test/src/test/resources/org/apache/knox/gateway/SecureKnoxShellTest/SecureWebHdfsPutGet.groovy
+++ b/gateway-test-release/webhdfs-kerb-test/src/test/resources/org/apache/knox/gateway/SecureKnoxShellTest/SecureWebHdfsPutGet.groovy
@@ -30,6 +30,7 @@ session = KnoxSession.kerberosLogin(gateway, jaasConf, krb5conf, true)
 status = Hdfs.status(session).file( "/" ).now().string
 
 Hdfs.put( session ).file( file ).to( dataDir + "/" + dataFile ).now()
+Thread.sleep(1000)
 Hdfs.put( session ).file( file ).to( dataDir + "/" + dataFile ).overwrite(true).permission(777).now()
 
 fetchedFile = Hdfs.get( session ).from( dataDir + "/" + dataFile).now().string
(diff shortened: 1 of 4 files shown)


zeroflag (approved):

gateway-test-release/webhdfs-kerb-test/src/test/resources/org/apache/knox/gateway/SecureKnoxShellTest/SecureWebHdfsPutGet.groovy
```diff
@@ -30,6 +30,7 @@ session = KnoxSession.kerberosLogin(gateway, jaasConf, krb5conf, true)
 status = Hdfs.status(session).file( "/" ).now().string
 
 Hdfs.put( session ).file( file ).to( dataDir + "/" + dataFile ).now()
+Thread.sleep(1000)
```
risdenk, line 33: Why are these `Thread.sleep` necessary? Does the hdfs put not actually return when the file is created? Is there a different HDFS call that we can check the file exists? Like `Hdfs.status` as done on line 30?

risdenk (changes requested):

gateway-test-release/webhdfs-kerb-test/src/test/resources/org/apache/knox/gateway/SecureKnoxShellTest/SecureWebHdfsPutGet.groovy
moresandeep: Thanks for the review @risdenk! We have recently started seeing this issue where `MiniDFSCluster` takes some time for the PUT operation. When I run these tests in debug mode with debugger on they work and this is where I noticed where they fail. 

Are you suggesting using status to see if the file exists if not wait for given time then perform fetch operation?

risdenk: I'm suggesting don't just do a `Thread.sleep` when there is some timing issue here.

I would expect `Hdfs.put( session ).file( file ).to( dataDir + "/" + dataFile ).now()` to return when the file is actually created. So there is some bug there along the line that it returns without the file actually being there somehow. (No idea if this is a bug in HDFS, WebHDFS, Knox, etc).

The fix you put in could work, but it might make sense to check if the file is created in a loop (ie: while not created `Thread.sleep(200)` or something like that). This would avoid an overloaded CI machine from artificially failing due to file not being ready yet.

moresandeep: `Hdfs.put( session ).file( file ).to( dataDir + "/" + dataFile ).now()` does return 200 but intermittently the following calls fail. Tried checking for response still blows up. It also appears that others are not seeing it as often as me. I am shelving this part of the fix for now until we know what's wrong. Thanks for your inputs @risdenk.

PR #531, opened 2022-01-14
KNOX-2693 - Redeploying a topology if it was changed regardless of its timestamp

smolnar82:

If there are subsequent changes to a topology and Knox still processes the one that came earlier it might happen that the second change is disregarded due to timing issues.
The solution is to make sure Knox always triggers an `UPDATE` event regardless of
- the topology file's timestamp
- or the `is change required` condition.

## How was this patch tested?

Updated and ran unit tests as well as conducted manual testing.


moresandeep (approved):
Loog good.

PR #532, opened 2022-01-18
KNOX-2703 - Configuring JWT type at creation/verification time

smolnar82:

Knox supports two new topology-level properties from now on:
- `knox.token.type` - this can be set as a `KNOXTOKEN` service parameter. If this is configured the generated JWT header will have this as the `typ` property
- `knox.token.allowed.jws.types` - this can be set as a `JWTProvider` provider parameter. Defaults to `JWT`. Please note, this configuration is only applied if token verification goes through the JWKS verification path as described in #216 .

## How was this patch tested?

Unit tested as well as executed manual test runs w/ different combinations of these new params.

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
@@ -26,6 +26,7 @@ import java.util.Base64;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import javax.security.auth.Subject;
 import javax.servlet.FilterChain;
@@ -46,6 +47,8 @@ import org.apache.knox.gateway.services.security.token.impl.JWTToken;
 import org.apache.knox.gateway.util.AuthFilterUtils;
 import org.apache.knox.gateway.util.CertificateUtils;
 
+import com.nimbusds.jose.JOSEObjectType;
+
 public class JWTFederationFilter extends AbstractJWTFilter {
 
   private static final JWTMessages LOGGER = MessagesFactory.get( JWTMessages.class );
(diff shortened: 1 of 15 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -90,6 +94,14 @@ public void init( FilterConfig filterConfig ) throws ServletException {
       expectedJWKSUrl = oidcjwksurl;
     }
 
+    allowedJwsTypes = new HashSet<>();
+    final String allowedTypes = filterConfig.getInitParameter(ALLOWED_JWS_TYPES);
+    if (allowedTypes != null) {
+      Stream.of(allowedTypes.trim().split(",", -1)).forEach(allowedType -> allowedJwsTypes.add(new JOSEObjectType(allowedType.trim())));
```
moresandeep, line 100: nit, -1 limit for split appears redundant.

moresandeep (approved):
Looks good.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/impl/JWTToken.java
```diff
@@ -122,6 +129,10 @@ public JWTToken(String alg, String[] claimsArray, List<String> audiences, boolea
     jwt = new SignedJWT(header, claims);
   }
 
+  private String getClaimValue(String[] claims, int index) {
+    return claims == null || claims.length < index ? null : claims[index];
```
zeroflag, line 133: There seems to be an off by one in the validation. Shouldn't this be `<=` ?

```java
return claims == null || claims.length <= index ? null : claims[index];
```

smolnar82: Nice catch! Fixed :)

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
smolnar82: Nice catch! Fixed :)

PR #534, opened 2022-01-25
KNOX-2704 - Honor request id (correlation id) set by upstream loadbalancer

moresandeep:
The PR adds a feature where, when a request comes to Knox with  header `X-Request-Id` then use that as a correlation id. This is useful to track requests that originate from other services or LBs and users need ability to trace requests.

## How was this patch tested?
This patch was tested locally on a local cluster.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayFilter.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayFilter.java
@@ -38,6 +38,7 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletRequestWrapper;
 import javax.servlet.http.HttpServletResponse;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.audit.api.Action;
 import org.apache.knox.gateway.audit.api.ActionOutcome;
 import org.apache.knox.gateway.audit.api.AuditContext;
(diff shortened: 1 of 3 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/GatewayFilter.java
```diff
@@ -150,7 +153,13 @@ public void doFilter( ServletRequest servletRequest, ServletResponse servletResp
       }
     }
 
-    assignCorrelationRequestId();
+    /* If request contains X-Request-Id header use it else use random uuid as correlation id */
+    final String req_id = ((HttpServletRequest) servletRequest).getHeader(REQUEST_ID_HEADER_NAME);
```
zeroflag, line 157: LGTM, just a nitpick here, why don't we use camel case names here?

zeroflag (approved):

gateway-server/src/main/java/org/apache/knox/gateway/GatewayFilter.java
smolnar82: +1

smolnar82 (approved):

gateway-server/src/main/java/org/apache/knox/gateway/GatewayFilter.java
moresandeep: Yeah, I should, I'll update the PR, thanks for catching that!

smolnar82 (approved):
LGTM

PR #537, opened 2022-02-16
KNOX-2707 - Virtual Group Mapping Provider

zeroflag:

A more flexible way to map principals to groups than the existing `group.principal.mapping` mechanism in `CommonIdentityAssertionFilter`.

This patch extends the existing `CommonIdentityAssertionFilter` with a new way to map virtual groups to users.

The syntax allows us to map a virtual group to a user, using arbitrary predicates expressed in a parenthesized prefix notation language.

```
<name>virtual.group.mapping.VIRTUAL-GROUP-NAME</name>
<value>PREDICATE</value>
```

The existing mappings like `principal.mapping` and `group.principal.mapping` are still supported by the provider.

## How was this patch tested?

Manually tested with the following shiro + provider config.

### Shiro config

```xml
        <provider>
            <role>authentication</role>
            <name>ShiroProvider</name>
            <enabled>true</enabled>
            <param>
                <name>sessionTimeout</name>
                <value>30</value>
            </param>
            <param>
                <name>main.ldapRealm</name>
                <value>org.apache.knox.gateway.shirorealm.KnoxLdapRealm</value>
            </param>
            <param>
                <name>main.ldapContextFactory</name>
                <value>org.apache.knox.gateway.shirorealm.KnoxLdapContextFactory</value>
            </param>
            <param>
                <name>main.ldapRealm.contextFactory</name>
                <value>$ldapContextFactory</value>
            </param>
            <param>
                <name>main.ldapRealm.userDnTemplate</name>
                <value>uid={0},ou=people,dc=hadoop,dc=apache,dc=org</value>
            </param>
            <param>
                <name>main.ldapRealm.contextFactory.url</name>
            </param>
            <param>
                <name>main.ldapRealm.contextFactory.authenticationMechanism</name>
                <value>simple</value>
            </param>
            <param>
                <name>urls./**</name>
                <value>authcBasic</value>
            </param>
            <!-- needed to get the groups from ldap -->
            <param>
                <name>main.ldapRealm.authorizationEnabled</name>
                <value>true</value>
            </param>
            <param>
                <name>main.ldapRealm.contextFactory.systemUsername</name>
                <value>uid=admin,ou=people,dc=hadoop,dc=apache,dc=org</value>
            </param>
            <param>
                <name>main.ldapRealm.contextFactory.systemPassword</name>
                <value>admin-password</value>
            </param>
            <param>
                <name>main.ldapRealm.groupSearchBase</name>
                <value>ou=groups,dc=hadoop,dc=apache,dc=org</value>
            </param>
            <param>
                <name>main.ldapRealm.groupObjectClass</name>
                <value>groupofnames</value>
            </param>
            <param>
                <name>main.ldapRealm.groupIdAttribute</name>
                <value>cn</value>
            </param>
            <param>
                <name>main.ldapRealm.memberAttribute</name>
                <value>member</value>
            </param>
            <param>
                <name>main.ldapRealm.memberAttributeValueTemplate</name>
                <value>uid={0},ou=people,dc=hadoop,dc=apache,dc=org</value>
            </param>
```

### Identity assertion provider config

```xml
     <provider>
            <name>Default</name>
            <enabled>true</enabled>
           <!-- existing principal to principal mapping already supported by CommonIdentityAssertionFilter -->
            <param>
                <name>principal.mapping</name>
                <value>tom=mapped-tom;</value>
            </param>
           <!-- existing principal to group mapping already supported by CommonIdentityAssertionFilter -->     
            <param>
                <name>group.principal.mapping</name>
                <value>mapped-tom=lg1;sam=lg2,lg3</value>
            </param>
            <!-- newly added syntax for virtual group -->
            <param>
                <name>virtual.group.mapping.sam-or-admin</name>
                <value>(or (username 'sam') (username 'admin'))</value>
            </param>
            <param>
                <name>virtual.group.mapping.tom-or-admin</name>
                <value>(or (username 'mapped-tom') (username 'admin'))</value>
            </param>
            <param>
                <name>virtual.group.mapping.regexp-group</name>
                <value>(and (match username 'mapped-tom|sam') (match groups 'analyst|scientist'))</value>
            </param>
            <param>
                <name>virtual.group.mapping.group-same-as-user</name>
                <value>(member username)</value>
            </param>
            <param>
                <name>virtual.group.mapping.at-least-one-group</name>
                <value>(!= 0 (size groups))</value>
            </param>
            <param>
                <name>virtual.group.mapping.everybody</name>
                <value>true</value>
            </param>
            <param>
                <name>virtual.group.mapping.both-analyst-and-scientist</name>
                <value>(and (member 'analyst') (member 'scientist'))</value>
            </param>
        </provider>
```

### DIFF
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
@@ -33,18 +33,18 @@ public interface IdentityAsserterMessages {
   @Message( level = MessageLevel.WARN, text = "Invalid mapping parameter name: Missing required group name.")
   void missingVirtualGroupName();
 
-  @Message( level = MessageLevel.WARN, text = "Invalid mapping, parse error: {2}. At {0}={1}")
+  @Message( level = MessageLevel.WARN, text = "Invalid mapping {0}={1}, Parse error: {2}")
   void parseError(String key, String script, SyntaxException e);
 
-  @Message( level = MessageLevel.WARN, text = "Invalid result: {2}. Expected boolean when evaluating: {1}. For virtualGroup: {0}")
+  @Message( level = MessageLevel.WARN, text = "Invalid result: {2}. Expected boolean when evaluating group {0} mapping value {1}.")
   void invalidResult(String virtualGroupName, Ast ast, Object result);
 
-  @Message( level = MessageLevel.DEBUG, text = "Adding user: {0} to virtual group: {1} using predicate: {2}")
+  @Message( level = MessageLevel.DEBUG, text = "Adding user {0} to group {1} based on predicate {2}")
   void addingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
 
-  @Message( level = MessageLevel.DEBUG, text = "Checking user: {0} (with groups: {1}) whether to add virtualGroup: {2} using predicate: {3}")
+  @Message( level = MessageLevel.DEBUG, text = "Checking whether user {0} (with group(s) {1}) should be added to group {2} based on predicate {3}")
   void checkingVirtualGroup(String userName, Set<String> userGroups, String virtualGroupName, Ast ast);
 
-  @Message( level = MessageLevel.DEBUG, text = "User: {0} (with groups: {1}) added to virtual groups: {2}")
-  void virtualGroups(String userName, Set<String> strings, Set<String> virtualGroups);
+  @Message( level = MessageLevel.DEBUG, text = "User {0} (with group(s) {1}) added to group(s) {2}")
+  void virtualGroups(String userName, Set<String> userGroups, Set<String> virtualGroups);
 }
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
@@ -22,7 +22,7 @@ import java.util.Set;
 import org.apache.knox.gateway.i18n.messages.Message;
 import org.apache.knox.gateway.i18n.messages.MessageLevel;
 import org.apache.knox.gateway.i18n.messages.Messages;
-import org.apache.knox.gateway.plang.Ast;
+import org.apache.knox.gateway.plang.AbstractSyntaxTree;
 import org.apache.knox.gateway.plang.SyntaxException;
 
 @Messages(logger="org.apache.knox.gateway")
@@ -37,13 +37,13 @@ public interface IdentityAsserterMessages {
   void parseError(String key, String script, SyntaxException e);
 
   @Message( level = MessageLevel.WARN, text = "Invalid result: {2}. Expected boolean when evaluating group {0} mapping value {1}.")
-  void invalidResult(String virtualGroupName, Ast ast, Object result);
+  void invalidResult(String virtualGroupName, AbstractSyntaxTree ast, Object result);
 
   @Message( level = MessageLevel.DEBUG, text = "Adding user {0} to group {1} based on predicate {2}")
-  void addingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
+  void addingUserToVirtualGroup(String username, String virtualGroupName, AbstractSyntaxTree ast);
 
   @Message( level = MessageLevel.DEBUG, text = "Checking whether user {0} (with group(s) {1}) should be added to group {2} based on predicate {3}")
-  void checkingVirtualGroup(String userName, Set<String> userGroups, String virtualGroupName, Ast ast);
+  void checkingVirtualGroup(String userName, Set<String> userGroups, String virtualGroupName, AbstractSyntaxTree ast);
 
   @Message( level = MessageLevel.DEBUG, text = "User {0} (with group(s) {1}) added to group(s) {2}")
   void virtualGroups(String userName, Set<String> userGroups, Set<String> virtualGroups);
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
@@ -30,10 +30,10 @@ public interface IdentityAsserterMessages {
   @Message( level = MessageLevel.ERROR, text = "Required subject/identity not available.  Check authentication/federation provider for proper configuration." )
   void subjectNotAvailable();
 
-  @Message( level = MessageLevel.WARN, text = "Virtual group name is missing after dot character.")
+  @Message( level = MessageLevel.WARN, text = "Invalid mapping parameter name: Missing required group name.")
   void missingVirtualGroupName();
 
-  @Message( level = MessageLevel.WARN, text = "Parse error: {2}. At {0}={1}")
+  @Message( level = MessageLevel.WARN, text = "Invalid mapping, parse error: {2}. At {0}={1}")
   void parseError(String key, String script, SyntaxException e);
 
   @Message( level = MessageLevel.WARN, text = "Invalid result: {2}. Expected boolean when evaluating: {1}. For virtualGroup: {0}")
@@ -42,12 +42,9 @@ public interface IdentityAsserterMessages {
   @Message( level = MessageLevel.DEBUG, text = "Adding user: {0} to virtual group: {1} using predicate: {2}")
   void addingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
 
-  @Message( level = MessageLevel.DEBUG, text = "Not adding user: {0} to virtual group: {1} using predicate: {2}")
-  void notAddingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
-
   @Message( level = MessageLevel.DEBUG, text = "Checking user: {0} (with groups: {1}) whether to add virtualGroup: {2} using predicate: {3}")
   void checkingVirtualGroup(String userName, Set<String> userGroups, String virtualGroupName, Ast ast);
 
-  @Message( level = MessageLevel.INFO, text = "User: {0} (with groups: {1}) added to virtual groups: {2}")
+  @Message( level = MessageLevel.DEBUG, text = "User: {0} (with groups: {1}) added to virtual groups: {2}")
   void virtualGroups(String userName, Set<String> strings, Set<String> virtualGroups);
 }
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
@@ -17,12 +17,37 @@
  */
 package org.apache.knox.gateway;
 
+import java.util.Set;
+
 import org.apache.knox.gateway.i18n.messages.Message;
 import org.apache.knox.gateway.i18n.messages.MessageLevel;
 import org.apache.knox.gateway.i18n.messages.Messages;
+import org.apache.knox.gateway.plang.Ast;
+import org.apache.knox.gateway.plang.SyntaxException;
 
 @Messages(logger="org.apache.knox.gateway")
 public interface IdentityAsserterMessages {
   @Message( level = MessageLevel.ERROR, text = "Required subject/identity not available.  Check authentication/federation provider for proper configuration." )
   void subjectNotAvailable();
+
+  @Message( level = MessageLevel.WARN, text = "Virtual group name is missing after dot character.")
+  void missingVirtualGroupName();
+
+  @Message( level = MessageLevel.WARN, text = "Parse error: {2}. At {0}={1}")
+  void parseError(String key, String script, SyntaxException e);
+
+  @Message( level = MessageLevel.WARN, text = "Invalid result: {2}. Expected boolean when evaluating: {1}. For virtualGroup: {0}")
+  void invalidResult(String virtualGroupName, Ast ast, Object result);
+
+  @Message( level = MessageLevel.DEBUG, text = "Adding user: {0} to virtual group: {1} using predicate: {2}")
+  void addingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
+
+  @Message( level = MessageLevel.DEBUG, text = "Not adding user: {0} to virtual group: {1} using predicate: {2}")
+  void notAddingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
+
+  @Message( level = MessageLevel.DEBUG, text = "Checking user: {0} (with groups: {1}) whether to add virtualGroup: {2} using predicate: {3}")
+  void checkingVirtualGroup(String userName, Set<String> userGroups, String virtualGroupName, Ast ast);
+
+  @Message( level = MessageLevel.INFO, text = "User: {0} (with groups: {1}) added to virtual groups: {2}")
+  void virtualGroups(String userName, Set<String> strings, Set<String> virtualGroups);
 }
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
@@ -40,7 +40,7 @@ import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.IdentityAsserterMessages;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
-import org.apache.knox.gateway.plang.Ast;
+import org.apache.knox.gateway.plang.AbstractSyntaxTree;
 import org.apache.knox.gateway.plang.Parser;
 import org.apache.knox.gateway.plang.SyntaxException;
 import org.apache.knox.gateway.security.GroupPrincipal;
@@ -77,8 +77,8 @@ public class CommonIdentityAssertionFilter extends AbstractIdentityAssertionFilt
     virtualGroupMapper = new VirtualGroupMapper(loadVirtualGroups(filterConfig));
   }
 
-  private Map<String, Ast> loadVirtualGroups(FilterConfig filterConfig) {
-    Map<String, Ast> predicateToGroupMapping = new HashMap<>();
+  private Map<String, AbstractSyntaxTree> loadVirtualGroups(FilterConfig filterConfig) {
+    Map<String, AbstractSyntaxTree> predicateToGroupMapping = new HashMap<>();
     loadVirtualGroupConfig(filterConfig, predicateToGroupMapping);
     if (predicateToGroupMapping.isEmpty() && filterConfig.getServletContext() != null) {
       loadVirtualGroupConfig(filterConfig.getServletContext(), predicateToGroupMapping);
@@ -89,10 +89,10 @@ public class CommonIdentityAssertionFilter extends AbstractIdentityAssertionFilt
     return predicateToGroupMapping;
   }
 
-  private void loadVirtualGroupConfig(FilterConfig config, Map<String, Ast> result) {
+  private void loadVirtualGroupConfig(FilterConfig config, Map<String, AbstractSyntaxTree> result) {
     for (String paramName : virtualGroupParameterNames(config.getInitParameterNames())) {
       try {
-        Ast ast = parser.parse(config.getInitParameter(paramName));
+        AbstractSyntaxTree ast = parser.parse(config.getInitParameter(paramName));
         result.put(paramName.substring(VIRTUAL_GROUP_MAPPING_PREFIX.length()).trim(), ast);
       } catch (SyntaxException e) {
         LOG.parseError(paramName, config.getInitParameter(paramName), e);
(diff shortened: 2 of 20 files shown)


gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
```diff
@@ -59,6 +74,55 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
+      loadVirtualGroupConfig(filterConfig.getServletContext(), predicateToGroupMapping);
+    }
+    if (predicateToGroupMapping.keySet().stream().anyMatch(StringUtils::isBlank)) {
+      LOG.missingVirtualGroupName();
+    }
+    return predicateToGroupMapping;
+  }
+
+  private void loadVirtualGroupConfig(FilterConfig config, Map<String, Ast> result) {
+    for (String paramName : virtualGroupParameterNames(config.getInitParameterNames())) {
```
pzampino, line 93: Seems like unnecessarily iterating over the group mapping params twice; It's not clear how much value is added by the virtualGroupParameterNames() method.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
```diff
@@ -0,0 +1,87 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.plang;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class Ast {
```
pzampino, line 23: What is Ast?

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Parser.java
```diff
@@ -0,0 +1,109 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.plang;
+
+import java.io.IOException;
+import java.io.PushbackReader;
+import java.io.StringReader;
+
+public class Parser {
```
pzampino, line 24: Should all of this parsing-related code go into the gateway-util-common module or its own module?

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
```diff
@@ -17,12 +17,37 @@
...
 import org.apache.knox.gateway.i18n.messages.Messages;
+import org.apache.knox.gateway.plang.Ast;
+import org.apache.knox.gateway.plang.SyntaxException;
 
 @Messages(logger="org.apache.knox.gateway")
 public interface IdentityAsserterMessages {
   @Message( level = MessageLevel.ERROR, text = "Required subject/identity not available.  Check authentication/federation provider for proper configuration." )
   void subjectNotAvailable();
+
+  @Message( level = MessageLevel.WARN, text = "Virtual group name is missing after dot character.")
```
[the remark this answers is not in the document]
zeroflag: We put the users into these groups based on their LDAP/OS level groups. If we refer both of the virtual groups and LDAP/OS groups simply as groups, it can be confusing, isn't it? I'm not sure what would be the proper naming.

> Is this in reference to param names? If so, perhaps something like "Invalid mapping parameter name: Missing required group name."

Ok. I changed the the message.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
```diff
@@ -17,12 +17,37 @@
...
+  void missingVirtualGroupName();
+
+  @Message( level = MessageLevel.WARN, text = "Parse error: {2}. At {0}={1}")
```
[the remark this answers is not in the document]
zeroflag: Few parsing error message examples:

* unterminated string (when someone forgets the closing ')
* Unexpected closing (when there are more closing parentheses then open parentheses)
* Missing closing ) (when there are more open parentheses then closing parentheses)

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
```diff
@@ -17,12 +17,37 @@
...
+  void parseError(String key, String script, SyntaxException e);
+
+  @Message( level = MessageLevel.WARN, text = "Invalid result: {2}. Expected boolean when evaluating: {1}. For virtualGroup: {0}")
```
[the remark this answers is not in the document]
zeroflag: I think this is a serious, but unlikely problem. So the predicate should always evaluate to a boolean value.

For example:

```(= username 'admin')```

But if the user specifies an expression that returns a string or a number, we won't know what to to and whether we should put the user into the virtual group or not.

File: gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java [hunk 7]
```diff
@@ -17,12 +17,37 @@
...
+  void invalidResult(String virtualGroupName, Ast ast, Object result);
+
+  @Message( level = MessageLevel.DEBUG, text = "Adding user: {0} to virtual group: {1} using predicate: {2}")
+  void addingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
+
+  @Message( level = MessageLevel.DEBUG, text = "Not adding user: {0} to virtual group: {1} using predicate: {2}")
+  void notAddingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
```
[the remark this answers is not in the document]
zeroflag: Right, this is probably overkill. I was helpful during testing and development but no longer needed. I'll remove it.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
```diff
@@ -17,12 +17,37 @@
...
+
+  @Message( level = MessageLevel.DEBUG, text = "Checking user: {0} (with groups: {1}) whether to add virtualGroup: {2} using predicate: {3}")
+  void checkingVirtualGroup(String userName, Set<String> userGroups, String virtualGroupName, Ast ast);
+
+  @Message( level = MessageLevel.INFO, text = "User: {0} (with groups: {1}) added to virtual groups: {2}")
```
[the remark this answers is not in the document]
zeroflag: Yes, audit logs shows the end result. This is only useful for showing the original groups as well. How about changing it to DEBUG?

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
zeroflag: This happens only in deployment time once, so from performance point of view it's not a big problem.

At first we try to load the groups from the FilterConfig then from the ServletContext (this is how we load other properties from other filters too).

Unfortunately FilterConfig and ServletContext doesn't have the same interface, that's why we have this duplication. The `virtualGroupParameterNames` was an attempt to reduce the duplication so at least the `if` statement is not duplicated in both methods.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/VirtualGroupMapper.java
```diff
@@ -0,0 +1,97 @@
...
+    private final Map<String, Ast> virtualGroupToPredicateMap;
+
+    public VirtualGroupMapper(Map<String, Ast> virtualGroupToPredicateMap) {
+        this.virtualGroupToPredicateMap = virtualGroupToPredicateMap;
+    }
+
+    /**
+     *  @return all virtual groups where the corresponding predicate matches
+     */
+    public Set<String> virtualGroupsOfUser(String username, Set<String> groups, ServletRequest request) {
```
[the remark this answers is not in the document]
zeroflag: Ok, renamed it.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
zeroflag: AST stands for Abstract Syntax Tree, it is language implementation terminology. The parser produces the AST, which is a tree structure that represents the source code. In our cases it's list of lists (which is also a tree). 

Fr example this is the text:

```
(or true (and false true))
```

This is the AST emitted by the parser. `See the AST>>toString`

```
[or, true, [and, false, true]]
```

It's a homoiconic language so there is 1 to 1 mapping between the text and the AST. It has the exact same structure, we just put everything into Java types, like `java.util.ArrayLists` and `java.lang.Booleans`.

We could have represented the AST as arbitrarily nested a `java.util.List`, but the Java's type system doesn't handle this nesting very well: `List<List<List<..> or Atom> or Atom>`, that's why we have a dedicated class.

The interpreter gets the AST and evaluates it. In our case the interpreter is a simple tree walking interpreter so it just recursively  evaluates the tree.

In the filter we store the AST not the text, so every time a request comes in, we just need to interpret, without using the parser.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Parser.java
zeroflag: Good questions. We can extract in into its own module. Although it's only few hundred lines of code. We also need to come up with a good name for the module if we choose to extract it out.

lmccay: I'd like to see it as reusable as possible. I see this AST language approach being applicable across multiple provider types: identity assertion for group mapping, a new authorization provider for richer ACLs/rules, perhaps a new provider type for routing or being used within existing dispatches for the same. I don't know that the initial implementation needs to account for this though as we can refactor later. Putting into gateway-util-common will go a long way though. It will also make it available in a lot of other consumer modules without having to add a new dependency.

zeroflag: The parser covers the full syntax, it is unlikely we need to change that unless we want to add new literal syntax for something (e.g.: set/map/vector). So I think it's both stable and reusable. The interpreter is extensible as it allows us to add functions from the outside depending on the use case. We can keep it under gateway-util-common for now and extract later only if needed.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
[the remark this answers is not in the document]
pzampino: Ok, I can buy that argument.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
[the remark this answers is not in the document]
pzampino: I guess with the referenced context (e.g., mapping name, predicate and actual result), that works.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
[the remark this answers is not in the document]
pzampino: DEBUG would be a good compromise.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
pzampino: Perhaps, renaming the class to AbstractSyntaxtTree and/or some javadoc would be helpful?

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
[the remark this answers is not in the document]
zeroflag: I changed it to DEBUG.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
```diff
@@ -17,12 +17,34 @@
...
   @Message( level = MessageLevel.ERROR, text = "Required subject/identity not available.  Check authentication/federation provider for proper configuration." )
   void subjectNotAvailable();
+
+  @Message( level = MessageLevel.WARN, text = "Invalid mapping parameter name: Missing required group name.")
+  void missingVirtualGroupName();
+
+  @Message( level = MessageLevel.WARN, text = "Invalid mapping, parse error: {2}. At {0}={1}")
+  void parseError(String key, String script, SyntaxException e);
+
+  @Message( level = MessageLevel.WARN, text = "Invalid result: {2}. Expected boolean when evaluating: {1}. For virtualGroup: {0}")
```
pzampino, line 39: Suggestion: "Invalid result: {2}. Expected boolean when evaluating group {0} mapping value {1}."

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
```diff
@@ -17,12 +17,34 @@
...
+  void invalidResult(String virtualGroupName, Ast ast, Object result);
+
+  @Message( level = MessageLevel.DEBUG, text = "Adding user: {0} to virtual group: {1} using predicate: {2}")
```
pzampino, line 42: Suggestion: "Adding user {0} to group {1} based on predicate {2}"

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
```diff
@@ -17,12 +17,34 @@
...
+  void addingUserToVirtualGroup(String username, String virtualGroupName, Ast ast);
+
+  @Message( level = MessageLevel.DEBUG, text = "Checking user: {0} (with groups: {1}) whether to add virtualGroup: {2} using predicate: {3}")
```
pzampino, line 45: Suggestion: "Checking whether user {0} (with group(s) {1}) should be added to group {2} based on predicate {3}"

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/IdentityAsserterMessages.java
```diff
@@ -17,12 +17,34 @@
...
+  void checkingVirtualGroup(String userName, Set<String> userGroups, String virtualGroupName, Ast ast);
+
+  @Message( level = MessageLevel.DEBUG, text = "User: {0} (with groups: {1}) added to virtual groups: {2}")
```
pzampino, line 48: Suggestion: "User {0} (with group(s) {1}) added to group(s) {2}"

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
```diff
@@ -17,30 +17,45 @@
...
+import org.apache.knox.gateway.plang.SyntaxException;
+import org.apache.knox.gateway.security.GroupPrincipal;
 import org.apache.knox.gateway.security.principal.PrincipalMappingException;
 import org.apache.knox.gateway.security.principal.SimplePrincipalMapper;
 
-import java.io.IOException;
-import java.security.AccessController;
-
 public class CommonIdentityAssertionFilter extends AbstractIdentityAssertionFilter {
+  public static final String VIRTUAL_GROUP_MAPPING_PREFIX = "virtual.group.mapping.";
```
pzampino, line 51: I thought we agreed to remove the "virtual" language from the configuration. Did we revert that agreement?
I think it's fine in the code, but from a configuration-perspective, is there anything preventing mapping a user to a "real" group with this mechanism?

pzampino (commented):
It's difficult for me to tell, but is there any testing of the combined (old/new) mapping configuration syntax?

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
zeroflag: I removed the virtual prefix, now it's just "group.mapping"

The provider supports the following:

* group.principal.mapping -> kminder=group1;lmccay=mrgroup,mrducks
* principal.mapping -> lmccay,kminder=hdfs;newuser=mapred
* group.mapping -> (username 'admin')

The 3rd one is the new one.

gateway-provider-identity-assertion-common/src/test/java/org/apache/knox/gateway/identityasserter/common/filter/VirtualGroupMapperTest.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.apache.knox.gateway.plang.Ast;
+import org.apache.knox.gateway.plang.Parser;
+import org.junit.Test;
+
+@SuppressWarnings("PMD.NonStaticInitializer")
+public class VirtualGroupMapperTest {
```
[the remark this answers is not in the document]
zeroflag: I added a new one.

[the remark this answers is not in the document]
zeroflag: I added a new one.

zeroflag:
> It's difficult for me to tell, but is there any testing of the combined (old/new) mapping configuration syntax?

Yes, in the `CommonIdentityAssertionFilterTest` there is one for testing the combination of both. Testing this feature at the Filter level is cumbersome because of the infrastructure related code and dependencies (servlet api and such), so the more extensive tests are in `VirtualGroupMapperTests`, `ParserTest` and `InterpreterTest`. Those are more focused and easier to write tests for.

gateway-provider-identity-assertion-common/src/test/java/org/apache/knox/gateway/identityasserter/filter/CommonIdentityAssertionFilterTest.java
```diff
@@ -74,6 +76,7 @@ public String mapUserPrincipal(String principalName) {
         return super.combineGroupMappings(mappedGroups, groups);
       }
     };
+    ThreadContext.put(MDC_AUDIT_CONTEXT_KEY, "dummy");
```
pzampino, line 79: I'm assuming this is for backward-compatibility. Is that correct?

zeroflag: The test fails with an NPE if there is no AuditContext (auditService.getContext() returns null). It is just for making the test work.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
pzampino: I would still like to see this class name spelled out as AbstractSyntaxtTree.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
```diff
@@ -0,0 +1,87 @@
...
+    public String toString() {
+        return isAtom() ? token : children.toString();
+    }
+
+    public String token() {
+        return token;
+    }
+
+    public boolean isStr() {
+        return token != null && token.startsWith("'") && token.endsWith("'");
```
pzampino, line 54: Are tokens allowed to be empty (i.e., '')?

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public boolean isNumber() {
+        if (token == null) {
```
pzampino, line 58: I personally prefer single return statements whenever possible.
```suggestion
boolean result = false;
if (token !=null) {
  try {
     numValue();
     result = true;
   } catch (NumberFormatException e) {
     // NaN
   }
}
return result;

File: gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java [hunk 19]
```diff
@@ -0,0 +1,87 @@
...
+    public Number numValue() {
+        try {
+            return Long.parseLong(token);
+        } catch (NumberFormatException e) {
+            return Double.parseDouble(token);
+        }
+    }
+
+    public String strValue() {
+        return token.substring(1, token.length() -1);
```
pzampino, line 78: Would it be overkill to wrap this in a call to isStr() for validation?

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
zeroflag: Ok, I'll rename it.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
zeroflag: Yes, that's an empty string. It is allowed like in other languages.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
zeroflag: Ok, I can add a check and throw an exception if it's not a string. There is already a check at the outside so it's not strictly necessary but I understand the concern.

pzampino: I noticed that, and given the current usage, it's not concerning, but strValue() is a public method which could be called by anyone in theory, so better to be safer in that case IMO.

gateway-util-common/src/main/java/org/apache/knox/gateway/plang/Ast.java
zeroflag: Interesting. I like to get over the easy checks (like if the argument is null) first and do an early return if something is trivially invalid. Then everything below that can expect valid arguments. You would probably like Lisp since it's expression based (there are no statements - including explicit `return` - in it). :)

I'll change this too.

pzampino (approved):
LGTM

PR #554, opened 2022-04-12
KNOX-2731 Allow group membership information to be included in issued JWTs

zeroflag:

The KnoxToken service should allow for the option to have group membership information for the authenticated user included in the issued JWT.

Refactor: I removed (redundant and unnecessary) claims array.

## How was this patch tested?

1. Enabled include groups in KNOXTOKEN service:

````
         <param>
            <name>knox.token.include.groups</name>
            <value>true</value>
        </param>
````

2. Added the admin user to admin-group1 and admin-group2

```
      <provider>
         <name>Default</name>
         <enabled>true</enabled>
         <param>
           <name>group.mapping.admin-group1</name>
           <value>true</value>
         </param>
         <param>
           <name>group.mapping.admin-group2</name>
           <value>true</value>
         </param>
      </provider>
```

3. Generated a token on the UI (gateway/homepage/tokengen/index.html) and checked the groups in the jwt payload:

```
{
  "sub": "admin",
  "jku": "https://localhost:8443/gateway/homepage/knoxtoken/api/v1/jwks.json",
  "kid": "oigA7mZCwA2d7oimQyUaB0oDAfhI-1Bjq9y1n-Mw_OU",
  "iss": "KNOXSSO",
  "exp": 1649777837,
  "knox.groups": [
    "admin-group2",
    "admin-group1"
  ],
  "managed.token": "true",
  "knox.id": "dfeb8979-7f00-4938-bbff-1bc7574bb53d"
}
```

4. Repeated the same with knox.token.include.groups = false

### CODE REVIEW

File: gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java [hunk 1]
```diff
@@ -104,26 +103,15 @@ public void setAliasService(AliasService as) {
 
   @Override
   public JWT issueToken(JWTokenAttributes jwtAttributes) throws TokenServiceException {
-    String[] claimArray = new String[6];
```
pzampino, line 107: JWTokenAttributes has replaced the claimsArray throughout?

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/JWTokenAttributesBuilder.java
```diff
@@ -32,6 +34,9 @@ public class JWTokenAttributesBuilder {
   private boolean managed;
   private String jku;
   private String type;
+  private Set<String> groups;
+  private String kid;
+  private String issuer = "KNOXSSO";
```
pzampino, line 39: I think this was like this before, but should the issuer really be KNOXSSO in ALL cases?

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -104,26 +103,15 @@ public void setAliasService(AliasService as) {
...
-      claimArray[3] = null;
-    }
-    else {
-      claimArray[3] = String.valueOf(jwtAttributes.getExpires());
-    }
     final String algorithm = jwtAttributes.getAlgorithm();
     if(SUPPORTED_HMAC_SIG_ALGS.contains(algorithm)) {
-      claimArray[4] = null;
-      claimArray[5] = null;
+      jwtAttributes.setKid(null);
```
pzampino, line 108: Do these need to be explicitly set to null, or could this code be simplified?

gateway-server/src/test/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateServiceTest.java
```diff
@@ -369,16 +370,7 @@ protected TokenStateService createTokenStateService() throws Exception {
...
-    claims[0] = "KNOXSSO";
-    claims[1] = "john.doe@example.com";
-    claims[2] = "https://login.example.com";
-    if(expiry > 0) {
-      claims[3] = Long.toString(expiry);
-    }
-    claims[4] = "E0LDZulQ0XE_otJ5aoQtQu-RnXv8hU-M9U4dD7vDioA";
-    claims[5] = null;
-    JWT token = new JWTToken("RS256", claims);
+    JWT token = new JWTToken(new JWTokenAttributesBuilder().setExpires(expiry).setAlgorithm("RS256").build());
```
pzampino, line 373: This test is slightly different in that the test token has differing contents. Has the coverage of this test changed as a result? Or, does it only reflect the change in handling of claims, whereas now, previously necessary claims are no longer necessary?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -106,6 +111,7 @@ public class TokenResource {
   private static final String TOKEN_TTL_PARAM = "knox.token.ttl";
   private static final String TOKEN_TYPE_PARAM = "knox.token.type";
   private static final String TOKEN_AUDIENCES_PARAM = "knox.token.audiences";
+  private static final String TOKEN_INCLUDE_GROUPS_IN_JWT = "knox.token.include.groups";
```
pzampino, line 114: MINOR: It may be good to define a TOKEN_PARAM_PREFIX="knox.token.", and reference this from the other config property name values to avoid accidental variations across these properties. I realize it was this way before your changes, but while you're here, it may be worth doing.
For example:
private static final String TOKEN_PARAM_PREFIX = "knox.token.";
private static final String TOKEN_TTLE_PARAM = TOKEN_PARAM_PREFIX + "ttl";
private static final String TOKEN_TYPE_PARAM = TOKEN_PARAM_PREFIX + "type";
private static final String TOKEN_AUDIENCES_PARAM = TOKEN_PARAM_PREFIX + "audiences";

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -228,6 +235,9 @@ public void init() throws AliasServiceException, ServiceLifecycleException, KeyL
       }
     }
 
+    String shouldIncludeGroupsParam = context.getInitParameter(TOKEN_INCLUDE_GROUPS_IN_JWT);
+    shouldIncludeGroups = shouldIncludeGroupsParam == null ? false : Boolean.parseBoolean(shouldIncludeGroupsParam);
```
pzampino, line 239: It looks like this is defaulting to false, and the explicit false in the homepage topology should be unnecessary.

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
```diff
@@ -1396,8 +1444,8 @@ private Subject createTestSubject(final String username) {
     return s;
   }
 
-  private static Map<String, String> parseJSONResponse(final String response) throws IOException {
-    return (new ObjectMapper()).readValue(response, new TypeReference<Map<String, String>>(){});
+  private static Map<String, Object> parseJSONResponse(final String response) throws IOException {
```
pzampino, line 1447: Why was this changed to return a Map of Object values? What is the response including that isn't a String?

pzampino (changes requested):

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/JWTokenAttributesBuilder.java
lmccay: I thought we made this configurable at some point. That change would likely be better as a separate PR though.

gateway-release/home/conf/topologies/homepage.xml
```diff
@@ -113,6 +113,10 @@
           <name>knox.token.proxyuser.admin.hosts</name>
           <value>*</value>
       </param>
+      <param>
+         <name>knox.token.include.groups</name>
+         <value>false</value>
```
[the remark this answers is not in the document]
zeroflag: Yes, the default is false so it is not needed. It was added here just to show the user that this option now available (it is easier to turn it on if we already have this param in the config).

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
zeroflag: Yes, I discussed it with @smolnar82 and we decided to remove the claimsArray, and use the JWTTokenAttributes instead.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
zeroflag: We can remove it from there. It only serves as "documentation" of a new config parameter.

gateway-service-knoxtoken/src/test/java/org/apache/knox/gateway/service/knoxtoken/TokenServiceResourceTest.java
zeroflag: Now the map can contain a Set value for the groups.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
zeroflag: Ok, good idea, I'll change this.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/JWTokenAttributesBuilder.java
zeroflag: We have a `jwt.expected.issuer` config param which is used in `validateToken`. But the it seems the issuer used by the `issueToken` has always been `KNOXSSO`

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
zeroflag: It follows the existing behaviour. @moresandeep do you remember why we need to null out the the jku + kid when the sign algorithm is "HS256", "HS384" or "HS512" ?

gateway-server/src/test/java/org/apache/knox/gateway/services/token/impl/DefaultTokenStateServiceTest.java
zeroflag: Putting it back or removing it makes no difference. These attributes are not checked in this test. `JWTTokenTest` covers the attribute checks.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/JWTokenAttributesBuilder.java
pzampino: I agree that it need not be part of this PR, but I think it should be a configurable value.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -827,6 +837,10 @@ private Response getAuthenticationToken() {
     return Response.ok().entity("{ \"Unable to acquire token.\" }").build();
   }
 
+  private boolean shouldIncludeGroups() {
+    return Boolean.parseBoolean(request.getParameter(KNOX_TOKEN_INCLUDE_GROUPS));
```
pzampino, line 841: What happens if someone specifies a non-boolean value for the request param? No one SHOULD do that, but what happens if they do?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -144,6 +144,7 @@ public class TokenResource {
   private static final String TARGET_ENDPOINT_PULIC_CERT_PEM = TOKEN_PARAM_PREFIX + "target.endpoint.cert.pem";
   static final String QUERY_PARAMETER_DOAS = "doAs";
   static final String PROXYUSER_PREFIX = TOKEN_PARAM_PREFIX + "proxyuser";
+  public static final String KNOX_TOKEN_INCLUDE_GROUPS = "knox.token.include.groups";
```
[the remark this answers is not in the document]
zeroflag: I fixed it.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
pzampino: Ok, cool.

pzampino (approved):

smolnar82 (approved):
LGTM

PR #560, opened 2022-04-19
KNOX-2732 Issuer claim in Knox JWTs should be configurable

zeroflag:

The issuer of the JWT tokens is now configurable.

## How was this patch tested?

1. Added `knox.token.issuer` to KNOXTOKEN service.

```
      <param>
          <name>knox.token.issuer</name>
          <value>test 1 2 3</value>
      </param>
````

2. Created a token on the UI and checked the issuer.
3. Removed the `knox.token.issuer` from the config and checked the issuer (default KNOXSSO) again

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
@@ -51,10 +51,12 @@ public class JWTAccessTokenAssertionFilter extends AbstractIdentityAssertionFilt
   private static final String TOKEN_TYPE = "token_type";
   private static final String ACCESS_TOKEN = "access_token";
   private static final String BEARER = "Bearer ";
+  public static final String ISSUER = "knox.token.issuer";
   private static JWTMessages log = MessagesFactory.get( JWTMessages.class );
   private long validity;
   private JWTokenAuthority authority;
   private ServiceRegistry sr;
+  private String tokenIssuer;
 
   @Override
   public void init( FilterConfig filterConfig ) throws ServletException {
@@ -68,6 +70,10 @@ public class JWTAccessTokenAssertionFilter extends AbstractIdentityAssertionFilt
     GatewayServices services = (GatewayServices) filterConfig.getServletContext().getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
     authority = services.getService(ServiceType.TOKEN_SERVICE);
     sr = services.getService(ServiceType.SERVICE_REGISTRY_SERVICE);
+
+    this.tokenIssuer = filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER) != null
+            ? filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER)
+            : JWTokenAttributes.DEFAULT_ISSUER;
   }
 
   @Override
@@ -141,7 +147,13 @@ public class JWTAccessTokenAssertionFilter extends AbstractIdentityAssertionFilt
 
     JWT token;
     try {
-      final JWTokenAttributes jwtAttributes = new JWTokenAttributesBuilder().setUserName(principalName).setAudiences(serviceName).setAlgorithm(signatureAlgorithm).setExpires(expires).build();
+      final JWTokenAttributes jwtAttributes = new JWTokenAttributesBuilder()
+              .setIssuer(tokenIssuer)
+              .setUserName(principalName)
+              .setAudiences(serviceName)
+              .setAlgorithm(signatureAlgorithm)
+              .setExpires(expires)
+              .build();
       token = authority.issueToken(jwtAttributes);
       // Coverity CID 1327961
       if( token != null ) {
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
@@ -31,6 +31,7 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.filter.security.AbstractIdentityAssertionFilter;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.provider.federation.jwt.JWTMessages;
@@ -71,9 +72,9 @@ public class JWTAccessTokenAssertionFilter extends AbstractIdentityAssertionFilt
     authority = services.getService(ServiceType.TOKEN_SERVICE);
     sr = services.getService(ServiceType.SERVICE_REGISTRY_SERVICE);
 
-    this.tokenIssuer = filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER) != null
-            ? filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER)
-            : JWTokenAttributes.DEFAULT_ISSUER;
+    this.tokenIssuer = StringUtils.isEmpty(filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER))
+            ? JWTokenAttributes.DEFAULT_ISSUER
+            : filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java [hunk 1]
```diff
@@ -68,6 +70,10 @@ public void init( FilterConfig filterConfig ) throws ServletException {
     GatewayServices services = (GatewayServices) filterConfig.getServletContext().getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
     authority = services.getService(ServiceType.TOKEN_SERVICE);
     sr = services.getService(ServiceType.SERVICE_REGISTRY_SERVICE);
+
+    this.tokenIssuer = filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER) != null
```
moresandeep, line 74: We should check for a blank string too.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
```diff
@@ -141,7 +147,13 @@ private String getAccessToken(final String principalName, String serviceName, lo
 
     JWT token;
     try {
-      final JWTokenAttributes jwtAttributes = new JWTokenAttributesBuilder().setUserName(principalName).setAudiences(serviceName).setAlgorithm(signatureAlgorithm).setExpires(expires).build();
+      final JWTokenAttributes jwtAttributes = new JWTokenAttributesBuilder()
```
moresandeep, line 150: Thanks :)

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAuthCodeAssertionFilter.java
```diff
@@ -55,6 +57,10 @@ public void init( FilterConfig filterConfig ) throws ServletException {
     GatewayServices services = (GatewayServices) filterConfig.getServletContext().getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
     authority = services.getService(ServiceType.TOKEN_SERVICE);
     sr = services.getService(ServiceType.SERVICE_REGISTRY_SERVICE);
+
+    this.tokenIssuer = filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER) != null
```
moresandeep, line 61: We should check for blank string.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAuthCodeAssertionFilter.java
```diff
@@ -42,6 +43,7 @@ public class JWTAuthCodeAssertionFilter extends AbstractIdentityAssertionFilter
   private JWTokenAuthority authority;
 
   private ServiceRegistry sr;
+  private String tokenIssuer;
 
   @Override
   public void init( FilterConfig filterConfig ) throws ServletException {
```
moresandeep, line 49: Not related to this change but we have some code duplication here with `JWTAccessTokenAssertionFilter` We can prolly combine it into a common base class. Not required here but would be nice.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAuthCodeAssertionFilter.java
```diff
@@ -55,6 +57,10 @@ public void init( FilterConfig filterConfig ) throws ServletException {
     GatewayServices services = (GatewayServices) filterConfig.getServletContext().getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
     authority = services.getService(ServiceType.TOKEN_SERVICE);
     sr = services.getService(ServiceType.SERVICE_REGISTRY_SERVICE);
+
+    this.tokenIssuer = filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER) != null
+            ? filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER)
```
moresandeep, line 62: As per RFC https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1 issuer is a `principal` and not principals but could there be cases where we might need to support multiple issuers? if not we are golden. Just wanted to raise this.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -242,6 +243,9 @@ public void init() throws AliasServiceException, ServiceLifecycleException, KeyL
             ? true
             : Boolean.parseBoolean(includeGroupsInTokenAllowedParam);
 
+    this.tokenIssuer = context.getInitParameter(KNOX_TOKEN_ISSUER) != null
```
moresandeep, line 246: Check for blank string.

moresandeep (changes requested):

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
zeroflag: I replaced the null check with `StringUtils.isEmpty`.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAuthCodeAssertionFilter.java
zeroflag: I don't know how likely this is, but the doc says it's a StringOrURI not an array. If we ever need to support multiple issuers we can still use coma separated values I guess, or introduce new application specific claim which is an array.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAuthCodeAssertionFilter.java
zeroflag: Yes, or maybe factor out some parts to some other classes and reference those from them. But this is probably out of scope of this change.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAccessTokenAssertionFilter.java
```diff
@@ -71,9 +72,9 @@ public void init( FilterConfig filterConfig ) throws ServletException {
     authority = services.getService(ServiceType.TOKEN_SERVICE);
     sr = services.getService(ServiceType.SERVICE_REGISTRY_SERVICE);
 
-    this.tokenIssuer = filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER) != null
-            ? filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER)
-            : JWTokenAttributes.DEFAULT_ISSUER;
+    this.tokenIssuer = StringUtils.isEmpty(filterConfig.getInitParameter(JWTAccessTokenAssertionFilter.ISSUER))
```
moresandeep, line 75: StringUtils.isBlank() is a little better since it checks for whitespace too.

lmccay (changes requested):
It seems to me that we likely need to add this logic to the JWTFederation provider and SSOCookieProvider as well. If we get an issuer other than KNOXSSO then we need to know whether we should expect something else and likely whether we have public keys for that issuer. If we don't do we need to rely on something in the token to retrieve the public keys for signature verification? How do we know we can trust those endpoints?

zeroflag:
@lmccay 

The token validator used by `SSOCookieFederationFilter` and `JWTFederationFilter` already has a config param called `jwt.expected.issuer` (which defaults to KNOXSSO).


This should be in sync with the newly added `knox.token.issuer`

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTAuthCodeAssertionFilter.java
pzampino: I can't imagine a case for multiple issuers. In that case, we would use multiple topologies IMO.

pzampino (approved):
LGTM

zeroflag:
Hey @lmccay does my above comment make sense and address your concern?

PR #578, opened 2022-05-18
KNOX-2736 Knox clients should support retry/failover - addendum

zeroflag:

A small change that was left out from KNOX-2736. 

## How was this patch tested?

N/A


smolnar82 (approved):
LGTM

PR #579, opened 2022-05-18
KNOX-2726 - Impersonation Params should be configurable

moresandeep:
Impersonation Params should be configurable so that Knox can scrub these params before sending it out to the backend services. There could be cases where some services might use their own impersonation params (similar to doAs used by Knox). Including these parameters in request could have unknown side effects. In order to prevent these issue, Impersonation Params should be configurable. 

The PR proposes adding `impersonation.params` parameter to the `identity-assertion` provider. 
This can be added in `service.xml` file or in topology file.

e.g. `service.xml`

```
 <policies>
	  <policy role="authentication"/>
	  <policy role="rewrite"/>
      <policy role="identity-assertion" name="Default">
	      <param>
	           <name>impersonation.params</name>
	           <value>customID, customServiceID</value>
	       </param>
	  </policy>
  </policies>
```
**NOTE: `authentication` and `rewrite` policies are required for proper functioning.**

e.g. `topology.xml`
```
      <provider>
            <name>Default</name>
            <enabled>true</enabled>
			<param>
			        <name>impersonation.params</name>
			        <value>customID</value>
			    </param>
        </provider>
```

As a side effect of this, we now have ability to specify parameters for providers in `service.xml` file, providers as defined in policy. e.g.

```
 <policies>
	  <policy role="authentication"/>
	  <policy role="rewrite"/>
      <policy role="identity-assertion" name="Default">
	      <param>
	           <name>impersonation.params</name>
	           <value>customID, customServiceID</value>
	       </param>
	  </policy>
  </policies>
```

**NOTE: `authentication` and `rewrite` policies are required for proper functioning.**

Adding this parameter tells Knox to scrub this parameter from the dispatched request. 
e.g. 
```
2022-05-18 07:06:12,733 cde0df3d-5fc8-4b7a-8b96-0f828f4e524e WARN  knox.gateway (IdentityAsserterHttpServletRequestWrapper.java:scrubOfExistingPrincipalParams(212)) - Possible identity spoofing attempt - impersonation parameter removed: customID
```

## How was this patch tested?
This patch was tested locally

Logs without `impersonation.params` 

```
2022-05-18 07:11:02,332 aadcfc0f-5159-4c50-bacb-0308a496ad32 DEBUG knox.gateway (UrlRewriteProcessor.java:rewrite(162)) - Rewrote URL: https://localhost:8443/gateway/sandbox/weather/data/2.5/forecast/city?id=524901&APPID=54557732afcfe106bfc955b9da04fb14&customID=somerandomestring, direction: IN via implicit rule: WEATHER/weather/inbound to URL: http://api.openweathermap.org:80/data/2.5/forecast/city?id=524901&APPID=54557732afcfe106bfc955b9da04fb14&customID=somerandomestring
2022-05-18 07:11:02,333 aadcfc0f-5159-4c50-bacb-0308a496ad32 DEBUG knox.gateway (DefaultDispatch.java:executeOutboundRequest(157)) - Dispatch request: GET http://api.openweathermap.org:80/data/2.5/forecast/city?id=524901&APPID=54557732afcfe106bfc955b9da04fb14&customID=somerandomestring&user.name=guest
```

Logs with `impersonation.params`

```
557732afcfe106bfc955b9da04fb14&customID=somerandomestring, direction: IN via implicit rule: WEATHER/weather/inbound to URL: http://api.openweathermap.org:80/data/2.5/forecast/city?id=524901&APPID=54557732afcfe106bfc955b9da04fb14&customID=somerandomestring
2022-05-18 07:06:12,733 cde0df3d-5fc8-4b7a-8b96-0f828f4e524e WARN  knox.gateway (IdentityAsserterHttpServletRequestWrapper.java:scrubOfExistingPrincipalParams(212)) - Possible identity spoofing attempt - impersonation parameter removed: customID
2022-05-18 07:06:12,738 cde0df3d-5fc8-4b7a-8b96-0f828f4e524e DEBUG knox.gateway (DefaultDispatch.java:executeOutboundRequest(157)) - Dispatch request: GET http://api.openweathermap.org:80/data/2.5/forecast/city?id=524901&APPID=54557732afcfe106bfc955b9da04fb14&user.name=guest
```

### DIFF
--- a/gateway-provider-identity-assertion-common/pom.xml
+++ b/gateway-provider-identity-assertion-common/pom.xml
@@ -88,12 +88,10 @@
         <dependency>
             <groupId>org.jboss.shrinkwrap.descriptors</groupId>
             <artifactId>shrinkwrap-descriptors-api-javaee</artifactId>
-            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>org.jboss.shrinkwrap</groupId>
             <artifactId>shrinkwrap-api</artifactId>
-            <scope>compile</scope>
         </dependency>
 
 
--- a/gateway-provider-identity-assertion-common/pom.xml
+++ b/gateway-provider-identity-assertion-common/pom.xml
@@ -85,8 +85,23 @@
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-util</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap.descriptors</groupId>
+            <artifactId>shrinkwrap-descriptors-api-javaee</artifactId>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-api</artifactId>
+            <scope>compile</scope>
+        </dependency>
 
 
+        <dependency>
+            <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-server</artifactId>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>org.apache.knox</groupId>
             <artifactId>gateway-test-utils</artifactId>
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
@@ -27,6 +27,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.StringTokenizer;
 import java.util.stream.Collectors;
 import javax.security.auth.Subject;
 import javax.servlet.FilterChain;
@@ -49,15 +50,22 @@ import org.apache.knox.gateway.security.GroupPrincipal;
 import org.apache.knox.gateway.security.principal.PrincipalMappingException;
 import org.apache.knox.gateway.security.principal.SimplePrincipalMapper;
 
+import static org.apache.knox.gateway.identityasserter.common.filter.AbstractIdentityAsserterDeploymentContributor.IMPERSONATION_PARAMS;
+
 public class CommonIdentityAssertionFilter extends AbstractIdentityAssertionFilter {
   public static final String VIRTUAL_GROUP_MAPPING_PREFIX = "group.mapping.";
   private IdentityAsserterMessages LOG = MessagesFactory.get(IdentityAsserterMessages.class);
 
   public static final String GROUP_PRINCIPAL_MAPPING = "group.principal.mapping";
   public static final String PRINCIPAL_MAPPING = "principal.mapping";
+
+  private static final String PRINCIPAL_PARAM = "user.name";
+  private static final String DOAS_PRINCIPAL_PARAM = "doAs";
   private SimplePrincipalMapper mapper = new SimplePrincipalMapper();
   private final Parser parser = new Parser();
   private VirtualGroupMapper virtualGroupMapper;
+  /* List of all default and configured impersonation params */
+  protected final List<String> impersonationParamsList = new ArrayList<>();
 
   @Override
   public void init(FilterConfig filterConfig) throws ServletException {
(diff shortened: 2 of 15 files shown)


gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
```diff
@@ -55,9 +56,16 @@ public class CommonIdentityAssertionFilter extends AbstractIdentityAssertionFilt
 
   public static final String GROUP_PRINCIPAL_MAPPING = "group.principal.mapping";
   public static final String PRINCIPAL_MAPPING = "principal.mapping";
+  /* Service specific impersonation params that needs to be scrubbed */
+  public static final String IMPERSONATION_PARAMS = "impersonation.params";
```
zeroflag, line 60: Can't we use the same constant here that is defined in the abstract class?

gateway-provider-identity-assertion-common/pom.xml
```diff
@@ -85,8 +85,23 @@
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>jetty-util</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap.descriptors</groupId>
```
zeroflag, line 89: Do we only use these dependencies from the test? If yes, can we change the scope?

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
moresandeep: @zeroflag which constant are you referring to? or are you suggesting moving this to `AbstractIdentityAssertionFilter`

gateway-provider-identity-assertion-common/pom.xml
moresandeep: Good point, I don't know, let me check.

moresandeep: Actually, this is compile time, that's why I added scope compile.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
zeroflag: @smolnar82 there is already a constant like this in AbstractIdentityAsserterDeploymentContributor.java

moresandeep: I see, I will change it, thanks!

gateway-provider-identity-assertion-common/pom.xml
zeroflag: Isn't compile the default scope in maven? I thought if we don't specify any scope it will be compile. My suggestion was to change it to test scope, if we only use this library from the test code.

moresandeep: ahh, let me change it to test scope and see if it runs.

moresandeep: Has to be compile time 
```
[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] Total time:  20.275 s
[INFO] Finished at: 2022-05-20T14:30:24-04:00
[WARNING] The requested profile "install" could not be activated because it does not exist.
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project gateway-provider-identity-assertion-common: Compilation failure
[ERROR] /knox/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAsserterDeploymentContributor.java:[89,32] cannot access org.jboss.shrinkwrap.descriptor.api.webapp30.WebAppDescriptor
[ERROR]   class file for org.jboss.shrinkwrap.descriptor.api.webapp30.WebAppDescriptor not found
```

zeroflag: ah ok, thanks for trying

zeroflag (approved):

gateway-provider-identity-assertion-common/pom.xml
```diff
@@ -85,8 +85,23 @@
...
+            <artifactId>shrinkwrap-descriptors-api-javaee</artifactId>
+            <scope>compile</scope>
```
risdenk, line 91: Don't set scope to compile. Not needed by default. Thats why you don't see `<scope>compile</scope>` anywhere else.

moresandeep: So this is just to be in-line with other declaration in the POM file?

PR #581, opened 2022-05-24
KNOX-2747 RemoteAliasService generates password without checking if it already exists

zeroflag:

RemoteAliasService always regenerates the password if generates=true, unlike the other implementation. This causes problems with HA deployments where the RemoteAliasService is used but the zookeeper based keystore is turned off. Each knox instance ends up having a different pac4j.password.

## How was this patch tested?

Using the following configs:

```
gateway.remote.alias.service.enabled=true
gateway.remote.config.monitor.client=zookeeper-client
gateway.service.alias.impl=org.apache.knox.gateway.services.security.impl.RemoteAliasService
gateway.remote.alias.service.config.type=zookeeper
gateway.remote.config.registry.zookeeper-client=type=ZooKeeper;address=ZKHOST:2181;authType=Kerberos;principal=knox@ROOT.HWX.SITE;keytab=/cdep/keytabs/knox.keytab;useKeyTab=true;useTicketCache=false
```

* I verified that redeployments and restarts don't change a user generated password. 
* I verified that after a clean start passwords are synchronized to both host.


lmccay (approved):
+1

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/RemoteAliasServiceTestProvider.java
```diff
@@ -88,7 +88,8 @@ public void removeAliasesForCluster(String clusterName, Set<String> aliases) thr
 
     @Override
     public char[] getPasswordFromAliasForCluster(String clusterName, String alias) {
-      return aliases.getOrDefault(clusterName, new HashMap<>()).get(alias).toCharArray();
+      String value = aliases.getOrDefault(clusterName, new HashMap<>()).get(alias);
+      return value == null ? null : value.toCharArray();
```
pzampino, line 92: This is only needed now because the password may be null since it's not ALWAYS being generated?

zeroflag: It is because of the new test (this is also test code). We simulate that there is no password yet and check if the password was generated.

pzampino (approved):

PR #594, opened 2022-06-14
KNOX-2762 

harshiljhaveri:

Whitespaces around delimiters were causing problems while specifying composite auth provider names. The parseProviderNames function has been modified to trim provider names before passing it to the Deployment Contributor function. 

## How was this patch tested?
Ran automated unit/integration tests to ensure provider names did not contain whitespaces. Also ran manual tests to ensure no malfunctions.

### DIFF
--- a/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
+++ b/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
@@ -69,7 +69,6 @@ public class CompositeAuthzDeploymentContributor extends ProviderDeploymentContr
 
   String[] parseProviderNames(String providerNames) {
     String[] b = providerNames.split("\\s*,\\s*");
-
     for (int i = 0; i < b.length; i++) {
       b[i] = b[i].trim();
     }
--- a/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
+++ b/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
@@ -56,17 +56,24 @@ public class CompositeAuthzDeploymentContributor extends ProviderDeploymentContr
 
     Map<String, String> providerParams = provider.getParams();
     String providerNames = providerParams.get("composite.provider.names");
+    if (!providerNames.isEmpty()) {
     String[] names = parseProviderNames(providerNames);
     for (String name : names) {
       getProviderSpecificParams(resource, params, providerParams, name);
       DeploymentFactory.getProviderContributor("authorization", name)
-        .contributeFilter(context, provider, service, resource, params);
+              .contributeFilter(context, provider, service, resource, params);
       params.clear();
+    }
     }
   }
 
   String[] parseProviderNames(String providerNames) {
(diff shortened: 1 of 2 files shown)


moresandeep (approved):

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
```diff
@@ -56,17 +56,23 @@ public void contributeFilter( DeploymentContext context, Provider provider, Serv
 
     Map<String, String> providerParams = provider.getParams();
     String providerNames = providerParams.get("composite.provider.names");
+    if (!providerNames.isEmpty()) {
```
smolnar82, line 59: I'd use StringUtils over String here because the first one is `null` safe. Even better, I think `parseProviderNames` should return a collection and not a String array.

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
```diff
@@ -56,17 +56,23 @@ public void contributeFilter( DeploymentContext context, Provider provider, Serv
...
     String[] names = parseProviderNames(providerNames);
     for (String name : names) {
       getProviderSpecificParams(resource, params, providerParams, name);
       DeploymentFactory.getProviderContributor("authorization", name)
-        .contributeFilter(context, provider, service, resource, params);
+              .contributeFilter(context, provider, service, resource, params);
       params.clear();
     }
+    }
```
smolnar82, line 67: This additional `if` statement modifies the indentation here so that content within that condition should be intended more.

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
```diff
@@ -56,17 +56,23 @@ public void contributeFilter( DeploymentContext context, Provider provider, Serv
...
   }
 
   String[] parseProviderNames(String providerNames) {
-    return providerNames.split(",\\s*");
+    String[] b = providerNames.split("\\s*,\\s*");
```
smolnar82, line 71: Please avoid using parameter names like `b`. It's just against clean code principles.

gateway-provider-security-authz-composite/src/test/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzProviderTest.java
```diff
@@ -64,4 +64,16 @@ public void testParsingProviderNames() throws Exception {
     assertEquals(providerNames[1], "SomeOther");
     assertEquals(providerNames[2], "TheOtherOne");
   }
+
+  @Test
+  public void testingParsingProviderNames() throws Exception {
+    String testnames = "   AclsAuthz  ,   SomeOther   ,   TheOtherOne   ,";
```
smolnar82, line 70: I'd modify this sample string to cover all cases:
- whitespace just before the name
- whitespace just after the name
- whitespace before and after the name
- more whitespaces
- no whitespace at all

moresandeep:
Sure, thanks for the comments @smolnar82!

moresandeep:
@harshiljhaveri can you post a followup PR with suggested changes?

PR #596, opened 2022-06-17
KNOX-2762 Bug fixes for spaces around delimiters with reviewed commen…

harshiljhaveri:

Issue arising from NPE due to whitespaces around delimiters in Composite Authz names fixed with review comments addressed.

## How was this patch tested?

Specific unit/integration tests covering most possible test cases conducted. Manual tests also conducted to ensure no build failures.

### DIFF
--- a/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
+++ b/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Arrays;
 
 public class CompositeAuthzDeploymentContributor extends ProviderDeploymentContributorBase {
   @Override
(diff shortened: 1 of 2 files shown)


gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
```diff
@@ -57,22 +58,23 @@ public void contributeFilter( DeploymentContext context, Provider provider, Serv
...
+      }
     }
   }
 
-  String[] parseProviderNames(String providerNames) {
-    String[] b = providerNames.split("\\s*,\\s*");
-    for (int i = 0; i < b.length; i++) {
-      b[i] = b[i].trim();
+  List parseProviderNames(String providerNames) {
+    String[] providerNamesList = providerNames.split("\\s*,\\s*");
```
moresandeep, line 72: You can just convert the array to List here and made the code a bit more simpler 
`List<String> providerNamesList = Arrays.asList(providerNames.split("\\s*,\\s*"))`

then you can just do something like `.forEach` and trim and process it then return.

moresandeep (approved):
Few minor changes.

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
smolnar82: +1 to @moresandeep 's comment!

smolnar82 (changes requested):

PR #597, opened 2022-06-20
KNOX-2762 Bug fixes for spaces around delimiters with all reviewed comments addressed

harshiljhaveri:

Fix for NullPointer Exception arising due to whitespaces around delimiters in Composite Auth Provider Names. All review comments addressed.
## How was this patch tested?

Comprehensive Unit tests used to cover most possible test cases. Also manually tested.

### DIFF
--- a/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
+++ b/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Collections;
 import java.util.Arrays;
 
 public class CompositeAuthzDeploymentContributor extends ProviderDeploymentContributorBase {
@@ -69,12 +70,12 @@ public class CompositeAuthzDeploymentContributor extends ProviderDeploymentContr
   }
 
   List parseProviderNames(String providerNames) {
-    String[] providerNamesList = providerNames.split("\\s*,\\s*");
-    for (int i = 0; i < providerNamesList.length; i++) {
-      providerNamesList[i] = providerNamesList[i].trim();
+    if (providerNames==null){
+      return Collections.singletonList("");
     }
-    List<String> providerNamesCollection = Arrays.asList(providerNamesList);
-    return providerNamesCollection;
+    List<String> providerNamesList = Arrays.asList(providerNames.split("\\s*,\\s*"));
+    providerNamesList.replaceAll(String::trim);
+    return providerNamesList;
   }
 
   void getProviderSpecificParams(ResourceDescriptor resource, List<FilterParamDescriptor> params,
--- a/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
+++ b/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
@@ -71,7 +71,7 @@ public class CompositeAuthzDeploymentContributor extends ProviderDeploymentContr
   }
 
   List parseProviderNames(String providerNames) {
-    if (providerNames==null || StringUtils.isBlank(providerNames)){
+    if (StringUtils.isBlank(providerNames)){
       return Collections.emptyList();
     }
     List<String> providerNamesList = Arrays.asList(providerNames.split("\\s*,\\s*"));
--- a/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
+++ b/gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
@@ -59,16 +59,14 @@ public class CompositeAuthzDeploymentContributor extends ProviderDeploymentContr
 
     Map<String, String> providerParams = provider.getParams();
     String providerNames = providerParams.get("composite.provider.names");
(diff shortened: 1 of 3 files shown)


smolnar82 (changes requested):

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
```diff
@@ -56,23 +59,24 @@ public void contributeFilter( DeploymentContext context, Provider provider, Serv
...
+      }
     }
   }
 
-  String[] parseProviderNames(String providerNames) {
-    String[] b = providerNames.split("\\s*,\\s*");
-    for (int i = 0; i < b.length; i++) {
-      b[i] = b[i].trim();
+  List parseProviderNames(String providerNames) {
+    if (providerNames==null || StringUtils.isEmpty(providerNames) || StringUtils.isBlank(providerNames)){
```
smolnar82, line 74: `StringUtils.isBlank(providerNames)` is enough here. It is `null safe` and will also cover `StringUtils.isEmpty(providerNames)`.

smolnar82 (changes requested):
Sorry for requesting another change, but I think this would be the last one. Instead of using 3 different conditions 1 one be enough.

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
```diff
@@ -71,7 +71,7 @@ public void contributeFilter( DeploymentContext context, Provider provider, Serv
   }
 
   List parseProviderNames(String providerNames) {
-    if (providerNames==null || StringUtils.isEmpty(providerNames) || StringUtils.isBlank(providerNames)){
+    if (providerNames==null || StringUtils.isBlank(providerNames)){
```
smolnar82, line 74: As I explained earlier, `StringUtils.isBlank(providerNames)` is enough here. It's null-safe.

smolnar82 (changes requested):

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
smolnar82: See the JavaDoc here: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank-java.lang.CharSequence-

smolnar82 (approved):

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
```diff
@@ -71,7 +71,7 @@ public void contributeFilter( DeploymentContext context, Provider provider, Serv
   }
 
   List parseProviderNames(String providerNames) {
```
zeroflag, line 73: What's the reason for not having a type parameter for the List? (E.g. List<String>)

gateway-provider-security-authz-composite/src/test/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzProviderTest.java
```diff
@@ -83,13 +85,19 @@ public void testingNullandEmptyProviderNames() throws Exception {
     String testnames = "";
     CompositeAuthzDeploymentContributor c = new CompositeAuthzDeploymentContributor();
     List providerNames = c.parseProviderNames(testnames);
-    assertEquals(providerNames.size(), 1);
-    assertEquals(providerNames.get(0), "");
+    assertSame(providerNames.size(),0);
```
zeroflag, line 88: Here the order of parameters are reversed. The first is the expected and the second is the actual. 
This is not a correctness issue, but the error message will misleading if the test fails.

For example if providerNames has a size of 3 it'll say: expected 3 got 0, but it is the other way around.

More importantly what's the reason for using assertSame instead of assertEquals? My understanding is that assertSame checks identity (`==`) not equality (`.equals`).

zeroflag (changes requested):

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
```diff
@@ -58,7 +59,7 @@ public void contributeFilter( DeploymentContext context, Provider provider, Serv
 
     Map<String, String> providerParams = provider.getParams();
     String providerNames = providerParams.get("composite.provider.names");
-    if (!providerNames.isEmpty()) {
+    if (!StringUtils.isEmpty(providerNames)) {
```
zeroflag, line 62: Good job and thank you for the changes @harshiljhaveri. One last thing is that this conditional seems to be unnecessary. The `parseProviderNames` returns an empty list if the parameter is blank so the for loop won't do anything in that case. I would just remove this `if` statement and rely on that.

zeroflag (changes requested):

gateway-provider-security-authz-composite/src/main/java/org/apache/knox/gateway/deploy/impl/CompositeAuthzDeploymentContributor.java
smolnar82: Nice catch Attila! I missed it previously. Thanks for spotting this out!

zeroflag (approved):
LGTM

smolnar82 (approved):
LGTM

moresandeep (approved):
Looks good.

PR #602, opened 2022-07-06
KNOX-2771 - Log HTTP client config parameters such as socket timeouts with info level

MrtnBalazs:

Changed log level of http connection timeout and socket timeout to from DEBUG to INFO and added INFO level log for retry count and retry non safe request parameters, because sometimes clients were confused why the timeout parameters haven't changed even though they changed it in gateway-site.xml, but service.xml override-s it. Now they can see in the log what is the value of these parameters, only after the first request after topology deployment.

## How was this patch tested?

I tested it manually, using ```curl -k -u tom:tom-password https://localhost:8443/gateway/sandbox/hive``` command from terminal.

1. Nothing is set:

semmiben

2. Timeout params set in gateway-site.xml and retry params set in sandbox.xml:

gatewayben
globalbeallitas
Screenshot 2022-07-06 at 8 56 18

3. Timeout params set in gateway-site.xml and in service.xml, retry params set like 2.:

servicebenesgatewayben
globalbeallitas
servicekod

We can see that service.xml overrides gateway-site.xml settings.

The logs only appear after the first request after the topology is deployed.
After that it is not logged until the topology is deployed again.
masodszorra nem logol

### DIFF
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/SpiGatewayMessages.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/SpiGatewayMessages.java
@@ -91,10 +91,10 @@ public interface SpiGatewayMessages {
             text = "The dispatch to {0} was disallowed because it fails the dispatch whitelist validation. See documentation for dispatch whitelisting." )
   void dispatchDisallowed(String uri);
 
-  @Message( level = MessageLevel.DEBUG, text = "HTTP client connection timeout is set to {0} for {1}" )
+  @Message( level = MessageLevel.INFO, text = "HTTP client connection timeout is set to {0} for {1}" )
   void setHttpClientConnectionTimeout(int connectionTimeout, String serviceRole);
 
-  @Message( level = MessageLevel.DEBUG, text = "HTTP client socket timeout is set to {0} for {1}" )
+  @Message( level = MessageLevel.INFO, text = "HTTP client socket timeout is set to {0} for {1}" )
   void setHttpClientSocketTimeout(int csocketTimeout, String serviceRole);
 
   @Message( level = MessageLevel.DEBUG, text = "replayBufferSize is set to {0} for {1}" )
@@ -108,4 +108,10 @@ public interface SpiGatewayMessages {
 
   @Message( level = MessageLevel.DEBUG, text = "Skipped adding outbound header {0} and value {1}" )
   void skippedOutboundHeader(String header, String value);
+
+  @Message( level = MessageLevel.INFO, text = "HTTP client retry count is set to {0} for {1}" )
+  void setRetryCount(int retryCount, String serviceRole);
+
+  @Message( level = MessageLevel.INFO, text = "HTTP client retry non safe request is set to {0} for {1}" )
+  void setRetryNonIndependent(boolean retryNonIndependent, String serviceRole);
 }
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/SpiGatewayMessages.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/SpiGatewayMessages.java
@@ -91,11 +91,11 @@ public interface SpiGatewayMessages {
             text = "The dispatch to {0} was disallowed because it fails the dispatch whitelist validation. See documentation for dispatch whitelisting." )
   void dispatchDisallowed(String uri);
 
-  @Message( level = MessageLevel.INFO, text = "HTTP client connection timeout is set to {0} msec for {1}" )
+  @Message( level = MessageLevel.INFO, text = "HTTP client connection timeout is set to {0} ms for {1}" )
   void setHttpClientConnectionTimeout(int connectionTimeout, String serviceRole);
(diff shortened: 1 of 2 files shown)


MrtnBalazs:
@smolnar82 @zeroflag

gateway-spi/src/main/java/org/apache/knox/gateway/SpiGatewayMessages.java
```diff
@@ -91,10 +91,10 @@ public interface SpiGatewayMessages {
             text = "The dispatch to {0} was disallowed because it fails the dispatch whitelist validation. See documentation for dispatch whitelisting." )
   void dispatchDisallowed(String uri);
 
-  @Message( level = MessageLevel.DEBUG, text = "HTTP client connection timeout is set to {0} for {1}" )
+  @Message( level = MessageLevel.INFO, text = "HTTP client connection timeout is set to {0} for {1}" )
   void setHttpClientConnectionTimeout(int connectionTimeout, String serviceRole);
 
-  @Message( level = MessageLevel.DEBUG, text = "HTTP client socket timeout is set to {0} for {1}" )
+  @Message( level = MessageLevel.INFO, text = "HTTP client socket timeout is set to {0} for {1}" )
```
zeroflag, line 97: Looks good. Maybe adding the unit of measurements (msec I guess) for both the connection and socket timeouts would useful.

gateway-spi/src/main/java/org/apache/knox/gateway/SpiGatewayMessages.java
```diff
@@ -91,10 +91,10 @@ public interface SpiGatewayMessages {
             text = "The dispatch to {0} was disallowed because it fails the dispatch whitelist validation. See documentation for dispatch whitelisting." )
   void dispatchDisallowed(String uri);
 
-  @Message( level = MessageLevel.DEBUG, text = "HTTP client connection timeout is set to {0} for {1}" )
+  @Message( level = MessageLevel.INFO, text = "HTTP client connection timeout is set to {0} msec for {1}" )
```
smolnar82, line 94: It should either be `ms` or `milliseconds`

gateway-spi/src/main/java/org/apache/knox/gateway/SpiGatewayMessages.java
```diff
@@ -91,10 +91,10 @@ public interface SpiGatewayMessages {
...
   void setHttpClientConnectionTimeout(int connectionTimeout, String serviceRole);
 
-  @Message( level = MessageLevel.DEBUG, text = "HTTP client socket timeout is set to {0} for {1}" )
+  @Message( level = MessageLevel.INFO, text = "HTTP client socket timeout is set to {0} msec for {1}" )
   void setHttpClientSocketTimeout(int csocketTimeout, String serviceRole);
```
smolnar82, line 98: nit: typo in `csocketTimeout`: either `clientSocketTimeout` or `socketTimeout` (I prefer the 2nd one).

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
```diff
@@ -128,12 +128,14 @@ public HttpClient createHttpClient(FilterConfig filterConfig) {
...
     if (doesRetryParamExist(filterConfig)) {
       int retryCount = Integer.parseInt(filterConfig.getInitParameter(PARAMETER_RETRY_COUNT));
+      LOG.setRetryCount(retryCount, serviceRole);
       /* do we want to retry non-idempotent requests? default no */
       boolean retryNonIdempotent = DEFAULT_PARAMETER_RETRY_NON_SAFE_REQUEST;
       if (filterConfig.getInitParameter(PARAMETER_RETRY_NON_SAFE_REQUEST)
           != null) {
         retryNonIdempotent = Boolean.parseBoolean(
             filterConfig.getInitParameter(PARAMETER_RETRY_NON_SAFE_REQUEST));
+        LOG.setRetryNonIndependent(retryNonIdempotent, serviceRole);
```
smolnar82, line 138: IMO, this should be outside of this `if` statement. Otherwise, we would assume it's set to default because the log entry was not shown.

smolnar82 (changes requested):
LGTM, only small changes are required.

smolnar82:
One more general ask: instead of using screenshots of text blocks (configuration, logs, etc...) in the description or comments, please use the 'text block' representation of GitHub PRs. The way it's impossible to copy-paste that content if I wanted to use it. For instance, instead of using this screenshot
Screenshot 2022-07-06 at 15 18 57
Please do this:
```
    <property>
        <name>gateway.httpclient.connectionTimeout</name>
        <value>8m</value>
    </property>
    <property>
        <name>gateway.httpclient.socketTimeout</name>
        <value>10m</value>
    </property>
```

zeroflag (approved):

smolnar82 (approved):
LGTM

PR #605, opened 2022-07-07
KNOX-2772 - add configuration for jetty renegotiation

nanhuirong:
disable renegotiation for embedded jetty

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
@@ -177,6 +177,7 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   private static final String SSL_EXCLUDE_PROTOCOLS = "ssl.exclude.protocols";
   private static final String SSL_INCLUDE_CIPHERS = "ssl.include.ciphers";
   private static final String SSL_EXCLUDE_CIPHERS = "ssl.exclude.ciphers";
+  private static final String SSL_RENEGOTIATION = "ssl.renegotiation";
   // END BACKWARD COMPATIBLE BLOCK
 
   public static final String DEFAULT_HTTP_PORT = "8888";
@@ -590,6 +591,11 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
       list = Arrays.asList(value.trim().split("\\s*,\\s*"));
     }
     return list;
+  }
+
+  @Override
+  public boolean isSSLRenegotiationAllowed() {
+    return getBoolean(SSL_RENEGOTIATION, true);
   }
 
   @Override
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
@@ -224,6 +224,8 @@ public class JettySSLService implements SSLService {
     if (sslExcludeProtocols != null && !sslExcludeProtocols.isEmpty()) {
       sslContextFactory.setExcludeProtocols( sslExcludeProtocols.toArray(new String[0]) );
     }
+
+    sslContextFactory.setRenegotiationAllowed(config.isSSLRenegotiationAllowed());
     return sslContextFactory;
   }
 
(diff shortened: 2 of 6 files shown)


lmccay (changes requested):
See my comment requesting a unit test.
Otherwise, LGTM.

smolnar82 (changes requested):
Please address my review comments as well as add a unit test s @lmccay requested. Thanks!

smolnar82:
Thanks for the new patchset, @nanhuirong !
I've just checked [Jetty's API](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/util/ssl/SslContextFactory.html#isRenegotiationAllowed()) and found that the default value is `true`:
```
[@ManagedAttribute](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/util/annotation/ManagedAttribute.html)("Whether renegotiation is allowed") public boolean isRenegotiationAllowed()
Returns:
whether TLS renegotiation is allowed (true by default)
```
Hence the question: why do we want to change the default to `false`?

Cc. @lmccay

nanhuirong:
> https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/util/annotation/ManagedAttribute.html)("Whether renegotiation is allowed"

In our scenario，we should provide better security for the gateway. After all, the IP maybe a public IP.

smolnar82:
> > [https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/util/annotation/ManagedAttribute.html)("Whether](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/util/annotation/ManagedAttribute.html)(%22Whether) renegotiation is allowed"
> 
> In our scenario，we should provide better security for the gateway. After all, the IP maybe a public IP.

What prevents you to set this new config to `false` in your environment and leave the default as `true` as intended by Jetty developers? They also introduced a limit (defaults to 5) that would fix the original CVE of causing a denial of service (CPU consumption) by performing many renegotiations within a single connection.

nanhuirong:
> > > [https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/util/annotation/ManagedAttribute.html)("Whether](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/util/annotation/ManagedAttribute.html)(%22Whether) renegotiation is allowed"
> > 
> > 
> > In our scenario，we should provide better security for the gateway. After all, the IP maybe a public IP.
> 
> What prevents you to set this new config to `false` in your environment and leave the default as `true` as intended by Jetty developers? They also introduced a limit (defaults to 5) that would fix the original CVE of causing a denial of service (CPU consumption) by performing many renegotiations within a single connection.

I got it. I don't notice the limit and update the patchset.

lmccay:
Disabling renegotiation is the more secure mode and I would expect that to actually be the default. Renegotiation had an attack vector a while ago whereby a middle man could renegotiate to a lower - like NONE - algorithm. While this may have been fixed, I don't know of any specific functionality that is blocked by this that requires it to be enabled by default. If this was recently changed to enabled by default in jetty then we should preserve backward compatibility and leave it false. If it was enabled by default previously then making it true by default preserves previous behanvior.

nanhuirong:
> Disabling renegotiation is the more secure mode and I would expect that to actually be the default. Renegotiation had an attack vector a while ago whereby a middle man could renegotiate to a lower - like NONE - algorithm. While this may have been fixed, I don't know of any specific functionality that is blocked by this that requires it to be enabled by default. If this was recently changed to enabled by default in jetty then we should preserve backward compatibility and leave it false. If it was enabled by default previously then making it true by default preserves previous behav

nanhuirong:
> Disabling renegotiation is the more secure mode and I would expect that to actually be the default. Renegotiation had an attack vector a while ago whereby a middle man could renegotiate to a lower - like NONE - algorithm. While this may have been fixed, I don't know of any specific functionality that is blocked by this that requires it to be enabled by default. If this was recently changed to enabled by default in jetty then we should preserve backward compatibility and leave it false. If it was enabled by default previously then making it true by default preserves previous behavior.

The build for arm reports "no sapce". Do you have any idea for it?

nanhuirong:
> Disabling renegotiation is the more secure mode and I would expect that to actually be the default. Renegotiation had an attack vector a while ago whereby a middle man could renegotiate to a lower - like NONE - algorithm. While this may have been fixed, I don't know of any specific functionality that is blocked by this that requires it to be enabled by default. If this was recently changed to enabled by default in jetty then we should preserve backward compatibility and leave it false. If it was enabled by default previously then making it true by default preserves previous behavior.

The default value is ture for jetty now.

smolnar82 (approved):
LGTM

lmccay (approved):
+1
LGTM

nanhuirong:
@smolnar82 @lmccay , Can you merge the PR? This is the first PR for me.

lmccay:
@nanhuirong - thank you for your PR and improvement!
I've merged this to master and do not think the build issue is related to this patch. If it does indeed continue and is determined to be related, we may need to revert it.

In the meantime, congrats on your first PR to Apache Knox and thanks again!

PR #608, opened 2022-07-14
KNOX-2777 - Add configurations for concurrent session verifier feature

MrtnBalazs:

I’ve added implementation in `GatewayConfig` , `GatewayConfigImpl` and `GatewayTestConfig` to receive the following parameters from the gateway-site.xml file. 
`gateway.non.privileged.users`
`gateway.privileged.users`
`gateway.privileged.user.concurrent.session.limit`
`gateway.non.privileged.user.concurrent.session.limit`

An example:
```
   <property>
        <value>tom,guest</value>
    </property>
    <property>
        <value>admin</value>
    </property>
    <property>
        <name>gateway.privileged.user.concurrent.session.limit</name>
        <value>3</value>
    </property>
    <property>
        <name>gateway.non.privileged.user.concurrent.session.limit</name>
        <value>2</value>
    </property>
```
It is not yet used, but it is necessary for the jira 2778 which depends on this one.

I’ve also created `ConcurrentSessionVerifier` class which will be used to keep track of the concurrent sessions of the users and verify whether they fit in the limit (based on their group) or not.
It is not yet wired in, but also necessary for the jira 2778. 

## How was this patch tested?

I’ve written unit tests for the new parameters in `GatewayConfigImplTest` class where I test the default parameters and what if the user adds normal and abnormal parameters. Result of the tests:
```
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.knox.gateway.config.impl.GatewayConfigImplTest
[INFO] Tests run: 20, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 1.055 s - in org.apache.knox.gateway.config.impl.GatewayConfigImplTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 20, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  8.831 s
[INFO] Finished at: 2022-07-14T16:49:26+02:00
```

I’ve also written unit tests for the `ConcurrentSessionVerifier` class in the `ConcurrentSessionVerifierTest` class. These tests test the following:
-The user is in neither of the two groups
-The user is in both groups
-User is privileged and goes over the session limit, then ends a session and goes over the limit again
-Same with non privileged users
-One of the limits is 0
-We end sessions before even creating them

Result of the tests:
```
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.knox.gateway.util.ConcurrentSessionVerifierTest
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.003 s - in org.apache.knox.gateway.util.ConcurrentSessionVerifierTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] BUILD SUCCESS
[INFO] Total time:  5.304 s
[INFO] Finished at: 2022-07-14T17:07:34+02:00
```


moresandeep:
Would be better to update the title indicating this is just adding configs to support this feature i.e. "Add configurations for concurrent session verifier feature" - feel free to change the wordings.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -293,6 +293,14 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   private static final String GATEWAY_DATABASE_VERIFY_SERVER_CERT =  GATEWAY_CONFIG_FILE_PREFIX + ".database.ssl.verify.server.cert";
   private static final String GATEWAY_DATABASE_TRUSTSTORE_FILE =  GATEWAY_CONFIG_FILE_PREFIX + ".database.ssl.truststore.file";
 
+  // Concurrent session properties
+  private static final String GATEWAY_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT =  GATEWAY_CONFIG_FILE_PREFIX + ".privileged.user.concurrent.session.limit";
```
moresandeep, line 297: `priviledged.user` can be a static variable.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -293,6 +293,14 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
...
+  private static final String GATEWAY_NON_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT =  GATEWAY_CONFIG_FILE_PREFIX + ".non.privileged.user.concurrent.session.limit";
+  private static final int GATEWAY_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT_DEFAULT = 3;
+  private static final int GATEWAY_NON_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT_DEFAULT = 2;
+  private static final String GATEWAY_PRIVILEGED_USERS = GATEWAY_CONFIG_FILE_PREFIX + ".privileged.users";
+  private static final String GATEWAY_NON_PRIVILEGED_USERS = GATEWAY_CONFIG_FILE_PREFIX + ".non.privileged.users";
```
moresandeep, line 302: nit: `non.priviledged.user` can be a static variable (using the previous static variable) :)

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -1335,4 +1343,49 @@ public int getJettyMaxFormKeys() {
     return getInt(JETTY_MAX_FORM_KEYS, ContextHandler.DEFAULT_MAX_FORM_KEYS);
   }
 
+  @Override
+  public int getPrivilegedUserConcurrentSessionLimit(){
+    int limit = getInt(GATEWAY_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT, GATEWAY_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT_DEFAULT);
+    if(limit < 0)
```
moresandeep: nit: brackets :)

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -1335,4 +1343,49 @@ public int getJettyMaxFormKeys() {
...
+      return GATEWAY_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT_DEFAULT;
+    else
+      return limit;
+  }
+
+  @Override
+  public int getNonPrivilegedUserConcurrentSessionLimit(){
+    int limit = getInt(GATEWAY_NON_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT, GATEWAY_NON_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT_DEFAULT);
+    if(limit < 0)
```
moresandeep: nit: brackets

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -1335,4 +1343,49 @@ public int getJettyMaxFormKeys() {
...
+      return GATEWAY_NON_PRIVILEGED_USER_CONCURRENT_SESSION_LIMIT_DEFAULT;
+    else
+      return limit;
+  }
+
+  @Override
+  public Set<String> getPrivilegedUsers(){
+    Set<String> set = new HashSet<>();
```
moresandeep, line 1368: set is not a good name since it is too close to the built in type Set

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -1335,4 +1343,49 @@ public int getJettyMaxFormKeys() {
...
+      List<String> temp = new ArrayList<>(Arrays.asList(value.split(",")));
+      temp.forEach(e -> { temp.set(temp.indexOf(e), e.trim()); });
+      set.addAll(temp);
+    }
+
+    return set;
+  }
+
+  @Override
+  public Set<String> getNonPrivilegedUsers(){
```
moresandeep, line 1381: All the same comments as above.

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,76 @@
...
+public class ConcurrentSessionVerifier {
+    private static Set<String> privilegedUsers = new HashSet<>();
+    private static Set<String> nonPrivilegedUsers = new HashSet<>();;
+    private static int privilegedUserConcurrentSessionLimit = 3;
+    private static int nonPrivilegedUserConcurrentSessionLimit = 2;
+    private static Map<String, Integer> concurrentSessionCounter = new HashMap<>();
+
+    private ConcurrentSessionVerifier(){}
+
+    public static void init(Set<String> _privilegedUsers, Set<String> _nonPrivilegedUsers, int _privilegedUserConcurrentSessionLimit, int _nonPrivilegedUserConcurrentSessionLimit){
```
moresandeep, line 35: We don't use _ in front of variable names, I see that you did not use it elsewhere so avoid using it just to be more consistant.

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+public class ConcurrentSessionVerifier {
+    private static Set<String> privilegedUsers = new HashSet<>();
+    private static Set<String> nonPrivilegedUsers = new HashSet<>();;
+    private static int privilegedUserConcurrentSessionLimit = 3;
+    private static int nonPrivilegedUserConcurrentSessionLimit = 2;
+    private static Map<String, Integer> concurrentSessionCounter = new HashMap<>();
```
moresandeep, line 31: The name suggests this is not thread safe, should this be `ConcurrentHashMap` ?

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private ConcurrentSessionVerifier(){}
+
+    public static void init(Set<String> _privilegedUsers, Set<String> _nonPrivilegedUsers, int _privilegedUserConcurrentSessionLimit, int _nonPrivilegedUserConcurrentSessionLimit){
+        privilegedUsers = _privilegedUsers;
+        nonPrivilegedUsers = _nonPrivilegedUsers;
+        privilegedUserConcurrentSessionLimit = _privilegedUserConcurrentSessionLimit;
+        nonPrivilegedUserConcurrentSessionLimit = _nonPrivilegedUserConcurrentSessionLimit;
+        concurrentSessionCounter = new HashMap<>();
```
moresandeep, line 40: The name suggests this is not thread safe, should this be `ConcurrentHashMap` ?

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public  static boolean verifySessionForUser(String username){
```
moresandeep, line 43: i am scared of this method. This is static, so we have to be very careful about the state here especially since we are dealing with usernames. Should this be synchronized? the data structures used in the method are not thread safe which will cause issues.

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,76 @@
...
+        return true;
+    }
+
+    private static void incrementConcurrentSessionCount(String username){
+        int count = concurrentSessionCounter.get(username);
+        count++;
+        concurrentSessionCounter.put(username, count);
+    }
+
+    public static void sessionEndedForUser(String username){
```
moresandeep, line 65: same comment as above.

moresandeep (changes requested):

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
MrtnBalazs: Is there any other conventions, you use, when you can't put `this` in front of the variable when giving it the same value as the parameter or should i come up with different names for the parameters?

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
MrtnBalazs: Yes, this class should be thread safe

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
MrtnBalazs: Changed it.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
MrtnBalazs: Changed it.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
MrtnBalazs: Added them.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
MrtnBalazs: Added them.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -1335,4 +1343,49 @@ public int getJettyMaxFormKeys() {
...
+    else
+      return limit;
+  }
+
+  @Override
+  public Set<String> getPrivilegedUsers(){
+    Set<String> set = new HashSet<>();
+
+    String value = get( GATEWAY_PRIVILEGED_USERS );
+    if ((value != null) && (!value.trim().equals("")) ) {
```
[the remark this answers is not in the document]
MrtnBalazs: Fixed it.

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
MrtnBalazs: Yes changed it to ConcurrentHashMap.

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
MrtnBalazs: Yes changed it to ConcurrentHashMap.

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
MrtnBalazs: I modified it to be thread safe.

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+package org.apache.knox.gateway.util;
+
+
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class ConcurrentSessionVerifier {
```
zeroflag, line 26: Instead of using static methods and static fields, let's have a normal non static class with instance variables and non static methods. Constants like `privilegedUserConcurrentSessionLimit`/`privilegedUserConcurrentSessionLimit` should be kept as statics but `concurrentSessionCounter` should be non static.

Then we can create one instance of this class and use that at each places where needed. Either by creating an instance somewhere in GatewayServer's `main()` and passing that around or by making a singleton. I prefer the former but that's up to how the code is structured. We can decide that later when we start using it.

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,79 @@
...
+        ConcurrentSessionVerifier.nonPrivilegedUserConcurrentSessionLimit = nonPrivilegedUserConcurrentSessionLimit;
+        concurrentSessionCounter = new ConcurrentHashMap<>();
+    }
+
+    public static boolean verifySessionForUser(String username){
+        if(!privilegedUsers.contains(username) && !nonPrivilegedUsers.contains(username)) {
+            return true;
+        }
+        synchronized(lock) {
+            if (!concurrentSessionCounter.containsKey(username)) {
```
zeroflag, line 49: This is probably can be simplified by using `computeIfAbsent` / `putIfAbsent` / `compute` methods.

For example:

```java
int count = concurrentSessionCounter.computeIfAbsent(username, name -> 0);
if (...) {
}
```

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,79 @@
...
+            if ((privilegedUsers.contains(username) && !(concurrentSessionCounter.get(username) < privilegedUserConcurrentSessionLimit)) || (nonPrivilegedUsers.contains(username) && !(concurrentSessionCounter.get(username) < nonPrivilegedUserConcurrentSessionLimit))) {
+                return false;
+            }
+            incrementConcurrentSessionCount(username);
+        }
+        return true;
+    }
+
+    private static synchronized void incrementConcurrentSessionCount(String username){
+        int count = concurrentSessionCounter.get(username);
```
zeroflag, line 61: Similarly, see if you can use `compute` to simplify this. For example:

```java
concurrentSessionCounter.compute(username, (key ,value) -> value + 1);
```

zeroflag (changes requested):

gateway-util-common/src/main/java/org/apache/knox/gateway/util/ConcurrentSessionVerifier.java
MrtnBalazs: Same

gateway-spi-common/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+public class ConcurrentSessionVerifier {
+  public static final ConcurrentSessionVerifier INSTANCE = new ConcurrentSessionVerifier();
+  private Set<String> privilegedUsers;
+  private Set<String> nonPrivilegedUsers;
+  private int privilegedUserConcurrentSessionLimit;
+  private int nonPrivilegedUserConcurrentSessionLimit;
+  private Map<String, Integer> concurrentSessionCounter;
+  private Lock sessionCountModifyLock = new ReentrantLock();
```
smolnar82, line 36: nit: this can be final.

smolnar82 (approved):
LGTM

gateway-spi-common/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
MrtnBalazs: Then we would have to give them value in the constructor which is private, because of the singleton pattern and in the test cases we would not be able to initialize the same verifier multiple times.

MrtnBalazs: Ok. Made it final.

gateway-spi-common/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,102 @@
...
+      return false;
+    }
+    return nonPrivilegedUsers.contains(username) && (concurrentSessionCounter.get(username) >= nonPrivilegedUserConcurrentSessionLimit);
+  }
+
+  public void sessionEndedForUser(String username) {
+    if (concurrentSessionCounter.containsKey(username)) {
+      sessionCountModifyLock.lock();
+      try {
+        int count = concurrentSessionCounter.get(username);
```
zeroflag, line 89: We can probably use something like this here:

```java
m.computeIfPresent(username, (user,counter) -> counter - 1);
```

But if the counter reaches zero, I think we should delete the user from the map altogether.

zeroflag (changes requested):

gateway-spi-common/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
smolnar82: I agree; this is a better approach. Nice catch, Attila! :)

zeroflag (approved):

smolnar82 (approved):

PR #614, opened 2022-07-26
KNOX-2786 - Upgrade spring and json-smart due to CVEs

zeroflag:

json-smart + spring version upgrade.

## How was this patch tested?

```bash
$ mvn dependency:tree | grep json-smart
$ mvn dependency:tree | grep spring
```


smolnar82 (approved):
LGTM

PR #615, opened 2022-08-01
KNOX-2778 - Enforce concurrent session limit in KnoxSSO

MrtnBalazs:

`ConcurrentSessionVerifier` is changed to be used and initilaized like a service, because initializing it in `GatewayServer` `main()` felt out of place. It is also modified to store the tokens issued at login so the expiry can be tracked since we don t have an event to subscribe to when the tokens expire. The verifier was put in `WebSSOResource` and `WebSSOutResource` to count the issued tokens when logged in or out.

## How was this patch tested?

I tested it manually by configuring `gateway-site.xml` and logging in and out from different browsers.
Configuration:
```
<property>
        <value>tom,guest</value>
 </property>
 <property>
        <value>admin</value>
  </property>
  <property>
        <name>gateway.privileged.users.concurrent.session.limit</name>
        <value>2</value>
  </property>
  <property>
        <name>gateway.non.privileged.users.concurrent.session.limit</name>
        <value>1</value>
  </property>
```
I also wrote unit tests for `WebSSOResource` which test reaching privileged and non-privileged limit when logging in.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
@@ -21,7 +21,12 @@ package org.apache.knox.gateway.session.control;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.services.Service;
 import org.apache.knox.gateway.services.ServiceLifecycleException;
+import org.apache.knox.gateway.services.security.token.impl.JWT;
+import org.apache.knox.gateway.services.security.token.impl.JWTToken;
 
+import java.text.ParseException;
+import java.util.Date;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
@@ -33,57 +38,82 @@ public class ConcurrentSessionVerifier implements Service {
   private Set<String> nonPrivilegedUsers;
   private int privilegedUserConcurrentSessionLimit;
   private int nonPrivilegedUserConcurrentSessionLimit;
-  private Map<String, Integer> concurrentSessionCounter;
+  private Map<String, Set<String>> concurrentSessionCounter;
   private final Lock sessionCountModifyLock = new ReentrantLock();
 
-  public boolean verifySessionForUser(String username) {
+  public boolean verifySessionForUser(String username, String token) {
     if (!privilegedUsers.contains(username) && !nonPrivilegedUsers.contains(username)) {
       return true;
     }
 
     sessionCountModifyLock.lock();
     try {
-      concurrentSessionCounter.putIfAbsent(username, 0);
-      if (privilegedUserCheckLimitReached(username) || nonPrivilegedUserCheckLimitReached(username)) {
+      int validTokenNumber = countValidTokensForUser(username);
+      if (privilegedUserCheckLimitReached(username, validTokenNumber) || nonPrivilegedUserCheckLimitReached(username, validTokenNumber)) {
         return false;
       }
-      concurrentSessionCounter.compute(username, (key, value) -> value + 1);
+      concurrentSessionCounter.putIfAbsent(username, new HashSet<>());
+      concurrentSessionCounter.compute(username, (key, tokenSet) -> addTokenForUser(tokenSet, token));
     } finally {
       sessionCountModifyLock.unlock();
     }
     return true;
   }
 
-  private boolean privilegedUserCheckLimitReached(String username) {
+  private int countValidTokensForUser(String username) {
+    int result = 0;
+    Set<String> tokens = concurrentSessionCounter.getOrDefault(username, null);
+    if (tokens != null && !tokens.isEmpty()) {
+      for (String token : tokens) {
+        try {
+          JWT jwtToken = new JWTToken(token);
+          Date expire = jwtToken.getExpiresDate();
+          if (expire == null || expire.after(new Date())) {
+            result++;
+          }
+        } catch (ParseException ignore) {
+        }
+      }
+    }
+    return result;
+  }
+
+  private boolean privilegedUserCheckLimitReached(String username, int validTokenNumber) {
     if (privilegedUserConcurrentSessionLimit < 0) {
       return false;
     }
-    return privilegedUsers.contains(username) && (concurrentSessionCounter.get(username) >= privilegedUserConcurrentSessionLimit);
+    return privilegedUsers.contains(username) && (validTokenNumber >= privilegedUserConcurrentSessionLimit);
   }
 
-  private boolean nonPrivilegedUserCheckLimitReached(String username) {
+  private boolean nonPrivilegedUserCheckLimitReached(String username, int validTokenNumber) {
     if (nonPrivilegedUserConcurrentSessionLimit < 0) {
       return false;
     }
-    return nonPrivilegedUsers.contains(username) && (concurrentSessionCounter.get(username) >= nonPrivilegedUserConcurrentSessionLimit);
+    return nonPrivilegedUsers.contains(username) && (validTokenNumber >= nonPrivilegedUserConcurrentSessionLimit);
   }
 
-  public void sessionEndedForUser(String username) {
-    sessionCountModifyLock.lock();
-    try {
-      concurrentSessionCounter.computeIfPresent(username, (key, counter) -> decreaseCounter(counter));
-    } finally {
-      sessionCountModifyLock.unlock();
+  public void sessionEndedForUser(String username, String token) {
+    if (!token.isEmpty()) {
+      sessionCountModifyLock.lock();
+      try {
+        concurrentSessionCounter.computeIfPresent(username, (key, tokenSet) -> removeTokenFromUser(tokenSet, token));
+      } finally {
+        sessionCountModifyLock.unlock();
+      }
     }
   }
 
-  private Integer decreaseCounter(Integer counter) {
-    counter--;
-    if (counter < 1) {
+  private Set<String> removeTokenFromUser(Set<String> tokenSet, String newToken) {
+    tokenSet.remove(newToken);
+    if (tokenSet.size() == 0) {
       return null;
-    } else {
-      return counter;
     }
+    return tokenSet;
+  }
+
+  private Set<String> addTokenForUser(Set<String> tokens, String newToken) {
+    tokens.add(newToken);
+    return tokens;
   }
 
   @Override
@@ -106,6 +136,7 @@ public class ConcurrentSessionVerifier implements Service {
   }
 
   Integer getUserConcurrentSessionCount(String username) {
-    return concurrentSessionCounter.get(username);
+    int result = countValidTokensForUser(username);
+    return (result == 0) ? null : result;
   }
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
@@ -19,6 +19,8 @@ package org.apache.knox.gateway.session.control;
 
 
 import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.services.Service;
+import org.apache.knox.gateway.services.ServiceLifecycleException;
 
 import java.util.Map;
 import java.util.Set;
@@ -26,29 +28,13 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
-public class ConcurrentSessionVerifier {
-  public static final ConcurrentSessionVerifier INSTANCE = new ConcurrentSessionVerifier();
+public class ConcurrentSessionVerifier implements Service {
   private Set<String> privilegedUsers;
   private Set<String> nonPrivilegedUsers;
   private int privilegedUserConcurrentSessionLimit;
   private int nonPrivilegedUserConcurrentSessionLimit;
   private Map<String, Integer> concurrentSessionCounter;
   private final Lock sessionCountModifyLock = new ReentrantLock();
-
-  private ConcurrentSessionVerifier() {
-  }
-
-  public static ConcurrentSessionVerifier getInstance() {
-    return INSTANCE;
-  }
-
-  public void init(GatewayConfig config) {
-    this.privilegedUsers = config.getPrivilegedUsers();
-    this.nonPrivilegedUsers = config.getNonPrivilegedUsers();
-    this.privilegedUserConcurrentSessionLimit = config.getPrivilegedUsersConcurrentSessionLimit();
-    this.nonPrivilegedUserConcurrentSessionLimit = config.getNonPrivilegedUsersConcurrentSessionLimit();
-    this.concurrentSessionCounter = new ConcurrentHashMap<>();
-  }
 
   public boolean verifySessionForUser(String username) {
     if (!privilegedUsers.contains(username) && !nonPrivilegedUsers.contains(username)) {
@@ -98,6 +84,25 @@ public class ConcurrentSessionVerifier {
     } else {
       return counter;
     }
+  }
+
+  @Override
+  public void init(GatewayConfig config, Map<String, String> options) throws ServiceLifecycleException {
+    this.privilegedUsers = config.getPrivilegedUsers();
+    this.nonPrivilegedUsers = config.getNonPrivilegedUsers();
+    this.privilegedUserConcurrentSessionLimit = config.getPrivilegedUsersConcurrentSessionLimit();
+    this.nonPrivilegedUserConcurrentSessionLimit = config.getNonPrivilegedUsersConcurrentSessionLimit();
+    this.concurrentSessionCounter = new ConcurrentHashMap<>();
+  }
+
+  @Override
+  public void start() throws ServiceLifecycleException {
+
+  }
+
+  @Override
+  public void stop() throws ServiceLifecycleException {
+
   }
 
   Integer getUserConcurrentSessionCount(String username) {
--- a/gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
@@ -105,7 +105,7 @@ public class ConcurrentSessionVerifier implements Service {
 
   private Set<String> removeTokenFromUser(Set<String> tokenSet, String newToken) {
     tokenSet.remove(newToken);
-    if (tokenSet.size() == 0) {
+    if (tokenSet.isEmpty()) {
       return null;
     }
     return tokenSet;
--- a/gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
@@ -22,9 +22,8 @@ import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.services.Service;
 import org.apache.knox.gateway.services.ServiceLifecycleException;
 import org.apache.knox.gateway.services.security.token.impl.JWT;
-import org.apache.knox.gateway.services.security.token.impl.JWTToken;
 
-import java.text.ParseException;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.Map;
@@ -38,10 +37,10 @@ public class ConcurrentSessionVerifier implements Service {
   private Set<String> nonPrivilegedUsers;
   private int privilegedUserConcurrentSessionLimit;
   private int nonPrivilegedUserConcurrentSessionLimit;
-  private Map<String, Set<String>> concurrentSessionCounter;
+  private Map<String, Set<SessionJWT>> concurrentSessionCounter;
   private final Lock sessionCountModifyLock = new ReentrantLock();
 
-  public boolean verifySessionForUser(String username, String token) {
+  public boolean verifySessionForUser(String username, JWT JWToken) {
     if (!privilegedUsers.contains(username) && !nonPrivilegedUsers.contains(username)) {
       return true;
     }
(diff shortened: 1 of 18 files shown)


MrtnBalazs:
@zeroflag

gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,142 @@
...
+
+  private int countValidTokensForUser(String username) {
+    int result = 0;
+    Set<String> tokens = concurrentSessionCounter.getOrDefault(username, null);
+    if (tokens != null && !tokens.isEmpty()) {
+      for (String token : tokens) {
+        try {
+          JWT jwtToken = new JWTToken(token);
+          Date expire = jwtToken.getExpiresDate();
+          if (expire == null || expire.after(new Date())) {
```
zeroflag, line 71: I would put a `hasExpired()` method into JWTToken to handle this check internally.

gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,142 @@
...
+      concurrentSessionCounter.compute(username, (key, tokenSet) -> addTokenForUser(tokenSet, token));
+    } finally {
+      sessionCountModifyLock.unlock();
+    }
+    return true;
+  }
+
+  private int countValidTokensForUser(String username) {
+    int result = 0;
+    Set<String> tokens = concurrentSessionCounter.getOrDefault(username, null);
```
zeroflag, line 65: The null check can be avoided here by using an emptySet as a default. The isEmpty check is not needed.

```java
for (String token: concurrentSessionCounter.getOrDefault(username, Collections.emptySet())) {
  ..
}
```

gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,142 @@
...
+    if (tokens != null && !tokens.isEmpty()) {
+      for (String token : tokens) {
+        try {
+          JWT jwtToken = new JWTToken(token);
+          Date expire = jwtToken.getExpiresDate();
+          if (expire == null || expire.after(new Date())) {
+            result++;
+          }
+        } catch (ParseException ignore) {
```
zeroflag, line 74: Under what circumstances do we get the ParseException. Depending on what it means we should either log it or throw it out.

gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,142 @@
...
+    return true;
+  }
+
+  private int countValidTokensForUser(String username) {
+    int result = 0;
+    Set<String> tokens = concurrentSessionCounter.getOrDefault(username, null);
+    if (tokens != null && !tokens.isEmpty()) {
+      for (String token : tokens) {
+        try {
+          JWT jwtToken = new JWTToken(token);
```
zeroflag, line 69: We need to consider the cost of `new JWToken(token)`. I assume it parses the token everyt ime. Depending on how expensive is this, we might need to store the preparsed tokens or some metadata about the token.

gateway-service-knoxssout/pom.xml
```diff
@@ -68,5 +68,9 @@
             <artifactId>gateway-test-utils</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.knox</groupId>
```
zeroflag, line 72: Which class is references from gateway-server from gateway-service-knoxssout?

gateway-service-knoxsso/pom.xml
```diff
@@ -81,5 +81,9 @@
             <artifactId>gateway-test-utils</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.knox</groupId>
+            <artifactId>gateway-server</artifactId>
```
zeroflag, line 86: Which class is references from gateway-server from gateway-service-knoxssou?

zeroflag (changes requested):
Just a few comments/questions. I haven't finished going through everything yet.

gateway-server/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
MrtnBalazs: Yes, thats way better.

gateway-service-knoxssout/pom.xml
MrtnBalazs: The ConcurrentSessionVerifier is in gateway-server.

gateway-service-knoxsso/pom.xml
MrtnBalazs: The ConcurrentSessionVerifier is in gateway-server.

gateway-service-knoxssout/pom.xml
zeroflag: We can solve this by inverting the dependencies. Let's introduce an interface fo the ConcurrentSessionVerifier and put it under gateway-spi. In fact the ConcurrentSessionVerifier is already a good interface name, I would rename the implementation to something like InMemoryConcurrentSessionVerifier. So the implementation can stay in gateway-server and the interface in gateway-spi. WebSSOutResource and WebSSOResource only needs to know the interface and not the implementation. After that we can delete this dependency.

gateway-service-knoxsso/pom.xml
zeroflag: Same as about. This can be deleted by inverting the dependencies.

gateway-service-knoxssout/pom.xml
MrtnBalazs: Thats a good idea thank you, i implemented it.

gateway-service-knoxsso/pom.xml
MrtnBalazs: Thats a good idea thank you, i implemented it.

gateway-service-knoxssout/pom.xml
zeroflag: Cool, thanks.

gateway-service-knoxssout/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOutResource.java
```diff
@@ -106,6 +111,21 @@ private boolean removeAuthenticationToken(HttpServletResponse response) {
...
 
+    Cookie[] cookies = request.getCookies();
+    if (cookies != null) {
+      Optional<Cookie> SSOCookie = Arrays.stream(cookies).filter(cookie -> cookie.getName().equals(cookieName)).findFirst();
+      if (SSOCookie.isPresent()) {
+        GatewayServices gwServices =
+                (GatewayServices) request.getServletContext().getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
+        if (gwServices != null) {
+          ConcurrentSessionVerifier verifier = gwServices.getService(ServiceType.CONCURRENT_SESSION_VERIFIER);
+          verifier.sessionEndedForUser(request.getUserPrincipal().getName(), SSOCookie.get().getValue());
```
zeroflag, line 122: Do we always have a ConcurrentSessionVerifier? I see at other places we check if it's null or not. Probably we need to add a null check here too.

gateway-service-knoxssout/src/main/java/org/apache/knox/gateway/service/knoxsso/KnoxSSOutMessages.java
```diff
@@ -25,4 +25,7 @@
 public interface KnoxSSOutMessages {
   @Message( level = MessageLevel.INFO, text = "There was a problem determining the SSO cookie domain - using default domain.")
   void problemWithCookieDomainUsingDefault();
+
+  @Message(level = MessageLevel.DEBUG, text = "Could not find cookie with the name: {0} in the request to be removed from the concurrent session counter for user: {1}. ")
+  void couldNotFindCookieWithTokenToRemove(String cookieName, String username);
```
zeroflag, line 30: I think this should be at least INFO or WARNING.

zeroflag (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,168 @@
...
+public class InMemoryConcurrentSessionVerifier implements ConcurrentSessionVerifier {
+  private Set<String> privilegedUsers;
+  private Set<String> nonPrivilegedUsers;
+  private int privilegedUserConcurrentSessionLimit;
+  private int nonPrivilegedUserConcurrentSessionLimit;
+  private Map<String, Set<SessionJWT>> concurrentSessionCounter;
+  private final Lock sessionCountModifyLock = new ReentrantLock();
+
+  @Override
+  public boolean verifySessionForUser(String username, JWT JWToken) {
```
zeroflag, line 44: Let's rename this parameter name to jwtToken

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,168 @@
...
+
+  private Set<SessionJWT> removeTokenFromUser(Set<SessionJWT> sessionTokenSet, String token) {
+    sessionTokenSet.removeIf(sessionToken -> sessionToken.getToken().equals(token));
+    if (sessionTokenSet.isEmpty()) {
+      return null;
+    }
+    return sessionTokenSet;
+  }
+
+  private Set<SessionJWT> addTokenForUser(Set<SessionJWT> sessionTokenSet, JWT JWToken) {
```
zeroflag, line 105: Let's rename the parameter name to jwtToken.

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,168 @@
...
+  private boolean nonPrivilegedUserCheckLimitReached(String username, int validTokenNumber) {
+    if (nonPrivilegedUserConcurrentSessionLimit < 0) {
+      return false;
+    }
+    return nonPrivilegedUsers.contains(username) && (validTokenNumber >= nonPrivilegedUserConcurrentSessionLimit);
+  }
+
+  @Override
+  public void sessionEndedForUser(String username, String token) {
+    if (!token.isEmpty()) {
```
zeroflag, line 87: Let's use StringUtils.isNotBlank instead. This one also checks if the token is null.

gateway-service-knoxssout/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOutResource.java
```diff
@@ -106,6 +111,23 @@ private boolean removeAuthenticationToken(HttpServletResponse response) {
     }
     response.addCookie(c);
 
+    Cookie[] cookies = request.getCookies();
+    if (cookies != null) {
+      Optional<Cookie> ssoCookie = Arrays.stream(cookies).filter(cookie -> cookie.getName().equals(cookieName)).findFirst();
```
zeroflag, line 116: Could you extract the cookie finding logic in its own method?

Something like this:

```java
Optional<Cookie> findCookie(String cookieName) {
    Cookie[] cookies = request.getCookies();
    if (cookies != null) {
      return Arrays.stream(cookies).filter(cookie -> cookie.getName().equals(cookieName)).findFirst();
   } else {
      return Optional.empty();
   }
}
```

zeroflag (changes requested):

zeroflag (approved):

gateway-server/src/test/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifierTest.java
```diff
@@ -106,25 +110,25 @@ public void userIsNotPrivileged() {
...
   }
 
   @Test
-  public void sessionsDoNotGoToNegative() {
+  public void sessionsDoNotGoToNegative() throws ServiceLifecycleException {
     GatewayConfig config = mockConfig(new HashSet<>(Arrays.asList("admin")), new HashSet<>(Arrays.asList("tom", "guest")), 2, 2);
-    verifier.init(config);
+    verifier.init(config, options);
 
     Assert.assertNull(verifier.getUserConcurrentSessionCount("admin"));
```
pzampino: Why does this return null instead of zero?

pzampino (commented):
I'm still looking, but wanted to share the comments I have thus far, but also want to ask if there is config for enabling/disabling this behavior. Can this functionality be disabled? It does not look that way to me, but it's a lot to digest.

gateway-server/src/test/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifierTest.java
```diff
@@ -48,25 +52,25 @@ private GatewayConfig mockConfig(Set<String> privilegedUsers, Set<String> nonPri
 
 
   @Test
-  public void userIsInNeitherOfTheGroups() {
+  public void userIsInNeitherOfTheGroups() throws ServiceLifecycleException {
```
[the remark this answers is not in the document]
MrtnBalazs: Okay, i will keep that in mind.

gateway-server/src/test/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifierTest.java
```diff
@@ -18,22 +18,26 @@
...
 public class ConcurrentSessionVerifierTest {
-
   private ConcurrentSessionVerifier verifier;
+  private Map<String, String> options;
 
   @Before
   public void setUp() {
-    verifier = ConcurrentSessionVerifier.getInstance();
+    verifier = new ConcurrentSessionVerifier();
+    options = new HashMap<>();
```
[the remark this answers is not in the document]
MrtnBalazs: It is not populated, good idea to change it to Collections.emptyMap().

gateway-server/src/test/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifierTest.java
MrtnBalazs: Originally it returned 0, but me and @smolnar82 reviewed the code and decided to return null instead. It was in the previous pull request: https://github.com/apache/knox/pull/608/commits/af16b58d8c3bcbd9c40841b7c92416ab0bff97f6#diff-cf1fb21462bed96d5f2087f2729a659a2b4141b20536e114a72b24f9d82d4356.
Unfortunately, we talked via Zoom so i can not remember the exact reason, but right now i can't see any reason why returning 0 would not be good.

zeroflag: As far as I remember, It was related to removing the session info from the map altogether so that it won't take any space in the Map when the counter goes zero, isn't it?

gateway-server/src/test/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifierTest.java
```diff
@@ -48,25 +52,25 @@ private GatewayConfig mockConfig(Set<String> privilegedUsers, Set<String> nonPri
 
 
   @Test
-  public void userIsInNeitherOfTheGroups() {
+  public void userIsInNeitherOfTheGroups() throws ServiceLifecycleException {
     GatewayConfig config = mockConfig(new HashSet<>(Arrays.asList("admin")), new HashSet<>(Arrays.asList("tom", "guest")), 3, 2);
-    verifier.init(config);
+    verifier.init(config, options);
     for (int i = 0; i < 4; i++) {
```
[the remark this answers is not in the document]
MrtnBalazs: The idea was to prove that if the user is in neither of the groups, neither of the limits apply to the user. To test that i verify the user one more time than the higher limit. In a later commit i took out the for cycle because we give the function a token as well, and the tokens are class variables and pre issued in the setUp() function. I'll add some comments to clarify the test and rename it "testuserIsInNeitherOfTheGroupsCanBeLoggedInUnlimitedTimes".

gateway-server/src/test/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifierTest.java
MrtnBalazs: Yes, because originally there were no tokens stored and we didn't have to count the not expired ones, so we just returned the value from the HashMap for the user, which returned null if the user was not in the HashMap. Now we have to count the not expired tokens so this function might be unnecessary, we could add countValidTokensForUser(username) function package visibility and use it in the test instead of getUserConcurrentSession(username).

MrtnBalazs:
The plan is to implement the enabling/disabling in the next pull request.

gateway-server/src/test/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifierTest.java
pzampino: Whether or not the count is stored in the map (implementation detail), the contract of the method could be to return zero. If there is no entry in the map, that also means there are zero sessions, correct?

gateway-spi/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.session.control;
+
+import org.apache.knox.gateway.services.Service;
+import org.apache.knox.gateway.services.security.token.impl.JWT;
+
+public interface ConcurrentSessionVerifier extends Service {
+  boolean verifySessionForUser(String username, JWT JWToken);
```
pzampino, line 24: This verify method answers the question of whether or not the specified user is permitted to have a[nother] session? Some javadoc may be helpful here.

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+    return nonPrivilegedUsers.contains(username) && (validTokenNumber >= nonPrivilegedUserConcurrentSessionLimit);
+  }
+
+  @Override
+  public void sessionEndedForUser(String username, String token) {
+    if (StringUtils.isNotBlank(token)) {
+      sessionCountModifyLock.lock();
+      try {
+        concurrentSessionCounter.computeIfPresent(username, (key, sessionTokenSet) -> removeTokenFromUser(sessionTokenSet, token));
```
pzampino, line 91: I think this could result in a NullPointerException if the username is null. It may be good to validate that parameter along with the token?

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
```diff
@@ -0,0 +1,169 @@
...
+
+  private Set<SessionJWT> addTokenForUser(Set<SessionJWT> sessionTokenSet, JWT jwtToken) {
+    sessionTokenSet.add(new SessionJWT(jwtToken));
+    return sessionTokenSet;
+  }
+
+  @Override
+  public void init(GatewayConfig config, Map<String, String> options) throws ServiceLifecycleException {
+    this.privilegedUsers = config.getPrivilegedUsers();
+    this.nonPrivilegedUsers = config.getNonPrivilegedUsers();
```
pzampino, line 114: Does this represent a configured set of non-privileged users for whom a concurrent session limit will be applied?
If so, why can't any user who is NOT configured as a member of the privileged users set be treated as a non-privileged user? Why do we need a separate configuration for these non-privileged users? What happens to those users who are in neither configured set? Do they get as many sessions as they want?

gateway-server/src/test/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifierTest.java
MrtnBalazs: The users who aren't configured for either of the groups could have sessions and they won't have entry in the map, but i think too that the method could return zero.

gateway-spi/src/main/java/org/apache/knox/gateway/session/control/ConcurrentSessionVerifier.java
MrtnBalazs: Yes, okay.

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
MrtnBalazs: Yes, good idea.

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
MrtnBalazs: Yes, there are two groups privileged and non-privileged, each has a limit. Users who are in neither of the groups can have unlimited sessions. @smolnar82 mentioned at the begging of the task that there might be overlapping between the groups so we need to have two groups configured, but if you suggest that configuring one group is enough, than there might have been some misunderstanding in this topic.

smolnar82: https://issues.apache.org/jira/browse/KNOX-2789

gateway-server/src/main/resources/META-INF/services/org.apache.knox.gateway.services.ServiceFactory
```diff
@@ -30,4 +30,5 @@ org.apache.knox.gateway.services.factory.ServiceRegistryServiceFactory
 org.apache.knox.gateway.services.factory.SslServiceFactory
 org.apache.knox.gateway.services.factory.TokenServiceFactory
 org.apache.knox.gateway.services.factory.TokenStateServiceFactory
-org.apache.knox.gateway.services.factory.TopologyServiceFactory
\ No newline at end of file
+org.apache.knox.gateway.services.factory.TopologyServiceFactory
+org.apache.knox.gateway.services.factory.ConcurrentSessionVerifierFactory
```
smolnar82, line 34: nit: this list is an alphabetically-ordered list. The new service factory should have been added in its appropriate line (line 20)

smolnar82 (approved):
Looks good to me except for a token creation concern which we'll address in another follow-up JIRA.
I'm going to file it and ask @MrtnBalazs to implement the required changes.

smolnar82:
The follow-up JIRA is created here: https://issues.apache.org/jira/browse/KNOX-2790

PR #620, opened 2022-08-18
KNOX-2788 - Implementing EmptyVerifier and Cleaning background thread

MrtnBalazs:

I added an empty implementation of the `ConcurrentSessionVerifier` (`EmptyConcurrentSessionVerifier`), now it is the default so users do not waste resources when they forget to configure the verifier. If they need the `InMemoryConcurrentSessionVerifier` implementation they need to configure it and pay attention to configure at least one user in at least one of the groups.
For example:
```
<property>
        <value>admin</value>
</property>
<property>
        <name>gateway.service.concurrentsessionverifier.impl</name>
        <value>org.apache.knox.gateway.session.control.InMemoryConcurrentSessionVerifier</value>
</property>
```

I also added a background thread which cleans out the expired tokens from the `InMemoryConcurrentSessionVerifier` to save some resources. The default cleaning time is 30 minutes, but it can be configured by adding how many minutes we want the cleaning period to be:
```
<property>
        <name>gateway.session.verification.expired.tokens.cleaning.period</name>
        <value>80</value>
</property>
<property>
        <value>admin</value>
</property>
<property>
        <name>gateway.service.concurrentsessionverifier.impl</name>
        <value>org.apache.knox.gateway.session.control.InMemoryConcurrentSessionVerifier</value>
</property>
```

## How was this patch tested?

I have written unit tests for the `ConcurrentSessionVeriferFactory` to test that the factory gives back the right implementation for the right configuration.
I also added a unit test into `InMemoryConcurrentSessionVerifierTest` to test the background thread removing the expired tokens.
I also added stress tests to `InMemoryConcurrentSessionVerifierTest` to test whether the locking is working and the `InMemoryConcurrentSessionVerifier` is truly thread safe.

I also tested the behaviour with the following configurations:

1. Nothing
2. Empty String
```
<property>
        <name>gateway.service.concurrentsessionverifier.impl</name>
        <value></value>
</property>
```
3. Empty implementation
```
<property>
        <name>gateway.service.concurrentsessionverifier.impl</name>
        <value>org.apache.knox.gateway.session.control.EmptyConcurrentSessionVerifier</value>
</property>
```
4. InMemory implementation without groups configured
```
<property>
        <name>gateway.service.concurrentsessionverifier.impl</name>
        <value>org.apache.knox.gateway.session.control.InMemoryConcurrentSessionVerifier</value>
</property>
```
5. InMemory fully configured
```
<property>
        <value>tom,guest</value>
</property>
<property>
        <value>admin</value>
</property>
<property>
        <name>gateway.privileged.users.concurrent.session.limit</name>
        <value>2</value>
</property>
<property>
        <name>gateway.non.privileged.users.concurrent.session.limit</name>
        <value>1</value>
</property>
<property>
        <name>gateway.session.verification.expired.tokens.cleaning.period</name>
        <value>80</value>
</property>
<property>
        <name>gateway.service.concurrentsessionverifier.impl</name>
        <value>org.apache.knox.gateway.session.control.InMemoryConcurrentSessionVerifier</value>
</property>
```

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
@@ -42,7 +42,7 @@ public class ConcurrentSessionVerifierFactory extends AbstractServiceFactory {
         if (isThereGroupConfigured(gatewayConfig)) {
           service = new InMemoryConcurrentSessionVerifier();
         } else {
-          throw new ServiceLifecycleException("Error, neither privileged nor non-privileged groups are configured!");
+          throw new ServiceLifecycleException("Error creating InMemoryConcurrentSessionVerifier, at least one user should be added in either the privileged group or the non-privileged group!");
         }
       }
       logServiceUsage(implementation, serviceType);
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
@@ -17,6 +17,12 @@
  */
 package org.apache.knox.gateway.services.factory;
 
+import static java.util.Arrays.asList;
+import static java.util.Collections.unmodifiableList;
+
+import java.util.Collection;
+import java.util.Map;
+
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.services.GatewayServices;
 import org.apache.knox.gateway.services.Service;
@@ -24,12 +30,6 @@ import org.apache.knox.gateway.services.ServiceLifecycleException;
 import org.apache.knox.gateway.services.ServiceType;
 import org.apache.knox.gateway.session.control.EmptyConcurrentSessionVerifier;
 import org.apache.knox.gateway.session.control.InMemoryConcurrentSessionVerifier;
-
-import java.util.Collection;
-import java.util.Map;
-
-import static java.util.Arrays.asList;
-import static java.util.Collections.unmodifiableList;
 
 public class ConcurrentSessionVerifierFactory extends AbstractServiceFactory {
   @Override
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
@@ -22,16 +22,32 @@ import org.apache.knox.gateway.services.GatewayServices;
 import org.apache.knox.gateway.services.Service;
 import org.apache.knox.gateway.services.ServiceLifecycleException;
 import org.apache.knox.gateway.services.ServiceType;
+import org.apache.knox.gateway.session.control.EmptyConcurrentSessionVerifier;
 import org.apache.knox.gateway.session.control.InMemoryConcurrentSessionVerifier;
 
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Map;
+
+import static java.util.Arrays.asList;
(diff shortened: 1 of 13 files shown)


MrtnBalazs:
@zeroflag @smolnar82

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
```diff
@@ -122,17 +129,34 @@ public void init(GatewayConfig config, Map<String, String> options) throws Servi
...
+    expiredTokenRemover.shutdown();
+  }
 
+  private void removeExpiredTokens() {
+    sessionCountModifyLock.lock();
+    try {
+      Iterator<Map.Entry<String, Set<SessionJWT>>> concurrentSessionCounterIterator = concurrentSessionCounter.entrySet().iterator();
+      while (concurrentSessionCounterIterator.hasNext()) {
+        Map.Entry<String, Set<SessionJWT>> concurrentSessionCounterMapEntry = concurrentSessionCounterIterator.next();
+        concurrentSessionCounterMapEntry.setValue(concurrentSessionCounterMapEntry.getValue().stream().filter(sessionJWT -> !sessionJWT.hasExpired()).collect(Collectors.toSet()));
```
zeroflag, line 152: This might be shorter and easer to read than the filter expression. Plus it's probably faster too.

```java
concurrentSessionCounterMapEntry.getValue().removeIf(session -> session.hasExpired());
```

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
```diff
@@ -17,21 +17,37 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Map;
-
 public class ConcurrentSessionVerifierFactory extends AbstractServiceFactory {
   @Override
   protected Service createService(GatewayServices gatewayServices, ServiceType serviceType, GatewayConfig gatewayConfig, Map<String, String> options, String implementation) throws ServiceLifecycleException {
-    return shouldCreateService(implementation) ? new InMemoryConcurrentSessionVerifier() : null;
+    Service service = null;
+    if (shouldCreateService(implementation)) {
```
zeroflag, line 38: If `shouldCreateService` returns false we don't create any implementation and we return null? Later at other place we have assumption that there is always an implementation available (either the null object or the real one).

zeroflag (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
MrtnBalazs: As far as i can see, it is impossible to add null as a service. When we add the service with the `addService` method of `AbstractGatewayServices` class a `NullPointerException` is thrown by these lines:
```
// Ensure the service is not null
if (service == null) {
   throw new NullPointerException("A null service may not be added.");
}
```

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
MrtnBalazs: Yes, thats an easier way. Thank you.

gateway-server/src/main/java/org/apache/knox/gateway/services/factory/ConcurrentSessionVerifierFactory.java
smolnar82: I agree with @MrtnBalazs on this. It's perfectly ok to return a `null` service from any of the service factories. That situation is handled by the caller.

smolnar82 (approved):
LGTM

zeroflag (approved):

PR #624, opened 2022-08-24
KNOX-2790 - Added a new funtion to verifier, this way the session lim…

MrtnBalazs:

Changed the verifier, introduced a new function called `registerToken`, this function is checking the session limits for the given user and stores the  given token. Changed `verifySessionForUser` function, now this function only checks the limit but do not store token. These changes were needed, because previously in case of an attack we would generate tokens needlessly before verifying the session and checking the limit. Now we check the limit without a token before token generation, so in case of an attack we do not waste resources. And we add the token after token generation and also check the limit because thread safety requires it.

## How was this patch tested?

I have changed the unit tests in `InMemoryConcurrentSessionVerifierTest` and `WebSSOResourceTest` to test the new usage of the verifier.
I also tested it manually with this configuration:
```
<property>
        <name>gateway.session.verification.unlimited.users</name>
        <value>admin</value>
</property>
<property>
        <name>gateway.session.verification.privileged.users</name>
        <value>tom</value>
</property>
<property>
        <name>gateway.session.verification.privileged.user.limit</name>
        <value>2</value>
</property>
<property>
        <name>gateway.session.verification.non.privileged.user.limit</name>
        <value>1</value>
</property>
<property>
        <name>gateway.session.verification.expired.tokens.cleaning.period</name>
        <value>80</value>
</property>
<property>
        <name>gateway.service.concurrentsessionverifier.impl</name>
        <value>org.apache.knox.gateway.session.control.InMemoryConcurrentSessionVerifier</value>
</property>
```

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
@@ -51,8 +51,37 @@ public class InMemoryConcurrentSessionVerifier implements ConcurrentSessionVerif
   private long cleaningPeriod;
   private final ScheduledExecutorService expiredTokenRemover = Executors.newSingleThreadScheduledExecutor();
 
+  /**
+   * This function is used after the verifySessionForUser function.
+   * It checks the session limits even though verifySessionForUser already checked them,
+   * because in high stress situations there might be some threads which get verified even though they are over the limit in
+   * verifySessionForUser function, so we catch them here.
+   */
   @Override
-  public boolean verifySessionForUser(String username, JWT jwtToken) {
+  public boolean registerToken(String username, JWT jwtToken) {
+    if (StringUtils.isBlank(username)) {
+      LOG.errorRegisteringTokenForBlankUsername();
+      return false;
+    }
+    if (unlimitedUsers.contains(username)) {
+      return true;
+    }
+
+    sessionCountModifyLock.lock();
+    try {
+      if (checkLimitReached(username)) {
+        return false;
+      }
+      concurrentSessionCounter.putIfAbsent(username, new HashSet<>());
+      concurrentSessionCounter.compute(username, (key, sessionTokenSet) -> addTokenForUser(sessionTokenSet, jwtToken));
+    } finally {
+      sessionCountModifyLock.unlock();
+    }
+    return true;
+  }
+
+  @Override
+  public boolean verifySessionForUser(String username) {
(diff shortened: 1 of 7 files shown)


MrtnBalazs:
@smolnar82 @zeroflag

gateway-server/src/main/java/org/apache/knox/gateway/session/control/InMemoryConcurrentSessionVerifier.java
```diff
@@ -51,10 +51,16 @@ public class InMemoryConcurrentSessionVerifier implements ConcurrentSessionVerif
...
 
+  /**
+   * This function is used after the verifySessionForUser function.
+   * It checks the session limits even though verifySessionForUser already checked them,
+   * because in high stress situations there might be some threads which get verified even though they are over the limit in
+   * verifySessionForUser function, so we catch them here.
+   */
   @Override
-  public boolean verifySessionForUser(String username, JWT jwtToken) {
+  public boolean registerToken(String username, JWT jwtToken) {
```
smolnar82, line 61: This could be simplified like that:
```
if (verifySessionForUser(userName)) {
    //lock
    concurrentSessionCounter.putIfAbsent(username, new HashSet<>());
    //unlock
    return true;
}
return false;
```

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
```diff
@@ -264,9 +264,14 @@ private Response getAuthenticationToken(int statusCode) {
       original = originalUrlCookies.get(0).getValue();
     }
 
+    Principal p = request.getUserPrincipal();
+    ConcurrentSessionVerifier verifier = services.getService(ServiceType.CONCURRENT_SESSION_VERIFIER);
+    if (!verifier.verifySessionForUser(p.getName())) {
```
smolnar82, line 269: Now we have this same 3-line code block in two places; I'd create a new private void method that throws the web app exception if needed (to avoid code duplication).

smolnar82 (changes requested):

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
smolnar82: oh...my bad. These are not the same lines.

smolnar82 (approved):
Please ignore my previous messages; LGTM

zeroflag (approved):
LGTM

PR #626, opened 2022-08-25
KNOX-2795 Handling missing OIDC client name parameter

zeroflag:

The client name config parameter was handled in an inconsistent manner.

 * When it was null we threw an exception saying that this parameter is mandatory
 * However when it was an empty string, we did fallback using the default client name coming from the provider

I removed the null check so that we use the default in that case too.

## How was this patch tested?

Manually tested with a topology like this:

```xml
      <provider>
          <role>federation</role>
          <name>pac4j</name>
          <enabled>true</enabled>
          <param>
            <name>pac4j.callbackUrl</name>
            <value>https://localhost:8443/gateway/knoxsso/api/v1/websso</value>
          </param>
<!-- client name is commented out
          <param>
            <name>clientName</name>
            <value>OidcClient</value>
          </param>
-->
          <param>
            <name>oidc.id</name>
            <value>....</value>
          </param>
          <param>
            <name>oidc.secret</name>
            <value>SECRET</value>
          </param>
          <param>
            <name>oidc.discoveryUri</name>
            <value>...</value>
          </param>
          <param>
            <name>oidc.preferredJwsAlgorithm</name>
            <value>RS256</value>
          </param>
      </provider>
```

### DIFF
--- a/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/Pac4jMessages.java
+++ b/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/Pac4jMessages.java
@@ -32,9 +32,6 @@ public interface Pac4jMessages {
   @Message( level = MessageLevel.ERROR, text = "pac4j callback URL required")
   void ssoAuthenticationProviderUrlRequired();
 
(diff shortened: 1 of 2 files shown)


smolnar82 (approved):
LGTM

PR #629, opened 2022-09-06
KNOX-2799 - When exporting certificate using Knox CLI, the certificate is coming with wrong name when type=PEM

zeroflag:

There is an inconsistency in naming of the output file of the export cert command. The doc says it should be `gateway-client-trust.<type>` but in case of pem it's `gateway-identity.pem`.

## How was this patch tested?

```bash
$ bin/knoxcli.sh export-cert --type pem
Certificate gateway-identity has been successfully exported to: /Users/attilamagyar/development/test/data/security/keystores/gateway-client-trust.pem
$  ls -al  /Users/attilamagyar/development/test/data/security/keystores/gateway-client-trust.pem
-rw-r--r--  1 attilamagyar  staff  1246 Sep  6 10:28 /Users/attilamagyar/development/test/data/security/keystores/gateway-client-trust.pem
````


smolnar82 (approved):
LGTM

PR #630, opened 2022-09-06
KNOX-2801 Error Message should come in virtual group mapping provider if there are multiple occurrence's of a virtual group.

zeroflag:

It turned out to be a general problem that effects other filters as well. I added a warning message if there are filter config params with the same name.

## How was this patch tested?

Tested with the following config:

```xml
        <provider>
            <name>HadoopGroupProvider</name>
            <enabled>true</enabled>
           <param>
                <name>group.mapping.abhi</name>
                <value>true</value>
            </param>
           <param>
                <name>group.mapping.abhi</name>
                <value>false</value>
            </param>
        </provider>
```

```bash
$ grep WARN logs/gateway.log 
2022-09-06 18:52:46,777  WARN  knox.gateway (Provider.java:addParam(67)) - Duplicated filter param key: group.mapping.abhi
```

### DIFF
--- a/gateway-release/home/conf/topologies/homepage.xml
+++ b/gateway-release/home/conf/topologies/homepage.xml
@@ -47,10 +47,6 @@
             <value>false</value>
          </param>
          <param>
-            <name>xframe.options.enabled</name>
-            <value>true</value>
-         </param>
-         <param>
             <name>xframe.options</name>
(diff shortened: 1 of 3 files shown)


smolnar82 (approved):
LGTM

PR #636, opened 2022-09-23
KNOX-2807 - Hive's SSL_ENABLED flag is processed on the service level with the correct service configuration name in CM

smolnar82:

Knox's automated service discovery had a bug when it came to HIVE(_ON_TEZ) service discovery. These services have their SSL-enabled flag on the service level, but the generated service model (by Knox) stored that information in the HS2 role configs.
In addition to this, the CM configuration name referred to the so-called `related name` configuration field which Knox does not consider when comparing previous and current service/role configs.

## How was this patch tested?

Updated JUnit tests and executed manual testing in a cluster with KNOX, HIVE and HIVE_ON_TEZ.


smolnar82:
Cc. @MrtnBalazs

zeroflag (approved):

PR #639, opened 2022-09-28
KNOX-2631 - Webshell

moresandeep:
This feature enables shell access to the machine running Apache Knox. Users can SSO into Knox and then access shell using the KnoxShell URL on knox homepage.

image

image

**Configuration:**
Webshell is not turned on by default. To enable Webshell following properties needs to be changed in gateway-site.xml
```
 <property>
        <name>gateway.websocket.feature.enabled</name>
        <value>true</value>
        <description>Enable/Disable websocket feature.</description>
    </property>

<property>
        <name>gateway.webshell.feature.enabled</name>
        <value>true</value>
        <description>Enable/Disable webshell feature.</description>
    </property>
<!-- in case JWT cookie validation for websockets is needed -->
<property>
        <name>gateway.websocket.JWT.validation.feature.enabled</name>
        <value>true</value>
        <description>Enable/Disable websocket JWT validation feature.</description>
    </property>
```
Create a sudoers file `/etc/sudoers.d/knox` with entries for all the users that need webshell acess. e.g. the following config file let's user `sam` login to webshell. Further restrictions on what user `sam`
 can do can be performed with sudoers file. More info: https://linux.die.net/man/5/sudoers

```
Defaults env_keep += JAVA_HOME
Defaults always_set_home
knox ALL=(sam:ALL) NOPASSWD: /bin/bash
knox ALL=(knoxui:ALL) NOPASSWD: /bin/bash
```

## How was this patch tested?
This patch was tested on my local machine.


gateway-server/src/main/java/org/apache/knox/gateway/deploy/DeploymentFactory.java
```diff
@@ -260,7 +260,13 @@ private static void addMissingDefaultProviders(Topology topology) {
...
     // and exceptions will be thrown as topologies are deployed even though they will
     // work fine at actual server runtime.
-    if (PROVIDER_CONTRIBUTOR_MAP.get("identity-assertion") != null) {
+
+    // SRM: Only add "Default" provider iff it is found in the provider contributor map
+    // There could be cases where service loader might find other
+    // identity-assertion providers e.g. JWTAuthCodeAsserter
+    if (PROVIDER_CONTRIBUTOR_MAP.get("identity-assertion") != null &&
+            PROVIDER_CONTRIBUTOR_MAP.get("identity-assertion").keySet() != null &&
+            PROVIDER_CONTRIBUTOR_MAP.get("identity-assertion").keySet().contains("Default")) {
```
lmccay, line 269: This doesn't seem to be specific to WebShell - is this a generic issue that should be pulled out separately?

gateway-server/src/main/java/org/apache/knox/gateway/websockets/GatewayWebsocketHandler.java
```diff
@@ -109,18 +115,48 @@ public void configure(final WebSocketServletFactory factory) {
...
+  private Boolean isWebshellRequest(URI requestURI){
+    return requestURI.toString().matches(REGEX_WEBSHELL_REQUEST_PATH);
+  }
+
+  private WebshellWebSocketAdapter handleWebshellRequest(ServletUpgradeRequest req){
+      if (config.isWebShellEnabled()){
+        if (concurrentWebshells.get() >= config.getMaximumConcurrentWebshells()){
+          throw new RuntimeException("Number of allowed concurrent Web Shell sessions exceeded");
+        }
+        JWTValidator jwtValidator = JWTValidatorFactory.create(req,services,config);
```
lmccay, line 127: spaces following param names

gateway-server/src/main/java/org/apache/knox/gateway/websockets/JWTValidator.java
```diff
@@ -0,0 +1,193 @@
...
+import org.apache.knox.gateway.services.security.token.TokenUtils;
+import org.apache.knox.gateway.services.security.token.UnknownTokenException;
+import org.apache.knox.gateway.services.security.token.impl.JWT;
+import org.apache.knox.gateway.util.Tokens;
+
+import java.security.interfaces.RSAPublicKey;
+import java.text.ParseException;
+import java.util.Date;
+
+public class JWTValidator {
```
lmccay, line 37: Not seeing anything websocket specific in this class and there seems to likely be redundant implementations of verification. Is there some reason that an existing class couldn't be used or that this isn't made more common?

gateway-server/src/main/java/org/apache/knox/gateway/websockets/JWTValidatorFactory.java
```diff
@@ -0,0 +1,131 @@
...
+import org.eclipse.jetty.websocket.servlet.ServletUpgradeRequest;
+import javax.servlet.ServletException;
+import java.net.HttpCookie;
+import java.security.interfaces.RSAPublicKey;
+import java.text.ParseException;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class JWTValidatorFactory {
```
lmccay, line 42: Again, there is very little websocket specific in here and I think that the ServletUpgradeRequest can likely be made more generic depending on its hierarchy. It seems like this Factory should be more generic and common JWT code rather than websocket specific.

lmccay: if ServletUpgradeRequest can't be cast to or used to retrieve the HttpServletRequest to be more generic then maybe there should be an abstract based class that can override extractToken here.

pom.xml
```diff
@@ -280,7 +284,13 @@
         <xml-matchers.version>0.10</xml-matchers.version>
         <zookeeper.version>3.5.7</zookeeper.version>
     </properties>
-
+    <repositories>
+        <repository>
+            <id>jcenter</id>
```
lmccay, line 289: what is this for?

pom.xml
```diff
@@ -248,13 +249,16 @@
...
+        <nodejs.version>v12.20.2</nodejs.version>
         <okhttp.version>2.7.5</okhttp.version>
         <opensaml.version>3.4.5</opensaml.version>
         <pac4j.version>4.5.6</pac4j.version>
         <postgresql.version>42.3.3</postgresql.version>
         <mysql.version>8.0.28</mysql.version>
         <protobuf.version>3.16.1</protobuf.version>
+        <powermock.version>2.0.9</powermock.version>
+        <purejavacomm.version>0.0.11.1</purejavacomm.version>
+        <pty4j.version>0.11.4</pty4j.version>
```
lmccay, line 261: Note that this is EPLv1 licensed. This is a weak copy-left licence under what ASLv2 considers Category B. We will need to meet the conditions in https://www.apache.org/legal/resolved.html#category-b

lmccay: Adding pty4j and anything else with this license to the entry in LICENSE file where we are calling out Jetty and Jerico is likely enough.

lmccay (changes requested):
Looks good, asked some questions and we need to update the LICENSE file for pty4j EPLv1 license.

gateway-release/home/conf/gateway-site.xml
```diff
@@ -88,6 +88,35 @@ limitations under the License.
...
+    <!-- must have websocket enabled to use webshell --> 
+    <property>
+        <name>gateway.webshell.feature.enabled</name>
+        <value>false</value>
+        <description>Enable/Disable webshell feature.</description>
+    </property>
+    <property>
+        <name>webshell.max.concurrent.sessions</name>
+        <value>3</value>
+        <description>maximum number of concurrent webshell sessions</description>
```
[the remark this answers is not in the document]
moresandeep: These are total :( what number would you recommend? does 20 sound good?

gateway-release/home/conf/gateway-site.xml
```diff
@@ -88,6 +88,35 @@ limitations under the License.
...
+    </property>
+    <property>
+        <name>gateway.webshell.audit.logging.enabled</name>
+        <value>false</value>
```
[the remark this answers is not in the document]
moresandeep: Right, audit feature is not fully ready yet, it is still in POC phase. We need to think about how to prevent logging secrets, which like you pointed out is lacking as of now. Add a description is a good idea. I am inclined to keeping Audit   in the property name for two reasons 

1. Logs to audit log file
2. For compatibility reasons, if someone actively turns on this feature knowing implications then next time we change property name the feature would stop working after upgrade.

gateway-release/home/conf/gateway-site.xml
```diff
@@ -88,6 +88,35 @@ limitations under the License.
...
+        <name>gateway.webshell.read.buffer.size</name>
+        <value>1024</value>
+        <description>Web Shell buffer size for reading</description>
+    </property>
+
+    <!-- @since 2.0.0 websocket JWT validation configs -->
+    <property>
+        <name>gateway.websocket.JWT.validation.feature.enabled</name>
+        <value>false</value>
+        <description>Enable/Disable websocket JWT validation feature.</description>
```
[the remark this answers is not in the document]
moresandeep: This should be on by default, thanks for pointing this out. This property is used to disable JWT token validation at the websocket layer. This is needed for cases where we want the session the terminate when JWT expires. Turning it off does not mean WS is less secure, WS connections are only established after successful authz and are always SSL.

gateway-server/src/main/java/org/apache/knox/gateway/deploy/DeploymentFactory.java
moresandeep: This was a cause of of NPE when I was working on Webshell. I think I encountered this when working on JWT validation stuff.

gateway-server/src/main/java/org/apache/knox/gateway/websockets/JWTValidator.java
moresandeep: This class leverages classes from `gateway-provider-security-jwt`. There were some features that we needed that were not directly accessible from `gateway-provider-security-jwt` package, mostly because `gateway-provider-security-jwt` implementation is filter specific and I wanted to avoid refactoring it. `JWTValidator ` has a `validate` method that checks issuer and token validity in one place.

moresandeep: I should have started with this comment, this class is used at websocket layer to validate JWT tokens, so this is websocket specific. The implementation is not websocket specific but we could not leverage `gateway-provider-security-jwt` since implementation there was specific to servlet filters.

gateway-server/src/main/java/org/apache/knox/gateway/websockets/JWTValidatorFactory.java
moresandeep: Unfortunately `ServletUpgradeRequest` is part of [Websocket implementation](https://www.eclipse.org/jetty/javadoc/jetty-9/org/eclipse/jetty/websocket/servlet/ServletUpgradeRequest.html) so difficult to make it generic. This idea is that factory is exposes a more generic `JWTValidator` class which is more generic and WS agnostic.

moresandeep: `ServletUpgradeRequest `  is a special request, this is the point where HTTP to WS handoff takes place, hence a logical place to grab JWT. This makes it WS specific unfortunately.

pom.xml
moresandeep: This is for pty4j. Interestingly, [Pty4j was not part of maven central until Nov 2021 ](https://mvnrepository.com/artifact/org.jetbrains.pty4j/pty4j)(when this was introduced). I can now see [Pty4j in maven central ](https://mvnrepository.com/artifact/org.jetbrains.pty4j/pty4j) 

I'll use the new version of Pty4j from maven central and see if it works.
Thank you soo much for catching this!

pom.xml
moresandeep: Thanks, I thought I had checked the license. Let me add those entries.

moresandeep: So, we already have Eclipse license for Jetty, I just added Pty4j to the list using that license (without duplicating the entire license text)

gateway-server/src/test/java/org/apache/knox/gateway/websockets/GatewayWebsocketHandlerTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+import static org.easymock.EasyMock.isA;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({JWTValidatorFactory.class, MessagesFactory.class, GatewayWebsocketHandler.class, AuditServiceFactory.class})
+public class GatewayWebsocketHandlerTest {
+    @BeforeClass
+    public static void setUpBeforeClass() throws Exception {
+        // setup MessageFactory for static variable
+        PowerMock.mockStatic(MessagesFactory.class);
```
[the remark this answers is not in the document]
moresandeep: This was a frustrating exercise :( we have MessagesFactory referenced in statically which was causing tests to blow up (NPEs)

gateway-server/src/test/java/org/apache/knox/gateway/websockets/GatewayWebsocketHandlerTest.java
```diff
@@ -0,0 +1,148 @@
...
+        EasyMock.expect(MessagesFactory.get(GatewaySpiMessages.class)).andReturn(EasyMock.createNiceMock(GatewaySpiMessages.class)).anyTimes();
+        EasyMock.expect(MessagesFactory.get(JWTMessages.class)).andReturn(EasyMock.createNiceMock(JWTMessages.class)).anyTimes();
+        EasyMock.expect(MessagesFactory.get(WebsocketLogMessages.class)).andReturn(EasyMock.createNiceMock(WebsocketLogMessages.class)).anyTimes();
+        PowerMock.replay(MessagesFactory.class);
+        // setup Auditor for static variable
+        PowerMock.mockStatic(AuditServiceFactory.class);
```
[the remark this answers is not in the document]
moresandeep: Same as above.

moresandeep:
> 

Thanks @lmccay ! For license I added Pty4j to the existing EPLv1 license text, is that fine or should I be copying the entire license text? There is no language in the license that indicates I need a separate copy as long as we clearly indicate that Pty4j uses EPLv1 license (followed by the license text)

pom.xml
moresandeep: unfortunately this is needed for [purejavacomm](https://mvnrepository.com/artifact/com.github.purejavacomm/purejavacomm) which is only in jcenter.

moresandeep: Unfortunately, this is needed for [PurejavaComm library](https://mvnrepository.com/artifact/org.jetbrains.pty4j/purejavacomm?repo=jcenter)  that is only available in jcenter.

gateway-release/home/conf/gateway-site.xml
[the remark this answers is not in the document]
lmccay: Good question. We should tease out what might be an upper limit rather than arbitrarily assign some number. Perhaps an inventory of resources allocated per session would be useful: threads, websockets, etc?

gateway-release/home/conf/gateway-site.xml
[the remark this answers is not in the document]
lmccay: Will enabling this also enforce it for other WS usecases like CM, Hue, Zeppelin, etc? Is there any issue in that regard?

gateway-release/home/conf/gateway-site.xml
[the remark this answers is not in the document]
moresandeep: Good point. So, each Webshell session will create 1 WS thread and one HTTP connection. There are reports of Knox comfortably  handling more than 20 WS threads (but I will test this out). About HTTP connections, for jetty 9 min thread count is 10 and max is 1000. I need to see what we have and also test this out. I will do some a bit more load testing on this and come up with a number.

gateway-release/home/conf/gateway-site.xml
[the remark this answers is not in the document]
moresandeep: There are two parts to the answer
1. For all services using WS (Webshell, Zeppelin etc.)
1.2  Yes, by enabling this we will check for valid JWT at the time of **WS connection is created**. 
2. For Webshell only
2.1 We check for valid JWT for each command that is transmitted from UI to Knox
2.2 For other services like Zeppelin we do not check for a valid JWT.

gateway-release/home/conf/gateway-site.xml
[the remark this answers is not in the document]
moresandeep: After testing it appears that this feature causes issues with Zeppelin. Since Zeppelin and other UI might not always use SSOCookie provider there will not be hadoop-jwt cookie to extract JWT token. I will rename this property to indicate this is specific to webshell and take out code that intercepts general WS connections.

knox-homepage-ui/home/app/generalProxyInformation/general.proxy.information.component.ts
```diff
@@ -108,6 +116,7 @@ export class GeneralProxyInformationComponent implements OnInit {
         this['showAdminAPI'] = (profile['gpi_admin_api'] === 'true');
         this['showMetadataAPI'] = (profile['gpi_md_api'] === 'true');
         this['showTokens'] = (profile['gpi_tokens'] === 'true');
+        this['showWebShell'] = (profile['gpi_webshell'] === 'true');
```
smolnar82, line 119: It will not work since you are missing the relevant profile declaration from `org.apache.knox.gateway.dto.HomePageProfile`

knox-webshell-ui/webshell-ui/assets/.gitkeep
```diff
@@ -0,0 +1,17 @@
+##########################################################################
```
smolnar82, line 1: What's the purpose of this file/folder? If there are no assets now, we should not add the folder and remove all references.

pom.xml
```diff
@@ -2473,11 +2524,17 @@
 
             <dependency>
                 <groupId>org.apache.velocity</groupId>
-                <artifactId>velocity-engine-core</artifactId>
+                <artifactId>velocity</artifactId>
```
smolnar82, line 2527: We need to remove `velocity`; see KNOX-2741 for more information.

smolnar82 (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/deploy/DeploymentFactory.java
moresandeep: Done. KNOX-2825

PR #647, opened 2022-10-12
KNOX-2824 - Make SameSite attribute on KnoxSSO Cookie Configurable

lmccay:

The attribute for KnoxSSO cookie is currently hardocded.
This improvement will make its value configurable to better accommodate various deployment scenarios.

## How was this patch tested?

Added new unit tests and ran existing


smolnar82 (approved):
LGTM

moresandeep (commented):
This introduces breaking change, changing to `strict` from `none`. it would be difficult to figure out what went wrong given the errors are thrown on the UI so debugging this would be challenging. Perhaps changing it to `Lax` would be a compromise?

lmccay:
> This introduces breaking change, changing to `strict` from `none`. it would be difficult to figure out what went wrong given the errors are thrown on the UI so debugging this would be challenging. Perhaps changing it to `Lax` would be a compromise?

This will actually only break deployments that are relying on an insecure setting. Making this Strict by default is the same as the change that added Secure=true many years ago. In modern deployments, the SSO cookie is only being presented to Knox itself and to therefore to the same domain. This should not break deployments that are configured properly to do this. Now, we could consider this a 2.0 release change and change the default for a new 1.6.x release. 2.0 is a major release and can carry incompatible changes.

PR #650, opened 2022-10-20
KNOX-2827 - isDispatchAllowed should cut off path segments from the URL

zeroflag:

In isDispatchAllowed we check the full URL against the whitelist regexp (including /path, but not including query parameters). This makes complicated and error prone to configure the regular expression.

## How was this patch tested?

gateway-site.xml

```xml
<property>
      <name>gateway.dispatch.whitelist.services</name>
      <value>HBASEUI,DATANODE,HBASEUI,HDFSUI,JOBHISTORYUI,NODEUI,YARNUI,knoxauth</value>
      <description>The comma-delimited list of service roles for which the gateway.dispatch.whitelist should be applied.</description>
</property>
```

Tested manually.

### DIFF
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/GatewayDispatchFilter.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/GatewayDispatchFilter.java
@@ -33,7 +33,9 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
+import java.net.MalformedURLException;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/GatewayDispatchFilter.java
```diff
@@ -145,14 +147,21 @@ private boolean isDispatchAllowed(HttpServletRequest request) {
...
           decodedURL = URLDecoder.decode(requestURI, StandardCharsets.UTF_8.name());
         } catch (UnsupportedEncodingException e) {
           //
         }
+        String baseUrl;
+        try {
+          URL url = new URL(decodedURL);
+          baseUrl = new URL(url.getProtocol(), url.getHost(), url.getPort(), "").toString();
+        } catch (MalformedURLException e) {
+          throw new RuntimeException(e);
```
moresandeep, line 161: We should log the error before wrapping the exception and throwing.

moresandeep (approved):
Looks good to me, I just have one nit about logging.

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/GatewayDispatchFilter.java
zeroflag: Hi @moresandeep, if we log it here then the message will appear in the log two times. I tested this with a malformed URL and the exception was still visible in the gateway.log, because at the outer level we catch and log everything. I don't see a reason to log it here too.

smolnar82 (approved):
LGTM

PR #659, opened 2022-10-26
fix validation errors of getPID: return 0 when the process started

anyorey:

In some cases  getPID will return 0 even if the process started. e.g. pid file is deleted or file is modified and we can not get 
correct pid from the file.  So I add a new function getPidByCommand. It will give the correct pid (APP_GATEWAY_PID ) if process is started. And we'll check the process whether it is started according to APP_PID obtained from pid file ,if it return false ,use APP_GATEWAY_PID  . set APP_PID=$APP_GATEWAY_PID.In this way,we can get correct messagge whether the gateway or ldap process is started

## How was this patch tested?
1.  start gateway or ldap process 
2.  delete pid file or modify the content in file
3.  execute gateway.sh status or ldap.sh status . It will return correct message 
 

### DIFF
--- a/gateway-release-common/home/bin/knox-functions.sh
+++ b/gateway-release-common/home/bin/knox-functions.sh
@@ -232,12 +232,20 @@ function getPID {
       printf "Can't find PID dir.\n"
       exit 1
    fi
-   if [ ! -f "$APP_PID_FILE" ]; then
+
+   getPidByCommand
+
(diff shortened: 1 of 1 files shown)


anyorey (commented):
fix validation errors of getPID: return 0 when the process started

pzampino:
@anyorey Are you experiencing deletion/modification of the Knox PID files as a regular occurrence?

smolnar82:
Let's close this now and have it re-opened once @anyorey has time to reply to the above question and can rebase this PR on top of the current `master` branch.

PR #670, opened 2022-11-09
KNOX-2835 SQL DB based topology monitor

zeroflag:

This is a similar implementation as the existing zookeeper based monitor, but it uses the SQL DB. Zookeeper is not always available and the DB is already used by the token state service.

The purpose of this monitor is to periodically check the DB for shared-providers and descriptors and update the local file system with the changes from the DB.

1. If a provider/descriptor is deleted from the db but exists on the file system the monitor is going to delete it from the file system
2. If a provider/descriptor exists in the DB but missing from the file system the monitor is going to download it from the DB to the FS.
3. If a provider/descriptor in the DB has a different content than the corresponding local file, then the monitor is going to update the local file with the content from he DB. However to avoid too frequent and unnecessary IO operations we only change the local files if the remote change happened after the last synchronization time (if it happened in the past, we assume the changes were already synced to the disk). So  One can change the local files on the disk without worrying about losing their local changes (until someone updates the DB).

The provider logic is implementeed in DbRemoteConfigurationMonitor.

The original DefaultRemoteConfigurationMonitor was renamed to ZkRemoteConfigurationMonitor.

Since the original zookeeper based monitor was already hooked in the Admin UI and Admin API, the new implementation also supports both the UI + API.

## How was this patch tested?

gateway-site.xml

```
<property>
    <name>gateway.service.remoteconfigurationmonitor.impl</name>
    <value>org.apache.knox.gateway.topology.monitor.db.DbRemoteConfigurationMonitorService</value>
</property>
<property>
    <value>mysql</value>
</property>
<property>
    <value>jdbc:mysql://root:root@localhost:3306/knox</value>
</property>
```

Creating provider with curl:

```bash
$    curl -X PUT -H "Content-Type: application/json" -v -k -u admin:admin-password -d "{'x':'y'}" https://localhost:8443/gateway/admin/api/v1/providerconfig/newprov
```

Checking the DB:

```
mysql> select * from knox_providers;
+--------------+-----------+---------------------+
| name         | content   | last_modified_time  |
+--------------+-----------+---------------------+
| newprov.json | {'x':'y'} | 2022-11-09 09:45:28 |
+--------------+-----------+---------------------+
1 row in set (0.00 sec)
```

Checking the local FS:

```bash
$ cat conf/shared-providers/newprov.json 
{'x':'y'} 
```

Creating descriptor with curl:

```bash
$ curl -X PUT -H "Content-Type: application/json" -v -k -u admin:admin-password -d "{'x':'y'}" https://localhost:8443/gateway/admin/api/v1/descriptors/newdesc
```

Checking the DB:

```
mysql> select * from knox_descriptors;
+--------------+---------------+---------------------+
| name         | content       | last_modified_time  |
+--------------+---------------+---------------------+
| newdesc.json | {'abc':'def'} | 2022-11-09 09:47:06 |
+--------------+---------------+---------------------+
```

Checking the local FS:

```bash
$ cat conf/descriptors/newdesc.json 
{'abc':'def'}
```

Deleting provider with curl:

```bash
$ curl -X DELETE -H "Content-Type: application/json" -v -k -u admin:admin-password  https://localhost:8443/gateway/admin/api/v1/providerconfig/newprov.json
```

Checking the DB and FS:

```bash
$ ls conf/shared-providers/

mysql> select * from knox_providers;
Empty set (0.00 sec)
```

Deleting descriptor with curl:

```bash
$ curl -X DELETE -H "Content-Type: application/json" -v -k -u admin:admin-password  https://localhost:8443/gateway/admin/api/v1/descriptors/newdesc.json
```

Checking the DB and FS:

```bash
$ cat conf/descriptors/newdesc.json 

mysql> select * from knox_descriptors;
Empty set (0.00 sec)
```

Repeated the same tests using the Admin UI.


gateway-server/src/main/java/org/apache/knox/gateway/topology/monitor/DefaultConfigurationMonitorProvider.java
```diff
@@ -16,16 +16,60 @@
...
+import org.apache.knox.gateway.services.security.AliasService;
+import org.apache.knox.gateway.services.security.AliasServiceException;
+import org.apache.knox.gateway.topology.monitor.db.DbRemoteConfigurationMonitor;
+import org.apache.knox.gateway.topology.monitor.db.LocalDirectory;
+import org.apache.knox.gateway.topology.monitor.db.RemoteConfigDatabase;
+import org.apache.knox.gateway.util.JDBCUtils;
 
 
-public class DefaultConfigurationMonitorProvider implements RemoteConfigurationMonitorProvider {
+public class DefaultConfigurationMonitorProvider extends AbstractServiceFactory {
```
smolnar82, line 38: I think this should be called a service factory from now on: `RemoteConfigurationMonitorServiceFactory`.

gateway-spi/src/main/java/org/apache/knox/gateway/topology/monitor/RemoteConfigurationMonitor.java
```diff
@@ -16,14 +16,11 @@
...
-public interface RemoteConfigurationMonitor {
-
-    void start() throws Exception;
-
-    void stop() throws Exception;
-
-    RemoteConfigurationRegistryClient getClient();
+import org.apache.knox.gateway.services.Service;
 
+public interface RemoteConfigurationMonitor extends Service {
```
smolnar82, line 21: This should be called a service to align with current naming conventions: `RemoteConfigurationMonitorService`

gateway-server/src/main/java/org/apache/knox/gateway/topology/monitor/DefaultConfigurationMonitorProvider.java
```diff
@@ -16,16 +16,60 @@
...
+    protected RemoteConfigurationMonitor createService(GatewayServices gatewayServices,
+                                    ServiceType serviceType,
+                                    GatewayConfig gatewayConfig,
+                                    Map<String, String> options,
+                                    String implementation) throws ServiceLifecycleException {
+        RemoteConfigurationMonitor service = null;
+        if (matchesImplementation(implementation, ZkRemoteConfigurationMonitor.class)) {
+            service = new ZkRemoteConfigurationMonitor(gatewayConfig, gatewayServices.getService(ServiceType.REMOTE_REGISTRY_CLIENT_SERVICE));
+        } else if (matchesImplementation(implementation, DbRemoteConfigurationMonitor.class)) {
+            service = createDbBasedMonitor(gatewayConfig, gatewayServices.getService(ServiceType.ALIAS_SERVICE));
```
smolnar82, line 49: The abstract class has a protected method called `getAliasService` for the same; please re-use that.

gateway-server/src/main/java/org/apache/knox/gateway/topology/monitor/db/DbRemoteConfigurationMonitor.java
```diff
@@ -0,0 +1,144 @@
...
+  @Override
+  public boolean deleteDescriptor(String name) {
+    return db.deleteDescriptor(name);
+  }
+
+  public void sync() {
+    try {
+      syncLocalWithRemote(db.selectProviders(), providersDir);
+      syncLocalWithRemote(db.selectDescriptors(), descriptorsDir);
+      lastSyncTime = Instant.now();
```
smolnar82, line 92: You may add a DEBUG-level log message with the fact that sync was done.

gateway-server/src/main/java/org/apache/knox/gateway/topology/monitor/db/DbRemoteConfigurationMonitor.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @Override
+  public void stop() throws ServiceLifecycleException {
+    executor.shutdown();
+  }
+
+  @Override
+  public boolean createProvider(String name, String content) {
+    return db.putProvider(name, content);
```
smolnar82, line 70: DEBUG-level log about creating the provider in the DB.

gateway-server/src/main/java/org/apache/knox/gateway/topology/monitor/db/DbRemoteConfigurationMonitor.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @Override
+  public boolean createDescriptor(String name, String content) {
+    return db.putDescriptor(name, content);
```
smolnar82, line 75: DEBUG-level log about creating the descriptor in the DB.

gateway-server/src/main/java/org/apache/knox/gateway/topology/monitor/db/DbRemoteConfigurationMonitor.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @Override
+  public boolean deleteProvider(String name) {
+    return db.deleteProvider(name);
```
smolnar82, line 80: DEBUG-level log about removing the provider from the DB.

gateway-server/src/main/java/org/apache/knox/gateway/topology/monitor/db/DbRemoteConfigurationMonitor.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @Override
+  public boolean deleteDescriptor(String name) {
+    return db.deleteDescriptor(name);
```
smolnar82, line 85: DEBUG-level log about removing a descriptor from the DB.

smolnar82 (changes requested):

smolnar82 (approved):
LGTM

PR #674, opened 2022-11-14
KNOX-2841 Oozie "root" rewrite rule's pattern is too open

dionusos:

Oozie's "root" rewrite rule must be more strict making it not to match to HDFS paths like "hdfs://namespace/oozie/libraries"

## How was this patch tested?

I did manual test steps in my environment:
 * I run an Oozie job using Hue having a workflow property: prop1=hdfs://namespace/oozie/libraries
 * checked the value is broken
 * then I edited /var/lib/knox/gateway/data/services/oozie/5.0.0/rewrite.xml to apply the change
 * then restarted Knox
 * then waited two minutes
 * then started the same Oozie workflow via Hue
then I experienced on the issue is gone. I did revert the change, tun the whole again then the issue came again. Then repeated the whole with the fix and the issue has gone again.

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/oozie/5.0.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/oozie/5.0.0/rewrite.xml
@@ -16,7 +16,7 @@
 -->
 <rules>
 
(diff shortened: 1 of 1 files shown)


moresandeep (approved):
Looks good, thanks @dionusos!

PR #680, opened 2022-11-17
KNOX-2844 - Remote config monitor should not automatically delete local files if they're missing from the DB

zeroflag:

[KNOX-2835](https://issues.apache.org/jira/browse/KNOX-2835) introduced an unwanted side effect after enabling the SQL based topology monitor.

The intent of  synchronization which was added in KNOX-2835 is:
 * If a descriptor/provider was deleted from the DB the monitor should delete the descriptor/provider from the local file system
 
But what really happens, implementation wise is that:

* If a descriptor/provider exists on the file system but doesn't exist in the DB the topology monitor is going to delete it.

These two are not exactly the same. If a user creates a descriptor/provider on the file system, it will be deleted by the topology monitor even though it never existed in the DB. The Zookeeper implementation uses onDeleteEvents which are only triggered when something was previously added but later it was deleted. But the SQL based topology uses polling.

So if one enable this feature with an empty DB, all of the descriptors/providers are going to be deleted.

To address this issue, this patch introduces logical deletes. Deleting a descriptor/provider only sets the delete=true column in the DB. The monitor only deletes a local file if a corresponding record exists in the DB and the delete column is set to true. If something never existed in the DB it'll never be deleted.

The logically deleted records are eventually cleared up to.

## How was this patch tested?

```xml
<property>
    <name>gateway.service.remoteconfigurationmonitor.impl</name>
    <value>org.apache.knox.gateway.topology.monitor.db.DbRemoteConfigurationMonitorService</value>
</property>
<property>
    <value>mysql</value>
</property>
<property>
    <value>jdbc:mysql://root:root@localhost:3306/knox</value>
</property>
```

Tested manually with 3 instances of knox. I created a custom descriptor + provider via the admin UI and checked the file content on each host. I deleted a descriptor/provider and checked if the file disappeared from the hosts. I checked the DB content to see the delete=true flag. I waited until the record was physically deleted. 
I checked that existing topologies were not effected by the deletes.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayMessages.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayMessages.java
@@ -767,4 +767,12 @@ public interface GatewayMessages {
   @Message(level = MessageLevel.DEBUG,
           text = "Creating local {0} with name {1}")
   void creatingLocalDescriptorProvider(String type, String name);
+
+  @Message(level = MessageLevel.DEBUG,
+          text = "Cleaning remote configuration db. Deleting logically deleted rows which are older than {0} seconds.")
+  void cleaningRemoteConfigTables(int cleanUpPeriodSeconds);
+
+  @Message(level = MessageLevel.INFO,
+          text = "Initializing remote configuration db. Sync interval={0} seconds. Clean up interval={1} seconds.")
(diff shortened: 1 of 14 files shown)


smolnar82 (approved):
LGTM

PR #687, opened 2022-12-01
KNOX-2848: Prevent overwriting generated descriptors/providers

zeroflag:

When creating a new descriptor or provider config, knox should check if there is an existing read-only descriptor/provider with the same name.

## How was this patch tested?

Using this config:

```xml
<property>
        <name>gateway.read.only.override.topologies</name>
        <value>admin,manager,xxx</value>    
    </property>
   <property>
	<name>gateway.read.only.override.providers</name>
	<value>prov1,prov2</value>
   </property>
```

Checking descriptor creation (negative case):

Screenshot 2022-12-02 at 14 41 56

Plus positive case.

### DIFF
--- a/gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/TopologiesResource.java
+++ b/gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/TopologiesResource.java
@@ -361,6 +361,14 @@ public class TopologiesResource {
     // extension.
     String filename = isUpdate ? existing.getName() : getFileNameForResource(name, headers);
 
+    GatewayConfig config =
+            (GatewayConfig) request.getServletContext().getAttribute(GatewayConfig.GATEWAY_CONFIG_ATTRIBUTE);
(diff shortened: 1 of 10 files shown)


moresandeep (approved):
Looks good.

gateway-service-admin/src/main/java/org/apache/knox/gateway/service/admin/TopologiesResource.java
```diff
@@ -424,6 +429,14 @@ public Response uploadSimpleDescriptor(@PathParam("name") String name,
     return response;
   }
 
+  private boolean existingGeneratedTopology(String fileName, TopologyService topologyService, GatewayConfig config) {
+    for (org.apache.knox.gateway.topology.Topology topology : topologyService.getTopologies()) {
```
smolnar82, line 433: I do not see where we check if the given topology is `readOnly`. This implementation will prevent creating/overwriting all topologies, won't it?

PR #688, opened 2022-12-02
KNOX-2850 - Bumped shiro, jetty, postgresql, hadoop, woodstox dependencies

MrtnBalazs:

The following dependencies were upgraded:
- hadoop: `3.2.3` -> `3.2.4`
- jetty: `9.4.45.v20220203` -> `9.4.48.v20220622`
- postgresql: `42.3.3` -> `42.4.1`
- woodstox: `6.1.1` -> `6.4.0`
- shiro: `1.8.0` -> `1.10.0`

## How was this patch tested?

I have configured `postgresql` for jdbc token state service.
I have created test topology on admin ui:
```
<topology>
   <uri>https://localhost:8443/gateway/test</uri>
   <name>test</name>
   <timestamp>1669992953127</timestamp>
   <generated>true</generated>
   <redeployTime>0</redeployTime>
   <gateway>
      <provider>
         <role>federation</role>
         <name>JWTProvider</name>
         <enabled>true</enabled>
         <param>
            <name>knox.token.exp.server-managed</name>
            <value>true</value>
         </param>
      </provider>
   </gateway>
   <service>
      <role>HIVE</role>
      <url>http://www.blankwebsite.com/</url>
   </service>
</topology>
```
Generated normal and impersonated token.
Tested both of them with curl request, disabled/enabled, revoked them.

I have also done some manual testing using the knox ui  pages:

Token-management:
* generated normal token
* generated impersonated token
* disabled impersonation
* set server-managed to false
* disabled and revoked the created tokens

Token-generation:
* set server-managed to true and false
* set impersonation to true and false
* set lifespan.input.enabled to true and false
* tried to generate token over the max lifespan limit
* generated tokens until reached token limit
* authenticated with a generated token

Homepage:
* Tried the links
* Mouse on service name -> description shows up
* When UI services are clicked they navigate to UI URL
* When API services are clicked description shows up
* Dropdown menus open and close
* New topologies appear under the topologies menu

Admin-ui:
* Created, deleted, modified service definitions
* Created, deleted, modified provider configuration
* Created, deleted, modified descriptor
* Copied, deleted, modified topology

WebShell-UI:
* Configured webshell and link appeared on homepage
* Signed in to knox with the user tom (who is a user on my local computer) and was able to connect to my computer

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -231,7 +231,7 @@
         <jee-pac4j.version>5.0.0</jee-pac4j.version>
         <jericho-html.version>3.4</jericho-html.version>
         <jersey.version>2.6</jersey.version>
-        <jetty.version>9.4.45.v20220203</jetty.version>
+        <jetty.version>9.4.48.v20220622</jetty.version>
         <jline.version>2.14.6</jline.version>
         <jna.version>5.6.0</jna.version>
         <joda-time.version>2.10.8</joda-time.version>
@@ -255,14 +255,14 @@
         <okhttp.version>2.7.5</okhttp.version>
         <opensaml.version>3.4.5</opensaml.version>
         <pac4j.version>4.5.6</pac4j.version>
-        <postgresql.version>42.3.3</postgresql.version>
+        <postgresql.version>42.4.1</postgresql.version>
         <mysql.version>8.0.28</mysql.version>
         <protobuf.version>3.16.1</protobuf.version>
         <powermock.version>2.0.9</powermock.version>
         <purejavacomm.version>0.0.11.1</purejavacomm.version>
         <pty4j.version>0.11.4</pty4j.version>
         <rest-assured.version>4.3.3</rest-assured.version>
-        <shiro.version>1.8.0</shiro.version>
+        <shiro.version>1.10.0</shiro.version>
         <shrinkwrap.version>1.2.6</shrinkwrap.version>
         <shrinkwrap.descriptors.version>2.0.0</shrinkwrap.descriptors.version>
         <slf4j.version>1.7.30</slf4j.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -1565,6 +1565,10 @@
                         <groupId>org.slf4j</groupId>
                         <artifactId>slf4j-log4j12</artifactId>
                     </exclusion>
+                    <exclusion>
```
smolnar82, line 1568: Could you please elaborate on why this exclusion is needed?

MrtnBalazs: I got this error message after building the project:
```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/mbalazs/projects/test/bin/../dep/slf4j-reload4j-1.7.35.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/mbalazs/projects/test/bin/../dep/log4j-slf4j-impl-2.17.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.impl.Reload4jLoggerFactory]
log4j:WARN No appenders could be found for logger (org.apache.hadoop.metrics2.lib.MutableMetricsFactory).
log4j:WARN Please initialize the log4j system properly.
```
I checked with `mvn dependency:tree` and the hadoop-auth and hadoop-common dependency pulled in the new `slf4j`.

smolnar82: @zeroflag - Does this exclusion seem OK from logging perspective?

zeroflag: The `slf4j-reload4j` is the same as `slf4j-log4j12` but for reload4j (which is a up to date version of log4j one). Given that hadoop or other components might use slf4j for logging instead log4j (what knox uses) we need some kind of bridging library between those two. But this is for log4j1. So slf4j knows about how to log using Log4j/reload4j. But knox uses log4j2. So I don't think this is needed. I think the exclusion is fine.

smolnar82 (approved):
LGTM

zeroflag (approved):

PR #706, opened 2023-01-03
KNOX-2860 - Build the knox-webshell-ui project the same way as we build other UIs

smolnar82:

I modified the Angular settings within `knox-webshell-ui` to conform to the same configuration we have in other UI projects.

## How was this patch tested?

Built the entire Knox deliverable locally and asked @zeroflag to do the same with my patch before I opened this PR.
After that I tested that WebShell is actually working.

### DIFF
--- a/knox-webshell-ui/angular.json
+++ b/knox-webshell-ui/angular.json
@@ -54,18 +54,8 @@
           },
           "configurations": {
             "production": {
-              "budgets": [
-                {
-                  "type": "initial",
-                  "maximumWarning": "500kb",
-                  "maximumError": "1mb"
-                },
-                {
-                  "type": "anyComponentStyle",
-                  "maximumWarning": "2kb",
-                  "maximumError": "4kb"
-                }
-              ],
+              "buildOptimizer": false,
+              "aot": false,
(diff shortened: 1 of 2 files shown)


moresandeep (approved):
Looks good, thanks!

knox-webshell-ui/angular.json
```diff
@@ -54,18 +54,8 @@
...
-                  "maximumError": "1mb"
-                },
-                {
-                  "type": "anyComponentStyle",
-                  "maximumWarning": "2kb",
-                  "maximumError": "4kb"
-                }
-              ],
+              "buildOptimizer": false,
+              "aot": false,
```
zeroflag, line 58: How does disabling ahead of time compilation affects the runtime performance?

smolnar82: We have the same config in AdminUI and TokenManagementUI and there were no complaints so far about their performance. It's true that we may want AOT to be turned on, but I think it has to be in the scope of a different JIRA; right now fixing the build is more important.
What do you think?

zeroflag (approved):

PR #713, opened 2023-01-06
KNOX-2859 - Token Management UI improvements

smolnar82:

Replaced the old Angular2 Datatable with a more modern Material Table implementation and configured filtering, sorting, and pagination on both tables on the Token Management UI.

## How was this patch tested?

Manual testing with a huge amount of tokens created (2k) to ensure the new implementation still performs.

EnhancedTokenManagementUI

### DIFF
--- a/knox-token-management-ui/token-management/app/token.management.component.html
+++ b/knox-token-management-ui/token-management/app/token.management.component.html
@@ -20,90 +20,116 @@
             <span class="glyphicon glyphicon-refresh"></span>
         </button>
     </div>
+
     <div class="table-responsive" style="width:100%; overflow: auto; overflow-y: scroll; padding: 10px 0px 0px 0px;">
         <label>My Knox Tokens</label>
-        <table class="table table-hover" [mfData]="knoxTokens" #tokens="mfDataTable" [mfRowsOnPage]="10">
-            <thead>
-            <tr>
-                <th>Token ID</th>
-                <th>Issued</th>
-                <th>Expires</th>
-                <th>Comment</th>
-                <th>Additional Metadata</th>
-                <th>Actions</th>
-            </tr>
-            </thead>
-            <tbody>
-            <tr *ngFor="let knoxToken of tokens.data">
-                <td>{{knoxToken.tokenId}}</td>
-                <td>{{formatDateTime(knoxToken.issueTimeLong)}}</td>
-                <td *ngIf="!isTokenExpired(knoxToken.expirationLong)" style="color: green">{{formatDateTime(knoxToken.expirationLong)}}</td>
-                <td *ngIf="isTokenExpired(knoxToken.expirationLong)" style="color: red">{{formatDateTime(knoxToken.expirationLong)}}</td>
-                <td>{{knoxToken.metadata.comment}}</td>
-                <td>
-                  <ul>
-                    <li *ngFor="let metadata of getCustomMetadataArray(knoxToken)">
-                      {{metadata[0]}} = {{metadata[1]}}
-                    </li>
-                  </ul>
-                </td>
-                <td>
+
+        <mat-form-field [floatLabel]="always" appearance="fill" #search>
+            <mat-label>Search by Token ID, Comment or Metadata...</mat-label>
+            <input matInput (keyup)="applyFilter(false, $event.target.value)">
+        </mat-form-field>
+
+        <mat-table [dataSource]="knoxTokens" matSort #ownSort="matSort">
+            <ng-container matColumnDef="tokenId">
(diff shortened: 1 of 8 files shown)


knox-token-management-ui/token-management/app/token.management.component.ts
```diff
@@ -57,13 +117,29 @@ export class TokenManagementComponent implements OnInit {
...
     }
 
     fetchKnoxTokens(impersonated: boolean): void {
         this.tokenManagementService.getKnoxTokens(this.userName, impersonated)
-            .then(tokens => impersonated ? this.doAsKnoxTokens = tokens : this.knoxTokens = tokens);
+            .then(tokens => this.populateTokens(impersonated, tokens));
+    }
+
+    populateTokens(impersonated: boolean, tokens: KnoxToken[]) {
+	    if (impersonated) {
```
zeroflag, line 130: nit: missing indentation

knox-token-management-ui/token-management/app/token.management.component.ts
```diff
@@ -42,6 +55,53 @@ export class TokenManagementComponent implements OnInit {
...
+
+        this.knoxTokens.sortingDataAccessor = (item, property) => {
+           switch(property) {
+             case 'metadata.comment': return item.metadata.comment;
+             default: return item[property];
+           }
+        };
+
+        this.doAsKnoxTokens.sortingDataAccessor = (item, property) => {
+	       let normalizedPropertyName = property.replace('impersonation.', '');
```
zeroflag, line 98: nit: wrong indentation

knox-token-management-ui/token-management/app/token.management.component.ts
smolnar82: Fixed.

knox-token-management-ui/token-management/app/token.management.component.ts
smolnar82: Fixed.

zeroflag (approved):

knox-token-management-ui/token-management/app/token.management.component.ts
```diff
@@ -57,13 +117,29 @@ export class TokenManagementComponent implements OnInit {
     }
 
     fetchAllKnoxTokens(): void {
-        this.fetchKnoxTokens(true);
         this.fetchKnoxTokens(false);
+        this.fetchKnoxTokens(true);
```
pzampino, line 121: Is the order important here? Just trying to understand the change.

knox-token-management-ui/token-management/app/token.management.component.html
```diff
@@ -20,90 +20,116 @@
...
-            </tbody>
-		    <tfoot>
-		    <tr>
-		        <td colspan="6">
-		            <mfBootstrapPaginator [rowsOnPageSet]="[5,10,15]"></mfBootstrapPaginator>
-		        </td>
-		    </tr>
-		    </tfoot>
-        </table>
+    <div class="table-responsive" style="width:100%; overflow: auto; overflow-y: scroll; padding: 10px 0px 0px 0px;">
```
pzampino, line 83: Why are tokens acquired with impersonation presented in a distinct table from "normal" tokens? Couldn't it be a single table? The relationship between the tables isn't entirely clear to me, and you can only act on rows in the My Knox Tokens table.

pzampino (commented):
This is a really nice improvement, especially if it performs as well with high token volume as your test have indicated. I just have some questions.

knox-token-management-ui/token-management/app/token.management.component.ts
smolnar82: Nope, the order is irrelevant. I just made to change to reflect the order of the tables on te GUI: my tokens first then the impersonation tokens.

knox-token-management-ui/token-management/app/token.management.component.html
smolnar82: In theory, they could be, but I made this decision when token impersonation was introduced (so this is not a new thing here). This is easier to understand and the UX is better this way (IMO, at least)
The relationship is straightforward:
- in the first table we list all tokens generated by the logged-in user for himself/herself: `userName=logged-in user`
- in the second one we list all tokens generated by the logged-in user on behalf of other users: `username=otherUser; createdBy=logged-in user`

The same is described [here](https://knox.apache.org/books/knox-2-0-0/user-guide.html#Token+Management).

pzampino: I recognize that this is not directly related to this PR.
I think my confusion may be due to the inability to disable/revoke the tokens in the second table, but that is a separate topic.

pzampino (approved):
LGTM

PR #717, opened 2023-01-24
KNOX-2864 - TLS cipher suites and protocols are configured for CM service discovery

smolnar82:

There are 2 changes in this PR:
- TLS cipher suites and protocols are now customizable in CM service discovery:
  - end-users can define the TLS protocol(s) and cipher suite(s) to use while creating a secure connection to a CM instance
  - if none is defined, Knox will pick up the relevant supported values from Java's own `java.security` thru the acquired SSL context.
- Added a new method to indicate the TLS protocols to be included when creating SSL context in the embedded Jetty server. Prior to this change, end-users could only tell what to exclude.

## How was this patch tested?

Unit testing and running service discovery in a secure cluster with updated SSL configs.

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/DiscoveryApiClient.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/DiscoveryApiClient.java
@@ -23,6 +23,7 @@ import static org.apache.knox.gateway.topology.discovery.cm.ClouderaManagerServi
 import java.security.KeyStore;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Set;
 
 import javax.net.ssl.SSLContext;
 import javax.security.auth.Subject;
@@ -55,10 +56,10 @@ public class DiscoveryApiClient extends ApiClient {
 
   private ServiceDiscoveryConfig config;
 
-  public DiscoveryApiClient(ServiceDiscoveryConfig discoveryConfig, AliasService aliasService,
+  public DiscoveryApiClient(GatewayConfig gatewayConfig, ServiceDiscoveryConfig discoveryConfig, AliasService aliasService,
       KeyStore trustStore) {
     this.config = discoveryConfig;
-    configure(aliasService, trustStore);
+    configure(gatewayConfig, aliasService, trustStore);
   }
 
   ServiceDiscoveryConfig getConfig() {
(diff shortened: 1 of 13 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/DiscoveryApiClient.java
```diff
@@ -157,12 +161,26 @@ private String getUsername() {
...
   }
 
-  private void configureTruststore(KeystoreService keystoreService) {
-    SSLContext truststoreSSLContext = TruststoreSSLContextUtils.getTruststoreSSLContext(keystoreService);
+  private void configureSsl(GatewayConfig gatewayConfig, KeyStore trustStore) {
+    final SSLContext truststoreSSLContext = TruststoreSSLContextUtils.getTruststoreSSLContext(trustStore);
+
     if (truststoreSSLContext != null) {
+      final ConnectionSpec.Builder connectionSpecBuilder = new ConnectionSpec.Builder(ConnectionSpec.MODERN_TLS);
+      if (gatewayConfig.getIncludedSSLCiphers().isEmpty()) {
```
zeroflag, line 169: Can this method `getIncludedSSLCiphers()` return null if no `ssl.include.ciphers` is configured?

smolnar82: Nope; it cannot be null (either an empty collection or one with the declared items).

zeroflag: I might miss something but I don't see how is this enforced. What if there is no `SSL_INCLUDE_CIPHERS` in the config?

```
  @Override
  public List<String> getIncludedSSLCiphers() {
    List<String> list = null;
    String value = get(SSL_INCLUDE_CIPHERS);
    if (value != null && !value.isEmpty() && !"none".equalsIgnoreCase(value.trim())) {
      list = Arrays.asList(value.trim().split("\\s*,\\s*"));
    }
    return list;
  }
```

smolnar82: Ohh...and you are right. I added a new method - `getIncludedSSLProtocols` - I misread and thought your question was about that one. Let me submit a new PS soon.

smolnar82: Fixed.

zeroflag (approved):

PR #718, opened 2023-01-25
KNOX-2841 Oozie "root" rewrite rule's pattern is too open

dionusos:

We had to change the Oozie rewrite rules more strict not allowing them to mach to any protocol but only http and https. Besides this change additional service routing configuration have to be added not to break the Oozie UI usability.

## How was this patch tested?

On a Hadoop cluster with Knox installed I edited /var/lib/knox/gateway/data/services/oozie/5.0.0/rewrite.xml and /var/lib/knox/gateway/data/services/oozie/5.0.0/service.xml files to have the final content. After that I deleted the Knox topology deployments then restarted Knox.
After that I re-submitted the Oozie workflow which I had issues with earlier. The job submission was successful, the Oozie job executed without any error, the properties were set to the desired values and I was able to see the expected values on the Oozie UI too via Knox.


smolnar82 (approved):
LGTM

moresandeep (approved):
Looks good, thanks @dionusos!

PR #719, opened 2023-01-26
KNOX-2865 -  Accessing parameters of a x-www-form-urlencoded request consumes the request body

zeroflag:

In section SRV.3.1.1 of Servlet spec states that any access to request parameters of an "x-www-form-urlencoded" request (e.g. using HttpServletRequest#getParameter()) can trigger the early consumption of the request InputStream.

For example:

```java
request.getParameter("doAs");
IOUtils.toString(request.getInputStream()); // <= This will be empty because getParameter() already consumed the stream
```

This causes knox to dispatch an empty request body to the service (HadoopAuthFilter always check the doAs param).

How is it fixed?

Knox as a proxy doesn't care about the request params coming from the request body. This patch introduces a wrapper class that only gets the parameters of the  "x-www-form-urlencoded" requests from the query string, ignoring the body altogether.

## How was this patch tested?

Tested it manually with curl.

```bash
$ curl -u admin:admin-password -H "X-XSRF-Header: whatever" -k -X POST -d "a=1&b=2" "https://localhost:8443/gateway/sandbox/hive?a=3"
```

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayFilter.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayFilter.java
@@ -184,7 +184,11 @@ public class GatewayFilter implements Filter {
       Chain chain = match.getValue();
       servletRequest.setAttribute( AbstractGatewayFilter.TARGET_SERVICE_ROLE, chain.getResourceRole() );
       try {
-        chain.doFilter( servletRequest, servletResponse );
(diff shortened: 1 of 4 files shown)


smolnar82 (approved):
LGTM. Excellent work, Attila!

pzampino (approved):
LGTM

moresandeep (approved):
LGTM

PR #721, opened 2023-02-02
KNOX-2871 - Refine should perform discovery check

zeroflag:

The `discovery-type` is defined set to default `"ClouderaManager"` if it's missing, so a discovery is always triggered even if the `discovery-address` is missing.

This patch adds an additional check to see if the url is set or not.

## How was this patch tested?

descriptor: test.json

```json
{
  "provider-config-ref":"ldap",
  "discovery-type": "ClouderaManager",
  "discovery-address": "http://localhost/test",
  "cluster": "test",
  "services":[
    {"name":"NAMENODE"},
    {"name":"WEBHDFS"}
  ]
}
```

provider: ldap.json

```json
{
  "providers": [
    {
      "role": "authentication",
      "name": "ShiroProvider",
      "enabled": "true",
      "params": {
        "sessionTimeout": "20",
        "main.ldapRealm": "org.apache.knox.gateway.shirorealm.KnoxLdapRealm",
        "main.ldapContextFactory": "org.apache.knox.gateway.shirorealm.KnoxLdapContextFactory",
        "main.ldapRealm.contextFactory": "$ldapContextFactory",
        "main.ldapRealm.userDnTemplate": "uid={0},ou=people,dc=hadoop,dc=apache,dc=org",
        "main.ldapRealm.contextFactory.url": "ldap://localhost:33389",
        "main.ldapRealm.contextFactory.authenticationMechanism": "simple",
        "urls./**": "authcBasic"
      }
    }
  ]
}
```

Checked that discovery was only triggered if both discovery address + type was defined.

### DIFF
--- a/gateway-test/src/test/java/org/apache/knox/gateway/SimpleDescriptorHandlerFuncTest.java
+++ b/gateway-test/src/test/java/org/apache/knox/gateway/SimpleDescriptorHandlerFuncTest.java
@@ -276,7 +276,7 @@ public class SimpleDescriptorHandlerFuncTest {
 
   private static final class NoOpServiceDiscovery implements ServiceDiscovery {
     static final String TYPE = "NO_OP";
-    static int discoveryCalled = 0;
+    static int discoveryCalled;
 
     @Override
     public String getType() {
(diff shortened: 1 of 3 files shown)


smolnar82 (approved):
LGTM

PR #722, opened 2023-02-02
KNOX-2869 - Handling the case when previously persisted CM discovery config is empty

smolnar82:

As described in the corresponding JIRA, Knox should handle the situation when a previously stored CM discovery config file exists, but it's empty. This may prevent the Knox Gateway from starting up.

## How was this patch tested?

Added new JUnit test cases to cover this event and ran manual testing:
- Stopped the Knox GW in a cluster where I had descriptors with CM service discovery enabled
- truncted the `$KNOX_GW_DATA_DIR/cm_clusters/xyz.conf` file
- restarted the Knox Gateway

Prior to my changes, the Knox GW did not start; After my fix, everything is working as expected.

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
@@ -142,6 +142,12 @@ public interface ClouderaManagerServiceDiscoveryMessages {
            text = "Terminating monitoring of {1} @ {0} for configuration changes because there are no referencing descriptors.")
   void stoppingConfigMonitoring(String discoverySource, String clusterName);
 
+  @Message(level = MessageLevel.WARN, text = "Missing property in previously saved service discovery configuration {0}")
+  void missingServiceDiscoveryConfigProperty(String propertyName);
+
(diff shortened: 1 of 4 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
```diff
@@ -142,6 +142,12 @@ void failedToInstantiateJAASConfigurationFileImplementation(String implementatio
            text = "Terminating monitoring of {1} @ {0} for configuration changes because there are no referencing descriptors.")
   void stoppingConfigMonitoring(String discoverySource, String clusterName);
 
+  @Message(level = MessageLevel.WARN, text = "Missing property in previously saved service discovery configuration {0}")
+  void missingServiceDiscoveryConfigProperty(String propertyName);
+
+  @Message(level = MessageLevel.DEBUG, text = "There is no cluster configuration to check yet.")
```
zeroflag, line 148: There aren't any format string parameters in the message, but the method has 2 arguments.

smolnar82: Nice catch; fixed!

zeroflag (approved):

PR #729, opened 2023-02-07
KNOX-2873 - Upgrade curator version to 5.4.0 and zookeeper to 3.8.1

zeroflag:

Upgrading curator and zookeeper.

## How was this patch tested?

```bash
$ mvn test -Dtest=ZookeeperTokenStateServiceTest
```


smolnar82:
@zeroflag - the changes themselves look good to me. There is only one thing I'd like to clarify: what's the driver of this change? The old versions are affected by some CVEs?

smolnar82 (approved):

PR #733, opened 2023-02-21
KNOX-2883 Missing attach logging context from HadoopAuthFilter

zeroflag:

`auditService.attachContext(context);` should be called so that the logging information is stored in the thread context.

## How was this patch tested?

Tested manually.


smolnar82 (approved):
LGTM

PR #734, opened 2023-02-28
KNOX-2884 - Skip generating read-only descriptor and provider generation from hxr

zeroflag:

Descriptors and providers which are managed externally and marked as read only should not be regenerated by the hadoop resource monitor.

## How was this patch tested?

Tested the following scenarios:

Topology with static URL.

```
	name  = custom
	value = providerConfigRef=pam#WEBHDFS:url=http://dummy
```

Topology with discovered service URL.

```
	name  = custom
	value = providerConfigRef=pam#discoveryAddress=http://disco.address:7180#discoveryType=ClouderaManager#cluster=Cluster 1#NAMENODE#WEBHDFS
```

Mixing static URL with discovery.

```
	name  = custom
	value = providerConfigRef=pam#discoveryAddress=http://disco.address:7180#discoveryType=ClouderaManager#cluster=Cluster 1#NAMENODE#WEBHDFS:url=dummy	
```

I checked that the topology ( /var/lib/knox/gateway/conf/topologies/custom.xml ) was only created if it was not added to `gateway.read.only.override.topologies`.

I checked the same for shared providers.

Single provider:

```
	name  = providerConfigs:mypam
	value = role=authentication#authentication.name=ShiroProvider	
```

Multiple providers:

```
	name  = providerConfigs:mypam,mypam2
	role=authentication#authentication.name=ShiroProvider
```

I added `mypam` to `gateway.read.only.override.providers` so that only mypam2 was regenerated.


smolnar82 (changes requested):
Thanks, @zeroflag for submitting this cool change.
As discussed offline, I think it'd be better if this logic was added in the HadoopXmlResourceParser for the following two reasons:
- it's easy to add new unit test cases
- we may save some parser rounds

zeroflag:
> Thanks, @zeroflag for submitting this cool change. As discussed offline, I think it'd be better if this logic was added in the HadoopXmlResourceParser for the following two reasons:
> 
> * it's easy to add new unit test cases
> * we may save some parser rounds

Good points. I made the change.

smolnar82 (approved):
LGTM

PR #747, opened 2023-04-11
KNOX-2897 - Using Expect:100-continue wherever possible to upload big content

smolnar82:

This commit consists of the following changes:
- the new-style "Expect:100-continue" approach for PUT/POST requests is disabled by default
- the new-style "Expect:100-continue" approach is enabled for the ATLAS, ATLAS-API, HIVE, RANGER and RANGER-UI services
- make sure that this feature for the WEBHDFS service is turned off even in its own service definition file (so that if the default is changed back to true, we'll be safe)
- HTTP entity creation relies on the 'addExpect100Continue' boolean flag
- implemented the missing method overload in HdfsHttpClientDispatch
- KnoxBufferedHttpEntity shall not read content at entity creation time (content reading is postponed until dispatch time when the back-end service actually needs the data)

## How was this patch tested?

Manual testing on a live cluster with the affected services (ATLAS, HIVE, RANGER)

### DIFF
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/KnoxBufferedHttpEntity.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/KnoxBufferedHttpEntity.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/KnoxBufferedHttpEntity.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * A {@link BufferedHttpEntity} implementation with the following differences:
+ * <ul>
+ * <li>Reading the content of the given entity does not happen at instance creation time (i.e not in the constructor)
+ * <li>Content length calculation is not based on the buffer but a delegate to the wrapped entity
+ * </ul>
+ */
+public class KnoxBufferedHttpEntity extends HttpEntityWrapper {
+
+  private byte[] buffer;
```
risdenk, line 40: There are no bounds checks for copying to this buffer? Can someone just crash Knox by sending a really large content stream that is not repeatable?

I thought that was part of the theory behind why there was a replay buffer size - to avoid copying the entire stream into memory.

smolnar82:
We may want to reconsider this PR and create another implementation that handles large inputs in a streaming fashion. That is, another `HttpEntity` implementation (replacing the one suggested in this change), with
```
    @Override
    public boolean isRepeatable() {
        return false;
    }

    @Override
    public boolean isStreaming() {
        return true;
    }
```

PR #748, opened 2023-04-14
KNOX-2893 - Prevent generating topology from read only descriptor

zeroflag:

DescriptorMonitor should also skip generating topology when it's read only.

## How was this patch tested?

Manually.


smolnar82 (approved):
LGTM

PR #749, opened 2023-04-18
KNOX-2899 - Disable service-based discovery filter in Cloudera Manager service discovery

smolnar82:

Service-based filtering is temporarily disabled while discovering a CM cluster.

## How was this patch tested?

Manually tested using a large CM cluster with lots of services.


smolnar82:
For reviewers: I wanted to do the least intrusive change possible, that's why the private method and the `includedServices` parameter are still in place and in use. Since everything happens in memory, it's not a real performance issue.

zeroflag (approved):
LGTM, I made a followup jira for the permanent fix: KNOX-2900

PR #752, opened 2023-04-26
KNOX-2904 -  Mark the endpoint as failed when we failover AND request is non-idempotent

moresandeep:

Make sure we mark the endpoint as failed in case of non-idempotent request failures.

## How was this patch tested?
Make sure unit tests pass


smolnar82:
Hey @moresandeep !

Please forgive me for my dummy question(s), but there is something I'd like to clarify: in the current `failoverRequest` method, there are other actions in terms of URL-management such as `setupUrlHashLookup` and ` activeURL.set(outboundRequest.getURI().toString())`.
Aren't they required along with the new `markEndpointFailed`?

moresandeep:
> Hey @moresandeep !
> 
> Please forgive me for my dummy question(s), but there is something I'd like to clarify: in the current `failoverRequest` method, there are other actions in terms of URL-management such as `setupUrlHashLookup` and `activeURL.set(outboundRequest.getURI().toString())`. Aren't they required along with the new `markEndpointFailed`?

Thank you for checking @smolnar82! I was debating whether to include them since we did not actually failover. But, reading your comments and the code again, we should be including it! I will update it.

PR #753, opened 2023-04-27
KNOX-2905 - GatewayDescriptorImporter is not thread safe

zeroflag:

The gateway descriptor importer is not thread safe and it is called from two different threads.

## How was this patch tested?

* I verified that calling the importer from multiple threads cause various parser exceptions (see the Jira ticket for more details).
* Tested with multiple knox restarts + curl requests at the same time.


pzampino (approved):
LGTM

PR #754, opened 2023-04-28
KNOX-2906 - SSO whitelist checker should cut off path segments and query params

zeroflag:

Similarly to [KNOX-2827](https://issues.apache.org/jira/browse/KNOX-2827), we should cut off query params as well as path segments from the originalUrl before checking it against the whitelist.

## How was this patch tested?

Manually using this whitelist

```xml
    <property>
        <name>gateway.dispatch.whitelist</name>
        <value>^https?:\/\/(.*localhost)(?::[0-9]+)?(?:\/.*)?$</value>
    </property>
```


smolnar82 (approved):
LGTM

PR #756, opened 2023-05-24
KNOX-2898 - Reconsider the usage of sso.unauthenticated.path.list

zeroflag:

Unauthenticated path list doesn't work when hadoop-jwt cookie exists but it's invalid / expired.

## How was this patch tested?

```xml
<topology>
   <gateway>
      <provider>
         <role>federation</role>
         <name>SSOCookieProvider</name>
         <enabled>true</enabled>
         <param>
              <name>sso.unauthenticated.path.list</name>
              <value>/favicon.ico;test;unsafepath</value>
         </param>
      </provider>
      <provider>
            <name>Default</name>
            <enabled>true</enabled>
      </provider>

        <provider>
            <role>hostmap</role>
            <name>static</name>
            <enabled>true</enabled>
            <param>
                <name>localhost</name>
                <value>sandbox,sandbox.hortonworks.com</value>
            </param>
        </provider>
    </gateway>
    <service>
        <role>CM-UI</role>
        <url>https://www.google.com</url>
    </service>
</topology>
```

```bash
 curl --cookie "hadoop-jwt=invalid" -k https://localhost:8443/gateway/sandbox/favicon.ico
```

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/SSOCookieFederationFilter.java
@@ -134,17 +134,18 @@ public class SSOCookieFederationFilter extends AbstractJWTFilter {
     HttpServletRequest req = (HttpServletRequest) request;
     HttpServletResponse res = (HttpServletResponse) response;
 
+    /* check for unauthenticated paths to bypass */
+    if(AuthFilterUtils.doesRequestContainUnauthPath(unAuthenticatedPaths, request)) {
+      /* This path is configured as an unauthenticated path let the request through */
+      final Subject sub = new Subject();
+      sub.getPrincipals().add(new PrimaryPrincipal("anonymous"));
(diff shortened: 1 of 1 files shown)


moresandeep (commented):
The concern I have is that one can add any path to that list and bypass authentication. Feature or Bug? :)

zeroflag:
@moresandeep 
If you omit the `hadoop-jwt` completely you can still bypass the authentication, I thought that was the goal to bypass the autnetication. However if your token expires you'll be redirected to the login page when accessing an unauthenticated path, such as /favicon.

In other filters such as `HadoopAuthFilter` we check the unauthenticated path first and do an early return if the path is there.

smolnar82 (approved):
LGTM.
@moresandeep - Have you had the time to think about @zeroflag 's previous comment?

moresandeep (approved):
Thanks @zeroflag! looks good!

moresandeep:
> @moresandeep If you omit the `hadoop-jwt` completely you can still bypass the authentication, I thought that was the goal to bypass the autnetication. However if your token expires you'll be redirected to the login page when accessing an unauthenticated path, such as /favicon.
> 
> In other filters such as `HadoopAuthFilter` we check the unauthenticated path first and do an early return if the path is there.

I see, that makes sense. Thanks !

PR #757, opened 2023-05-31
KNOX-2912 - Don't fail over non idempotent requests unless it's a connect exception

zeroflag:

By default Knox doesn't fail over in case of non-idempotent requests. In case of a connect exception (connection refused, unknown host, etc), it is safe to failover since the connection hasn't been established yes and the request body was not consumed.

## How was this patch tested?

Testet it manually with the following config:

```
    <provider>
         <role>ha</role>
         <name>HaProvider</name>
         <enabled>true</enabled>
         <param>
            <name>HIVE</name>
            <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
         </param>
      </provider>
    </gateway>

    <service>
        <role>HIVE</role>
        <!-- <url>http://localhost:1700</url> -->
        <url>http://10.255.255.1:1234</url>
    </service>
```


smolnar82 (approved):
LGTM

PR #763, opened 2023-06-13
KNOX-2912 - Don't fail over non idempotent requests unless it's a connect exception - ConnectionTimeout

zeroflag:

ConnectionTimeout was missing from the original patch

## How was this patch tested?

```xml
    <provider>
         <role>ha</role>
         <name>HaProvider</name>
         <enabled>true</enabled>
         <param>
            <name>HIVE</name>
            <value>enabled=true;maxFailoverAttempts=3;failoverSleep=1000</value>
         </param>
      </provider>
    </gateway>

    <service>
        <role>HIVE</role>
        <url>http://google.com:81</url>
    </service>
```

### DIFF
--- a/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
+++ b/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
@@ -220,7 +220,7 @@ public class ConfigurableHADispatch extends ConfigurableDispatch {
     } catch ( IOException e ) {
       /* if non-idempotent requests are not allowed to failover, unless it's a connection error */
       if(!isConnectionError(e.getCause()) && isNonIdempotentAndNonIdempotentFailoverDisabled(outboundRequest)) {
-        LOG.cannotFailoverNonIdempotentRequest(outboundRequest.getMethod(), e.toString());
+        LOG.cannotFailoverNonIdempotentRequest(outboundRequest.getMethod(), e.getCause());
         /* mark endpoint as failed */
         markEndpointFailed(outboundRequest, inboundRequest);
         throw e;
--- a/gateway-service-webhdfs/src/main/java/org/apache/knox/gateway/hdfs/dispatch/AbstractHdfsHaDispatch.java
+++ b/gateway-service-webhdfs/src/main/java/org/apache/knox/gateway/hdfs/dispatch/AbstractHdfsHaDispatch.java
@@ -62,7 +62,7 @@ public abstract class AbstractHdfsHaDispatch extends ConfigurableHADispatch {
       } catch (StandbyException | SafeModeException | IOException e) {
         /* if non-idempotent requests are not allowed to failover */
         if(!failoverNonIdempotentRequestEnabled && nonIdempotentRequests.stream().anyMatch(outboundRequest.getMethod()::equalsIgnoreCase)) {
-          LOG.cannotFailoverNonIdempotentRequest(outboundRequest.getMethod(), e.toString());
+          LOG.cannotFailoverNonIdempotentRequest(outboundRequest.getMethod(), e.getCause());
           throw e;
         } else {
           printExceptionLogMessage(e, outboundRequest.getURI().toString());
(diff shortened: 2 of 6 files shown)


gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
```diff
@@ -220,7 +220,7 @@ protected void executeRequest(HttpUriRequest outboundRequest, HttpServletRequest
     } catch ( IOException e ) {
       /* if non-idempotent requests are not allowed to failover, unless it's a connection error */
       if(!isConnectionError(e.getCause()) && isNonIdempotentAndNonIdempotentFailoverDisabled(outboundRequest)) {
-        LOG.cannotFailoverNonIdempotentRequest(outboundRequest.getMethod(), e.toString());
+        LOG.cannotFailoverNonIdempotentRequest(outboundRequest.getMethod(), e.getCause());
```
moresandeep, line 223: In case cause is null this will throw NPE, e.toString() saves us from that.

gateway-service-webhdfs/src/main/java/org/apache/knox/gateway/hdfs/dispatch/AbstractHdfsHaDispatch.java
```diff
@@ -62,7 +62,7 @@ protected void executeRequest(HttpUriRequest outboundRequest, HttpServletRequest
       } catch (StandbyException | SafeModeException | IOException e) {
         /* if non-idempotent requests are not allowed to failover */
         if(!failoverNonIdempotentRequestEnabled && nonIdempotentRequests.stream().anyMatch(outboundRequest.getMethod()::equalsIgnoreCase)) {
-          LOG.cannotFailoverNonIdempotentRequest(outboundRequest.getMethod(), e.toString());
+          LOG.cannotFailoverNonIdempotentRequest(outboundRequest.getMethod(), e.getCause());
```
moresandeep, line 65: In case cause is null this will throw NPE, e.toString() saves us from that.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/ConfigurableHADispatch.java
zeroflag: Thanks for the review @moresandeep. I tried it with a null cause, but it did not cause an NPE, it just logged out null.

```
2023-06-13 15:16:05,205 8a432e79-6f55-4be0-8ef3-fcbe74773e19 ERROR knox.gateway (ConfigurableHADispatch.java:executeRequest(222)) - Request is non-idempotent POST, failover prevented, to allow non-idempotent requests to failover set failoverNonIdempotentRequestEnabled=true in HA config. Non connection related error: null
```

But this shouldn't happen because we always wrap the cause into an IOException, so the cause shouldn't be empty.

My problem with the `toString` was, that it didn't reveal the actual cause (which is used to make the decision whether to fail over or not).

Here is how the message looked like before (using the `toString`):

```
2023-06-12 15:09:28,865 30b4e772d8617006eca9b589eda16813 ERROR knox.gateway (ConfigurableHADispatch.java:executeRequest(223)) - Request is non-idempotent POST, failover prevented, to allow non-idempotent requests to failover set failoverNonIdempotentRequestEnabled=true in HA config. Non connection related error: java.io.IOException: Service connectivity error.
```

Basically the error message was always the same regardless the error.

From this message I can't see if the error happened because of a connection timeout or connection refused or any other reason.

Now it looks like:

```
2023-06-13 13:14:20,843 34dd13f0-324b-421e-b6e6-5947e0e38440 ERROR knox.gateway (ConfigurableHADispatch.java:executeRequest(223)) - Request is non-idempotent POST, failover prevented, to allow non-idempotent requests to failover set failoverNonIdempotentRequestEnabled=true in HA config. Non connection related error: javax.net.ssl.SSLException: java.net.SocketException: Socket closed
```

moresandeep: I see, thanks @zeroflag !

moresandeep (approved):
LGTM

PR #767, opened 2023-06-28
KNOX-2930 - Fix OM and SCM UIs to work in HA

dombizita:

Previously if there were multiple OMs/SCMs in a cluster on the Knox UI if you clicked on any of the OM/SCM UIs it was consistently showing one of the OMs/SCMs UI, because `$serviceUrl[OZONE-SCM]` was picking the first URL from the three in the `cdp-proxy.xml`.

In this fix I add a `?host={host}` to the ending of the URLs and used that `{host}` later when rewriting the requests, so it will go to the correct host. I added a separate route for the documentation to work in the UIs. 

## How was this patch tested?

I tested it in a cluster.


smolnar82 (approved):
LGTM
@zeroflag - Could you please check this out too?

zeroflag (approved):
LGTM

PR #768, opened 2023-06-28
KNOX-2887 - Add required changes to ozone-scm service definitions to handle DataNode WebUIs

dombizita:

In this change I added the datanode links correctly to the Ozone SCM UI and integrated the Datanode UI into Knox.

## How was this patch tested?

I tested it in a cluster and the datanode links on the SCM UI are redirecting to the correct Datanode UI and the content of the page is loaded correctly as well.

SCM UI when I am hovering over the datanode link it is showing the right URL (bottom of the page):
scm_dn_link

Datanode UI:
dn_ui


moresandeep (approved):

zeroflag (approved):
LGTM - 
Since #769 is merged which fixes the escaping problem,  could you open a jira about removing the workaround in the future?

PR #772, opened 2023-07-17
KNOX-2932 - Fix TLS Ciphers issue and add kerberos support to the docker image

moresandeep:
Currently docker image appears to be broken with TLS failures, this appears to be due to the underlying alpine base image. This update upgrades the alpine base image (which also pulls in few security fixes) and adds support for kerberos.

## How was this patch tested?

### DIFF
--- a/gateway-docker/src/main/resources/docker/Dockerfile
+++ b/gateway-docker/src/main/resources/docker/Dockerfile
@@ -13,11 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM openjdk:8-jre-alpine
+FROM openjdk:8-jre-alpine3.8
 MAINTAINER Apache Knox <dev@knox.apache.org>
 
 # Make sure required packages are available
-RUN apk --no-cache add bash procps ca-certificates && update-ca-certificates
+RUN apk --no-cache add bash procps ca-certificates krb5 && update-ca-certificates
 
 # Create an knox user
 RUN addgroup -S knox && adduser -S -G knox knox


zeroflag (approved):

gateway-docker/src/main/resources/docker/Dockerfile
```diff
@@ -13,11 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM openjdk:8-jre-alpine
+FROM openjdk:8-jre-alpine3.8
```
risdenk, line 16: openjdk images aren't updated anymore. see https://hub.docker.com/_/openjdk

you probably want to replace this with:

https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=11-alpine

```
FROM eclipse-temurin:11-alpine
```

moresandeep: Thanks for @risdenk, yeah, noticed that, unfortunately (and frustratingly) we need support for Java 8. So, until we move away from Java 8 this is the best alternative i could find.

moresandeep: Interesting, i was looking for images by openjdk this is how I landed on jre-alpine3.8.  I wasn't aware of the [Temurin](https://projects.eclipse.org/projects/adoptium.temurin), and reluctant to go with something that was not from openjdk. Looks like Temurin is a thing. I'll give this a try. Thanks for the inputs!

moresandeep: I tried using `eclipse-temurin:8-alpine`, there are TLS issues with 
```
*   Trying 127.0.0.1:8443...
* Connected to localhost (127.0.0.1) port 8443 (#0)
* ALPN: offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS alert, user canceled (346):
* TLSv1.3 (IN), TLS alert, close notify (256):
* OpenSSL SSL_connect: SSL_ERROR_ZERO_RETURN in connection to localhost:8443
* Closing connection 0
curl: (35) OpenSSL SSL_connect: SSL_ERROR_ZERO_RETURN in connection to localhost:8443
```

I am not sure what the different is between this base image and the openjdk one but this appears broken (atleast as far as Knox is concerned). 

Let me know if you have any concerns.

risdenk: `openjdk` just isn't even an option anymore. Its not updated period. Even switching the tag. openjdk image is just plain wrong. See https://hub.docker.com/_/openjdk

> The only tags which will continue to receive updates beyond July 2022 will be Early Access builds (which are sourced from [jdk.java.net](https://jdk.java.net/)), as those are not published/supported by any of the above projects.

so really can't use `openjdk`

related to the error you are getting - I don't know.

risdenk: What command are you running to get that error? Does it happen everytime?

It looks like maybe TLS 1.3? Does Knox support TLS 1.3?

moresandeep: `curl -vvv  -iku admin:admin-password  https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token`
I understand it is wrong, but better than broken, right now Knox docker is broken. We need to fix it so atleast we have a working image then we can test with different base images and see which works.

risdenk: I mean https://issues.apache.org/jira/browse/KNOX-2932 doesn't detail anything about the docker image being broken. It talks about adding support for Kerberos - which is a new feature not necessarily meaning things are broken.

The title of the PR was updated to include "TLS Ciphers Issue" but no details as to what that means or why moving to 3.8 alpine actually fixes that.

moresandeep: The TLS issue was found while adding kerberos, this is how I noticed the image is broken. I am going to commit the changes and look for an official image that works. If you find something that work let us know I'd be happy to test.

PR #786, opened 2023-08-15
KNOX-2946 - Cookie Path Scoping doesn't work when using default topology URL

zeroflag:

When cookie path scoping is enabled knox will update the Path attribute of the cookies with `/gateway/topology_name`.
This doesn't work when a service is accessed via the default topology name (there is no `gateway/topology_name` in the url).

## How was this patch tested?

Enable cookie scoping + set default topology to sandbox:

```xml
    <property>
        <name>default.app.topology.name</name>
        <value>sandbox</value>
    </property>
    <property>
        <name>gateway.scope.cookies.feature.enabled</name>
        <value>true</value>
    </property>
```

Set service url in sandbox.xml:

```xml
    <service>
        <role>HIVE</role>
    </service>
```

Set up a test http server:

```smalltalk
Teapot on
    GET: '/' -> [:req |  
        ZnResponse noContent
    	      addCookie: ((ZnCookie name: 'test' value:'val') path: '/original/path'); 
    	      yourself 
         ];
	start. 
```

Unblock cookies in service.xml (data/services/hive/0.13.0/service.xml):
```xml
       <param>
            <name>responseExcludeHeaders</name>
            <value>SET-COOKIE;asdfg</value>
      </param>
```

Cookie path is updated when using the full URL:

```
$ curl -u admin:admin-password -k -v https://localhost:8443/gateway/sandbox/hive

< HTTP/1.1 204 No Content
< Set-Cookie: test=val; Path=/gateway/sandbox/original/path
```

Cookie path is not updated when using the default URL:

```
$ curl -u admin:admin-password -k -v https://localhost:8443/hive

< HTTP/1.1 204 No Content
< Set-Cookie: test=val; path=/original/path
```

### DIFF
--- a/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/api/CookieScopeServletFilter.java
+++ b/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/api/CookieScopeServletFilter.java
@@ -27,9 +27,11 @@ import javax.servlet.http.HttpServletResponse;
 
 import org.apache.knox.gateway.filter.AbstractGatewayFilter;
 import org.apache.knox.gateway.filter.rewrite.impl.CookieScopeResponseWrapper;
+import org.apache.knox.gateway.i18n.GatewaySpiMessages;
+import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 
 public class CookieScopeServletFilter extends AbstractGatewayFilter {
-
+  private static final GatewaySpiMessages LOG = MessagesFactory.get(GatewaySpiMessages.class);
   private String gatewayPath;
   private String topologyName;
 
@@ -43,7 +45,12 @@ public class CookieScopeServletFilter extends AbstractGatewayFilter {
   @Override
   protected void doFilter( HttpServletRequest request, HttpServletResponse response, FilterChain chain )
       throws IOException, ServletException {
-    chain.doFilter(request, new CookieScopeResponseWrapper(response, gatewayPath, topologyName));
+    if ("true".equals(request.getAttribute(DEFAULT_TOPOLOGY_FORWARD_ATTRIBUTE_NAME))) {
+      LOG.ignoringCookiePathScopeForDefaultTopology();
+      chain.doFilter(request, response);
+    } else {
+      chain.doFilter(request, new CookieScopeResponseWrapper(response, gatewayPath, topologyName));
+    }
   }
 
 }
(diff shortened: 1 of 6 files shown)


gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/api/CookieScopeServletFilter.java
```diff
@@ -43,7 +45,12 @@ public void init( FilterConfig filterConfig ) throws ServletException {
   @Override
   protected void doFilter( HttpServletRequest request, HttpServletResponse response, FilterChain chain )
       throws IOException, ServletException {
-    chain.doFilter(request, new CookieScopeResponseWrapper(response, gatewayPath, topologyName));
+    if ("true".equals(request.getAttribute(DEFAULT_TOPOLOGY_FORWARD_ATTRIBUTE_NAME))) {
```
pzampino, line 48: Would Boolean.parseBoolean(request.getAttribute(DEFAULT_TOPOLOGY_FORWARD_ATTRIBUTE_NAME)) be better here?

gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/CookieScopeResponseWrapper.java
```diff
@@ -46,8 +46,8 @@ public CookieScopeResponseWrapper(HttpServletResponse response, String gatewayPa
     public void addHeader(String name, String value) {
         if (SET_COOKIE.equals(name)) {
             String updatedCookie;
-            if (value.contains(COOKIE_PATH)) {
-                updatedCookie = value.replace(COOKIE_PATH, scopePath);
+            if (hasCookiePathAttribute(value)) {
+                updatedCookie = value.replaceAll("(?i)" + COOKIE_PATH, scopePath);
```
pzampino, line 50: What is being done here?

pzampino (commented):
Couple of questions...

gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/api/CookieScopeServletFilter.java
zeroflag: That might be a bit better. I'll modify it.

pzampino (approved):
LGTM

PR #787, opened 2023-08-18
KNOX-2950 - Handling application path aliases

smolnar82:

With this change, our end-users have a way to define custom application path aliases next to the ones we have OOTB to reach certain apps, UIs.
I added the following `gateway-site.xml` property(ies) to support this feature:
```
    <property>
        <name>gateway.application.path.alias.$original-application-path-to-match</name>
        <value>$comma-separated-list-of-alias-names</value>
    </property>
```
where
- $original-application-path-to-match indicates the existing application path we would like to have an alias for
- $comma-separated-list-of-alias-names marks the list of alias names to register (pointing to the same application as the original path above), separated by a comma if there are more

For instance:
```
    <property>
        <name>gateway.application.path.alias.token-generation</name>
        <value>tokengen</value>
    </property>
```
will allow end-users to reach the Token Generation UI on both `.../token-generation/index.html` and `.../tokengen/index.html`.

## How was this patch tested?

Updated JUnit tests and conducted manual testing:
1. Added the following application path aliases in `gateway-site.xml` (under the existing `token-generation/tokengen` entry):
```
    <property>
        <name>gateway.application.path.alias.token-management</name>
        <value>tokenman</value>
    </property>

    <property>
        <name>gateway.application.path.alias.home</name>
        <value>knoxhome, knox-home</value>
    </property>
```
2. Validated that the Token Generation page is reachable on both paths:
Screenshot 2023-08-18 at 12 41 09
Screenshot 2023-08-18 at 12 41 20

3. Validated that the Token Management page is reachable on both paths:
Screenshot 2023-08-18 at 12 42 44
Screenshot 2023-08-18 at 12 42 53

4. Validated that the Knox Home page is reachable on the original path (`/home`) as well as on the two path aliases:
Screenshot 2023-08-18 at 12 45 09
Screenshot 2023-08-18 at 12 45 16
Screenshot 2023-08-18 at 12 45 26

### DIFF
--- a/gateway-release/home/conf/gateway-site.xml
+++ b/gateway-release/home/conf/gateway-site.xml
@@ -141,6 +141,13 @@ limitations under the License.
         <description>A duration (in seconds) beyond a token’s expiration to wait before evicting its state. This configuration only applies when server-managed token state is enabled either in gateway-site or at the topology level.</description>
     </property>
 
+    <!-- @since 2.1.0 application path aliases -->
+    <property>
+        <name>gateway.application.path.alias.token-generation</name>
+        <value>tokengen</value>
+    </property>
+
+
     <!-- Knox Admin related config -->
     <property>
         <name>gateway.knox.admin.groups</name>
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
@@ -926,6 +926,9 @@ public class GatewayServer {
         contexts.removeHandler( oldContext );
       }
       contexts.addHandler( newContext );
+
+      processApplicationPathAliases(warDir, topology);
+
       if( contexts.isRunning() && !newContext.isRunning() ) {
         newContext.start();
         if(!newContext.isAvailable()) {
@@ -937,6 +940,18 @@ public class GatewayServer {
       auditor.audit( Action.DEPLOY, topology.getName(), ResourceType.TOPOLOGY, ActionOutcome.FAILURE );
       log.failedToDeployTopology( topology.getName(), e );
     }
+  }
+
+  private void processApplicationPathAliases(File warDir, Topology topology) {
+    final Map<String, Collection<String>> applicationPathAliases = config.getApplicationPathAliases();
+    applicationPathAliases.forEach((appName, aliases) -> {
+      if (warDir.getName().contains(appName) && !aliases.isEmpty()) {
+        aliases.forEach(alias -> {
+          WebAppContext aliasContext = createWebAppContext(topology, warDir, alias);
+          contexts.addHandler(aliasContext);
+        });
+      }
+    });
   }
 
   private synchronized void internalDeactivateTopology( Topology topology ) {
--- a/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
@@ -1496,4 +1496,20 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
     return usersCanSeeAllTokens == null ? false : usersCanSeeAllTokens.contains(userName);
   }
 
+  @Override
+  public Map<String, Collection<String>> getApplicationPathAliases() {
+    return getPathAliases(".application");
+  }
+
+  private Map<String, Collection<String>> getPathAliases(String qualifier) {
+    final String prefix = GATEWAY_CONFIG_FILE_PREFIX + qualifier + DEPLOYMENT_PATH_ALIAS;
+    final Map<String, Collection<String>> pathAliases = new HashMap<>();
+    this.forEach(config -> {
+      if (config.getKey().startsWith(prefix)) {
+        pathAliases.put(config.getKey().substring(prefix.length()).toLowerCase(Locale.getDefault()), getTrimmedStringCollection(config.getKey()));
+      }
+    });
+    return pathAliases;
+  }
+
 }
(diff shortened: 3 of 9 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
```diff
@@ -933,6 +936,20 @@ private synchronized void internalActivateArchive( Topology topology, File warDi
     }
   }
 
+  private void processApplicationPathAliases(File warDir, Topology topology) {
+    final Map<String, Collection<String>> applicationPathAliases = config.getApplicationPathAliases();
+    if (!applicationPathAliases.isEmpty()) {
```
zeroflag, line 941: Why do we need the `isEmpty` before iterating over the map?

knox-homepage-ui/home/app/homepage.service.ts
```diff
@@ -27,8 +27,10 @@ import {SessionInformation} from './sessionInformation/session.information';
...
 export class HomepageService {
-    apiUrl = window.location.pathname.replace(new RegExp('home/.*'), 'api/v1/metadata/');
-    sessionUrl = window.location.pathname.replace(new RegExp('home/.*'), 'session/api/v1/sessioninfo');
+    pathParts = window.location.pathname.split('/');
+    topologyContext = '/' + this.pathParts[1] + '/' + this.pathParts[2] + '/';
+    apiUrl = this.topologyContext + 'api/v1/metadata/';
+    sessionUrl = this.topologyContext + 'session/api/v1/sessioninfo';
     generalProxyInformationUrl = this.apiUrl + 'info';
     publicCertUrl = this.apiUrl + 'publicCert?type=';
     topologiesUrl = this.apiUrl + 'topologies';
```
zeroflag, line 36: Are these changes relevant to this patch, or is this just refactoring of the existing code?

smolnar82: This is relevant. As you can see, other UIs were revised, and updated if necessary, to get their URLs relative to the topology context and not rely on hard-coded app names to be replaced.
Without this change the feature is not working.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
smolnar82: We don't need it, just defensive programming. I'd leave it for now, it does not hurt or cause any issues.

zeroflag: Then at least let's make it a null check instead of `isEmpty`. That would protect us against NPEs, while this doesn't defend against anything. 
Personally if `config.getApplicationPathAliases()` is guaranteed to be not null, I wouldn't even use a null check, but if I had to decide between those 2 then I would choose a null check.

zeroflag (approved):

gateway-release/home/conf/gateway-site.xml
```diff
@@ -134,6 +134,13 @@ limitations under the License.
         <description>A duration (in seconds) beyond a token’s expiration to wait before evicting its state. This configuration only applies when server-managed token state is enabled either in gateway-site or at the topology level.</description>
     </property>
 
+    <!-- @since 2.1.0 application path aliases -->
+    <property>
```
pzampino, line 138: I'm wondering if it's possible/beneficial to support this at the topology level.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -1489,4 +1489,20 @@ public boolean isAsyncSupported() {
     return getBoolean(GATEWAY_SERVLET_ASYNC_SUPPORTED, GATEWAY_SERVLET_ASYNC_SUPPORTED_DEFAULT);
   }
 
+  @Override
+  public Map<String, Collection<String>> getApplicationPathAliases() {
+    return getPathAliase(".application");
```
pzampino, line 1494: typo: see comment on method itself.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -1489,4 +1489,20 @@ public boolean isAsyncSupported() {
...
+  }
+
+  private Map<String, Collection<String>> getPathAliase(String qualifier) {
```
pzampino, line 1497: typo: getPathAliase --> getPathAliases

knox-token-management-ui/token-management/app/token.management.component.ts
```diff
@@ -30,7 +30,9 @@ import {MatSort} from '@angular/material/sort';
 
 export class TokenManagementComponent implements OnInit {
 
-    tokenGenerationPageURL = window.location.pathname.replace(new RegExp('token-management/.*'), 'token-generation/index.html');
+    pathParts = window.location.pathname.split('/');
```
pzampino, line 33: These lines smell like a utility method to me.

smolnar82: I couldn't agree more, but I did not find out (yet) how to do that properly in TypeScript between different Maven modules. Let's leave it as-is for now and refactor this out when I'll have time to investigate.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
smolnar82: Ack; I'll fix it soon.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
smolnar82: Ack; I'll fix it soon.

gateway-release/home/conf/gateway-site.xml
smolnar82: I filed https://issues.apache.org/jira/browse/KNOX-2964 to address this improvement.

pzampino (approved):

PR #788, opened 2023-08-21
KNOX-2951 - During discovery if cm is not reachable and throws SocketException then retry is not happening

zeroflag:

Service discovery was not retried after a SocketTimeoutException.

## How was this patch tested?

Unittest.


pzampino (approved):
LGTM

PR #793, opened 2023-09-08
KNOX-2948 - HXR parser can handle the new 'provisionEncryptQueryStringCredential' boolean field in SimpleDescriptor

smolnar82:

My previous commit added support for a new boolean variable called `provisionEncryptQueryStringCredential` in a descriptor file. This commit makes it possible to set that new field using a Hadoop XML resource.

## How was this patch tested?

Unit testing.


gateway-topology-hadoop-xml/src/main/java/org/apache/knox/gateway/topology/hadoop/xml/HadoopXmlResourceParser.java
```diff
@@ -266,6 +267,8 @@ private SimpleDescriptor parseXmlDescriptor(String name, String xmlValue) {
         case CONFIG_NAME_PROVIDER_CONFIG_REFERENCE:
           descriptor.setProviderConfig(parameterPairParts[1].trim());
           break;
+        case CONFIG_NAME_PROVISION_ENCRYPT_QUERY_STRING_CREDENTIAL:
+          descriptor.setProvisionEncryptQueryStringCredential(Boolean.valueOf(parameterPairParts[1].trim()));
```
zeroflag, line 271: A `break` statement is missing here.

smolnar82: Indeed; nice catch @zeroflag !
I submitted a new PS.

zeroflag (approved):

PR #794, opened 2023-09-13
KNOX-2957 - Update groovy version

moresandeep:

Update groovy version and version of other libs that depend on groovy.

## How was this patch tested?

Locally.

### DIFF
--- a/build-tools/src/main/resources/build-tools/spotbugs-filter.xml
+++ b/build-tools/src/main/resources/build-tools/spotbugs-filter.xml
@@ -90,4 +90,14 @@ limitations under the License.
     <Class name="org.apache.knox.gateway.audit.log4j.correlation.Log4jCorrelationService" />
     <Bug pattern="OBJECT_DESERIALIZATION" />
   </Match>
+
+  <Match>
+    <Class name="org.apache.knox.gateway.services.security.impl.ConfigurableEncryptor" />
+    <Bug pattern="DMI_RANDOM_USED_ONLY_ONCE" />
+  </Match>
+
+  <Match>
+    <Class name="org.apache.knox.gateway.services.registry.impl.DefaultServiceRegistryService" />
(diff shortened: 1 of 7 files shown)


build-tools/src/main/resources/build-tools/spotbugs-filter.xml
```diff
@@ -90,4 +90,14 @@ limitations under the License.
     <Class name="org.apache.knox.gateway.audit.log4j.correlation.Log4jCorrelationService" />
     <Bug pattern="OBJECT_DESERIALIZATION" />
   </Match>
+
+  <Match>
+    <Class name="org.apache.knox.gateway.services.security.impl.ConfigurableEncryptor" />
+    <Bug pattern="DMI_RANDOM_USED_ONLY_ONCE" />
```
risdenk, line 96: I think this is a real finding.

https://github.com/apache/knox/blob/master/gateway-spi/src/main/java/org/apache/knox/gateway/services/security/impl/ConfigurableEncryptor.java#L106

that random could be created once instead of every time for encrypting bytes.

build-tools/src/main/resources/build-tools/spotbugs-filter.xml
```diff
@@ -90,4 +90,14 @@ limitations under the License.
...
+  </Match>
+
+  <Match>
+    <Class name="org.apache.knox.gateway.services.registry.impl.DefaultServiceRegistryService" />
+    <Bug pattern="DMI_RANDOM_USED_ONLY_ONCE" />
```
risdenk, line 101: I think this is also a real finding - https://github.com/apache/knox/blob/master/gateway-server/src/main/java/org/apache/knox/gateway/services/registry/impl/DefaultServiceRegistryService.java#L74

risdenk (changes requested):

PR #796, opened 2023-09-28
KNOX-2959 - Auto discovery to support scaling scenarios

zeroflag:

After scaling up and down, the service list in the topologies should be regenerated. For example, if the user adds a new service component to a new host then the new URL should be added to the topology. Similarly if a service component is removed from a host, the URL should be deleted.

One small problem is that after adding a component, the new URL will appear in the topology right after the monitor picks up the change, even if the component is not yet started. Failover will handle this scenario. Plus this is not a new thing, since stopping a component has the same effect. But feel free to share you concerns regrading this.

## How was this patch tested?

1. Ensure that cluster config monitor is turned on in gateway-site.xml
2. Add a new component (I tested it with KAFKA_BROKER) to a host
3. Wait for the cluster config monitor to pick up the change
4. Check the topology and see if the new URL was added

Deletion event was tested similarly.


pzampino (approved):
LGTM

PR #799, opened 2023-10-06
KNOX-2961 - Knox SSO cookie Invalidation - Phase II

smolnar82:

In phase II, the following features are implemented:
- Pre-configured users can see all tokens on the Token Management page.
- End-users can execute batch operations on selected Knox Tokens.
- Allowing end-users to show/hide previously disabled KnoxSSO Cookies on the Token Management page.

There is a new Gateway-level configuration that controls who can see all tokens on the Token Management UI:
```
     <property>
        <name>gateway.knox.token.management.users.can.see.all.tokens</name>
        <value>admin</value>
        <description>A comma-separated list of user names who can see all tokens on the Token Management page</description>
    </property>
```
OOTB, the `admin` user acts as the superuser who has the power of controlling tokens created for others. The rest of the users can only see their own tokens only (impersonated or not).

This is how the Token Management UI looks like after my changes:
Screenshot 2023-10-06 at 10 06 44

As you can see, there is only one table (compared to the previous tables of own/impersonated tokens), where all the information is displayed in a compact form:
- Each row starts with a selection checkbox for batch operations (except for disabled KnoxSSO cookies, as there is no point in doing anything with them)
- The disabled token's `Token ID` value is shown in `orange`
- `Username` indicates the user for whom the token is created for
- `Impersonated` is a `boolean` flag indicating if this is an impersonated token:
  - `green check`: yes, this is impersonated. You'll see the user who created the token under the icon
  - `red cross`: no, this is not an impersonated token
- `KnoxSSO` is another `boolean` flag that indicates if this token is created by the `KNOXSSO` service if the feature was enabled
  - `green check`: yes, this is KnoxSSO cookie (token)
  - `red cross`: no, this is not a KnoxSSO cookie (it was created by a regular token API call or on the Token Generation page)
- In the `Actions` column you will see
  - the enable/disable/revoke actions are visible for impersonated tokens too
  - KnoxSSO cookies cannot be revoked nor re-enabled
  
Batch operations:
- when at least one token is selected, we show the following buttons under the table:
  - `Disable Selected Tokens`: when executed, all the selected tokens become disabled (if they were disabled originally, they will remain disabled)
  - `Enable Selected Tokens`: when executed, all the selected tokens become enabled (if they were enabled originally, they will remain enabled)
  - `Revoke Selected Tokens`: when executed, all the selected tokens will be revoked. Please note this option is shown only, if there is no KnoxSSO cookie (token) selected (i.e. batch revocation only works with regular tokens).

Toggles:
- `Show Disabled KnoxSSO Cookies`: this is `true` by default. Since disabled KnoxSSO cookies remain in the underlying token state service until they expire, it may bother users to see them in the tokens table. Flipping this toggle button helps to hide them.
- `Show My Tokens Only`: this toggle button is only visible to users, who can see all tokens. By default, this is `false`. Enabling it will filter the tokens table in a way such that it will contain tokens only that were generated for the logged in user (impersonated or not).

Token API changes:
- new API endpoints, `enableTokens` and `disableTokens`, to enable/disable tokens in one batch. Sample `curl` command:
```
curl -iku admin:admin-password -H "Content-Type: application/json" -d '["62445b09-adb3-4f18-a89d-e5091d87b815","c2ee4646-b695-49af-8910-371ae40f6f75","10c46471-0ff0-4e47-8dac-7c226255d80f"]' -X PUT https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token/disableTokens
```
When this API is used, the state check (if it's already enabled/disabled) is not in use. That is, you can include already enabled/disabled tokens in your batch operation.

-  new API endpoint, `revokeTokens`, to revoke tokens in one batch. Sample `curl` command:
```
curl -iku admin:admin-password -H "Content-Type: application/json" -d '["62445b09-adb3-4f18-a89d-e5091d87b815","c2ee4646-b695-49af-8910-371ae40f6f75","10c46471-0ff0-4e47-8dac-7c226255d80f"]' -X DELETE https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token/revokeTokens
```

- the `revoke` endpoint throws an error upon attempting to revoke a KnoxSSO cookie
```
$ curl -iku admin:admin-password  -H "Content-Type: application/json" -d 'c236d20c-4a05-4cfa-b35e-2ba6dc451de0' -X DELETE https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token/revoke
HTTP/1.1 403 Forbidden
Date: Fri, 06 Oct 2023 08:55:25 GMT
Set-Cookie: KNOXSESSIONID=node03e9y0cy8giy31rh00xc1mrcfx0.node0; Path=/gateway/sandbox; Secure; HttpOnly
Set-Cookie: rememberMe=deleteMe; Path=/gateway/sandbox; Max-Age=0; Expires=Thu, 05-Oct-2023 08:55:25 GMT; SameSite=lax
Content-Length: 113

{
  "revoked": "false",
  "error": "SSO cookie (c236d20c...2ba6dc451de0) cannot not be revoked.",
  "code": 20
}
```

- the `enable` endpoint throws an error upon attempting to enable a disabled KnoxSSO cookie:
```
$ curl -iku admin:admin-password  -H "Content-Type: application/json" -d '107824ab-c54d-4db3-b3b5-5c964892ad05' -X PUT https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token/enable
HTTP/1.1 400 Bad Request
Date: Fri, 06 Oct 2023 08:57:58 GMT
Set-Cookie: KNOXSESSIONID=node011ejmvgcjnlpl13mchqmqjtdjc1.node0; Path=/gateway/sandbox; Secure; HttpOnly
Set-Cookie: rememberMe=deleteMe; Path=/gateway/sandbox; Max-Age=0; Expires=Thu, 05-Oct-2023 08:57:58 GMT; SameSite=lax
Content-Length: 107

{
  "setEnabledFlag": "false",
  "error": "Disabled KnoxSSO Cookies cannot not be enabled",
  "code": 80
}
```
- the `getUserTokens` endpoint changed in a way that it accepts a new boolean query parameter called `allTokens`. If the value of this param is `true`, this API returns all tokens:
```
curl -iku admin:admin-password -X GET https://localhost:8443/gateway/sandbox/knoxtoken/api/v1/token/getUserTokens?allTokens=true
```

## How was this patch tested?

Updated relevant JUnit tests and executed comprehensive manual testing.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
@@ -52,11 +52,11 @@ public class TokenStateDatabase {
   private static final String ADD_METADATA_SQL = "INSERT INTO " + TOKEN_METADATA_TABLE_NAME + "(token_id, md_name, md_value) VALUES(?, ?, ?)";
   private static final String UPDATE_METADATA_SQL = "UPDATE " + TOKEN_METADATA_TABLE_NAME + " SET md_value = ? WHERE token_id = ? AND md_name = ?";
   private static final String GET_METADATA_SQL = "SELECT md_name, md_value FROM " + TOKEN_METADATA_TABLE_NAME + " WHERE token_id = ?";
-  private static final String GET_TOKENS_BY_USER_NAME_SQL = "SELECT kt.token_id, kt.issue_time, kt.expiration, kt.max_lifetime, ktm.md_name, ktm.md_value FROM " + TOKENS_TABLE_NAME
-      + " kt, " + TOKEN_METADATA_TABLE_NAME + " ktm WHERE kt.token_id = ktm.token_id AND kt.token_id IN (SELECT token_id FROM " + TOKEN_METADATA_TABLE_NAME + " WHERE md_name = '" + TokenMetadata.USER_NAME + "' AND md_value = ? )"
+  private static final String GET_ALL_TOKENS_SQL = "SELECT kt.token_id, kt.issue_time, kt.expiration, kt.max_lifetime, ktm.md_name, ktm.md_value FROM " + TOKENS_TABLE_NAME
+      + " kt, " + TOKEN_METADATA_TABLE_NAME + " ktm WHERE kt.token_id = ktm.token_id";
+  private static final String GET_TOKENS_BY_USER_NAME_SQL = GET_ALL_TOKENS_SQL + " AND kt.token_id IN (SELECT token_id FROM " + TOKEN_METADATA_TABLE_NAME + " WHERE md_name = '" + TokenMetadata.USER_NAME + "' AND md_value = ? )"
       + " ORDER BY kt.issue_time";
-  private static final String GET_TOKENS_CREATED_BY_USER_NAME_SQL = "SELECT kt.token_id, kt.issue_time, kt.expiration, kt.max_lifetime, ktm.md_name, ktm.md_value FROM " + TOKENS_TABLE_NAME
-      + " kt, " + TOKEN_METADATA_TABLE_NAME + " ktm WHERE kt.token_id = ktm.token_id AND kt.token_id IN (SELECT token_id FROM " + TOKEN_METADATA_TABLE_NAME + " WHERE md_name = '" + TokenMetadata.CREATED_BY + "' AND md_value = ? )"
+  private static final String GET_TOKENS_CREATED_BY_USER_NAME_SQL = GET_ALL_TOKENS_SQL + " AND kt.token_id IN (SELECT token_id FROM " + TOKEN_METADATA_TABLE_NAME + " WHERE md_name = '" + TokenMetadata.CREATED_BY + "' AND md_value = ? )"
       + " ORDER BY kt.issue_time";
 
   private final DataSource dataSource;
@@ -181,6 +181,10 @@ public class TokenStateDatabase {
     return metadataName.equals(TokenMetadata.PASSCODE) ? new String(Base64.decodeBase64(metadataValue.getBytes(UTF_8)), UTF_8) : metadataValue;
   }
 
+  Collection<KnoxToken> getAllTokens() throws SQLException {
+    return fetchTokens(null, GET_ALL_TOKENS_SQL);
+  }
+
   Collection<KnoxToken> getTokens(String userName) throws SQLException {
     return fetchTokens(userName, GET_TOKENS_BY_USER_NAME_SQL);
   }
@@ -192,7 +196,9 @@ public class TokenStateDatabase {
   private Collection<KnoxToken> fetchTokens(String userName, String sql) throws SQLException {
     Map<String, KnoxToken> tokenMap = new LinkedHashMap<>();
     try (Connection connection = dataSource.getConnection(); PreparedStatement getTokenIdsStatement = connection.prepareStatement(sql)) {
-      getTokenIdsStatement.setString(1, userName);
+      if (userName != null) {
+        getTokenIdsStatement.setString(1, userName);
+      }
       try (ResultSet rs = getTokenIdsStatement.executeQuery()) {
         while (rs.next()) {
           String tokenId = rs.getString(1);
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -452,8 +452,11 @@ public class TokenResource {
       final String userName = uriInfo.getQueryParameters().getFirst("userName");
       final String createdBy = uriInfo.getQueryParameters().getFirst("createdBy");
       final String userNameOrCreatedBy = uriInfo.getQueryParameters().getFirst("userNameOrCreatedBy");
+      final boolean allTokens = Boolean.parseBoolean(uriInfo.getQueryParameters().getFirst("allTokens"));
       final Collection<KnoxToken> userTokens;
-      if (userNameOrCreatedBy == null) {
+      if (allTokens) {
+        userTokens = tokenStateService.getAllTokens();
+      } else if (userNameOrCreatedBy == null) {
         userTokens = createdBy == null ? tokenStateService.getTokens(userName) : tokenStateService.getDoAsTokens(createdBy);
       } else {
         userTokens = new HashSet<>(tokenStateService.getTokens(userNameOrCreatedBy));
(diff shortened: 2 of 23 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
```diff
@@ -192,7 +196,9 @@ Collection<KnoxToken> getDoAsTokens(String userName) throws SQLException {
   private Collection<KnoxToken> fetchTokens(String userName, String sql) throws SQLException {
     Map<String, KnoxToken> tokenMap = new LinkedHashMap<>();
     try (Connection connection = dataSource.getConnection(); PreparedStatement getTokenIdsStatement = connection.prepareStatement(sql)) {
-      getTokenIdsStatement.setString(1, userName);
+      if (userName != null) {
+        getTokenIdsStatement.setString(1, userName);
```
zeroflag, line 200: What happens if we don't set the userName parameter of the prepared statement? Does it work too?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -452,8 +459,11 @@ public Response getUserTokens(@Context UriInfo uriInfo) {
       final String userName = uriInfo.getQueryParameters().getFirst("userName");
       final String createdBy = uriInfo.getQueryParameters().getFirst("createdBy");
       final String userNameOrCreatedBy = uriInfo.getQueryParameters().getFirst("userNameOrCreatedBy");
+      final boolean allTokens = Boolean.parseBoolean(uriInfo.getQueryParameters().getFirst("allTokens"));
```
zeroflag, line 462: What happens if there are no query parameters? Will this fail with an exception?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -559,6 +569,22 @@ public Response renew(String token) {
     return resp;
   }
 
+  @DELETE
+  @Path(BATCH_REVOKE_PATH)
+  @Produces({APPLICATION_JSON})
+  public Response revokeTokens(String tokenIds) {
+    final List<String> ids = GSON.fromJson(tokenIds, List.class);
```
zeroflag, line 576: I wonder if the JSON input parsing would work with `@Consumes(APPLICATION_JSON)` out of the box, without manually doing it with Gson.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -648,17 +673,47 @@ private String getTokenId(String token) throws ParseException {
...
+  @Consumes({ APPLICATION_JSON })
+  @Produces({ APPLICATION_JSON })
+  public Response disableTokens(String tokenIds) {
+    return setTokenEnabledFlags(tokenIds, false);
   }
 
-  private Response setTokenEnabledFlag(String tokenId, boolean enabled) {
+  @SuppressWarnings("unchecked")
+  private Response setTokenEnabledFlags(String tokenIds, boolean enabled) {
+    final List<String> ids = GSON.fromJson(tokenIds, List.class);
```
zeroflag, line 704: Same as above.

knox-token-management-ui/token-management/app/knox.token.ts
```diff
@@ -25,4 +25,8 @@ export class KnoxToken {
     expirationLong: number;
     maxLifetime: string;
     metadata: Metadata;
+
+    public toString = (): string => {
+        return `KnoxToken (tokenId: ${this.tokenId})`;
```
zeroflag, line 30: What is the backtick here?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
smolnar82: I thought the same way, however, I could not make it work this way:
```
  @DELETE
  @Path(BATCH_REVOKE_PATH)
  @Produces({APPLICATION_JSON})
  @Consumes(APPLICATION_JSON)
  public Response revokeTokens(List<String> tokenIds) {
  }
```
I always received 415 (Unsupported Media Type) every time no matter what I tried (added the required headers in the `curl` command).

knox-token-management-ui/token-management/app/knox.token.ts
smolnar82: Hm...that's wrong. Let me fix that.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
zeroflag: Hmm, I remember there was some arguments whether DELETE requests are allowed to have request bodies or not. Maybe the frameworks think it's not allowed therefore it rejects it.

I found this:

```
There is nothing in the specifications that prohibits the use of a body for a HTTP Delete request. However for some HTTP clients this may not be implemented and some servers may ignore it, work incorrectly or refuse the request.
```

Does it work in all major web browsers?

knox-token-management-ui/token-management/app/knox.token.ts
smolnar82: Fixed.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
smolnar82: It does work with our embedded Jetty. If it didn't, the previous `delete` endpoint would fail too.

smolnar82: Please also note I received 415 for the new `disableTokens` and `enableTokens` too, and they are `PUT` requests.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/TokenStateDatabase.java
smolnar82: An SQL error is thrown here which is handled in `JDBCTokenStateService`:
```2023-10-09 11:25:29,822 84376fc3-8a56-42f3-a399-9f81f0ffe2da ERROR token.state (JDBCTokenStateService.java:getTokens(312)) - An error occurred while fetching tokens for user null from the database : No value specified for parameter 1.```

smolnar82: The result is an empty token list in the response for `getUserTokens`.

PR #801, opened 2023-10-09
KNOX-2963 - CM service discovery should work when legacy mode is turned off

zeroflag:

If cloudera manager legacy mode is turned off then certain hdfs config values are moved under the CORE_SETTINGS service.

When the service model generator fetches `hdfs_hadoop_ssl_enabled` it will find a null value (since the real config is under CORE_SETTINGS) and it will generate a non-ssl URL even despite SSL is enabled.

Affected serives: hdfs ui, job history ui, yarn ui.

[Affected config properties](https://docs.cloudera.com/cloudera-manager/7.9.5/managing-clusters/topics/cm-core-configuration-paramters-migrated.html                    
): only `hdfs_hadoop_ssl_enabled`

## How was this patch tested?

Tested with HDFS in non legacy mode.


smolnar82 (approved):
LGTM

PR #811, opened 2023-10-26
KNOX-2978 - Race condition between Service Discovery and Polling Config Analyzer

zeroflag:

When a config change is detected by the Polling Config Analyzer then then the cache used by the service discovery will be cleared. If this happens when discovery is in progress then a NullPointerException will happen.

```java
  private ServiceDetails getServiceDetails(ServiceDiscoveryConfig serviceDiscoveryConfig, ApiService service) {
    return getClusterServices(serviceDiscoveryConfig).getIfPresent(service); // <= NPE
  }
```

```java
  @Override
  public void onConfigurationChange(String source, String clusterName) {
    log.clearServiceDiscoveryRepository();
    repository.clear(); // this will cause the NPE
  }
```
This was observed on a live cluster when certain cluster properties was changed during knox startup.

## How was this patch tested?

- unit test


smolnar82 (approved):
LGTM

PR #824, opened 2023-11-30
KNOX-2991 - Sanitise Oozie rewrite rules

dionusos:

In this change I remove 2-2 Oozie rewrite rules which are too general and they are causing Oozie action errors.

## How was this patch tested?

With the provided patch I build Knox and installed it onto a Hadoop environment. Then executed the `distcp` Oozie example workflow which was succeeded with the fix.


smolnar82 (approved):
LGTM

PR #826, opened 2023-12-14
KNOX-2990 - Using DerbyDatabaseTSS instead of AliasBasedTSS by default

smolnar82:
## What changes were proposed in this pull request?

Implemented the changes I listed in [KNOX-2990](https://issues.apache.org/jira/browse/KNOX-2990):

- deprecated the following TSS implementations:
  - AliasBasedTokenStateService
  - ZookeeperTokenStateService
  - JournalBasedTokenStateService
- implemented a DerbyDB storage that stores tokens in `$DATA_DIR/security/tokens` (this time it's not yet encrypted)
- file permissions are set on that folder to `700` (only the owner can access it)
- changed the default implementation in `TokenStateServiceFactory` to the new DerbyDatabaseTSS
- implemenedt a new KnoxCLI command that migrates existing tokens from credential stores to any JDBC-based TSS backend (tested it with the new DerbyDB TSS; see below)
- integrated this new KnoxCLI command in a way such that it runs when Knox Gateway is started: if token management is enabled, and the configured TSS implementation is a migration target (currently it's true for any JDBC-based TSS implementation)
- added some new `gateway-site.xml` properties:
  - `gateway.knox.token.migration.skip `: ensures that the previous automated step can be controlled (E.g. in case of unforeseen errors it can be turned off). Defaults
  - `gateway.knox.token.migration.archive.tokens `: indicates if migrated tokesn should be archived in another credential store called `__tokens`. Defaults to `false`.
  - `gateway.knox.token.migration.include.expired.tokens`: whether expired tokens should be migrated or skipped. Defaults to `false`.
  - `gateway.knox.token.migration.verbose`: if true, migrated/skipped tokens are added in the `[gateway|knoxcli].log` and, optionally, on the STDOUT (when running the KnoxCLI tool manually). Defaults to `true`, because it's very useful to have the chancee to cross-reference token IDs in case of error debugging.
  - `gateway.knox.token.migration.progress.count`: the number of tokens after the token migration tool displays progress in the logs and, optionally, on the STDOUT.
- modified the token generation page to accept the new DerbyDB TSS.

## How was this patch tested?

Configured Knox to have the `AliasBasedTSS` as the token state backend and to allow unlimited token creation:
```
    <property>
        <name>gateway.service.tokenstate.impl</name>
        <value>org.apache.knox.gateway.services.token.impl.AliasBasedTokenStateService</value>
    </property>

    <property>
        <name>gateway.knox.token.limit.per.user</name>
        <value>-1</value>
    </property>
[...]

PR #828, opened 2023-12-19
KNOX-2995 - Support json parsing NaN values

upczsh:
# What changes were proposed in this pull request?
Implemented the changes I listed in [KNOX-2995](https://issues.apache.org/jira/browse/KNOX-2995):
- Solve the problem of page parsing failure when Json returns NaN value.
> Added JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS.mappedFeature() to support this feature

I click on the page with return json and the content of Resopnse is empty. like this :

image

Checking the gateway.log log shows the following error message.
image

The display results after my repair are as follows：
image

# How was this patch tested?
I have added test cases. You can run the test cases to see if they are successful.
Run JsonFilterReaderTest.testNaN() function. 
Or look at the actual problems I encountered above to test.

### DIFF
--- a/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/json/JsonFilterReader.java
+++ b/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/json/JsonFilterReader.java
@@ -21,6 +21,7 @@ import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
+import com.fasterxml.jackson.core.json.JsonReadFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
@@ -81,7 +82,7 @@ class JsonFilterReader extends Reader {
   }
 
   private void jsonParserConfigInit() {
-    parser.enable(com.fasterxml.jackson.core.JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS);
+    parser.enable( JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS.mappedFeature() );
   }
 
   @Override
--- a/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/json/JsonFilterReader.java
+++ b/gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/json/JsonFilterReader.java
@@ -77,6 +77,11 @@ class JsonFilterReader extends Reader {
     bufferingLevel = null;
     bufferingConfig = null;
     this.config = config;
+    jsonParserConfigInit();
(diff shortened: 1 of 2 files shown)


upczsh:
you can see this issuse: [KNOX-2995](https://issues.apache.org/jira/browse/KNOX-2995)

gateway-provider-rewrite/src/main/java/org/apache/knox/gateway/filter/rewrite/impl/json/JsonFilterReader.java
```diff
@@ -77,6 +77,11 @@ class JsonFilterReader extends Reader {
     bufferingLevel = null;
     bufferingConfig = null;
     this.config = config;
+    jsonParserConfigInit();
+  }
+
+  private void jsonParserConfigInit() {
+    parser.enable(com.fasterxml.jackson.core.JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS);
```
smolnar82, line 84: Why don't you add this feature when creating the `factory` in line 69?
Moreover, the referenced feature has been deprecated since 2.10 according to Jackson docs (we use 2.11.4 in Knox already):
```
        /**
         * Feature that allows parser to recognize set of
         * "Not-a-Number" (NaN) tokens as legal floating number
         * values (similar to how many other data formats and
         * programming language source code allows it).
         * Specific subset contains values that
         * <a href="http://www.w3.org/TR/xmlschema-2/">XML Schema</a>
         * (see section 3.2.4.1, Lexical Representation)
         * allows (tokens are quoted contents, not including quotes):
         *<ul>
         *  <li>"INF" (for positive infinity), as well as alias of "Infinity"
         *  <li>"-INF" (for negative infinity), alias "-Infinity"
         *  <li>"NaN" (for other not-a-numbers, like result of division by zero)
         *</ul>
         *<p>
         * Since JSON specification does not allow use of such values,
         * this is a non-standard feature, and as such disabled by default.
          *
          * @deprecated Since 2.10 use {@link com.fasterxml.jackson.core.json.JsonReadFeature#ALLOW_NON_NUMERIC_NUMBERS} instead
          */
         @Deprecated
         ALLOW_NON_NUMERIC_NUMBERS(false),
```

upczsh: @smolnar82 Thanks for your guidance, I have used JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS instead of the deprecated JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS.
On the other hand, the reason for not adding it on line 69 is that this setting is the content of parse, and other json parse configurations may be enabled in the future, so I wrote a separate method.

smolnar82 (approved):
LGTM

smolnar82:
@upczsh - could you please be more verbose on the PR description above?
We do [provide a template](https://github.com/apache/knox/blob/master/.github/PULL_REQUEST_TEMPLATE.md) OOTB, which we require to be filled in when someone opens a PR.
See this one as a recent sample: https://github.com/apache/knox/pull/826

upczsh:
> @upczsh - We are about to release Knox v2.1.0 and I'd like to include your change. However, I cannot merge it until you update the PR description as requested above. Thanks!

terribly sorry!!! I have updated the PR description.It's my honor to contribute!

PR #833, opened 2024-01-17
KNOX-3000 - Add configurable socket / read timeout parameter to discovery client

zeroflag:

Adding configurable connect, read and write timeout parameters to CM's discovery client.

## How was this patch tested?

gateway-site.xml

```xml
<property>
    <name>gateway.cloudera.manager.service.discovery.connect.timeout.ms</name>
    <value>15000</value>
</property>

<property>
    <name>gateway.cloudera.manager.service.discovery.read.timeout.ms</name>
    <value>16000</value>
</property>

<property>
    <name>gateway.cloudera.manager.service.discovery.write.timeout.ms</name>
    <value>17000</value>
</property>
```    

```bash
cat conf/descriptors/test.json

{
  "name" : "cdp-proxy",
  "discovery-type" : "ClouderaManager",
  "discovery-address" : "http://HOST:7180",
  "discovery-user" : "test",
  "discovery-pwd-alias" : "pass",
  "provider-config-ref" : "default-providers",
  "read-only" : true,
  "cluster" : "Cluster 1",
  "services" : [ {
    "name" : "CM-API",
    "version" : null,
    "urls" : null
  }
 ]
}
```

```bash
./bin/knoxcli.sh create-alias pass --value pass
```

gateway.log

```
2024-01-17 12:52:05,839  DEBUG discovery.cm (DiscoveryApiClient.java:configureTimeouts(147)) - Service discovery client connect timeout 15,000ms, read timeout 16,000ms, write timeout 17,000ms
```

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
@@ -276,6 +276,6 @@ public interface ClouderaManagerServiceDiscoveryMessages {
   @Message(level = MessageLevel.DEBUG, text = "Found downscale event for role: {0} hosts: {1}")
   void foundDownScaleEvent(String role, Set<String> hosts);
 
-  @Message(level = MessageLevel.DEBUG, text = "Service discovery client connect timeout {0}ms, read timeout {1}ms, write timeout {2}ms")
+  @Message(level = MessageLevel.DEBUG, text = "Service discovery client connect timeout {0} ms, read timeout {1} ms, write timeout {2} ms")
   void discoveryClientTimeout(long connectTimeout, long readTimeout, long writeTimeout);
 }
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
@@ -19,6 +19,7 @@ package org.apache.knox.gateway.topology.discovery.cm;
 import java.util.Set;
 
 import com.cloudera.api.swagger.client.ApiException;
+import com.squareup.okhttp.OkHttpClient;
 import org.apache.knox.gateway.i18n.messages.Message;
 import org.apache.knox.gateway.i18n.messages.MessageLevel;
 import org.apache.knox.gateway.i18n.messages.Messages;
@@ -275,4 +276,7 @@ public interface ClouderaManagerServiceDiscoveryMessages {
 
   @Message(level = MessageLevel.DEBUG, text = "Found downscale event for role: {0} hosts: {1}")
   void foundDownScaleEvent(String role, Set<String> hosts);
+
+  @Message(level = MessageLevel.DEBUG, text = "Service discovery client connect timeout {0}ms, read timeout {1}ms, write timeout {2}ms")
+  void discoveryClientTimeout(long connectTimeout, long readTimeout, long writeTimeout);
 }
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
@@ -19,7 +19,6 @@ package org.apache.knox.gateway.topology.discovery.cm;
 import java.util.Set;
 
 import com.cloudera.api.swagger.client.ApiException;
-import com.squareup.okhttp.OkHttpClient;
 import org.apache.knox.gateway.i18n.messages.Message;
 import org.apache.knox.gateway.i18n.messages.MessageLevel;
 import org.apache.knox.gateway.i18n.messages.Messages;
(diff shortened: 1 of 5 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
```diff
@@ -275,4 +275,7 @@ void roleConfigurationPropertyHasChanged(String propertyName,
 
   @Message(level = MessageLevel.DEBUG, text = "Found downscale event for role: {0} hosts: {1}")
   void foundDownScaleEvent(String role, Set<String> hosts);
+
+  @Message(level = MessageLevel.DEBUG, text = "Service discovery client connect timeout {0}ms, read timeout {1}ms, write timeout {2}ms")
```
smolnar82, line 279: I would set this one on `INFO` level and add a space before all `ms` occurrences in the above `text` field.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -287,6 +287,13 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   private static final String CLOUDERA_MANAGER_ADVANCED_SERVICE_DISCOVERY_CONF_MONITOR_INTERVAL = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.advanced.service.discovery.config.monitor.interval";
   private static final String CLOUDERA_MANAGER_SERVICE_DISCOVERY_REPOSITORY_CACHE_ENTRY_TTL = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.service.discovery.repository.cache.entry.ttl";
   private static final String CLOUDERA_MANAGER_SERVICE_DISCOVERY_MAX_RETRY_ATTEMPS = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.service.discovery.maximum.retry.attemps";
+  private static final String CLOUDERA_MANAGER_SERVICE_DISCOVERY_CONNECT_TIMEOUT = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.service.discovery.connect.timeout.msec";
+  private static final String CLOUDERA_MANAGER_SERVICE_DISCOVERY_READ_TIMEOUT = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.service.discovery.read.timeout.msec";
+  private static final String CLOUDERA_MANAGER_SERVICE_DISCOVERY_WRITE_TIMEOUT = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.service.discovery.write.timeout.msec";
+
+  private static final long CLOUDERA_MANAGER_SERVICE_DISCOVERY_CONNECT_TIMEOUT_DEFAULT = 20000;
```
smolnar82, line 294: Do you know if the default was 20 seconds earlier too? Or where this default is coming from?

smolnar82 (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -287,6 +287,13 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   private static final String CLOUDERA_MANAGER_ADVANCED_SERVICE_DISCOVERY_CONF_MONITOR_INTERVAL = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.advanced.service.discovery.config.monitor.interval";
   private static final String CLOUDERA_MANAGER_SERVICE_DISCOVERY_REPOSITORY_CACHE_ENTRY_TTL = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.service.discovery.repository.cache.entry.ttl";
   private static final String CLOUDERA_MANAGER_SERVICE_DISCOVERY_MAX_RETRY_ATTEMPS = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.service.discovery.maximum.retry.attemps";
+  private static final String CLOUDERA_MANAGER_SERVICE_DISCOVERY_CONNECT_TIMEOUT = GATEWAY_CONFIG_FILE_PREFIX + ".cloudera.manager.service.discovery.connect.timeout.msec";
```
smolnar82, line 290: We may want to replace `msec` with `ms`, or, alternatively, omit the time qualifier here and have it documented.
What do you think?

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
zeroflag: The default is [only 10 seconds.](https://github.com/square/okhttp/blob/00ac57e8c3a7920e7752f213d526a776415bc457/okhttp/src/main/kotlin/okhttp3/OkHttpClient.kt#L247). Since this is causing problems at some deployments, I thought it's a good idea to double it.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
zeroflag: I like that it's self-documenting so users don't need to check the unit every time they change it. I can change it to `ms` if it makes more sense.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
smolnar82: What if we kept the 10 seconds in the code, to be backward compatible, and add new entries in the `gateway-site.xml` with 20 seconds?

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
zeroflag: The `getClient()` is called before every service discovery. I can change it to `INFO` but then probably we should put the logging elsewhere to ensure that it's not called too frequently.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
zeroflag: Ok, I'll change it back 10 seconds then.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
smolnar82: ACK.

smolnar82 (approved):
LGTM

PR #835, opened 2024-01-25
KNOX-3002 - KnoxCLI command for generating descriptor for a role type from a list of hosts

zeroflag:

Adding a command to knoxcli that can generate a topology descriptor from a list of URLs.

Usage:

```
Create Knox topology descriptor file for one service
Options are as follows: 
--service-urls-file (required) path to a text file containing service urls 
--service-name (required) the name of the service, such as WEBHDFS, IMPALAUI or HIVE 
--descriptor-name (required) name of descriptor to be created 
--provider-name (required) name of the referenced shared provider 
--output-dir (optional) output directory to save the descriptor file 
--force (optional) force rewriting of existing files, if not used, command will fail when the configs files with same name 
already exist. 
--param service-param-name1 service-param-value1
--param service-param-nameN service-param-valueN
```

## How was this patch tested?
```bash
cat /tmp/urls.txt

http://url1.site:5000
http://url2.site:5000
```

```bash
$ bin/knoxcli.sh generate-descriptor --service-urls-file /tmp/urls.txt --service-name IMPALAUI --provider-name pam.json --descriptor-name impalads.json

Descriptor impalads.json was successfully saved to /Users/attilamagyar/development/test/.
```

```bash
cat impalads.txt

{
  "name" : "impalads",
  "provider-config-ref" : "pam",
  "services" : [ {
    "params" : { },
    "name" : "IMPALAUI",
    "urls" : [ "http://url1.site:5000", "http://url2.site:5000" ]
  } ]
}
```

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/util/DescriptorGenerator.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/util/DescriptorGenerator.java
@@ -20,8 +20,11 @@ package org.apache.knox.gateway.util;
 import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -35,23 +38,25 @@ public class DescriptorGenerator {
   private final String providerName;
   private final String serviceName;
   private final ServiceUrls serviceUrls;
+  private final Map<String, String> params;
 
   static {
     /* skip printing out null fields */
     mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
   }
 
-  public DescriptorGenerator(String descriptorName, String providerName, String serviceName, ServiceUrls serviceUrls) {
+  public DescriptorGenerator(String descriptorName, String providerName, String serviceName, ServiceUrls serviceUrls, Map<String, String> params) {
     this.descriptorName = descriptorName;
     this.providerName = providerName;
     this.serviceName = serviceName.toUpperCase(Locale.ROOT);
     this.serviceUrls = serviceUrls;
+    this.params = params;
   }
 
   public void saveDescriptor(File outputDir, boolean forceOverwrite) {
     File outputFile = new File(outputDir, descriptorName);
     if (outputFile.exists() && !forceOverwrite) {
-      throw new IllegalArgumentException(outputFile + "already exists");
+      throw new IllegalArgumentException(outputFile + " already exists");
     }
     DescriptorConfiguration descriptor = new DescriptorConfiguration();
     descriptor.setName(FilenameUtils.removeExtension(descriptorName));
@@ -59,6 +64,7 @@ public class DescriptorGenerator {
     Topology.Service service = new Topology.Service();
     service.setRole(serviceName);
     service.setUrls(serviceUrls.toList());
+    setParams(service, params);
     descriptor.setServices(Arrays.asList(service));
     try {
       mapper.writerWithDefaultPrettyPrinter()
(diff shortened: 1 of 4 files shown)


gateway-server/src/test/java/org/apache/knox/gateway/util/DescriptorGeneratorTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+public class DescriptorGeneratorTest {
+  private static final ObjectMapper mapper = new ObjectMapper();
+  private static final String TEST_DESC_1 = "test_desc1.json";
+  private static final String TEST_PROV_1 = "test_prov1.json";
+  private static final String IMPALA_UI = "IMPALAUI";
+  private static final List<String> URLS =
+          Arrays.asList("http://amagyar-1.test.site:25000/", "http://amagyar-2.test.site:25000");
+  @Rule
+  public TemporaryFolder folder= new TemporaryFolder();
```
smolnar82, line 27: Instead of making this a temporary file, we could create a static file in `src/test/resources` and have the code read that file. Since this is not a dynamic one, I think it would make more sense and we would avoid creating a temporary folder every time this test is running (the read I/O operation is unavoidable, but creation can be).

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -2392,6 +2413,55 @@ public String getUsage() {
...
+                    "[--output-dir /path/to/output_dir] \n" +
+                    "[--force] \n";
+    public static final String DESC =
+            "Create Knox topology descriptor file for one service\n"
+                    + "Options are as follows: \n"
+                    + "--service-urls-file (required) path to a text file containing service urls \n"
+                    + "--service-name (required) the name of the service, such as WEBHDFS, IMPALAUI or HIVE \n"
+                    + "--descriptor-name (required) name of descriptor to be created \n"
+                    + "--provider-name (required) name of the referenced shared provider \n"
+                    + "--output-dir (optional) output directory to save the descriptor file \n"
```
smolnar82, line 2430: This is an option field. The default value is missing from the description.

gateway-server/src/main/java/org/apache/knox/gateway/util/DescriptorGenerator.java
```diff
@@ -0,0 +1,53 @@
+package org.apache.knox.gateway.util;
```
smolnar82, line 1: Missing header? The build should fail anyway...

gateway-server/src/main/java/org/apache/knox/gateway/util/ServiceUrls.java
```diff
@@ -0,0 +1,36 @@
+package org.apache.knox.gateway.util;
```
smolnar82, line 1: Missing header? The build should fail anyway...

gateway-server/src/test/java/org/apache/knox/gateway/util/DescriptorGeneratorTest.java
```diff
@@ -0,0 +1,55 @@
+package org.apache.knox.gateway.util;
```
smolnar82, line 1: Missing header? The build should fail anyway...

smolnar82 (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -2392,6 +2413,55 @@ public String getUsage() {
...
+        throw new IllegalArgumentException(urlsFilePath + " does not exist");
+      }
+      if (StringUtils.isBlank(serviceName)) {
+        throw new IllegalArgumentException("Missing --service-name");
+      }
+      File outputDir = StringUtils.isBlank(KnoxCLI.this.outputDir) ? new File(".") : new File(KnoxCLI.this.outputDir);
+
+      DescriptorGenerator generator =
+              new DescriptorGenerator(descriptorName, providerName, serviceName, ServiceUrls.fromFile(new File(urlsFilePath)));
+      generator.saveDescriptor(
```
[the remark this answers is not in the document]
smolnar82: I meant the `generator.saveDescriptor(...)` call, just be make it clear :)

gateway-server/src/test/java/org/apache/knox/gateway/util/DescriptorGeneratorTest.java
zeroflag: This is not for creating an input file, but for the resulting output file. The generrated descriptor will be placed into this temp folder (and junit will clean it automatically).

smolnar82 (approved):
LGTM

PR #840, opened 2024-02-07
KNOX-3006 - PAM module occasionally generates garbage group names

zeroflag:

PAM module returns garbage group names due to a thread safety problem:

```
2024-02-07 12:41:44,954 23da5c8b-2664-4c94-a356-fd569eff930c INFO  knox.gateway (KnoxPamRealm.java:doGetAuthorizationInfo(129)) - Computed roles/groups: [nifire39965:] for principal: knoxui
2024-02-07 12:41:44,956 6778ee4d-f655-4d09-b111-2e28ca3f1d4b INFO  knox.gateway (KnoxPamRealm.java:doGetAuthorizationInfo(129)) - Computed roles/groups: [mapredx] for principal: knoxui
2024-02-07 12:41:44,957 9829bbe7-ff7f-4c87-9a65-cbc589be7583 INFO  knox.gateway (KnoxPamRealm.java:doGetAuthorizationInfo(129)) - Computed roles/groups: [mail:x:12] for principal: knoxui
2024-02-07 12:42:03,082 a02943fc-2d6b-402f-b938-041fb64eee0a INFO  knox.gateway (KnoxPamRealm.java:doGetAuthorizationInfo(129)) - Computed roles/groups: [cdro] for principal: knoxui
```

Switching to org.glassfish.main.libpam4j:libpam4j solves the problem.

## How was this patch tested?

By using the following script:

```
#/bin/bash

task(){
   curl -vk -u knoxui:password https://HOST:8443/gateway/cdp-proxy-api/webhdfs/v1/?op=GETHOMEDIRECTORY
}

for i in {1..15000}
do
   echo "$i"
   task &
done
```

### DIFF
--- a/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxPamRealm.java
+++ b/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxPamRealm.java
@@ -87,7 +87,7 @@ public class KnoxPamRealm extends AuthorizingRealm {
 
   private static final AuditService auditService = AuditServiceFactory.getAuditService();
   private static final Auditor auditor = auditService.getAuditor(AuditConstants.DEFAULT_AUDITOR_NAME,
-          AuditConstants.KNOX_SERVICE_NAME, AuditConstants.KNOX_COMPONENT_NAME);
+      AuditConstants.KNOX_SERVICE_NAME, AuditConstants.KNOX_COMPONENT_NAME);
 
   private final HashService hashService = new DefaultHashService();
   private final KnoxShiroMessages shiroLog = MessagesFactory.get(KnoxShiroMessages.class);
@@ -132,7 +132,7 @@ public class KnoxPamRealm extends AuthorizingRealm {
 
   @Override
   protected AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken token)
-          throws AuthenticationException {
+      throws AuthenticationException {
     PAM pam = null;
     UnixUser user = null;
     try {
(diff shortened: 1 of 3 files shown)


gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxPamRealm.java
```diff
@@ -161,6 +149,23 @@ protected AuthenticationInfo doGetAuthenticationInfo(AuthenticationToken token)
         credentialsHash.getSalt(), getName());
   }
 
+  private synchronized UnixUser lookupUnixUser(AuthenticationToken token) {
```
risdenk, line 152: This single threads all requests that use groups then right?

I know that libpam4j used to have threading issues under the hood. It might be there are still issues there 

see https://github.com/kohsuke/libpam4j/issues/28

pom.xml
```diff
@@ -239,7 +239,7 @@
         <json-smart.version>2.4.8</json-smart.version>
         <junit.version>4.13.2</junit.version>
         <lang-tag.version>1.5</lang-tag.version>
-        <libpam4j.version>1.11</libpam4j.version>
+        <libpam4j.version>5.1.0</libpam4j.version>
```
risdenk, line 242: Looks like 8.0.0-M2 is the latest

https://mvnrepository.com/artifact/org.glassfish.main.libpam4j/libpam4j/8.0.0-M2

risdenk (commented):
Is there a bunch of whitespace changes with the latest commit?

pom.xml
zeroflag: Yes but unfortunately 8.0.0-M2 is incompatible on the binary level as it's compiled with java11 language level.

smolnar82 (commented):
LGTM.
Thanks, @zeroflag , for explaining the reason for an older version being used here.
@risdenk - Do you have any concerns with merging of this PR as-is?

smolnar82 (approved):

risdenk (approved):

wilbur4321:
Why did this change get reverted? `UnixUser` in libpam4j is still not thread-safe, and we are definitely running into related crashes.

pzampino:
> Why did this change get reverted? `UnixUser` in libpam4j is still not thread-safe, and we are definitely running into related crashes.

I believe there was an issue with the reverted version of the dependency whereby it was triggering segmentation faults under load.

PR #842, opened 2024-02-23
KNOX-3008 - Displaying hostname and custom banner text on the Knox Home page and other UIs

smolnar82:

As explained in [KNOX-3008](https://issues.apache.org/jira/browse/KNOX-3008), the following changes were implemented in this PR:
- the hostname where the Knox Gateway instance is running is added in the `General Proxy Information' section on the Knox Home page
- end-users can configure a custom banner text displayed on the top banner between the Knox logo and the logged-in user name. The new gateway-level configuration name is `gateway.ui.banner.text` and defaults to an `empty string`. Administrators can set this configuration to simple text, or they can also set HTML style elements to match their needs.

## How was this patch tested?

Manually tested with the following `gateway-site.xml` settings:
1. Simple text
```
    <property>
        <name>gateway.ui.banner.text</name>
        <value>Message of the day: give Knox a star on GitHub!</value>
    </property>
```
Screenshot 2024-02-23 at 14 13 14
Screenshot 2024-02-23 at 14 13 23
Screenshot 2024-02-23 at 14 13 32
Screenshot 2024-02-23 at 14 13 47

2. Message within a `span` with yellow color. Please note the XML-encoded entry in the `value` field:
```
    <property>
        <name>gateway.ui.banner.text</name>
        <value>&lt;span style=&quot;color: yellow&quot;&gt;This is a very long banner text without any meaningful content. This is a very long banner text without any meaningful content. This is a very long banner text without any meaningful content. This is a very long banner text without any meaningful content. This is a very long banner text without any meaningful content. This is a very long banner text without any meaningful content.&lt;/span&gt;</value>
    </property>
```
Screenshot 2024-02-23 at 14 15 52
Screenshot 2024-02-23 at 14 16 00
Screenshot 2024-02-23 at 14 16 08

### DIFF
--- a/gateway-service-metadata/src/main/java/org/apache/knox/gateway/service/metadata/KnoxMetadataResource.java
+++ b/gateway-service-metadata/src/main/java/org/apache/knox/gateway/service/metadata/KnoxMetadataResource.java
@@ -72,6 +72,8 @@ import org.apache.knox.gateway.topology.Topology;
 import org.apache.knox.gateway.util.JsonUtils;
 import org.apache.knox.gateway.util.X509CertificateUtil;
 
+import com.kstruct.gethostname4j.Hostname;
+
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiOperation;
 
@@ -101,6 +103,7 @@ public class KnoxMetadataResource {
       final ServerInfoService serviceInfoService = gatewayServices.getService(ServiceType.SERVER_INFO_SERVICE);
       final String versionInfo = serviceInfoService.getBuildVersion() + " (hash=" + serviceInfoService.getBuildHash() + ")";
       proxyInfo.setVersion(versionInfo);
+      proxyInfo.setHostname(Hostname.getHostname());
       proxyInfo.setAdminApiBookUrl(
           String.format(Locale.ROOT, "https://knox.apache.org/books/knox-%s/user-guide.html#Admin+API", getAdminApiBookVersion(serviceInfoService.getBuildVersion())));
       final GatewayConfig config = (GatewayConfig) request.getServletContext().getAttribute(GatewayConfig.GATEWAY_CONFIG_ATTRIBUTE);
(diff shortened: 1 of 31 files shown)


gateway-service-metadata/src/main/java/org/apache/knox/gateway/service/metadata/KnoxMetadataResource.java
```diff
@@ -101,6 +103,7 @@ public GeneralProxyInformation getGeneralProxyInformation() {
       final ServerInfoService serviceInfoService = gatewayServices.getService(ServiceType.SERVER_INFO_SERVICE);
       final String versionInfo = serviceInfoService.getBuildVersion() + " (hash=" + serviceInfoService.getBuildHash() + ")";
       proxyInfo.setVersion(versionInfo);
+      proxyInfo.setHostname(Hostname.getHostname());
```
zeroflag, line 106: Wouldn't a simple `InetAddress.getLocalHost().getHostName()` work? Do we know how is this library implemented?

smolnar82: See explanation here: https://github.com/mattsheppard/gethostname4j

zeroflag: We've been already using `InetAddress.getLocalHost().getHostName()` and `InetAddress.getLocalHost().getCanonicalHostName()` at a couple of places in the project. 

I wonder how critical is to have the real hostname in cases like this. If this is better we might need to consider replacing the existing references to these `InetAddress` calls. 

Otherwise if this is just a nice to have information, we might    need to think about using the existing idom, instead of introducing a new dependency.

Since adding a new dependency is always a liability, for future maintenance and possible CVE point of view.

smolnar82: > I wonder how critical is to have the real hostname in cases like this.

In the case of troubleshooting in a real cluster, this can be critical.

I agree, we may revisit the current usage of `InetAddress.getLocalHost()` and replace them, if needed, with this new library.

WRT. the CVE concern, I also had that in my mind, but at the end of the day, this new dependency does nothing special but delegate the `gethostname` call to the OS-native layer using JNA. This is a 1-class dependency, I cannot see any reported vulnerabilities so far.
Screenshot 2024-02-27 at 9 37 05

zeroflag (approved):

PR #873, opened 2024-02-28
KNOX-3012 - Fix the DN links on the Ozone SCM UI

dombizita:

In Ozone in [HDDS-9732](https://issues.apache.org/jira/browse/HDDS-9732) the datanode links were changed on the SCM UI. We need to reflect these changes in Knox. As Ozone is 1.4.0 since the last changes in Knox, I added a new folder for the `ozone-scm` service with the changes and left the 1.2.0 folder as it was. 

## How was this patch tested?

Tested the changes in a cluster. Before my changes it was redirecting to the DN UI without Knox, so the link was not rewritten:

before_change

With the changes Knox rewrites the link to the correct url:

after_change


smolnar82 (approved):
LGTM

PR #874, opened 2024-03-04
KNOX-3013 - Knox redirecting Yarn Node Manager URLs to http instead of https

K0K0V0K:

- YARNUI/yarn/outbound/node3 rule rewrites the https schemes to http
- To fix the issue we skip this rule in case if the schema is https

## How was this patch tested?

- unit tests
- manually deployed a cluster and verified the url stays https
   - run MapReduce job
   - after that i checked the ResourceManager UI behind Knox
   - verified the scheme stayed https which links to the NodeManager (befor the change it was http cause the node3 rule)
   
knox

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
@@ -220,13 +220,6 @@
     </content>
 </filter>
 
-<filter name="YARNUI/yarn/outbound/filter/nodes">
-    <content type="*/html">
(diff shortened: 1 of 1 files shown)


K0K0V0K:
Hi @smolnar82 !
May I ask for a review on this, please?
Thanks in advance!

K0K0V0K:
Hi @smolnar82 !

I updated the description with the screenshot.
May I ask for a review of it?
Thanks!

smolnar82 (approved):

smolnar82:
@K0K0V0K - Please let me know when your investigation is done and we can proceed with the merge. Thanks!

K0K0V0K:
Hi @smolnar82 !

I am done with it, no issues were found.
Thanks for the review!

PR #876, opened 2024-03-08
KNOX-3016 - add support for client credentials flow

lmccay:
Adding support for integrations to Knox proxied services and APIs via OAuth style cllient credentials flow. This allows an integration that is provided a CLIENT_ID and CLIENT_SECRET to authenticate to Knox and directly access proxied services with those or exchange those credentials for short lived JWT based access, id and refresh tokens.

This change introduces only the acceptance of the Knox TokenID and Passcode tokens as CLIENT_ID and CLIENT_SECRET in a standard OAuth 2.0 client credentials flow request body. This body will contain the following params:

1. grant_type and it will be "client_credentials"
2. client_id which will be the KnoxToken tokenId or KnoxID
3. client_secret which will be the passcode token for which we store the hash

Authentication using this flow will result in the effective user being what is provided as the CLIENT_ID.

## How was this patch tested?

Ran existing tests and added new unit tests

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
@@ -306,7 +306,7 @@ public abstract class AbstractJWTFilter implements Filter {
     if (metadata != null) {
       String type =  metadata.getMetadata(TYPE);
       // using tokenID and passcode as CLIENT_ID and CLIENT_SECRET will
-      // result in a metadata item called "type". If the valid is set
+      // result in a metadata item called "type". If the value is set
       // to CLIENT_ID then it will be assumed to be a CLIENT_ID and we
       // will use the token id as the username. Since we don't know the
       // token id until it is created, the username is always the same
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
@@ -96,6 +96,8 @@ public abstract class AbstractJWTFilter implements Filter {
    */
   public static final String JWT_EXPECTED_SIGALG = "jwt.expected.sigalg";
   public static final String JWT_DEFAULT_SIGALG = "RS256";
+  public static final String TYPE = "type";
+  public static final String CLIENT_ID = "CLIENT_ID";
 
   static JWTMessages log = MessagesFactory.get( JWTMessages.class );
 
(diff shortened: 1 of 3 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -238,10 +241,33 @@ public Pair<TokenType, String> getWireToken(final ServletRequest request) {
...
-          if (token != null) {
-            parsed = Pair.of(TokenType.JWT, token);
+          String grantType = request.getParameter(GRANT_TYPE);
+          if (CLIENT_CREDENTIALS.equals(grantType)) {
+            // this is indeed a client credentials flow client_id and
+            // client_secret are expected now the client_id will be in
+            // the token as the token_id so we will get that later
+            token = request.getParameter(CLIENT_SECRET);
+            parsed = Pair.of(TokenType.Passcode, token);
+          }
```
smolnar82, line 265: I think this code - along with the above comment that explains the different query parameters in a client credentials flow request - deserves to be in a private method and can be added before the original `parsed == null` check. This way it's easier to read, IMO.

For instance:
```
// Let's check whether this is a client credentials oauth request
parsed = parsed == null ? parseOauthRequestClientCredentials(request) : parsed;

// Finally, whether the token has been configured for another usecase specific header
if (parsed == null) {
   ...

```

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -300,8 +302,23 @@ protected Subject createSubjectFromToken(final JWT token) throws UnknownTokenExc
 
   public Subject createSubjectFromTokenIdentifier(final String tokenId) throws UnknownTokenException {
     TokenMetadata metadata = tokenStateService.getTokenMetadata(tokenId);
+    String username = null;
     if (metadata != null) {
-      return createSubjectFromTokenData(metadata.getUserName(), null);
+      String type =  metadata.getMetadata(TYPE);
```
smolnar82, line 307: As you indicated in the PR's description,
> This change introduces only the acceptance of the Knox TokenID and Passcode tokens as CLIENT_ID and CLIENT_SECRET in a standard OAuth 2.0 client credentials flow request body

Based on this, I've to ask if there is going to be a separate PR where the `type` metadata is created when OAuth clients get a new token (and replay the `tokenId` and `passcode` fields as `CLIENT_ID` and `CLIENT_SECRET`)?

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -300,8 +302,23 @@ protected Subject createSubjectFromToken(final JWT token) throws UnknownTokenExc
...
+      // using tokenID and passcode as CLIENT_ID and CLIENT_SECRET will
+      // result in a metadata item called "type". If the valid is set
```
smolnar82, line 309: I'm not sure I understand this sentence. Did you mean "if the `type` is set to CLIENT_ID"?

lmccay: ahh - yeah, I meant value actually.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
lmccay: I can buy that. I'll change that and fix the type in the other comment.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
smolnar82: As we discussed offline, there's going to be a follow-up JIRA to handle the `type` metadata.

smolnar82 (approved):
LGTM

PR #877, opened 2024-03-11
KNOX-3017 - Avoid showing lifetime adjustment popup when TTL is set to -1

smolnar82:

Two changes on the `Token Generation UI`:
- if the `homepage` topology is configured with `Unlimited token lifetime`, the lifetime adjustment popup will not shown. The logic was there before, this was a casting issue between strings and numbers in the JS code.
- added an `exclamation` icon to better emphasize the `Unlimited lifetime` setting on the UI.

## How was this patch tested?

Manually tested and confirmed that the popup is no longer displayed when creating a 1-year token with `Unlimited lifetime` settings. The new exclamation mark is shown as expected (tested with and without the `-1` TTL).

Screenshot 2024-03-11 at 14 13 13


zeroflag (approved):

moresandeep (commented):
Looks Good!

PR #878, opened 2024-03-11
KNOX-3018 - Tokens that never expire should not be evicted automatically and their expiration should be displayed properly

smolnar82:

The `Token Management` and `Token Generation` UIs are updated in a way such that they display `Never expires` in case a token never expires (that is, if the token's expiration is set to a negative number).
Moreover, the automated token eviction logic is modified to never remove those tokens from the underlying token state backend.

## How was this patch tested?

Updated existing unit tests to cover the token eviction logic change and ran manual testing on the UI.

Screenshot 2024-03-11 at 16 39 33
Screenshot 2024-03-11 at 17 14 10
Screenshot 2024-03-11 at 17 14 41
Screenshot 2024-03-11 at 17 17 46


zeroflag (approved):

moresandeep (approved):
Looks good.

PR #890, opened 2024-03-19
KNOX-3022 - Handling the case when previously persisted CM cluster config file is empty

smolnar82:

As described in the corresponding KNOX-3022 JIRA, it might happen - due to various IO reasons - that the CM cluster configuration file is empty at Knox startup time. This could prevent the Knox Gateway from starting.
In this PR we fix this issue.

## How was this patch tested?

Added a new JUnit test case as well as ran manual testing on a real cluster using CM discovery. Once my patch was applied, the Knox Gateway started properly, and the empty file log entry appeared as expected:
```
grep "is empty" /var/log/knox/gateway/gateway.log 
2024-03-19 02:56:57,434 WARN  discovery.cm (ClusterConfigurationFileStore.java:get(107)) - Previously saved cluster configuration file /var/lib/knox/gateway/data/cm-clusters/https___CM-HOST_CM-PORT-Cluster_1.ver is empty.
```


moresandeep (approved):
Looks good!

PR #905, opened 2024-05-04
KNOX-3036 - Add Primary Group Virtual Group

lmccay:

The virtual groups through predicate evaluations should include a means to dynamically add a group principal with the same name as the username.

This will require intercepting the configured mapping key name which usually ends with the literal virtual group name that will be added upon matching of the predicate logic.

For this, we will add an optional Logical Virtual Group which will need to be resolved rather than used as a literal. For this specific usecase, we can use syntax such as:

```
<param>
    <name>group.mapping.$PRIMARY_GROUP</name>
    <value>(not (member username))</value>
</param>
```

This will add a primary group for all authenticated users that don't already have one in the current groups list.

## How was this patch tested?

Existing unit tests were run and a new one added to prove existing capability to determine that a user is not a member of a group with the username.

curl command used to test it manually

curl -ivku guest:guest-password https://localhost:8443/gateway/sandbox/knoxtoken/v1/oauth/tokens

Audit entries show group added to match the name:

24/05/04 19:58:36 ||142cc739-e70e-494e-926c-0c0f6df64171|audit|[0:0:0:0:0:0:0:1]|KNOXTOKEN|guest|||authentication|uri|/gateway/sandbox/knoxtoken/v1/oauth/tokens|success|Groups: []
24/05/04 19:58:36 ||142cc739-e70e-494e-926c-0c0f6df64171|audit|[0:0:0:0:0:0:0:1]|KNOXTOKEN|guest|||identity-mapping|principal|guest|success|Groups: [guest]

### DIFF
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/VirtualGroupMapper.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/VirtualGroupMapper.java
@@ -32,6 +32,7 @@ import org.apache.knox.gateway.plang.AbstractSyntaxTree;
 import org.apache.knox.gateway.plang.Interpreter;
 
 public class VirtualGroupMapper {
+    public static final String PRIMARY_GROUP = "$PRIMARY_GROUP";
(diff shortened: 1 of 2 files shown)


smolnar82 (approved):
LGTM

moresandeep (approved):
Looks good!

PR #912, opened 2024-05-10
KNOX-3033 Add option to set the correct path for sticky session cookies

stoty:
… instead of appending the role name

Add a new dispatch property to generate the sticky session cookie path from the service routes, and use a default or configured sticky session cookie name without appending the service role to it.

This is expected to be necessary for integration with external load balancers like the AWS Application load balancer which expect a single sticky cookie name.

## How was this patch tested?

Deployed the changes on a live cluster, set the new parameter and checked the cookie name and path set by Knox.
Also added some unit tests.


stoty:
I have followed up on this, and this won't work with some service definitions.

A few service definition, like hbaseui 2.1.0 have a template in the metadata context field.
AFAICT this never gets used anywhere, as up until now context was just an informative field, but this means that my current patch is guaranteed not not work with those services.

Either 
* We need to change the context field for those context values for a non-templated one
* We need to find/define another field to use for the cookie context
* We need to determine the context some other way like calculating a prefix of all route entries.

moresandeep (changes requested):
I made some comments, can you please check those. Thanks!

stoty:
Thanks for the review @moresandeep .

At the moment this is on hold, as we are pursuing another solution.

I will come back to this and try to address your comments later.

moresandeep:
> Thanks for the review @moresandeep .
> 
> At the moment this is on hold, as we are pursuing another solution.
> 
> I will come back to this and try to address your comments later.

Sure, thanks for contributing :)

smolnar82:
@stoty , @moresandeep - I'm closing this now, feel free to re-open when needed.

PR #914, opened 2024-06-04
KNOX-3039 Add error message sanitization to GatewayServlet

kardolus:

This pull request introduces a mechanism to sanitize error messages in the `GatewayServlet` to improve security by hiding IP addresses from exception messages. The following changes were made:
- Added a `isErrorMessageSanitizationEnabled` flag to the `GatewayServlet` to control whether error messages should be sanitized.
- Implemented the `sanitizeException` and `sanitizeAndRethrow` methods in `GatewayServlet` to handle exception sanitization.
- Updated the `GatewayConfig` interface and its implementation `GatewayConfigImpl` to include a new method `isErrorMessageSanitizationEnabled`.
- Created the `GatewayServletTest` class to parameterize tests for scenarios where sanitization is enabled and disabled.

## How was this patch tested?

This patch was tested using the following methods:
- Parameterized unit tests were added to `GatewayServletTest` to cover both scenarios where error message sanitization is enabled and disabled.
- Manual review and inspection of the code changes to ensure accuracy and completeness.

Test steps:
1. Added unit tests in `GatewayServletTest` to check for sanitized and non-sanitized error messages.
2. Verified the new tests pass successfully, ensuring error messages are appropriately sanitized or left unchanged based on the configuration.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
@@ -42,6 +42,7 @@ import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.lang.reflect.Constructor;
 import java.net.URISyntaxException;
 import java.nio.charset.StandardCharsets;
 import java.util.Enumeration;
@@ -49,6 +50,8 @@ import java.util.Enumeration;
 public class GatewayServlet implements Servlet, Filter {
   public static final String GATEWAY_DESCRIPTOR_LOCATION_DEFAULT = "gateway.xml";
   public static final String GATEWAY_DESCRIPTOR_LOCATION_PARAM = "gatewayDescriptorLocation";
+
+  private static boolean isErrorMessageSanitizationEnabled = true;
 
   private static final GatewayResources res = ResourcesFactory.get( GatewayResources.class );
   private static final GatewayMessages LOG = MessagesFactory.get( GatewayMessages.class );
@@ -83,6 +86,8 @@ public class GatewayServlet implements Servlet, Filter {
 
   @Override
   public synchronized void init( ServletConfig servletConfig ) throws ServletException {
+    GatewayConfig gatewayConfig = (GatewayConfig) servletConfig.getServletContext().getAttribute(GatewayConfig.GATEWAY_CONFIG_ATTRIBUTE);
+    isErrorMessageSanitizationEnabled = gatewayConfig.isErrorMessageSanitizationEnabled();
     try {
       if( filter == null ) {
         filter = createFilter( servletConfig );
@@ -92,8 +97,7 @@ public class GatewayServlet implements Servlet, Filter {
         filter.init( filterConfig );
       }
     } catch( ServletException | RuntimeException e ) {
-      LOG.failedToInitializeServletInstace( e );
-      throw e;
+      throw sanitizeAndRethrow(e);
     }
   }
 
@@ -101,14 +105,15 @@ public class GatewayServlet implements Servlet, Filter {
   public void init( FilterConfig filterConfig ) throws ServletException {
     try {
       if( filter == null ) {
+        GatewayConfig gatewayConfig = (GatewayConfig) filterConfig.getServletContext().getAttribute(GatewayConfig.GATEWAY_CONFIG_ATTRIBUTE);
+        isErrorMessageSanitizationEnabled = gatewayConfig.isErrorMessageSanitizationEnabled();
         filter = createFilter( filterConfig );
       }
       if( filter != null ) {
         filter.init( filterConfig );
       }
     } catch( ServletException | RuntimeException e ) {
-      LOG.failedToInitializeServletInstace( e );
-      throw e;
+      throw sanitizeAndRethrow(e);
     }
   }
 
@@ -126,8 +131,7 @@ public class GatewayServlet implements Servlet, Filter {
         try {
           f.doFilter( servletRequest, servletResponse, null );
         } catch( IOException | RuntimeException | ServletException e ) {
-          LOG.failedToExecuteFilter( e );
-          throw e;
+          throw sanitizeAndRethrow(e);
         }
       } else {
         ((HttpServletResponse)servletResponse).setStatus( HttpServletResponse.SC_SERVICE_UNAVAILABLE );
@@ -153,10 +157,8 @@ public class GatewayServlet implements Servlet, Filter {
             //TODO: This should really happen naturally somehow as part of being a filter.  This way will cause problems eventually.
             chain.doFilter( servletRequest, servletResponse );
           }
-
-        } catch( IOException | RuntimeException | ServletException e ) {
-          LOG.failedToExecuteFilter( e );
-          throw e;
+        } catch (Exception e) {
+          throw sanitizeAndRethrow(e);
         }
       } else {
         ((HttpServletResponse)servletResponse).setStatus( HttpServletResponse.SC_SERVICE_UNAVAILABLE );
@@ -167,7 +169,6 @@ public class GatewayServlet implements Servlet, Filter {
       CorrelationServiceFactory.getCorrelationService().detachContext();
     }
   }
-
 
   @Override
   public String getServletInfo() {
@@ -276,5 +277,35 @@ public class GatewayServlet implements Servlet, Filter {
     public Enumeration<String> getInitParameterNames() {
       return config.getInitParameterNames();
     }
+  }
+
+  private Exception sanitizeException(Exception e) {
+    if (e == null || e.getMessage() == null) {
+      return e;
+    }
+    if (!isErrorMessageSanitizationEnabled || e.getMessage() == null) {
+      return e;
+    }
+    String sanitizedMessage = e.getMessage().replaceAll("\\b(?:\\d{1,3}\\.){3}\\d{1,3}\\b", "[hidden]");
+    return createNewException(e, sanitizedMessage);
+  }
+
+  private <T extends Exception> T createNewException(T e, String sanitizedMessage) {
+    try {
+      Constructor<? extends Exception> constructor = e.getClass().getConstructor(String.class, Throwable.class);
+      T sanitizedException = (T) constructor.newInstance(sanitizedMessage, e.getCause());
+      sanitizedException.setStackTrace(e.getStackTrace());
+      return sanitizedException;
+    } catch (Exception ex) {
+      Exception genericException = new Exception(sanitizedMessage, e.getCause());
+      genericException.setStackTrace(e.getStackTrace());
+      return (T) genericException;
+    }
+  }
+
+  private <T extends Exception> T sanitizeAndRethrow(Exception e) throws T {
+    Exception sanitizedException = sanitizeException(e);
+    LOG.failedToExecuteFilter(sanitizedException);
+    throw (T) sanitizedException;
   }
 }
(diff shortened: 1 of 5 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
```diff
@@ -277,4 +278,34 @@ public Enumeration<String> getInitParameterNames() {
...
+
+  private Exception sanitizeException(Exception e) {
+    if (e == null || e.getMessage() == null) {
+      return e;
+    }
+    if (!isErrorMessageSanitizationEnabled || e.getMessage() == null) {
+      return e;
+    }
+    String sanitizedMessage = e.getMessage().replaceAll("\\b(?:\\d{1,3}\\.){3}\\d{1,3}\\b", "[hidden]");
+    return createNewException(e, sanitizedMessage);
```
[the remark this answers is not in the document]
pzampino: I agree that a SanitizedException class to wrap other exceptions would be appropriate for this. The only reason for not doing this is if upstream catchers of these exceptions would ignore the SanitizedException because they're expecting what would be wrapped thereby.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
```diff
@@ -277,4 +278,34 @@ public Enumeration<String> getInitParameterNames() {
...
+      Exception genericException = new Exception(sanitizedMessage, e.getCause());
+      genericException.setStackTrace(e.getStackTrace());
+      return (T) genericException;
+    }
+  }
+
+  private <T extends Exception> T sanitizeAndRethrow(Exception e) throws T {
+    Exception sanitizedException = sanitizeException(e);
+    LOG.failedToExecuteFilter(sanitizedException);
+    throw (T) sanitizedException;
```
pzampino, line 309: Why does this have to throw rather than simply return the sanitized exception?

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -176,6 +176,8 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   public static final String CLUSTER_CONFIG_MONITOR_INTERVAL_SUFFIX = ".interval";
   public static final String CLUSTER_CONFIG_MONITOR_ENABLED_SUFFIX = ".enabled";
 
+  private static final String ERROR_MESSAGE_SANITIZATION_ENABLED = GATEWAY_CONFIG_FILE_PREFIX + ".error.sanitization.enabled";
+  private static final boolean DEFAULT_ERROR_MESSAGE_SANITIZATION_ENABLED = true;
```
pzampino, line 180: The convention would dictate ERROR_MESSAGE_SANITIZATION_ENABLED_DEFAULT

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
```diff
@@ -277,4 +278,34 @@ public Enumeration<String> getInitParameterNames() {
...
+      return e;
+    }
+    if (!isErrorMessageSanitizationEnabled || e.getMessage() == null) {
+      return e;
+    }
+    String sanitizedMessage = e.getMessage().replaceAll("\\b(?:\\d{1,3}\\.){3}\\d{1,3}\\b", "[hidden]");
+    return createNewException(e, sanitizedMessage);
+  }
+
+  private <T extends Exception> T createNewException(T e, String sanitizedMessage) {
```
pzampino, line 293: I would prefer this be named something like createSanitizedException, and encapsulate the sanitization as part of the creation of the new object. Is there any scenario where the message would be sanitized and not result in a new corresponding Exception object?

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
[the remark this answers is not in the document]
kardolus: @moresandeep @pzampino I see the merit in having a dedicated class for sanitizing and logging, which can offer a more maintainable approach moving forward.

However, I have some concerns regarding potential breaking changes. Specifically, I'm not certain if there are users who might be utilizing `GatewayServlet` as if it were a library. If that's the case, changing the exception signature could lead to breaking changes for them. This is because, by introducing a new `SanitizedKnoxException`, we would be altering the types of exceptions that upstream code is expected to catch. Currently, throwing back the same exception type ensures backward compatibility and constitutes a non-breaking change.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
```diff
@@ -277,4 +278,34 @@ public Enumeration<String> getInitParameterNames() {
...
   }
+
+  private Exception sanitizeException(Exception e) {
+    if (e == null || e.getMessage() == null) {
+      return e;
+    }
+    if (!isErrorMessageSanitizationEnabled || e.getMessage() == null) {
+      return e;
+    }
+    String sanitizedMessage = e.getMessage().replaceAll("\\b(?:\\d{1,3}\\.){3}\\d{1,3}\\b", "[hidden]");
```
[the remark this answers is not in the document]
kardolus: Sounds good to me!

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
kardolus: @pzampino I chose this approach because, unlike returning an exception, which requires upstream code to handle the `Exception` type, throwing it directly takes advantage of Java's compiler to manage error handling more effectively, avoiding the need for callers to explicitly deal with broad `Exception` handling.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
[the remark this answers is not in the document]
kardolus: @moresandeep Logging the original error makes sense. Just committed the fix.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
kardolus: @pzampino I renamed the method. By encapsulation I assume you mean wrapping it in the proposed `SanitizedException` class?

pzampino: Actually, I mean modifying the error message within this same method that creates the new Exception, rather than passing the modified message as an argument to this method.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
pzampino: It seems to me that all the callers of this method invoke it as `throw sanitizeAndRethrow(e) ` ; If this method returned the new exception object, it would be thrown by those invocations, no?

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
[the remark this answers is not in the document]
pzampino: +1 for do it now

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
kardolus: @pzampino Nope! I was wrong. Not sure what happened. My guess is that there was an issue with my IDE.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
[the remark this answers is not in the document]
kardolus: You got it!

moresandeep (approved):
Looks good!

moresandeep:
It would be nice to have a 4xx and 5xx page for errors. So instead of showing the ugly page with an exception we can show a custom page with a UUID corresponding the the error. This UUID would be the request-id for which the error is thrown so we can cross reference it in the logs. Just a though :) can be a new feature.

kardolus:
@moresandeep I think that is a great idea! I will start a discussion on the Apache mailing list to explore implementing custom 4xx and 5xx error pages.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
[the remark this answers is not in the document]
kardolus: Just implemented the SanitzedException.

kardolus:
@pzampino @moresandeep I implemented the `SanitizedException`. It makes the code a lot cleaner. I am not happy with the fact though that an `IOException` can be converted to a `ServletException` through the sanitization flow. But maybe I am overthinking things and overengineering. What do y'all think? Is it fine?

gateway-server/src/main/java/org/apache/knox/gateway/SanitizedException.java
```diff
@@ -0,0 +1,9 @@
+package org.apache.knox.gateway;
```
smolnar82, line 1: You need to add the apache license here.

smolnar82 (changes requested):

smolnar82 (approved):

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
```diff
@@ -282,33 +281,15 @@ public Enumeration<String> getInitParameterNames() {
...
-      Exception genericException = new Exception(sanitizedMessage, e.getCause());
-      genericException.setStackTrace(e.getStackTrace());
-      return (T) genericException;
-    }
-  }
-
-  private <T extends Exception> T logAndSanitizeException(Exception e) throws T {
-    LOG.failedToExecuteFilter(e);
-    throw (T) sanitizeException(e);
+    return new SanitizedException(sanitizedMessage);
```
pzampino, line 293: Why not perform the message sanitization in the SanitizationException class?

pzampino:
> @pzampino @moresandeep I implemented the `SanitizedException`. It makes the code a lot cleaner. I am not happy with the fact though that an `IOException` or a `RuntimeException` can be converted to a `ServletException` (a `SanitizedException` `IS-A` `ServletException`) through the sanitization flow.
> 
> But maybe I am overthinking things and overengineering. What do y'all think? Is it fine?

Unless the servlet engine treats them differently in some meaningful way, I think it's ok, so long as the exception does not get lost entirely, which in this case, it does not.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServlet.java
pzampino: I'm not sure I agree about the Single Responsibility Principle. What is the responsibility of the SanitizationException if not to provide a sanitized the message? Otherwise, the burden of message sanitization is on all entities that throw it, AND there is no guarantee that a thrower will have sanitized the message.

If there is a need to customize the sanitization, IMO that is what extensions are for. Nothing would prevent the subsequent creation of classes that extends SanitizationException to override the sanitizing logic.

Personally, I don't think level of effort for refactoring counts for much in these types of decisions, especially refactoring of test code.

kardolus: Yeah, I am bummed that I didn't receive any feedback on the tests. This class had none until I wrote some. Refactoring could mean I need tests on both classes, and since we didn't discuss whether this new pattern is actually better than the old one, it seems like wasted work.

PR #918, opened 2024-06-17
KNOX-3042 - Correct KnoxToken comparator

bonampak:

Correct KnoxToken comparator so that TokenStateService.fetchTokens() would return all the tokens, even if their issue time is the same. If two tokens have the same issue time, then one with the smaller token id (uuid) will be revoked.
## How was this patch tested?
So far only the existing tests are run and we could add a revocation test to TokenServiceResourceTest.


lmccay:
Hi @bonampak - could you add a unit test for this to clearly show what the failure was and that it is fixed and doesn't regress? Sorry for the delayed review here!

smolnar82 (approved):
LGTM

PR #920, opened 2024-06-20
KNOX-3045 - Adding the most recent service definitions for Ranger

bonampak:

Added service definitions for Ranger. 

## How was this patch tested?
The patch was tested locally.


moresandeep (approved):
Great, thanks you @bonampak!

PR #921, opened 2024-06-21
KNOX-3046 - Yarn REST API should support query parameters

slankka:

Knox Gateway for Yarn ResourceManager REST API should respect query parameters.

The conclusion is: `{**}?{**}` works and `{**}` will ignore query params.

## How was this patch tested?

There is already one can prove that patten `{**}?{**}` works with query parameters and without parameters.

https://github.com/apache/knox/blob/d21cf8f63944afdd5d4ffa0236a16cd788d9a14a/gateway-provider-rewrite/src/test/java/org/apache/knox/gateway/filter/rewrite/api/UrlRewriteProcessorTest.java#L274-L303

and rewrite.xml :

https://github.com/apache/knox/blob/d21cf8f63944afdd5d4ffa0236a16cd788d9a14a/gateway-provider-rewrite/src/test/resources/org/apache/knox/gateway/filter/rewrite/api/UrlRewriteProcessorTest/rewrite.xml#L22-L25

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
@@ -23,8 +23,8 @@
      e.g. http://host.com:8088
 -->
 
-<rule dir="IN" name="YARNUI/yarn/inbound/ws" pattern="*://*:*/**/yarn/ws/v1/{**}">
-    <rewrite template="{$serviceUrl[YARNUI]}/ws/v1/{**}"/>
+<rule dir="IN" name="YARNUI/yarn/inbound/ws" pattern="*://*:*/**/yarn/ws/v1/{**}?{**}">
+    <rewrite template="{$serviceUrl[YARNUI]}/ws/v1/{**}?{**}"/>
 </rule>
 <rule dir="IN" name="YARNUI/yarn/inbound/root" pattern="*://*:*/**/yarn/">
     <rewrite template="{$serviceUrl[YARNUI]}/cluster"/>


gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
```diff
@@ -23,8 +23,8 @@
      e.g. http://host.com:8088
 -->
 
-<rule dir="IN" name="YARNUI/yarn/inbound/ws" pattern="*://*:*/**/yarn/ws/v1/{**}">
-    <rewrite template="{$serviceUrl[YARNUI]}/ws/v1/{**}"/>
+<rule dir="IN" name="YARNUI/yarn/inbound/ws" pattern="*://*:*/**/yarn/ws/v1/{**}?{**}">
```
pzampino, line 26: Does this still work WITHOUT any query params?

moresandeep: I'd add a new rule for `{**}?{**}` case then updating the existing one.

moresandeep (changes requested):

gateway-service-definitions/src/main/resources/services/yarnui/2.7.0/rewrite.xml
slankka: > Does this still work WITHOUT any query params?

yes, without `?{**}`, the query params are ignored

slankka: Yes, I have test Yarn RM API
`/ws/v1/cluster/apps` works! and `/ws/v1/cluster/apps?states=running` works

[Cluster Applications API](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/ResourceManagerRest.html#Cluster_Applications_API)

slankka: > I'd add a new rule for `{**}?{**}` case then updating the existing one.

afaik It's not necessary to do that.

lmccay:
@pzampino , @moresandeep - it seems that the conversation topics are resolved here. If this is good to go, we should get it in prior to 2.1.0 branching which is will happen shortly.

pzampino (approved):
LGTM

moresandeep (approved):

PR #924, opened 2024-07-09
KNOX-3040 - Followup patch. Support for multiple JWKS endpoints + code cleanup

moresandeep:

This PR is a followup to KNOX-3040 and adds support for multiple JWKS endpoints. Users can specify one JWKS endpoint or multiple comma separated JWKS endpoints as shown in following example.

```
           <param>
                <name>knox.token.jwks.url</name>
                <value>https://example.com/oauth2/keys?accountId=1234567890, https://www.googleapis.com/oauth2/v3/certs</value>
            </param>
``` 
Note that the parameter name `knox.token.jwks.url` did not change. 

This PR also fixes some deprecated classes and adds supports for JWKS caching and retries (just one retry). 
TTL for JWKS caching is set to 2 hours. 

Two parameters were added one topology level `knox.token.jwks.urls` indicating support for multiple jwks urls (this is in addition to previous param which also supports multiple jwks urls but differs in name only) and one gateway level `gateway.jwks.outage.cache.ttl` where we can now configure JWKS outage TTL value.

## How was this patch tested?

Following is the log snippet of token verification from one valid and one invalid JWKS endpoints (https://example.com/oauth2/keys?accountId=1234567890, https://www.googleapis.com/oauth2/v3/certs)

```
2024-07-08 23:06:08,004 bfc003e6-5d86-4b09-bf0d-a9f1565e0d60 ERROR token.state (DefaultTokenAuthorityService.java:verifyToken(270)) - Failed to verify token using JWKS endpoint https://www.googleapis.com/oauth2/v3/certs, reason: org.apache.knox.gateway.services.security.token.TokenServiceException: Cannot verify token.
2024-07-08 23:06:13,977 bfc003e6-5d86-4b09-bf0d-a9f1565e0d60 INFO  federation.jwt (AbstractJWTFilter.java:verifyTokenSignature(514)) - Token verification result using provided JWKS Url, verified: true
```

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
@@ -125,5 +125,5 @@ public interface JWTMessages {
   void signingKeyVerificationResultMessage(boolean verified);
 
   @Message(level = MessageLevel.ERROR, text = "Invalid URL ignored. Not a valid JWKS url {0}")
-  void notValidJwksUrl(String jwksUrl);
+  void invalidJwksUrl(String jwksUrl);
 }
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
@@ -123,4 +123,7 @@ public interface JWTMessages {
 
   @Message( level = MessageLevel.INFO, text = "Token verification result using knox signing cert, verified: {0}" )
   void signingKeyVerificationResultMessage(boolean verified);
+
+  @Message(level = MessageLevel.ERROR, text = "Invalid URL ignored. Not a valid JWKS url {0}")
+  void notValidJwksUrl(String jwksUrl);
 }
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
@@ -116,7 +116,7 @@ public abstract class AbstractJWTFilter implements Filter {
   private String expectedIssuer;
   private String expectedSigAlg;
   protected String expectedPrincipalClaim;
-  protected Set<URI> expectedJWKSUrl = new HashSet();
+  protected Set<URI> expectedJWKSUrls = new HashSet();
   protected Set<JOSEObjectType> allowedJwsTypes;
 
   private TokenStateService tokenStateService;
@@ -517,8 +517,8 @@ public abstract class AbstractJWTFilter implements Filter {
           log.pemVerificationResultMessage(verified);
         }
 
-        if (!verified && expectedJWKSUrl != null && !expectedJWKSUrl.isEmpty()) {
-          verified = authority.verifyToken(token, expectedJWKSUrl, expectedSigAlg, allowedJwsTypes);
+        if (!verified && expectedJWKSUrls != null && !expectedJWKSUrls.isEmpty()) {
+          verified = authority.verifyToken(token, expectedJWKSUrls, expectedSigAlg, allowedJwsTypes);
           log.jwksVerificationResultMessage(verified);
         }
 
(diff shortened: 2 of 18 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -149,6 +153,30 @@ public void init( FilterConfig filterConfig ) throws ServletException {
     configureExpectedParameters(filterConfig);
   }
 
+  /**
+   * Helper function to extract URLs from given string
+   * in the form of https://url:port/contxt/.wellknown, https://url2:port/contxt/.wellknown
+   * into expectedJWKSUrl URL set.
+   * @param oidcjwksurl
+   */
+  private Set<URI> getJwksUrlsFromConfig(final String oidcjwksurl) {
```
smolnar82, line 162: The parameter should be renamed to `oidcJWKSUrls`.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -83,6 +84,8 @@ public class DefaultTokenAuthorityService implements JWTokenAuthority, Service {
   // https://tools.ietf.org/html/rfc7518
   private static final Set<String> SUPPORTED_PKI_SIG_ALGS = new HashSet<>(Arrays.asList("RS256", "RS384", "RS512", "PS256", "PS384", "PS512"));
   private static final Set<String> SUPPORTED_HMAC_SIG_ALGS = new HashSet<>(Arrays.asList("HS256", "HS384", "HS512"));
+  /* cache JWKS endpoint for 2 hrs in case of outage */
+  private static final long OUTAGE_TTL = 2*60*60*1000;
```
smolnar82, line 88: Having the default set to 2 hours is good, but, IMO, this should be configurable on the topology level.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/JWTMessages.java
```diff
@@ -123,4 +123,7 @@ public interface JWTMessages {
 
   @Message( level = MessageLevel.INFO, text = "Token verification result using knox signing cert, verified: {0}" )
   void signingKeyVerificationResultMessage(boolean verified);
+
+  @Message(level = MessageLevel.ERROR, text = "Invalid URL ignored. Not a valid JWKS url {0}")
+  void notValidJwksUrl(String jwksUrl);
```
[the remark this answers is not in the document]
moresandeep: Ahh, yes!

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -115,7 +116,7 @@ public abstract class AbstractJWTFilter implements Filter {
   private String expectedIssuer;
   private String expectedSigAlg;
   protected String expectedPrincipalClaim;
-  protected String expectedJWKSUrl;
+  protected Set<URI> expectedJWKSUrl = new HashSet();
```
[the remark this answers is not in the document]
moresandeep: Right, makes it more meaningful. Thanks!

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -110,7 +114,7 @@ public void init( FilterConfig filterConfig ) throws ServletException {
     //  JWKSUrl
     String oidcjwksurl = filterConfig.getInitParameter(JWKS_URL);
```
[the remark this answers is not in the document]
moresandeep: Yeah, I thought about that but we have other provisioning tools that are using this config. So depreciating it would mean changing a lot of supporting code. But, I would introduce `knox.token.jwks.urls` property as well.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
moresandeep: Ack

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
moresandeep: This is a service so the configurations are gateway level. I will add it at gateway level in case one wants to configure it.

PR #926, opened 2024-07-24
KNOX-3052 - Allow Multiple Issuers and with some and no Audience

lmccay:

While we have a change to introduce the ability to use multiple JWKS Urls to verify a token signature, without this change any tokens would need to have the same Issuer. This isn't ideal and limits the flexibility that we are looking for.

This change is only an iteration beyond that approach but still not ideal. We will want to have a better isolation of the expected claims, algorithms, etc - per token. This will suffice for now but we will revisit it in the near future for better isolation.

Here we will simply change the expectedIssuers param to be a List of Strings from a comma separated list and introduce a keyword "NONE" to indicate even though there are expected audiences for some tokens, it is also possible to accept a token with no audience as well. This is an opt-in only feature that requires the admin to configure "NONE" as an acceptable audience claim. This will pass when there are no audiences in the token or even if there is one called "NONE". Again, this will be revisited in the future and done better.

## How was this patch tested?
New unit tests were added and existing tests run along with the new tests.

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
@@ -33,6 +33,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.security.auth.Subject;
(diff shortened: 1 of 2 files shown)


pzampino (approved):

moresandeep (approved):

smolnar82 (approved):
LGTM

pzampino:
This is a more complete version of https://github.com/apache/knox/pull/926, addressing some test issues therewith.

smolnar82:
Closing this one because it was fixed in #1006 .

PR #929, opened 2024-08-20
KNOX-3058: Avoid 404 When Topology Is Being Redeployed

pzampino:

Modified error handling when a topology is being redeployed, such that the response is not HTTP 404 'Not Found', but rather HTTP 503 'Service Unavailable'. The 503 response is much more likely to be retried by clients than is a 404 response.

Added a Jetty ErrorHandler that checks whether or not the topology being requested is in a Set marked as inactive. Topology names are add to this inactive set when the associated topology is deactivated, and removed from this set when the topology is reactivated. In the case of topology deletion, the topology is marked as inactive, but then removed from the inactive set because we know it's being deleted.

## How was this patch tested?

I deployed a test topology with a demo LDAP provider and the Knox Token service.

I then ran the following script with 'https://localhost:8443/gateway/demo/knoxtoken/api/v2/token' and one of the demo LDAP username/pwd combinations, and piped the output to a file. This script outputs only the HTTP response status code for each invocation.
```
#!/bin/sh
#
#

ENDPOINT=$1
echo "Endpoint: $ENDPOINT"

if [ ! -z "$2" ] ; then
  USER=$2
fi

if [ ! -z "$3" ] ; then
  PWD=$3
fi

for i in {1..100000}
do
  curl -o /dev/null -s -w "%{http_code}\n" -ku ${USER}:${PWD} ${ENDPOINT}
done

```
Example:
`~/bin/resp-test.sh 'https://localhost:8443/gateway/demo/knoxtoken/api/v2/token' sam sam-password > ~/response-code-test.txt &
`
While this script is running, I copied the test topology to the conf/topologies directory, then subsequently "touched" the test topology to trigger redeployment many times over several minutes. Finally, I deleted the test topology.

Following this, I reviewed the output to verify that there were initially 404 responses, followed by some 503 responses while the topology was being activated. Then, I verified that there were no additional 404 responses until that time at which I deleted the topology. I also verified the periodic 503 responses which are expected, and the normal 200 responses in between.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
@@ -62,11 +62,13 @@ import org.eclipse.jetty.server.Handler;
 import org.eclipse.jetty.server.HttpConfiguration;
 import org.eclipse.jetty.server.HttpConnectionFactory;
 import org.eclipse.jetty.server.NetworkConnector;
+import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.SecureRequestCustomizer;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.server.handler.ContextHandler;
 import org.eclipse.jetty.server.handler.ContextHandlerCollection;
+import org.eclipse.jetty.server.handler.ErrorHandler;
 import org.eclipse.jetty.server.handler.HandlerCollection;
 import org.eclipse.jetty.server.handler.RequestLogHandler;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
@@ -84,6 +86,8 @@ import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
 import javax.servlet.SessionCookieConfig;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.transform.TransformerException;
 import java.io.File;
@@ -147,6 +151,8 @@ public class GatewayServer {
   private Map<String, WebAppContext> deployments;
   private AtomicBoolean stopped = new AtomicBoolean(false);
   private GatewayStatusService gatewayStatusService;
+
+  private final Set<String> inactiveTopologies = new HashSet<>();
 
   public static void main( String[] args ) {
     try {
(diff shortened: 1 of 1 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
```diff
@@ -962,10 +985,32 @@ private void processApplicationPathAliases(File warDir, Topology topology) {
...
+
+  private boolean isInactiveTopology(final String topologyName) {
+    boolean result = false;
+    synchronized (inactiveTopologies) {
+      result = inactiveTopologies.contains(topologyName);
+    }
+    return result;
+  }
+
   private synchronized void internalDeactivateTopology( Topology topology ) {
```
lmccay, line 1008: Does this get called for the first deployment on the initial start as well?
When does it get marked as active again? Does it incorporate discovery at all?
Should this be done outside of the TopologyService? Seems like state that should be managed by that service maybe.

lmccay (commented):
Are we able to add a test for this?
See my other comments as well.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
pzampino: Automated testing will be tricky given the small window of time when this is an issue. I'm open to suggestions though.

When you ask "Does this get called for the first deployment...", what is "this"? On first deployment, there had not been any opportunity for it to have been marked as inactive, so I think the answer is "no".

The topology gets marked active upon activation (see internalActivateTopology method). This is not connected to discovery because we're dealing directly with topologies here. Discovery will have happened prior to this, and the undeployment won't happen until the new version of the topology is written out.

I thought about that, but

1. The source of the 404 responses is Jetty, so overriding them seems appropriate to do with Jetty facilities.
2. It seems best to do it as close to the activation/deactivation logic as possible.

lmccay: "On first deployment, there had not been any opportunity for it to have been marked as inactive, so I think the answer is "no"."

If we don't automatically call deactivate prior to deployment and activate then 404's will still be available while topologies are being deployed. No?

I think the state and isTopologyActive to see if it is active or not can be in the TopologyService and used in the GatewayService error handler. Spreading that state out is what I am questioning.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
```diff
@@ -717,6 +723,21 @@ void createJetty() throws IOException, CertificateException, NoSuchAlgorithmExce
...
+    // Add a handler for the 404 responses when a topology is being redeployed (i.e., is inactive)
+    jetty.setErrorHandler(new ErrorHandler() {
+      @Override
+      public void doError(String target, Request baseRequest, HttpServletRequest request, HttpServletResponse response) throws IOException {
+        final int gatewayPrefixLength = ("/" + config.getGatewayPath() + "/").length();
+        String pathInfo = baseRequest.getPathInfo();
+        String topologyName = pathInfo.substring(gatewayPrefixLength, pathInfo.indexOf('/', gatewayPrefixLength));
+        if (isInactiveTopology(topologyName) && (response.getStatus() == HttpServletResponse.SC_NOT_FOUND)) {
+          request.setAttribute("javax.servlet.error.message", "Service Unavailable"); // The default ErrorHandler references this attribute
+          response.setStatus(HttpServletResponse.SC_SERVICE_UNAVAILABLE);
```
moresandeep, line 735: Just a thought, can we specify in the error message that the topology is inactive? This will be helpful in cases where we are debugging and knox the request failed because topology was inactive and not due to 500 error.

moresandeep (approved):

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
pzampino: I'm not convinced that the state should be part of the TopologyService. The TopologyService has more to do with handling topology artifacts than actual deployment/activation. It seems to me that the responsibility for deployment/activation is handled by GatewayServer. I think managing this active state in the TopologyService would broaden the window of time during which clients would unnecessarily receive a 503 response where they would have previously not received a 404 response. The 404 responses would not occur until the app was actually deactivated.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
pzampino: We could certainly log that fact, but I think the error messages (e.g., Service Unavailable) are pretty standard. Given the separation between status code and error message though, I could be persuaded otherwise.

gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
pzampino: Concerning the point about initial deployments, we could/should immediately mark topologies as inactive on deployment initiation, and then mark them as active upon completion. I believe this would further minimize the potential for clients to receive a 404 response for a "deployed" topology.

pzampino: I've added the initial deployment behavior you've suggested.

lmccay (approved):
LGTM
+1

PR #933, opened 2024-09-06
KNOX-3062 - Fix an issue where group headers were not added to POST and PUT requests

moresandeep:

Fix an issue where group headers were not added to POST and PUT requests

## How was this patch tested?

This patch was tested locally with POST and GET requests.


smolnar82:
Looking closer again, I found this one irrelevant. The reason is, that the submitted code here assumes that #747 is merged, which isn't the case.
Thus, this PR can be closed as we cover all REST methods in our dispatch properly in terms of group headers addition.

PR #934, opened 2024-09-09
KNOX-3063: Added monitorInterval to log4j2 configuration file

hanicz:

Added monitorInterval property to the Configuration tag in the log4j2 configuration file. With this change Knox can change log level at runtime there is no need to restart the service.

## How was this patch tested?

Ran existing tests locally.
Tested the new behaviour locally by changing the knox.gateway log level to TRACE and checking the gateway.log file for DEBUG and TRACE level logs.

### DIFF
--- a/gateway-release/home/conf/gateway-log4j2.xml
+++ b/gateway-release/home/conf/gateway-log4j2.xml
@@ -15,7 +15,7 @@
   See the License for the specific language governing permissions and
   limitations under the License.
 -->
-<Configuration>
+<Configuration monitorInterval="15">
     <Properties>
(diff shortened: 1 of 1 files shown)


smolnar82 (approved):
Nice additonion, @hanicz !

smolnar82:
For the sake of documentation, the given number should be interpreted as the 
```Number of seconds between polls for configuration changes```
(the 'seconds' information was not obvious from the description here or in the JIRA)

moresandeep (approved):
This is great, I did not know about this feature :)

pzampino (approved):

PR #936, opened 2024-10-04
KNOX-3067: Upgrade ehcache to 3.3.1

hanicz:

Upgrade ehcache from 2.x to 3.x

Currently with 2.x the CacheManager is a Singleton therefore every cache will be stored in the same CacheManager. Due to this caches with the same name will be created only once and different topologies are using the same cache. For example if a user authenticates in the sandbox topology the same user and password can be used for other topologies as well where ehcache is enabled and the cache name is the same. This could cause issues if the ldap configuration is different. 

With Ehcache 3.x creating the CacheManager is our responsibility. Different topologies will have different configurations and caches even with the same name. Every topology that enables ehcache needs a different configuration file. The persistence attribute has to be unique.

<persistence directory="${java.io.tmpdir}/TOPOLOGY_NAME"/>

## How was this patch tested?

Unit tests and manually on local setup

### DIFF
--- a/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/ShiroMessages.java
+++ b/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/ShiroMessages.java
@@ -20,12 +20,28 @@ package org.apache.knox.gateway;
 import org.apache.knox.gateway.i18n.messages.Message;
 import org.apache.knox.gateway.i18n.messages.MessageLevel;
 import org.apache.knox.gateway.i18n.messages.Messages;
+import org.apache.knox.gateway.i18n.messages.StackTrace;
 
 @Messages(logger="org.apache.knox.gateway")
 public interface ShiroMessages {
-  @Message(level = MessageLevel.INFO, text = "Request {0} matches unauthenticated path configured in topology, letting it through" )
+  @Message( level = MessageLevel.INFO, text = "Request {0} matches unauthenticated path configured in topology, letting it through" )
   void unauthenticatedPathBypass(String uri);
 
   @Message( level = MessageLevel.WARN, text = "Invalid URL pattern for rule: {0}" )
   void invalidURLPattern(String rule);
+
+  @Message( level = MessageLevel.TRACE, text = "Acquiring EhcacheShiro instance named {0}" )
+  void acquireEhcacheShiro(String name);
+
+  @Message( level = MessageLevel.INFO, text = "Cache with name {0} does not yet exist.  Creating now." )
+  void noCacheFound(String name);
+
+  @Message( level = MessageLevel.INFO, text = "Added EhcacheShiro named {0}" )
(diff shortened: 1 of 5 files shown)


gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/ShiroMessages.java
```diff
@@ -20,12 +20,28 @@
...
+  @Message( level = MessageLevel.INFO, text = "Cache with name {0} does not yet exist.  Creating now." )
+  void noCacheFound(String name);
+
+  @Message( level = MessageLevel.INFO, text = "Added EhcacheShiro named {0}" )
+  void ehcacheShiroAdded(String name);
+
+  @Message( level = MessageLevel.INFO, text = "Using existing EhcacheShiro named {0}" )
+  void usingExistingEhcacheShiro(String name);
+
+  @Message( level = MessageLevel.WARN, text = "The Shiro managed CacheManager threw an Exception while closing: {0}" )
```
moresandeep, line 45: Nit: Think of log messages are targeted to user, we do not need to tell them than exception occured. We can just note: "There was a error closing cache manager, reason: ". No need to disclose the underlying implementation i.e. shiro. Just an observation :)

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxCacheManager.java
```diff
@@ -17,11 +17,136 @@
...
+      } else {
+        LOG.usingExistingEhcacheShiro(name);
+      }
+      return new EhcacheShiro<>(cache);
+    } catch (MalformedURLException | ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+      throw new CacheException(e);
+    }
+  }
+
+  private synchronized org.ehcache.Cache<Object, Object> createCache(String name)
```
moresandeep, line 87: Why are we synchronizing?

moresandeep (changes requested):

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/ShiroMessages.java
hanicz: I modified the log message

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxCacheManager.java
hanicz: The implementation is based on the 2.x CacheManager. In our use-case there is no race condition currently however future changes might introduce. The synchronized block doesn't slow Knox's start-up because there is no contention.

pzampino (approved):
LGTM

PR #940, opened 2024-10-18
KNOX-3071: New ability in list-alias to list for multiple clusters. N…

hanicz:
…ew create-list-alias command to create multiple aliases for multiple clusters and also list them.

This PR proposes a few changes to the KnoxCLI list-alias and the create-aliases commands. Usage was missing for BatchAliasCreateCommand in printKnoxShellUsage so I added that as well.

With this change the list-alias command will be able to list for multiple clusters. The cluster names have to be concatenated by ','. It is backward compatible with the current usage so it is able to list for a single cluster as well.
Usage: list-alias --cluster1,cluster2,clusterN

New create-list-aliases command that creates aliases on multiple clusters and list them at the end.

create-list-aliases --alias alias1 --value value1 --cluster cluster1 --alias alias2 --value value2 --cluster2 --alias aliasN --value valueN --cluster clusterN

When there is a --cluster args every alias before it will be added to that cluster.

## How was this patch tested?

I wrote new unit tests to cover the new functionalities and tested manually on my local computer.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
@@ -136,7 +136,8 @@ public class KnoxCLI extends Configured implements Tool {
       "   [" + TopologyConverter.USAGE + "]\n" +
       "   [" + JWKGenerator.USAGE  + "]\n" +
       "   [" + GenerateDescriptorCommand.USAGE + "]\n" +
-      "   [" + TokenMigration.USAGE  + "]\n";
+      "   [" + TokenMigration.USAGE  + "]\n" +
+      "   [" + CreateListAliasesCommand.USAGE + "]\n";
   private static final String CLUSTER_STRING_SEPARATOR = ",";
 
   /** allows stdout to be captured if necessary */
@@ -163,7 +164,6 @@ public class KnoxCLI extends Configured implements Tool {
   private boolean migrateExpiredTokens;
   private boolean verbose;
   private String alias;
-  private boolean listAliases;
 
   private String remoteRegistryClient;
   private String remoteRegistryEntryName;
@@ -277,6 +277,12 @@ public class KnoxCLI extends Configured implements Tool {
           printKnoxShellUsage();
           return -1;
         }
+      } else if (args[i].equals("create-list-aliases")) {
+        command = new CreateListAliasesCommand();
+        if (args.length < 3 || "--help".equals(alias)) {
+          printKnoxShellUsage();
+          return -1;
+        }
       } else if (args[i].equals("create-cert")) {
         command = new CertCreateCommand();
         if ((args.length > i + 1) && args[i + 1].equals("--help")) {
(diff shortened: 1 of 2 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -136,7 +136,8 @@ public class KnoxCLI extends Configured implements Tool {
       "   [" + TopologyConverter.USAGE + "]\n" +
       "   [" + JWKGenerator.USAGE  + "]\n" +
       "   [" + GenerateDescriptorCommand.USAGE + "]\n" +
-      "   [" + TokenMigration.USAGE  + "]\n";
+      "   [" + TokenMigration.USAGE  + "]\n" +
+      "   [" + CreateListAliasesCreateCommand.USAGE + "]\n";
```
moresandeep, line 140: Great catch!

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -701,31 +717,34 @@ protected RemoteConfigurationRegistryClientService getRemoteConfigRegistryClient
...
-        out.println("Listing aliases for: " + cluster);
-        List<String> aliases = as.getAliasesForCluster(cluster);
-        for (String alias : aliases) {
-          out.println(alias);
-        }
-        out.println("\n" + aliases.size() + " items.");
-      } else {
-        out.println("Invalid cluster name provided: " + cluster);
-      }
+     String[] clusters = cluster.split(":");
```
moresandeep, line 733: 1. `:` should be a constant e.g. CLUSTER_STRING_SEPARATOR
2. I think `,` would be more obvious, no?
3. What if the cluster had `:` it should not but we had a case where alias name had special characters.

moresandeep (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
hanicz: Yes, it makes more sense to use `,`. I also added it as a constant.

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -275,6 +277,12 @@ private int init(String[] args) throws IOException {
           printKnoxShellUsage();
           return -1;
         }
+      } else if (args[i].equals("create-list-aliases")) {
```
pzampino, line 280: Does this need to be a separate command, or could it be another option to the existing create-alias(es) commands?

hanicz: It could be the same command. My thinking was that we are introducing two new functionalities which in my opinion could warrant a new command.

1. With this change we are able to add multiple aliases to multiple topologies. 
2. When the alias creation is done for a specific topology it will list all of the aliases for it.

pzampino: If the new functionality can be added as an option(s) to the existing command without changing the previous behavior of the existing command (without those options), then I think I would prefer this over an entirely separate command. Ultimately, we're just creating aliases.

hanicz: I moved the new functionalities into the existing command. This modification is not backward compatible because previously the --cluster parameter's position didn't matter, however from now on an alias will be created for a cluster if it is listed before the cluster.

moresandeep (commented):
Looks good, i am just worried about breaking backwards changes. There could be scripts that users might have been using and now this might break them. Folks might also be using it in say docker image setup a breakage there might be tricky to debug.

pzampino (approved):
LGTM. In the future, squashed commits are easier to review.

PR #949, opened 2024-11-05
KNOX-3073 - Token verification fallback to Knox keys behavior should be configurable

pzampino:

Added a topology-level JWTProvider param for enabling token signature verification fallback to Knox's key(s) when a PEM and/or JWKS URL(s) are explicitly configured for the verification. By default, if one or both of those explicit methods is configured, the verification will not fallback to Knox's key(s).

## How was this patch tested?

Unit tests and manual testing was performed to validate the flow and conditions.

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
@@ -31,6 +31,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
@@ -102,6 +103,9 @@ public abstract class AbstractJWTFilter implements Filter {
   public static final String JWT_EXPECTED_SIGALG = "jwt.expected.sigalg";
   public static final String JWT_DEFAULT_SIGALG = "RS256";
 
+  public static final String JWT_INSTANCE_KEY_FALLBACK = "jwt.instance.key.fallback";
+  public static final boolean JWT_INSTANCE_KEY_FALLBACK_DEFAULT = false;
+
   static JWTMessages log = MessagesFactory.get( JWTMessages.class );
 
   private static AuditService auditService = AuditServiceFactory.getAuditService();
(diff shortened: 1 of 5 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
```diff
@@ -512,17 +520,22 @@ protected boolean verifyTokenSignature(final JWT token) {
...
         }
 
         if (!verified && expectedJWKSUrls != null && !expectedJWKSUrls.isEmpty()) {
+          hasJWKS = true;
           verified = authority.verifyToken(token, expectedJWKSUrls, expectedSigAlg, allowedJwsTypes);
           log.jwksVerificationResultMessage(verified);
         }
 
-        if(!verified) {
+        if(!verified && ((!hasPem && !hasJWKS) || isJwtInstanceKeyFallback)) {
```
smolnar82, line 538: I think `hasPEM` and `hasJWKS` are redundant (or need a better name).
First, `hasJWKS` is misleading, because it's only set to `true` when
- no PEM configured
- PEM configured, but the PEM verification attempt failed

So even if JWKS is configured, it might remain `false`.

If I were you, I'd only rely on the already existing class members as follows:
- new private method: `private boolean configuredPEM() { return publicKey != null; }`
- new private method `private boolean configuredJWKS() { return  expectedJWKSUrls != null && !expectedJWKSUrls.isEmpty(); }`
- then, you could use these new methods in lines 526 and 532 as well as in the new condition in line 538. Even better, I'd create another new private method for that purpose:
```
private boolean verifyInstanceKeys() { //name might be changing
  return isJwtInstanceKeyFallback || (!hasPEM() && !hasJWKS());
}
```
So the updated condition would look like this: `if (!verified && verifyInstanceKeys()) {`
Thus, the new boolean variables can be removed.

smolnar82: Hi @pzampino!

Yes, I'm fine with those new proposed variable names, they sound more accurate.

Thanks for your reply!

pzampino: @smolnar82 Thank you for the thoughtful review. I've made the agreed upon changes.

PR #964, opened 2024-11-20
KNOX-3061: Upgrade Bouncy Castle to 1.78

hanicz:

Upgrade Bouncy Castle to 1.78 due to 
- CVE-2023-33202
- CVE-2023-33201
- CVE-2024-29857
- CVE-2024-30171
- CVE-2024-30172
- CVE-2024-34447

## How was this patch tested?

Ran tests locally
Checked the dependency tree with: `mvn dependency:tree -Dincludes=org.bouncycastle | grep "org.bouncycastle"`
Checked the dep folder after ant install-test-home with: `ls -la *bp*`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -168,7 +168,8 @@
         <asm.version>9.0</asm.version>
         <aspectj.version>1.9.6</aspectj.version>
         <asynchttpclient.version>4.1.5</asynchttpclient.version>
-        <bcprov-jdk15on.version>1.67</bcprov-jdk15on.version>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -2344,6 +2368,10 @@
                         <groupId>dom4j</groupId>
                         <artifactId>dom4j</artifactId>
                     </exclusion>
+                    <exclusion>
```
pzampino, line 2371: The pac4j version is compatible with this later version of bouncycastle?

PR #971, opened 2024-12-01
KNOX-3051: Ability to extend classpath with configurable paths

hanicz:

**Classpath extension for patches**

It would be convenient to have the ability to easily pre-pend classes/JARs to the Knox classpath for quickly applying/testing patches. This approach is completely configurable and multiple locations can be configured although their order has to be kept in mind to avoid overrides. Placing any class or JAR in these locations will take precedence on the classpath as it will come first. These locations can be within the Knox installation, external to it or even a mix of them. 

The logic is checking the gateway-site.xml file for 'gateway.server.classpath.extension' property, if it is there it will extend the classpath with its value. The launcher only modifies the classpath for the GatewayServer. There is a check for the main class that's in the properties and if it is 'org.apache.knox.gateway.GatewayServer' only then the classpath is extended.

Example extension:

```
<property>
    <name>gateway.server.classpath.extension</name>
    <value>/new/classpath/jars/*.jar;../patch/*;/new/classpath/classes;</value>
</property>
```

- The different locations should be separated by ',' or ';'.
- '*.jar': picks up all jar files in the folder.
- '*': picks up all files in the folder.
- '/folder': class files are picked up from various folder hierarchies (eq. put GatewayServer.class in org/apache/knox/gateway folder) 

## How was this patch tested?

- Unit tests
- Tested locally with class and JAR files

### DIFF
--- a/gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/Extender.java
+++ b/gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/Extender.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.launcher;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Extender {
+
+    private static final String CLASSPATH_EXTENSION_PROPERTY = "gateway.server.classpath.extension";
+    private static final String CLASSPATH_PROPERTY_PATTERN = "<property>\\s*<name>" + CLASSPATH_EXTENSION_PROPERTY + "</name>\\s*<value>(.*?)</value>\\s*</property>";
+    private static final String CONFIG_FILE = "gateway-site.xml";
+    private static final String CONFIG_PATH = "../conf/" + CONFIG_FILE;
+    private static final String CLASS_PATH_PROPERTY = "class.path";
+    private static final String MAIN_CLASS_PROPERTY = "main.class";
+    private static final String GATEWAY_SERVER_MAIN_CLASS = "org.apache.knox.gateway.GatewayServer";
+    private static final String[] CLASS_PATH_DELIMITERS = new String[]{",", ";"};
+
+    private final File base;
+    private final Properties properties;
+
+    public Extender(File base, Properties properties) {
+        this.base = base;
+        this.properties = properties;
+    }
+
+    public void extendClassPathProperty() throws IOException {
+        Path configFilePath = Paths.get(base.getPath(), CONFIG_PATH);
+        if (GATEWAY_SERVER_MAIN_CLASS.equals(properties.getProperty(MAIN_CLASS_PROPERTY)) && Files.isReadable(configFilePath)) {
+            String configContent = new String(Files.readAllBytes(configFilePath), StandardCharsets.UTF_8);
+            extractExtensionPathIntoProperty(configContent);
+        }
+    }
+
+    protected void extractExtensionPathIntoProperty(String configContent) {
+        Pattern pattern = Pattern.compile(CLASSPATH_PROPERTY_PATTERN, Pattern.DOTALL);
+        Matcher matcher = pattern.matcher(configContent);
+
+        if (matcher.find()) {
+            StringBuilder newClassPath = new StringBuilder(matcher.group(1).trim());
+            if (newClassPath.length() > 0) {
+                if (!endsWithDelimiter(newClassPath.toString())) {
+                    newClassPath.append(CLASS_PATH_DELIMITERS[1]);
+                }
+                newClassPath.append(properties.getProperty(CLASS_PATH_PROPERTY));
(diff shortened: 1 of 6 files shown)


gateway-util-launcher/src/test/java/org/apache/knox/gateway/launcher/ExtenderTest.java
```diff
@@ -0,0 +1,195 @@
...
+        properties.setProperty("class.path", "classpath");
+        properties.setProperty("main.class", "org.apache.knox.gateway.GatewayServer");
+        Extender extender = new Extender(confDir.toFile(), properties);
+
+        String configContent = this.getConfigContent("/new/classp/*");
+        Files.write(configFilePath, configContent.getBytes(StandardCharsets.UTF_8));
+        extender.extendClassPathProperty();
+
+        assertEquals("/new/classp/*;classpath", properties.getProperty("class.path"));
+        this.cleanUpDirs();
```
smolnar82, line 50: What if an exception occurred before this line? No cleanup?
Why not annotate `setupDirs` and `cleanUpDirs` with `Before/After`? Even if they are not needed for some of the test cases, they don't cause any harm, and error handling is managed by OOTB.
What do you think?

gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/Extender.java
```diff
@@ -0,0 +1,77 @@
...
+
+        if (matcher.find()) {
+            StringBuilder newClassPath = new StringBuilder(matcher.group(1).trim());
+            if (newClassPath.length() > 0) {
+                if (!endsWithDelimiter(newClassPath.toString())) {
+                    newClassPath.append(CLASS_PATH_DELIMITERS[1]);
+                }
+                newClassPath.append(properties.getProperty(CLASS_PATH_PROPERTY));
+                properties.setProperty(CLASS_PATH_PROPERTY, newClassPath.toString());
+            }
```
smolnar82, line 70: Maybe adding a WARN log message here. So that users who explicitly configured something will not be surprised their configuration doesn't take effect.

gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/Extender.java
```diff
@@ -0,0 +1,77 @@
...
+        Path configFilePath = Paths.get(base.getPath(), CONFIG_PATH);
+        if (GATEWAY_SERVER_MAIN_CLASS.equals(properties.getProperty(MAIN_CLASS_PROPERTY)) && Files.isReadable(configFilePath)) {
+            String configContent = new String(Files.readAllBytes(configFilePath), StandardCharsets.UTF_8);
+            extractExtensionPathIntoProperty(configContent);
+        }
+    }
+
+    protected void extractExtensionPathIntoProperty(String configContent) {
+        Pattern pattern = Pattern.compile(CLASSPATH_PROPERTY_PATTERN, Pattern.DOTALL);
+        Matcher matcher = pattern.matcher(configContent);
```
smolnar82, line 60: nit: they can be `final`.

gateway-util-launcher/src/test/java/org/apache/knox/gateway/launcher/ExtenderTest.java
```diff
@@ -0,0 +1,195 @@
...
+                        "        <description>Enable/Disable websocket JWT validation at websocket layer.</description>\n" +
+                        "    </property>\n" +
+                        "\n" +
+                        "    <!-- @since 1.5.0 homepage logout -->\n" +
+                        "    <property>\n" +
+                        "        <name>knox.homepage.logout.enabled</name>\n" +
+                        "        <value>true</value>\n" +
+                        "        <description>Enable/disable logout from the Knox Homepage.</description>\n" +
+                        "    </property>\n" +
+                        "</configuration>";
```
smolnar82, line 168: You can add static XML content in `src/test/resources` and load them here to make the code more readable.

smolnar82 (changes requested):

gateway-util-launcher/src/test/java/org/apache/knox/gateway/launcher/ExtenderTest.java
hanicz: Yes, that was my thinking. They are not needed for every case, however I didn't think of error handling. I changed them to `After/Before`.

gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/Extender.java
hanicz: Changed them to final

gateway-util-launcher/src/test/java/org/apache/knox/gateway/launcher/ExtenderTest.java
hanicz: I moved the XML to the gateway-site-test.xml file in the test/resources folder.

gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/Extender.java
hanicz: I agree it would be nice to have some kind of feedback for the user. However there is no any kind of logging setup in the launcher, there is no log dependencies added either to keep it lightweight. At this point the log4j properties are not loaded. Only the standard output is available.

smolnar82: Ack.

smolnar82 (approved):
LGTM

smolnar82:
Hi @moresandeep , @pzampino - Could you please review this one? Thanks!

gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/Extender.java
```diff
@@ -0,0 +1,77 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.Properties;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class Extender {
```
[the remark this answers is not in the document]
hanicz: Renamed to GatewayServerClasspathExtender

gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/Extender.java
```diff
@@ -0,0 +1,77 @@
...
+    private final File base;
+    private final Properties properties;
+    private final Pattern pattern = Pattern.compile(CLASSPATH_PROPERTY_PATTERN, Pattern.DOTALL);
+
+    public Extender(File base, Properties properties) {
+        this.base = base;
+        this.properties = properties;
+    }
+
+    public void extendClassPathProperty() throws IOException {
```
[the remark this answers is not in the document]
hanicz: I removed the reference to the properties and moved it to method parameter.
I'd rather not add logic to the constructor.

pzampino (approved):
LGTM

PR #972, opened 2024-12-12
KNOX-3077 - Add pac4j.cookie.max.age param

moresandeep:
This PR introduces a new parameter `pac4j.cookie.max.age` for the Pac4J provider that enforces cookie age on the cookies created by the pac4j provider. 
e.g. 
```
         <provider>
                  <role>federation</role>
                  <name>pac4j</name>
                  <enabled>true</enabled>
                  <param>
                      <name>pac4j.callbackUrl</name>
                      <value>https://www.local.com:8443/gateway/knoxsso/api/v1/websso</value>
                  </param>

                 <!-- Unit is seconds -->
                  <param>
                      <name>pac4j.cookie.max.age</name>
                      <value>180</value>
                  </param>
            .........

```
Note that these value needs to be same as `knoxsso.cookie.max.age` if  you want Knox session to expire at the same time.

## How was this patch tested?

### DIFF
--- a/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
+++ b/gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
@@ -122,6 +122,12 @@ public class Pac4jDispatcherFilter implements Filter, SessionInvalidator {
 
   private static final String PAC4J_OIDC_TYPE = "oidc.type";
 
+  /* property for specifying pac4j cookies ttl */
+  public static final String PAC4J_COOKIE_MAX_AGE = "pac4j.cookie.max.age";
+
+  /* default value is same is KNOXSSO token ttl default */
+  public static final int PAC4J_COOKIE_MAX_AGE_DEFAULT = -1;
+
   private CallbackFilter callbackFilter;
 
   private SecurityFilter securityFilter;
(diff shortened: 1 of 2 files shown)


gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
```diff
@@ -122,6 +122,12 @@ public class Pac4jDispatcherFilter implements Filter, SessionInvalidator {
 
   private static final String PAC4J_OIDC_TYPE = "oidc.type";
 
+  /* property for specifying pac4j cookies ttl */
+  public static final String PAC4J_COOKIE_MAX_AGE = "pac4j.cookie.max.age";
+
+  /* default value is same is KNOXSSO token ttl default */
+  public static final int PAC4J_COOKIE_MAX_AGE_DEFAULT = -1;
```
smolnar82, line 129: This could - should - be `private` (not used anywhere outside of this class) and `String` instead of `int` (i.e. `"-1"`, to avoid unnecessary conversion below)

smolnar82 (changes requested):

gateway-provider-security-pac4j/src/main/java/org/apache/knox/gateway/pac4j/filter/Pac4jDispatcherFilter.java
moresandeep: Agreed! thanks for the review!

moresandeep: @smolnar82 actually  `PAC4J_COOKIE_MAX_AGE` is used outside of this class in  `KnoxSessionStore` which needs this to be public. 
I did update the `int` to `String` thanks for pointing it out!

smolnar82: @moresandeep - yes, I know the parameter name is used elsewhere, that's why I added a comment on the default value only 😊
Thanks for applying the changes so quickly.

smolnar82 (approved):
LGTM

smolnar82:
Please don't forget to file a Doc Jira with this new config.

PR #982, opened 2025-01-15
KNOX-3086: Add JVM exports for the Ranger plug-in to work with Java 17

pzampino:

Added JVM exports to allow the Ranger plug-in to continue to function with Java 17.

## How was this patch tested?

Unit and manual testing.


pzampino:
It looks like KNOX-3088 already took care of this.

PR #983, opened 2025-01-15
Revert KNOX-3006

pzampino:

This change reverts KNOX-3006, which replaced the libpam4j implementation in an attempt to resolve the intermittent garbage group names issue. However, this change apparently does not completely resolve that issue AND it introduces intermittent authentication failures.

## How was this patch tested?

Unit and manual testing.


smolnar82:
@wilbur4321 - see explanation above for your question in #840.

PR #988, opened 2025-01-23
KNOX-3085: Metadata API should return the certificate chain (if any) instead of only the configured Knox instance certificate

pzampino:

Modify the Metadata API to return the certificate chain when a load balancer is between Knox and its clients.

## How was this patch tested?

Unit tests and manual tests with a LB in front of Knox.


smolnar82 (approved):
LGTM.

PR #989, opened 2025-02-11
KNOX-3095:Added HBase UI proxying for Startup progress

hingu8103:
KNOX-3095 : Added Hbase UI proxying for Startup progress

Added rewrite rule for "Startup Progress" on HBase UI.

## How was this patch tested?

Copied the rewrite.xml file to a cluster, restarted Knox, and verified the URL is redirecting correctly.

### DIFF
--- a/gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml
+++ b/gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml
@@ -144,6 +144,9 @@
   <rule dir="OUT" name="HBASEUI/hbase/outbound/master/children" pattern="/namedQueueLog.jsp?{**}">
     <rewrite template="{$frontend[url]}/hbase/webui/master/namedQueueLog.jsp?host={$inboundurl[host]}?port={$inboundurl[port]}?{**}"/>
   </rule>
+  <rule dir="OUT" name="HBASEUI/hbase/outbound/master/children" pattern="/startupProgress.jsp?{**}">
+    <rewrite template="{$frontend[url]}/hbase/webui/master/startupProgress.jsp?host={$inboundurl[host]}?port={$inboundurl[port]}?{**}"/>
+  </rule>
 
   <!-- RegionServer UI proxying -->
   <rule dir="IN" name="HBASEUI/hbase/inbound/regionserver/root/qualified" pattern="*://*:*/**/hbase/webui/regionserver?{host}?port}">


smolnar82:
A more recent version of the HBase UI service definition is 2.5.10. Why is your change only applied in 2.1.0?

Cc. @stoty @tr0k

tr0k:
Hey @smolnar82 indeed it has been added later to 2.4.17 and 2.5.4
https://issues.apache.org/jira/browse/HBASE-21521
I will propose a fix for this tmrw. Thanks for spotting.

smolnar82:
Thanks, @tr0k, for getting back so quickly.

@hingu8103 - based on the above, the only option you have is to create a separate service definition for 2.5.4 and add your change there. Or, alternatively, you may want to upgrade Hbase to at least 2.5.10. In any case, you may also want to ensure in your topology that you are using the appropriate service definition by applying its version (see sample [here](https://knox.apache.org/books/knox-2-1-0/dev-guide.html#service.xml)).

vinayakphegde:
> I did notice that the rewrite rule for 2.5.10 differs from the 2.1.0 addition in this PR. Is that intentional?

Hmm, correct. `?host={$inboundurl[host]}?port={$inboundurl[port]}?{**}` is additional in the rewrite template here.

In general, we have this same difference between 2.1.0 and 2.5.10 UI rules. for example
2.1.0: [hbaseui/2.1.0/rewrite.xml#L129-#L146](https://github.com/apache/knox/blob/master/gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml#L129-#L146)
2.5.10: [hbaseui/2.5.10/rewrite.xml#L152-#L178](https://github.com/apache/knox/blob/master/gateway-service-definitions/src/main/resources/services/hbaseui/2.5.10/rewrite.xml#L152-#L178)

Are these parameters no longer needed in the latest versions? If so, what was the reason for this change?

pzampino:
> > I did notice that the rewrite rule for 2.5.10 differs from the 2.1.0 addition in this PR. Is that intentional?
> 
> Hmm, correct. `?host={$inboundurl[host]}?port={$inboundurl[port]}?{**}` is additional in the rewrite template here.
> 
> In general, we have this same difference between 2.1.0 and 2.5.10 UI rules. for example 2.1.0: [hbaseui/2.1.0/rewrite.xml#L129-#L146](https://github.com/apache/knox/blob/master/gateway-service-definitions/src/main/resources/services/hbaseui/2.1.0/rewrite.xml#L129-#L146) 2.5.10: [hbaseui/2.5.10/rewrite.xml#L152-#L178](https://github.com/apache/knox/blob/master/gateway-service-definitions/src/main/resources/services/hbaseui/2.5.10/rewrite.xml#L152-#L178)
> 
> Are these parameters no longer needed in the latest versions? If so, what was the reason for this change?

That is a component-specific question. Knox itself does not dictate the patterns for these rules.

smolnar82:
@tr0k - Since you added the new 2.5.10 service definition version, I think it's best that you answer the above question.

moresandeep:
Thanks for the patch @tr0k and @vinayakphegde! the patch looks good, if we can reach to a conclusion on weather to move them to a separate version folder or in the existing version we can commit it.

tr0k:
Hi @vinayakphegde I made it simpler since I don't need to explicitly pass these two parameters. Let me know if it's causing any issues. I have just tested it again and with or without it it works the same for me in 2.5.10.
I noticed however an orthogonal issue that I may open sth later for. In HA when Active becomes backup Knox redirects to the backup one, then one needs to click the URL of the active one to be redirected to the correct node however most of the subpages don't work. Let me know if you face this problem in 2.1.0 or other versions too. Cheers

pzampino:
@tr0k Thank you for this contribution.

vinayakphegde:
Thanks Everyone!

PR #991, opened 2025-02-17
KNOX-3097 Add more redirect.whitelist Test Cases for KNOXSSO

lmccay:
Additional unit tests for KNOXSSO and redirect whitelist handling.
These tests are currently in the WebSSOResourceTest class for some reason.
We should move them to a new RegexUtilsTest at some point.

## How was this patch tested?
With the new and existing tests.


lmccay:
Inadvertently included other changes.

smolnar82:
My two cents:
- the description suggest tests in the `KNOXSSOUT` service but those tests are added in the test cases for `KNOXSSO` ([WebSSOutResourceTest ](https://github.com/apache/knox/blob/master/gateway-service-knoxssout/src/test/java/org/apache/knox/gateway/service/knoxsso/WebSSOutResourceTest.java) vs [WebSSOResourceTest](https://github.com/apache/knox/blob/master/gateway-service-knoxsso/src/test/java/org/apache/knox/gateway/service/knoxsso/WebSSOResourceTest.java))
- these tests - within `WebSSOResourceTest` - are actually testing [`RegexUtils`](https://github.com/apache/knox/blob/de3c78cbc44098d273f520b0510cad9519843849/gateway-util-common/src/main/java/org/apache/knox/gateway/util/RegExUtils.java). I wonder if it was better to have those test cases in a new `RegExUtilsTest` class :bulb:

moresandeep:
> My two cents:
> 
> * the description suggest tests in the `KNOXSSOUT` service but those tests are added in the test cases for `KNOXSSO` ([WebSSOutResourceTest ](https://github.com/apache/knox/blob/master/gateway-service-knoxssout/src/test/java/org/apache/knox/gateway/service/knoxsso/WebSSOutResourceTest.java) vs [WebSSOResourceTest](https://github.com/apache/knox/blob/master/gateway-service-knoxsso/src/test/java/org/apache/knox/gateway/service/knoxsso/WebSSOResourceTest.java))
> * these tests - within `WebSSOResourceTest` - are actually testing [`RegexUtils`](https://github.com/apache/knox/blob/de3c78cbc44098d273f520b0510cad9519843849/gateway-util-common/src/main/java/org/apache/knox/gateway/util/RegExUtils.java). I wonder if it was better to have those test cases in a new `RegExUtilsTest` class 💡

I think so too, i don't see a test for RegexUtils class, so perhaps create a new one and add these tests to it?

lmccay:
Ahh, very true on the WEBSSOUT point. I'll see if I can change that.
I just tracked down where the RegexUtils tests were already and added on there.
We can move them all in a follow on patch, I think.

pzampino (approved):

PR #992, opened 2025-02-19
KNOX-3094 - Update CM API swagger to 7.13.1

bonampak:

Upgrade com.cloudera.api.swagger:cloudera-manager-api-swagger to 7.13.1.
Update code for okhttp3. Resolve dependency convergence issues.

## How was this patch tested?

Unit and integration tests, deployed to a dev cluster and restarted Knox to verify discovery and knox ui works.

### DIFF
--- a/gateway-spi/src/main/java/org/apache/knox/gateway/util/TruststoreSSLContextUtils.java
+++ b/gateway-spi/src/main/java/org/apache/knox/gateway/util/TruststoreSSLContextUtils.java
@@ -20,8 +20,12 @@ import java.security.KeyManagementException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
 
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+import javax.net.ssl.X509TrustManager;
 
 import org.apache.http.ssl.SSLContextBuilder;
 import org.apache.http.ssl.SSLContexts;
@@ -46,6 +50,27 @@ public class TruststoreSSLContextUtils {
       LOGGER.failedToLoadTruststore(e.getMessage(), e);
     }
     return sslContext;
+  }
+
+  public static X509TrustManager getTrustManager(KeyStore truststore) {
+    try {
+      if (truststore != null) {
+        TrustManagerFactory trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+        trustManagerFactory.init(truststore);
+        TrustManager[] trustManagers = trustManagerFactory.getTrustManagers();
+        if (trustManagers != null) {
+          for (TrustManager tm : trustManagers) {
+            if (tm instanceof X509TrustManager) {
+              return (X509TrustManager) tm;
+            }
+          }
+        }
+        throw new IllegalStateException("Unexpected default trust managers:" + Arrays.toString(trustManagers));
+      }
+    } catch (KeyStoreException | NoSuchAlgorithmException | IllegalStateException e) {
+      LOGGER.failedToLoadTruststore(e.getMessage(), e);
+    }
+    return null;
   }
 
 }
--- a/gateway-discovery-cm/pom.xml
+++ b/gateway-discovery-cm/pom.xml
@@ -66,15 +66,9 @@
         <dependency>
             <groupId>com.cloudera.api.swagger</groupId>
             <artifactId>cloudera-manager-api-swagger</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>io.swagger</groupId>
-                    <artifactId>swagger-annotations</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
-            <groupId>com.squareup.okhttp</groupId>
+            <groupId>com.squareup.okhttp3</groupId>
             <artifactId>okhttp</artifactId>
         </dependency>
         <dependency>
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/DiscoveryApiClient.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/DiscoveryApiClient.java
@@ -26,14 +26,13 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import javax.net.ssl.SSLContext;
+import javax.net.ssl.X509TrustManager;
 import javax.security.auth.Subject;
 
 import com.cloudera.api.swagger.client.ApiClient;
-import com.cloudera.api.swagger.client.Pair;
-import com.cloudera.api.swagger.client.auth.Authentication;
-import com.cloudera.api.swagger.client.auth.HttpBasicAuth;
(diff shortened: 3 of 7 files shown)


gateway-spi/src/main/java/org/apache/knox/gateway/util/TruststoreSSLContextUtils.java
```diff
@@ -48,4 +52,25 @@ public static SSLContext getTruststoreSSLContext(KeyStore truststore) {
...
+        trustManagerFactory.init(truststore);
+        TrustManager[] trustManagers = trustManagerFactory.getTrustManagers();
+        if (trustManagers != null) {
+          for (TrustManager tm : trustManagers) {
+            if (tm instanceof X509TrustManager) {
+              return (X509TrustManager) tm;
+            }
+          }
+        }
+        throw new IllegalStateException("Unexpected default trust managers:" + Arrays.toString(trustManagers));
```
pzampino, line 68: What is the consequence of throwing this exception? The DiscoveryApiClient#configureSsl() method does not catch it, so it will bubble up. Do you know where it will eventually be handled and how?

smolnar82:
According to the PR description, this change is supposed to upgrade CM and OkHttp versions. However, I see that many others were touched:
image
Why are these new dependencies needed (particularly the Kotlin one)?

gateway-spi/src/main/java/org/apache/knox/gateway/util/TruststoreSSLContextUtils.java
bonampak: I took this part of the migration from here:

https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/-builder/ssl-socket-factory.html
I would expect this to work in a standard setup (and it's the case with JDK 8-17).
I have not tested it on a FIPS-enabled cluster though.

I throw and catch IllegalStateException and return null if the trust manager for the default algorithm and trust store is not an instance of X509TrustManager. The `getTruststoreSSLContext()` method also uses the same methods by calling SSLContextBuilder.loadTrustMaterial(): uses the default TrustManagerFactory algorithm, initializes and gets the trust managers; it does not check the instance type. But the okhttp API expects an implementation of javax.net.ssl.X509TrustManager to validate the server's certificates.

The deprecated method variant does not require it, okhttp would use reflection to get one from `sun.security.ssl.SSLContextImpl`:

https://github.com/square/okhttp/blob/4984568367caaf359b82c452bd28b5e192824d1c/okhttp/src/main/kotlin/okhttp3/internal/platform/Platform.kt#L88
But this was removed in Okhttp 5.

bonampak:
Okhttp 4.x is written in Kotlin and needs the kotlin-stdlib artifacts.

The com.cloudera.api.swagger:cloudera-manager-api-swagger:7.13.1 brings in
```
<gson-fire-version>1.8.5</gson-fire-version>
<swagger-core-version>2.0.0</swagger-core-version>
<okhttp-version>4.10.0</okhttp-version>
<gson-version>2.10.1</gson-version>
```

Okhttp 4.10 is not the latest, but the swagger-codegen library only has this option.

From the CM part, they would have liked to set the okhttp3 version to 4.12. but swagger-codegen is bringing internally version 4.10
https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/JavaClientCodegen.java#L89

https://github.com/swagger-api/swagger-codegen/issues/12337
(It seems that openapi-generator with okhttp 4.12 - gson or with httpclient 5 could be used instead of swagger-codegen.) 

So we need to bump okhttp to 4.12.0, and I was also asked to upgrade gson-fire to 1.9.0; the latter has a gson:2.10.1 dependency (we had 2.8.9).

The kotlin libraries are upgraded to 1.9.10 because of dependency convergence issues:

```
Dependency convergence error for org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.8.21 paths to dependency are:
+-org.apache.knox:gateway-discovery-cm:2.1.0-SNAPSHOT
  +-com.cloudera.api.swagger:cloudera-manager-api-swagger:7.13.1
    +-com.squareup.okhttp3:logging-interceptor:4.12.0
      +-org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.8.21
and
+-org.apache.knox:gateway-discovery-cm:2.1.0-SNAPSHOT
  +-com.squareup.okhttp3:okhttp:4.12.0
    +-com.squareup.okio:okio:3.6.0
      +-com.squareup.okio:okio-jvm:3.6.0
        +-org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.10
and
+-org.apache.knox:gateway-discovery-cm:2.1.0-SNAPSHOT
  +-com.squareup.okhttp3:okhttp:4.12.0
    +-org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.8.21
```

https://github.com/square/okhttp/issues/8288

"OkHttp 4.12.0 expects you to use any version of kotlin greater than or equal to 1.8.21. We don't rush out a release each time okio releases with a newer Kotlin. This isn't a thing we worry about at all." ... "It's already fixed in the 5.0 alphas.".

Due to these issues, teams either remain on the latest non-kotlin version or remove okhttp:
https://issues.apache.org/jira/browse/HADOOP-18496 upgrade kotlin-stdlib due to CVEs
https://issues.apache.org/jira/browse/HADOOP-18890 Remove use of okhttp in runtime code.

org.jetbrains:annotations also needed to be resolved:
```
Dependency convergence error for org.jetbrains:annotations:16.0.2 paths to dependency are:
+-org.apache.knox:gateway-discovery-cm:2.1.0-SNAPSHOT
  +-org.apache.knox:gateway-server:2.1.0-SNAPSHOT
    +-org.jetbrains.pty4j:pty4j:0.11.4
      +-org.jetbrains:annotations:16.0.2
and
+-org.apache.knox:gateway-discovery-cm:2.1.0-SNAPSHOT
  +-com.squareup.okhttp3:okhttp:4.12.0
    +-org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.10
      +-org.jetbrains.kotlin:kotlin-stdlib:1.9.10
        +-org.jetbrains:annotations:13.0
```
I excluded it from kotlin-stdlib because by simply adding to dependencyManagement would cause it to become a compile-time dependency on pty4j.

The okio dependency was pinned to 3.6.0 because of this:

```
Dependency convergence error for com.squareup.okio:okio:1.6.0 paths to dependency are:
+-org.apache.knox:gateway-test-release:2.1.0-SNAPSHOT
  +-org.apache.hadoop:hadoop-hdfs-client:3.2.4
    +-com.squareup.okhttp:okhttp:2.7.5
      +-com.squareup.okio:okio:1.6.0
and
+-org.apache.knox:gateway-test-release:2.1.0-SNAPSHOT
  +-org.apache.knox:gateway-release:2.1.0-SNAPSHOT
    +-org.apache.knox:gateway-discovery-cm:2.1.0-SNAPSHOT
      +-com.squareup.okhttp3:okhttp:4.12.0
        +-com.squareup.okio:okio:3.6.0
```
hadoop-hdfs-client is a test dependency (so the old okhttp:2.7.5 will not be in our release artifact dependencies) and okio is backwards-compatible:
https://github.com/square/okio/issues/1323

smolnar82:
> because of dependency convergence issues
Isn't it possible to exclude them everywhere? I'd like to avoid pulling in more 3rd party dependencies that are not essential for Knox to work. They are expensive to maintain in the future (CVEs, possible incompatibility, etc...).

PR #995, opened 2025-02-25
KNOX-3099: Ability to exclude topologies form client auth needed policy

hanicz:

Currently if the user adds `gateway.client.auth.needed` as true to the gateway-site.xml every topology will enforce this. There is no option to exclude from this policy. The user can also specify gateway.client.auth.wanted as true however the policy is not enforced in this case.

We need the ability to be able to exclude topologies in case the client auth is needed.

Example setup:

By adding the below to the gateway-site.xml we enable the client auth needed policy. To be able to exclude a topology from this we have to define a port mapping for that topology and add it to the `gateway.client.auth.exclude` as well.

    <property>
        <name>gateway.client.auth.needed</name>
        <value>true</value>
    </property>
    <property>
        <name>gateway.port.mapping.health</name>
        <value>9443</value>
    </property>
    <property>
        <name>gateway.client.auth.exclude</name>
        <value>health</value>
    </property>

## How was this patch tested?

Unit tests
Tested manually on my local setup

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
@@ -237,6 +237,17 @@ public class JettySSLService implements SSLService {
   }
 
   @Override
+  public void excludeTopologyFromClientAuth(SslContextFactory sslContextFactory, GatewayConfig config, String topologyName) {
+    if(config.getClientAuthExclude() != null && topologyName != null
+            && config.isClientAuthNeeded() && config.getClientAuthExclude().contains(topologyName)) {
+      SslContextFactory.Server sslContextFactoryServer = (SslContextFactory.Server) sslContextFactory;
+      sslContextFactoryServer.setNeedClientAuth(false);
+      sslContextFactoryServer.setWantClientAuth(true);
+      log.topologyExcludedFromClientAuth(topologyName);
+    }
+  }
+
+  @Override
   public void start() throws ServiceLifecycleException {
   }
 
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
@@ -238,8 +238,7 @@ public class JettySSLService implements SSLService {
 
   @Override
   public void excludeTopologyFromClientAuth(SslContextFactory sslContextFactory, GatewayConfig config, String topologyName) {
-    if(config.getClientAuthExclude() != null && topologyName != null
-            && config.isClientAuthNeeded() && config.getClientAuthExclude().contains(topologyName)) {
+    if(config.isClientAuthNeeded() && config.isTopologyExcludedFromClientAuth(topologyName)) {
       SslContextFactory.Server sslContextFactoryServer = (SslContextFactory.Server) sslContextFactory;
       sslContextFactoryServer.setNeedClientAuth(false);
       sslContextFactoryServer.setWantClientAuth(true);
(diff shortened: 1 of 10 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
```diff
@@ -236,6 +236,17 @@ public Object buildSslContextFactory(GatewayConfig config) throws AliasServiceEx
     return sslContextFactory;
   }
 
+  @Override
+  public void excludeTopologyFromClientAuth(SslContextFactory sslContextFactory, GatewayConfig config, String topologyName) {
+    if(config.getClientAuthExclude() != null && topologyName != null
+            && config.isClientAuthNeeded() && config.getClientAuthExclude().contains(topologyName)) {
```
smolnar82, line 242: Most of these conditions could be implemented in GatewayConfigImpl to be self-contained. For instance:
`config.isClientAuthExclude(topologyName)` could return `true` if the desired configs are set and the topology is meant to be excluded.

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
```diff
@@ -34,6 +34,7 @@
 import org.apache.knox.gateway.services.security.AliasService;
 import org.apache.knox.gateway.services.security.AliasServiceException;
 import org.apache.knox.gateway.services.security.KeystoreService;
+import org.easymock.EasyMock;
```
smolnar82, line 37: nit: all EasyMock methods are imported in a static way in this test class (see above) you may want to follow this pattern.

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
```diff
@@ -486,4 +544,19 @@ private GatewayConfig createGatewayConfig(boolean isClientAuthNeeded, boolean is
     return config;
   }
 
+  private GatewayConfig createGatewayConfigForExclude(boolean isClientAuthNeeded, String exclude) {
```
smolnar82, line 547: nit: the `exclude` parameter could be `topologiesToExclude` (or similar) and the method name could be `...ForExcludeTopologiesFromClientAuth`

smolnar82 (changes requested):

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
hanicz: done

gateway-server/src/test/java/org/apache/knox/gateway/services/security/impl/JettySSLServiceTest.java
hanicz: renamed the method

gateway-server/src/test/java/org/apache/knox/gateway/GatewayGlobalConfigTest.java
```diff
@@ -67,11 +69,20 @@ public void testSiteConfig() {
...
     assertThat( config.getTruststorePath(), is("./gateway-trust.jks"));
     assertThat( config.getTruststoreType(), is( "PKCS12" ) );
     assertThat( config.getKeystoreType(), is(KeyStore.getDefaultType()) );
   }
 
+  @Test
+  public void testSiteConfigWithDifferentTopologyExcluded() {
+    System.setProperty( GatewayConfigImpl.GATEWAY_HOME_VAR, getHomeDirName( "conf-site/conf/gateway-site.xml" ) );
+    GatewayConfig config = new GatewayConfigImpl();
+    assertThat( config.isClientAuthNeeded(), is( true ) );
```
smolnar82, line 82: nit: `assertTrue` might have been better,  but doesn't prevent the merge this as-is

lmccay (approved):
LGTM +1

This does assume that any existing consumers of something like the health check topology will need to be adjusted upon changing this. Let's make sure that this is called out in the docs for this improvement.

lmccay:
Thank you for your contribution, @hanicz this will be available in the next release!

PR #996, opened 2025-02-25
KNOX-3101 - hash key for RemoteAuthProvider cache

lmccay:

The initial implementation of RemoteAuthProvider caches authenticated Subjects locally based on the header that contained the credentials. While the cache is designed to provide only a few mins of caching, it is less than ideal to use the credentials as keys. This needs to be strengthened to use a hash as to not inadvertently risk leaking the credentials.

This will require some overhead involved in the hashing so we may need to find something else but we shouldn't use the credentials themselves. We would normally have to do a hash for implementing authentication for things like RDMS or LDAP based passwords, etc.

## How was this patch tested?

Refactored and ran existing unit tests to ensure no regression was introduced.


smolnar82 (approved):
LGTM

PR #997, opened 2025-02-26
KNOX-3100 - RemoteAuthProvider to accept multiple group headers

lmccay:

Initial iteration of the RemoteAuthProvider has support for a single HTTP Header containing the group memberships. The Pre API in the Knox Auth Service breaks memberships up across multiple headers to manage the size.

We need to determine a syntax as an explicit list or a range with some builtin syntax of a numeric suffix. Otherwise, it will need a list of possible header names with it accommodating the largest set which seems odd.

## How was this patch tested?

Ran existing unit tests plus new tests for the new functionality.


moresandeep (approved):
Looks good!

PR #998, opened 2025-02-27
KNOX-3103 - Move RegEx Tests from KnoxSSO module to RegExUtils

lmccay:

Recently added a test to the KnoxSSO suite for additional tests for the regex whitelist for redirects. The regex tests actually belong with the regexutils class. This patch will move them there as appropriate.

## How was this patch tested?

Unit tests run in new location and all existing tests run.


smolnar82 (approved):
LGTM; shipt it :)

PR #1002, opened 2025-03-05
KNOX-3107 - Alias Support for the RemoteAuthProvider truststore password

lmccay:

Added Alias support for the truststore password.
Also moved the creation and load of the truststore to the init method rather than the request processing path.
This required an adjustment in the way that the topology level truststore tests were written and the expectations to fail fast at init time which broke previous expectations of the request results reflecting truststore issues or success.

## How was this patch tested?

Existing unit tests and new test approaches for previously provided local truststore.
Manual run with separate Knox instance and export cert as the truststore.

### DIFF
--- a/gateway-provider-security-authc-remote/src/main/java/org/apache/knox/gateway/filter/RemoteAuthFilter.java
+++ b/gateway-provider-security-authc-remote/src/main/java/org/apache/knox/gateway/filter/RemoteAuthFilter.java
@@ -33,6 +33,8 @@ import org.apache.knox.gateway.security.GroupPrincipal;
 import org.apache.knox.gateway.security.PrimaryPrincipal;
 import org.apache.knox.gateway.services.GatewayServices;
 import org.apache.knox.gateway.services.ServiceType;
+import org.apache.knox.gateway.services.security.AliasService;
+import org.apache.knox.gateway.services.security.AliasServiceException;
 import org.apache.knox.gateway.services.security.KeystoreService;
 import org.apache.knox.gateway.services.security.KeystoreServiceException;
 import org.apache.logging.log4j.ThreadContext;
@@ -98,9 +100,7 @@ public class RemoteAuthFilter implements Filter {
           AuditConstants.DEFAULT_AUDITOR_NAME, AuditConstants.KNOX_SERVICE_NAME, AuditConstants.KNOX_COMPONENT_NAME );
   private final RemoteAuthMessages LOGGER = MessagesFactory.get( RemoteAuthMessages.class );
 
-  private String truststorePath;
-  private String truststorePassword;
-  private String truststoreType;
+  private KeyStore trustStore;
 
   @Override
   public void init(FilterConfig filterConfig) throws ServletException {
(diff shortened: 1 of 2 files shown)


gateway-provider-security-authc-remote/src/main/java/org/apache/knox/gateway/filter/RemoteAuthFilter.java
```diff
@@ -132,12 +132,69 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
+    ServletContext context = filterConfig.getServletContext();
+    if (context != null) {
+      String topologyName = (String) context.getAttribute(GatewayServices.GATEWAY_CLUSTER_ATTRIBUTE);
+      GatewayServices services = (GatewayServices) context.getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
+      if (services != null) {
+        try {
+          final AliasService aliasService =  services.getService(ServiceType.ALIAS_SERVICE);
+          if (truststorePath != null && !truststorePath.isEmpty()) {
+            if (truststorePassword == null || truststorePassword.isEmpty()) {
+              // let's check the for an alias given the intent to specify a truststore path
```
pzampino, line 155: nit: comment typo

gateway-provider-security-authc-remote/src/main/java/org/apache/knox/gateway/filter/RemoteAuthFilter.java
```diff
@@ -132,12 +132,69 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
+              }
+            }
+          }
+          KeystoreService keystoreService = services.getService(ServiceType.KEYSTORE_SERVICE);
+          trustStore = getTrustStore(truststorePath, truststoreType, truststorePassword, keystoreService);
+        } catch (AliasServiceException | IOException e) {
+          throw new ServletException("Error while initializing RemoteAuthProvider", e);
+        }
+      }
+    } else if (truststorePath != null && !truststorePath.isEmpty()) {
```
pzampino, line 172: Do we get here if the ServletContext cannot be resolved? Seems like perhaps, rather than an else clause, there should be a check to see if we were able to initialize the trust store.

lmccay: Good catch. Working on a fix.

gateway-provider-security-authc-remote/src/main/java/org/apache/knox/gateway/filter/RemoteAuthFilter.java
lmccay: done.

gateway-provider-security-authc-remote/src/main/java/org/apache/knox/gateway/filter/RemoteAuthFilter.java
lmccay: done.

PR #1004, opened 2025-03-06
KNOX-3108: Append classpath with additional paths

hanicz:

[KNOX-3047](https://issues.apache.org/jira/projects/KNOX/issues/KNOX-3047) Extended the classpath with /usr/share/java/*.jar however it could cause issues if other data-services use the same location for their jars. This PR iterates on the changes introduced in [KNOX-3051](https://issues.apache.org/jira/projects/KNOX/issues/KNOX-3051) and makes it possible to prepend and/or append paths to the classpath dynamically using the `gateway-site.xml` configuration file.

To append new path(s) to the classpath add the below property to `gateway-site.xml`:

```
<property>
        <name>gateway.server.append.classpath</name>
        <value>/new/append/path/*.jar</value>
</property>
```

To prepend new path(s) to the classpath add the below property to `gateway-site.xml`:

```
<property>
        <name>gateway.server.prepend.classpath</name>
        <value>/new/prepend/path/*.jar</value>
</property>
```

The different locations should be separated by ',' or ';'.
- '*.jar': picks up all jar files in the folder.
- '*': picks up all files in the folder.
- '/folder': class files are picked up from various folder hierarchies (eq. put GatewayServer.class in org/apache/knox/gateway folder)

## How was this patch tested?

Unit tests
Manual tested locally


moresandeep (approved):
Looks good.

PR #1007, opened 2025-03-20
KNOX-3111: Gobal config for HSTS headers

hanicz:

This PR introduces the ability to enable `Strict-Transport-Security` header globally for all responses (even 404). Currently there is a topology wide configuration in the WebAppSec provider. The users now can enable it for only specific topologies or globally as well. If both is enabled and there is a request for that specific topology the WebAppSec configuration will take precedence. 

To enable the global configuration the below is required in the gateway-site.xml configuration file. The `option` config is optional, the default value is `max-age=31536000` (Its the same for the WebAppSec provider as well).

```
<property>
    <name>gateway.strict.transport.enabled</name>
    <value>true</value>
</property>
<property>
    <name>gateway.strict.transport.option</name>
    <value>max-age=31</value>
</property>
```

## How was this patch tested?

Unit tests.
Manual tests locally.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayMessages.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayMessages.java
@@ -809,4 +809,7 @@ public interface GatewayMessages {
 
   @Message( level = MessageLevel.INFO, text = "Excluded \"{0}\" topology from client auth" )
   void topologyExcludedFromClientAuth( String topologyName );
+
+  @Message( level = MessageLevel.DEBUG, text = "Strict-Transport-Security header enabled with \"{0}\" option" )
+  void strictTransportHeaderEnabled(String option);
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/GatewayServer.java
@@ -33,6 +33,7 @@ import org.apache.knox.gateway.config.impl.GatewayConfigImpl;
 import org.apache.knox.gateway.deploy.DeploymentException;
 import org.apache.knox.gateway.deploy.DeploymentFactory;
 import org.apache.knox.gateway.filter.CorrelationHandler;
+import org.apache.knox.gateway.filter.HSTSHandler;
 import org.apache.knox.gateway.filter.PortMappingHelperHandler;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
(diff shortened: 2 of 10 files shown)


hanicz:
> How does this affect behavior when topology-level config exists for the same?

If both are enabled and there is a request for that specific topology the WebAppSec configuration will take precedence.

pzampino:
> > How does this affect behavior when topology-level config exists for the same?
> 
> If both are enabled and there is a request for that specific topology the WebAppSec configuration will take precedence.

Is there a test for that?

hanicz:
> > > How does this affect behavior when topology-level config exists for the same?
> > 
> > 
> > If both are enabled and there is a request for that specific topology the WebAppSec configuration will take precedence.
> 
> Is there a test for that?

No there isn't, I validated the behaviour manually. The handler and the StrictTransportFilter are in two different modules and are called at different points of the requests lifecycle. What I can do is mock a response object and call the handle and doFilter methods with it and verify after. 

The setHeader method is used in the StrictTransportFilter which will override the existing header.

pzampino:
> > > > How does this affect behavior when topology-level config exists for the same?
> > > 
> > > 
> > > If both are enabled and there is a request for that specific topology the WebAppSec configuration will take precedence.
> > 
> > 
> > Is there a test for that?
> 
> No there isn't, I validated the behaviour manually. The handler and the StrictTransportFilter are in two different modules and are called at different points of the requests lifecycle. What I can do is mock a response object and call the handle and doFilter methods with it and verify after.
> 
> The setHeader method is used in the StrictTransportFilter which will override the existing header.

I think the test you've proposed is better than only manual testing. Thank you.

hanicz:
> > > > > How does this affect behavior when topology-level config exists for the same?
> > > > 
> > > > 
> > > > If both are enabled and there is a request for that specific topology the WebAppSec configuration will take precedence.
> > > 
> > > 
> > > Is there a test for that?
> > 
> > 
> > No there isn't, I validated the behaviour manually. The handler and the StrictTransportFilter are in two different modules and are called at different points of the requests lifecycle. What I can do is mock a response object and call the handle and doFilter methods with it and verify after.
> > The setHeader method is used in the StrictTransportFilter which will override the existing header.
> 
> I think the test you've proposed is better than only manual testing. Thank you.

Added new tests for the scenario

pzampino (approved):
LGTM

pzampino:
Thank you for adding the tests.

PR #1011, opened 2025-03-30
KNOX-3112 - Add an API for CLIENT_ID and SECRET based on KNOXTOKEN API

lmccay:

This is a specialized extension of the KNOXTOKEN API to request CLIENT_ID and CLIENT_SECRET.
This requires the use of Passcode tokens as the CLIENT_SECRET and the TokenID as the CLIENT_ID.

One of the motivations for this is to provide this extension so that it can be deployed separately within the same topology as the KNOXTOKEN API. Another is to codify the conventions being used in order to use the pattern described above for CLIENT_ID and CLIENT_SECRET without having to leave it to the consumer to interpret how to do so.

## How was this patch tested?

New unit tests were added for the API extension and run along with all existing unit tests.
Manual testing of the use of CLIENT_ID and CLIENT_SECRET was done with the JWTProvider.

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ClientCredentialsResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ClientCredentialsResource.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file to
+ * you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.service.knoxtoken;
+
+import org.apache.knox.gateway.services.security.token.TokenMetadata;
+import org.apache.knox.gateway.services.security.token.TokenMetadataType;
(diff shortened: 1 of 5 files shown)


gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/deploy/ClientCredentialsServiceDeploymentContributor.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Override
+    protected String[] getPackages() {
+      return new String[]{ "org.apache.knox.gateway.service.knoxtoken" };
+    }
+
+    @Override
+    protected String[] getPatterns() {
+       return new String[]{ "clientid/api/**?**" };
```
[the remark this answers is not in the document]
lmccay: This is fine and aligned with our other pattern definitions across APIs in Knox

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ClientCredentialsResource.java
```diff
@@ -75,7 +75,7 @@ public Response getAuthenticationToken() {
 
         TokenResponseContext resp = getTokenResponse(context);
         if (resp.responseMap != null) {
-            String passcode = resp.responseMap.passcode;
+            String passcode = (String) resp.responseMap.map.get(PASSCODE);
```
pzampino, line 78: Why do you not need to reference the token id similarly?

lmccay: Yeah, this was the source of the initial bug in the implementation. For some reason the tokenid and the passcode are surfaced at the top level but the passcode there is just the passcode portion of what ends up being the client_secret which contains both. I didn't really dig into why there are two levels but the one inside the map is put together and base64 encoded properly for use where the other is just the UUID format which can't be used directly.

lmccay: I appears to be related to the fact that when there is no token state server configured that the passcode field isn't generated since it can't be persisted and it requires server side state. Why the actual passcode UUID is generated and returned in the out response is still not clear.

pzampino (approved):
LGTM

PR #1016, opened 2025-04-06
KNOX-3119 - Provide appropriate defaults for CLIENTID extension of KN…

lmccay:
Provide appropriate defaults for CLIENTID extension of KNOXTOKEN API

There are a number of params that should be set by default for the CLIENTID extension of the KNOXTOKEN API. This will reduce the likelihood of errors based on having to explicitly configure these params which are either required or the logical defaults.

1. knox.token.exp.server-managed should be enabled by default this service if based on passcode tokens and requires server state
2. knox.token.ttl - should be -1 by default to not expire - fight me! 
3. knox.token.hash.key should be set by default as well. We need to check whether this already exists via the AliasService and if not we should generate it automatically or fail to deploy with appropriate error/exception.

This required the creation of a hook into the KNOXTOKEN resource to allow extensions to wrap the servletContext in order to provide defaults appropriate to the extension that would otherwise require the deploying admin to configure the same properties as convention for every deployment.

## How was this patch tested?

New unit tests added and existing tests run.
Manually tested.

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ClientCredentialsResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ClientCredentialsResource.java
@@ -17,17 +17,31 @@
  */
 package org.apache.knox.gateway.service.knoxtoken;
 
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.jwk.OctetSequenceKey;
+import com.nimbusds.jose.jwk.gen.OctetSequenceKeyGenerator;
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.services.GatewayServices;
+import org.apache.knox.gateway.services.ServiceType;
+import org.apache.knox.gateway.services.security.AliasService;
+import org.apache.knox.gateway.services.security.AliasServiceException;
 import org.apache.knox.gateway.services.security.token.TokenMetadata;
 import org.apache.knox.gateway.services.security.token.TokenMetadataType;
+import org.apache.knox.gateway.services.security.token.TokenStateService;
+import org.apache.knox.gateway.services.security.token.impl.TokenMAC;
 import org.apache.knox.gateway.util.JsonUtils;
 
 import javax.inject.Singleton;
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Response;
 import java.util.HashMap;
+import java.util.UUID;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.MediaType.APPLICATION_XML;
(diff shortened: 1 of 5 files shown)


gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ClientCredentialsResource.java
```diff
@@ -40,6 +54,47 @@ public class ClientCredentialsResource extends TokenResource {
     public static final String CLIENT_ID = "client_id";
     public static final String CLIENT_SECRET = "client_secret";
 
+    private GatewayServices services;
+
+    @Override
+    protected ServletContext wrapContextForDefaultParams(ServletContext context) throws ServletException {
+        ServletContext wrapperContext = new ServletContextWrapper(context);
+        wrapperContext.setInitParameter(TokenStateService.CONFIG_SERVER_MANAGED, "true");
+        wrapperContext.setInitParameter(TokenResource.TOKEN_TTL_PARAM, "-1");
```
pzampino, line 63: This could never be set otherwise then, right?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ServletContextWrapper.java
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    @Override
+    public String getServerInfo() {
+        return delegate.getServerInfo();
+    }
+
+    @Override
+    public String getInitParameter(String name) {
+        String value = delegate.getInitParameter(name);
```
pzampino, line 152: So, anything configured in the topology should override the defaults explicitly set in the wrapContextForDefaultParams method?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ServletContextWrapper.java
```diff
@@ -0,0 +1,324 @@
...
+
+    @Override
+    public Enumeration<String> getInitParameterNames() {
+        // Add all keys from the overriddenParams map
+        Set<String> combinedNames = new HashSet<>(defaultParams.keySet());
+
+        // Add all parameter names from the delegate context
+        Enumeration<String> delegateNames = delegate.getInitParameterNames();
+        while (delegateNames.hasMoreElements()) {
+            combinedNames.add(delegateNames.nextElement());
```
pzampino, line 167: HashSet eliminates the possibility of duplicate entries, right?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ServletContextWrapper.java
lmccay: Yes.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ClientCredentialsResource.java
lmccay: Topology level config overrides the defaults.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -219,8 +220,13 @@ public int toInt() {
...
   }
 
+  protected ServletContext wrapContextForDefaultParams(ServletContext context) throws ServletException {
+    return context;
+  }
+
   @PostConstruct
-  public void init() throws AliasServiceException, ServiceLifecycleException, KeyLengthException {
+  public void init() throws AliasServiceException, ServiceLifecycleException, KeyLengthException, ServletException {
+    ServletContext context = wrapContextForDefaultParams(this.context);
```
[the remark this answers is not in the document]
lmccay: I guess not - I was considering whether there was some concurrency issue or the like but there shouldn't be.
I'll change that.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/ServletContextWrapper.java
lmccay: I'll add to the test to make sure.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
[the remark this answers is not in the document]
lmccay: addressed.

pzampino (approved):
LGTM

PR #1018, opened 2025-04-07
KNOX-3122:Gateway config path is read from system property first during classpath extension

hanicz:

This change is an improvement on [KNOX-3108](https://issues.apache.org/jira/browse/KNOX-3108) (classpath extension). The gateway config file's location is read from a system property first during classpath extension. The solution is similar to the one we already have when we look for the config file on GatewayConfigImpl class.

## How was this patch tested?

Unit tests
Manually tested


moresandeep (approved):
Looks good

gateway-util-launcher/src/main/java/org/apache/knox/gateway/launcher/GatewayServerClasspathExtender.java
```diff
@@ -99,4 +100,19 @@ private boolean startsWithDelimiter(String path) {
...
+
+    private String getVar(String variableName, String defaultValue) {
+        String value = System.getProperty(variableName);
+        if (value == null) {
+            value = System.getenv(variableName);
+        }
+        if (value == null) {
+            value = defaultValue;
+        }
+        return value;
```
smolnar82, line 116: nit:
```
return Optional.ofNullable(System.getProperty(variableName))
               .or(() -> Optional.ofNullable(System.getenv(variableName)))
               .orElse(defaultValue);
```
This might be more concise, but your code works too.

PR #1019, opened 2025-04-08
KNOX-3120 - Add a specialized use API for API KEY based on KNOXTOKEN API

lmccay:

Similar to [KNOX-3112](https://issues.apache.org/jira/browse/KNOX-3112) which added an API for CLIENTID for acquiring client credentials flow CLIENT_ID and CLIENT_SECRET we should add a specialized extension of KNOXTOKEN for an APIKEY API.
The intent here is to add an API that both frees up KNOXTOKEN to be deployed within the same topology for known token exchange patterns as well as to codify the conventions that are being used by some that are using the passcode token as an API Key.

We will make appropriate configuration and metadata defaults within the API implementation to reduce operations and config requirements and to return a translated response with the passcode as an

{api_key: xxxx}
or some similar response that meets standard or defacto standard expectations.

## How was this patch tested?
Added new unit tests and ran all existing and new tests.
Tested manually:

```
curl -ivku admin:admin-password -X POST "https://localhost:8443/gateway/sandbox/apikey/api/v1/auth/key"

{"key_id":"d859cf22-007c-40d6-832c-8ddf068f3606","api_key":"WkRnMU9XTm1Nakl0T....1tTXRPR1JrWmpBMk9HWXpOakEyOjpPV1l6WVRFM05tRXRORFpqWVMwME5HUmtMV0pr....Nek15T0dWa1pHUm0="}

```

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/APIKeyResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/APIKeyResource.java
@@ -32,13 +32,13 @@ import java.util.HashMap;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.MediaType.APPLICATION_XML;
 
-@Path(ClientCredentialsResource.RESOURCE_PATH)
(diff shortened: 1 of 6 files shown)


pzampino (approved):
LGTM

PR #1024, opened 2025-04-14
KNOX-3131 - Add addExpect100Continue for Hue service

moresandeep:

Add addExpect100Continue for Hue service


smolnar82:
Hi @moresandeep !

The change itself looks good, but it won't have any effect since #747 is not merged yet.

moresandeep:
> Hi @moresandeep !
> 
> The change itself looks good, but it won't have any effect since #747 is not merged yet.

Thanks @smolnar82 ! 
I thought we had merged that :( We need to resurrect that first. Good point, thanks!

smolnar82:
As discusse above, this cannot be merged because its pre-requisite change is not yet available. Closing for now, but keep the JIRA open and see what we can do about it in the near future.

PR #1029, opened 2025-04-28
KNOX-3134: Set Secure and HttpOnly attributes for pac4jCsrfToken Cookie

hanicz:

pac4jCsrfToken cookie Secure and HttpOnly attributes are not set

The fix is included in 5.1.5
https://www.pac4j.org/5.1.x/docs/release-notes.html

pac4j 5.x requires jdk11
https://www.pac4j.org/docs/alldocs.html

Added a new HttpServletResponseWrapper that overrides the addCookie method. It checks for the pac4jCsrfToken cookie and sets the Secure and HttpOnly attributes to true.

## How was this patch tested?

New unit tests
Manually tested with pac4j setup
image


moresandeep (commented):
Looks good. My only note is do we have a usecase where this needs to be false? i.e. when we are not using https, what happens then does it break? There could be cases where users might not want to use https.

hanicz:
> Looks good. My only note is do we have a usecase where this needs to be false? i.e. when we are not using https, what happens then does it break? There could be cases where users might not want to use https.

We might have such usecase. I can check the SSL enabled gateway config and only add the secure attribute if it is set to true.

hanicz:
@moresandeep The new solution is checking for the SSL enabled configuration. Could you please review it again?

moresandeep (approved):
Awesome, thanks @hanicz ! looks great.

PR #1030, opened 2025-04-28
KNOX-3135 Fixed incorrect string format

dk2k:
KNOX-3135 Fixed incorrect string format


smolnar82:
The change itself looks good, but I'm closing this one due the the following reasons:
- missing required PR template
- no proof of tests (or at least a unit test case with an invalid backend URL that proves the error message is no correct)

PR #1031, opened 2025-04-28
KNOX-3136 removed write-only object

dk2k:
KNOX-3136 removed write-only object


smolnar82:
losing this one because there is no proof for testing (at a minimum services with URLTemplate queries should have been tested) and the required PR description template is missing.

PR #1033, opened 2025-04-28
KNOX-3138 implemented try-with-resources for Closeables

dk2k:
KNOX-3138 implemented try-with-resources for Closeables


lmccay:
Hi @dk2k - Thank you for your contribution! While we wait for the checks to pass, can you please revisit the PR description to follow the template that we provide for contributions? That level of detail is important for reviews. Thanks again!

dk2k:
@lmccay try-with-resources (TwR) was introduced exactly for these cases. If a closeable has a non-empty close() method, it should be called. Of course, you can call close() explicitly, but it's boilerplate code.
I found this issue with static analysis tool.
From my experience: if code is buildable after such changes, it means the changes are safe.

PR #1039, opened 2025-05-03
KNOX-3145 - Ensure Client Credentials flow client_secret belongs to t…

lmccay:
…he presented client_id

Currently, the support for client_id and client_secret treats the inclusion of the CLIENT_ID as a formality of the client credentials flow and since the actual client_id is resolvable from the client_secret, it is ignored.

While there it may be arguable whether we need to enforce this, it seems a reasonable expectation that they should match. Let's close that gap.

## How was this patch tested?

New unit tests added and all existing and new tests run.
Manually tested.

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
@@ -249,21 +249,7 @@ public class JWTFederationFilter extends AbstractJWTFilter {
           passcode = decodeBase64(base64DecodedTokenIdAndPasscode[1]);
           // if this is a client credentials flow request then ensure the presented clientId is
           // the actual owner of the client_secret
-          final String requestBodyString = getRequestBodyString(request);
-          if (requestBodyString != null && !requestBodyString.isEmpty()) {
-            final String grantType = RequestBodyUtils.getRequestBodyParameter(requestBodyString, GRANT_TYPE);
-            if (grantType != null && !grantType.isEmpty()) {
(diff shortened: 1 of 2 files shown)


gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -240,17 +242,34 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+              // if there is no client_id then this is not a client credentials flow
+              if (clientID != null && !tokenId.equals(clientID)) {
+                prechecks = false;
+                log.wrongPasscodeToken(tokenId);
+                handleValidationError((HttpServletRequest) request, (HttpServletResponse) response,
+                        HttpServletResponse.SC_UNAUTHORIZED,
+                        MISMATCHING_CLIENT_ID_AND_CLIENT_SECRET);
+              }
+            }
+          }
```
smolnar82, line 266: nit: I might have created a new `boolean validateClientCredentialsFlow(...)` method and assigned that to `prechecks`:
```
try {
   ...
   prechecks = validateClientCredentialsFlow(...);
} catch (Exception e) {
   ...
}

...

// if this is a client credentials flow request, then ensure the presented clientId is
// the actual owner of the client_secret
private boolean validateClientCredentialsFlow(...) {
   final String requestBodyString = getRequestBodyString(request);
   if (StringUtils.isNotBlank(requestBodyString )) {
      ...
   
}
```

lmccay: Good call. I'll take care of it.

lmccay: Done.

PR #1042, opened 2025-05-08
KNOX-3148: Make SameSite configurable for pac4j session cookies

hanicz:

This change makes the SameSite attribute for the pac4j session cookies configurable. The default value is `Strict`.
The pac4j Cookie class doesn't support the SameSite attribute so I had to change the implementation to Set-Cookie headers.
Modified MockHttpServletResponse class to accommodate the testing of multiple Set-Cookie headers.

## How was this patch tested?

Manually tested, unit tests
image


smolnar82:
Thanks, @hanicz, for your contribution. It was a good call to build the Set-Cookie header and add all values. Since you are doing that, I think there is an opportunity to improve the code.

See the same (similar) code fragment in `WebSSOResource`:
image
I think it'd be beneficial to create a utility class that builds the `Set-Cookie` header, where all the params can be constants, the utility class can be tested, etc... Then resource, filter classes can use that class and its method(s).
What do you think?

smolnar82 (approved):
LGTM, thanks for your contribution, @hanicz !

PR #1043, opened 2025-05-08
KNOX-3048 - Add support for group-based impersonation

moresandeep:

Added support for Group Based impersonation provider. 

```
<!-- turn the feature on -->
<param>
     <name>hadoop.proxygroup.impersonation.enabled</name>
     <value>true</value>
</param>

<!-- proxy group configs -->
<param>
     <name>hadoop.proxygroup.KNOX_ADMIN.groups</name>
     <value>*</value>
</param>

<param>
     <name>hadoop.proxygroup.VIRTUAL_GROUP.groups</name>
     <value>*</value>
</param>

<param>
     <name>hadoop.proxygroup.VIRTUAL_GROUP.hosts</name>
     <value>*</value>
</param>

``` 

## How was this patch tested?
Unit tests and local testing

### DIFF
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
@@ -17,33 +17,6 @@
  */
 package org.apache.knox.gateway.identityasserter.common.filter;
 
-import static org.apache.knox.gateway.identityasserter.common.filter.AbstractIdentityAsserterDeploymentContributor.IMPERSONATION_PARAMS;
-import static org.apache.knox.gateway.identityasserter.common.filter.AbstractIdentityAsserterDeploymentContributor.ROLE;
-import static org.apache.knox.gateway.identityasserter.common.filter.VirtualGroupMapper.addRequestFunctions;
-
-import java.io.IOException;
-import java.security.AccessController;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.StringTokenizer;
-import java.util.stream.Collectors;
-import javax.security.auth.Subject;
-import javax.servlet.FilterChain;
-import javax.servlet.FilterConfig;
-import javax.servlet.ServletContext;
-import javax.servlet.ServletException;
-import javax.servlet.ServletRequest;
-import javax.servlet.ServletResponse;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletRequestWrapper;
-import javax.servlet.http.HttpServletResponse;
-
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.IdentityAsserterMessages;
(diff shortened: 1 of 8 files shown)


moresandeep:
Folks apology for formatting :( I accidentally formatted the code as i went along to prevent checkstyle issues and now going back to the old formatting is pain. If anyone has objections I can update the code (might take a while though :( )

moresandeep:
> > Folks apology for formatting :( I accidentally formatted the code as i went along to prevent checkstyle issues and now going back to the old formatting is pain. If anyone has objections I can update the code (might take a while though :( )
> 
> Hi, @moresandeep ! It'd be beneficial if you updated the code and include the 'real' changes here. Later on, we can have a formatting PR. This would help reviewers to understand the purpose of this one.

Thanks @smolnar82 I will provide a summary of the changes so it will be easy for folks to review.

moresandeep:
Following are the major changes in `CommonIdentityAssertionFilter` class

- Centralized “is impersonation enabled” logic
     •	Old: read IMPERSONATION_ENABLED_PARAM directly from filterConfig, parsed its boolean value inline
	•	New: delegated that check to a new helper 
-  Eliminated duplicate constant
	•	Removed the class-local static final String IMPERSONATION_ENABLED_PARAM = AuthFilterUtils.PROXYUSER_PREFIX + ".impersonation.enabled";
        •	Instead statically import AuthFilterUtils.IMPERSONATION_ENABLED_PARAM (and added GROUP_IMPERSONATION_ENABLED_PARAM for future use)

- Helper methods promoted & grouped
	•	Moved virtualGroupParameterNames(...) and unique(...) up near the top of the class, declared them private static so all utility logic is visible before the lifecycle methods.

- No other behavior changes
	•	All the parsing of advanced principal mappings, virtual-group loading, doFilter flow, etc., is untouched—only how and where the impersonation flag is determined was refactored.

moresandeep:
**Note**: _The following summary is generated using co-pilot_

This pull request refactors the `CommonIdentityAssertionFilter` class and its related tests to improve code maintainability, readability, and functionality. Key changes include the introduction of a new method for checking impersonation provider requirements, reorganization of imports, and updates to test cases to align with the refactored implementation.

### Refactoring and Code Simplification:
* Introduced the `shouldAddImpersonationProvider` method to encapsulate logic for determining if the impersonation provider should be added, replacing inline checks. (`[gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.javaR170-R180](diffhunk://#diff-4577bd8f7a7a8aa491f44533f5111e9deb9496962b2059d5e0cbb3f1b59c4bfdR170-R180)`)
* Removed redundant methods `virtualGroupParameterNames` and `unique`, and integrated their functionality directly where needed. (`[[1]](diffhunk://#diff-4577bd8f7a7a8aa491f44533f5111e9deb9496962b2059d5e0cbb3f1b59c4bfdL207-L211)`, `[[2]](diffhunk://#diff-4577bd8f7a7a8aa491f44533f5111e9deb9496962b2059d5e0cbb3f1b59c4bfdL299-R312)`)
* Reorganized imports in `CommonIdentityAssertionFilter` to improve clarity and remove unused dependencies. (`[[1]](diffhunk://#diff-4577bd8f7a7a8aa491f44533f5111e9deb9496962b2059d5e0cbb3f1b59c4bfdL20-L46)`, `[[2]](diffhunk://#diff-4577bd8f7a7a8aa491f44533f5111e9deb9496962b2059d5e0cbb3f1b59c4bfdL65-R90)`)

### Test Case Updates:
* Updated test cases in `CommonIdentityAssertionFilterTest` to use the refactored `shouldAddImpersonationProvider` method and replaced references to the removed `IMPERSONATION_ENABLED_PARAM` constant. (`[[1]](diffhunk://#diff-f1e853cfe04089b2ce1381bc562b0d6e96ace1db5ee2c0085fa1dd8bdaaae7dfL286-R264)`, `[[2]](diffhunk://#diff-f1e853cfe04089b2ce1381bc562b0d6e96ace1db5ee2c0085fa1dd8bdaaae7dfL141-R141)`)
* Reorganized imports in `CommonIdentityAssertionFilterTest` for consistency with the refactored implementation. (`[[1]](diffhunk://#diff-f1e853cfe04089b2ce1381bc562b0d6e96ace1db5ee2c0085fa1dd8bdaaae7dfL20-R40)`, `[[2]](diffhunk://#diff-f1e853cfe04089b2ce1381bc562b0d6e96ace1db5ee2c0085fa1dd8bdaaae7dfL41-R60)`)
* Added expectations for new impersonation-related parameters in `HadoopAuthFilterTest` to ensure compatibility with the updated implementation. (`[[1]](diffhunk://#diff-856be9bd8db07e7ffa4042e3f2943b254b2c6204e1bd7e50c6bfae8fa89e3cf3R148-R149)`, `[[2]](diffhunk://#diff-856be9bd8db07e7ffa4042e3f2943b254b2c6204e1bd7e50c6bfae8fa89e3cf3R214-R215)`, `[[3]](diffhunk://#diff-856be9bd8db07e7ffa4042e3f2943b254b2c6204e1bd7e50c6bfae8fa89e3cf3R266-R267)`, `[[4]](diffhunk://#diff-856be9bd8db07e7ffa4042e3f2943b254b2c6204e1bd7e50c6bfae8fa89e3cf3R329-R330)`, `[[5]](diffhunk://#diff-856be9bd8db07e7ffa4042e3f2943b254b2c6204e1bd7e50c6bfae8fa89e3cf3R392-R393)`)

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
```diff
@@ -0,0 +1,252 @@
...
+            return;
+        }
+
+        boolean isProxyUserAuthorized = false;
+        boolean isProxyGroupAuthorized = false;
+        if (isProxyUserEnabled) {
+            /* authorize proxy user */
+            isProxyUserAuthorized = checkProxyUserAuthorization(user, remoteAddress);
+            /* In case of AND no reason to check for proxy groups if isProxyUserAuthorized=false */
+            if("AND".equalsIgnoreCase(impersonationMode) && !isProxyUserAuthorized) {
```
zeroflag, line 143: I would either extract this to a string constant, or introduce an enum.

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
```diff
@@ -0,0 +1,252 @@
...
+                        + " is not allowed to impersonate " + user.getUserName());
+            }
+        }
+
+        if (isProxyGroupEnabled) {
+            /* check for proxy group impersonation */
+            isProxyGroupAuthorized = checkProxyGroupAuthorization(user, remoteAddress);
+        }
+
+        if("AND".equalsIgnoreCase(impersonationMode)) {
```
zeroflag, line 154: Same as above.

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
```diff
@@ -0,0 +1,252 @@
...
+                break;
+            }
+        }
+
+        if (!proxyGroupFound) {
+            LOG.failedToImpersonateGroups(realUser.getUserName(), realUserGroups.toString(), user.getUserName());
+            return false;
+        }
+
+        boolean proxyGroupHostFound = false;
```
zeroflag, line 232: That could be another candidate to extract into its own method.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
```diff
@@ -62,286 +35,318 @@
...
-    return impersonationEnabled;
-  }
+    /**
+     * Check if the impersonation provider should be added to the filter chain based on
+     * user and group impersonation settings.
+     *
+     * @param filterConfig
+     * @return
+     */
+    boolean shouldAddImpersonationProvider(final FilterConfig filterConfig) {
```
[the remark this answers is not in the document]
moresandeep: @zeroflag yup!

gateway-spi/src/main/java/org/apache/knox/gateway/util/AuthFilterUtils.java
```diff
@@ -27,227 +39,270 @@
...
-import javax.servlet.ServletContext;
-import javax.servlet.ServletRequest;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletRequestWrapper;
+public class AuthFilterUtils {
+    public static final String DEFAULT_AUTH_UNAUTHENTICATED_PATHS_PARAM = "/knoxtoken/api/v1/jwks.json";
+    public static final String PROXYUSER_PREFIX = "hadoop.proxyuser";
+    public static final String PROXYGROUP_PREFIX = "hadoop.proxygroup";
+    public static final String IMPERSONATION_MODE = "hadoop.impersonation.mode";
+    public static final String DEFAULT_IMPERSONATION_MODE = "OR";
```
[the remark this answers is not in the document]
moresandeep: ugh, i should have updated my PR description 
```
                 <param>
                    <name>hadoop.impersonation.mode</name>
                    <value>OR</value>
                  </param>
```
This would be an example.

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
```diff
@@ -0,0 +1,252 @@
...
+            return true;
+        }
+
+        // Get the real user's groups (both real and virtual)
+        Set<String> realUserGroups = new HashSet<>();
+        if (user.getRealUser().getGroupNames() != null) {
+            Collections.addAll(realUserGroups, user.getRealUser().getGroupNames());
+        }
+
+        boolean proxyGroupFound = false;
```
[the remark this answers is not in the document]
moresandeep: Yup

moresandeep:
@smolnar82 @zeroflag thank you so much for the valuable reviews! I going to make the changes and do some more refactoring in light of changes and some debugging. I will mark the PR as WIP for now and let you know when i have it back up. 
Thank You again!!

gateway-spi/src/main/java/org/apache/knox/gateway/util/AuthFilterUtils.java
```diff
@@ -42,212 +45,307 @@
...
+
+    public static boolean shouldDoGlobalLogout(HttpServletRequest request) {
+        return request.getAttribute(DO_GLOBAL_LOGOUT_ATTRIBUTE) == null ? false : Boolean.parseBoolean((String) request.getAttribute(DO_GLOBAL_LOGOUT_ATTRIBUTE));
+    }
+
+    /**
+     * Check if user or group impersonation is enabled based on filter configuration.
+     *
+     * @param filterConfig The filter configuration
+     * @return A boolean array where the first element indicates if user impersonation is enabled
```
zeroflag, line 325: Stale javadoc comment. It no longer returns a boolean array.

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
```diff
@@ -0,0 +1,228 @@
...
+            } else {
+                proxyGroupHostFound = true;
+                break;
+            }
+        }
+
+        if (!proxyGroupHostFound) {
+            LOG.failedToImpersonateGroupsFromAddress(realUser.getUserName(), realUserGroups.toString(), user.getUserName(), remoteAddress.toString());
+            throw new AuthorizationException("User: " + realUser.getUserName()
+                    + "with groups " + realUserGroups.toString()
```
zeroflag, line 221: Missing space before `"with groups`

lmccay (commented):
@moresandeep - why add a new provider for this and not just add it to common so that it is available everywhere that extends that?

moresandeep:
> @moresandeep - why add a new provider for this and not just add it to common so that it is available everywhere that extends that?

Good point, I thought about that but decided against it in favor of code separation. Since all of the implementations use Common any bug might affect all the providers. Not all providers need this the ones that need this feature are virtual groups (already included in default) and hadoop-group-lookup so i decided to create a new provider that supports these.

Putting this in common will be easy but this is not really common feature that can be used by all the other providers. 
Do you anticipate this to be used by others? i can move this common in that case.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
```diff
@@ -63,285 +64,304 @@
...
+        topologyName = (String) filterConfig.getServletContext().getAttribute(GatewayServices.GATEWAY_CLUSTER_ATTRIBUTE);
+        String principalMapping = filterConfig.getInitParameter(PRINCIPAL_MAPPING);
+        if (principalMapping == null || principalMapping.isEmpty()) {
+            principalMapping = filterConfig.getServletContext().getInitParameter(PRINCIPAL_MAPPING);
+        }
+        String groupPrincipalMapping = filterConfig.getInitParameter(GROUP_PRINCIPAL_MAPPING);
+        if (groupPrincipalMapping == null || groupPrincipalMapping.isEmpty()) {
+            groupPrincipalMapping = filterConfig.getServletContext().getInitParameter(GROUP_PRINCIPAL_MAPPING);
+        }
+        if (principalMapping != null && !principalMapping.isEmpty() || groupPrincipalMapping != null && !groupPrincipalMapping.isEmpty()) {
```
pzampino, line 98: Do these need to be grouped? You're checking whether there are valid userproxy configs or group proxy configs, correct? Something like ((hasPrincipalMapping) || (hasGroupMapping)):

`if ((principalMapping != null && !principalMapping.isEmpty()) || (groupPrincipalMapping != null && !groupPrincipalMapping.isEmpty()))`

Maybe that is how the logical operators are evaluated by default?

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
```diff
@@ -0,0 +1,257 @@
...
+    private void initGroupBasedProvider(String proxyGroupPrefix) {
+        groupConfigPrefix = proxyGroupPrefix +
+                (proxyGroupPrefix.endsWith(".") ? "" : ".");
+
+        // constructing regex to match the following patterns:
+        //   $configPrefix.[ANY].users
+        //   $configPrefix.[ANY].groups
+        //   $configPrefix.[ANY].hosts
+        //
+        String prefixRegEx = groupConfigPrefix.replace(".", "\\.");
```
pzampino: Could these regex defintions be constants?

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
```diff
@@ -0,0 +1,257 @@
...
+    public void authorize(UserGroupInformation user, InetAddress remoteAddress, List<String> groups) throws AuthorizationException {
+
+        /**
+         *  check for proxy user authorization only if PROXYUSER_PREFIX properties exist.
+         *  If proxy is configured then use those properties instead of group-based properties
+         *  given user based configs are more finegrained.
+         */
+        if (doesProxyUserConfigExist) {
+            super.authorize(user, remoteAddress);
+        } else{
```
pzampino: I would expect this to be evaluated in two cases:

1. There is no proxyuser config
2. There is no matching proxyuser config for this specific user

I would not expect this to be an if-else relationship.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/CommonIdentityAssertionFilter.java
moresandeep: Good point, the do behave the same (the existing statement and the example you mentioned) this is because `&&` has higher precedence than `||` in Java [1]. Style wise i prefer the one you suggested but this is not part of my changes this is how it was before the change, not sure why Github is flagging this as a new change though. If you prefer i change it let me know!

[1] https://stackoverflow.com/questions/4436669/how-to-prove-the-precedence-of-and-by-coding-in-java

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
moresandeep: I think i see what you mean with 1 and 2. i do not follow the if-else part though. Are you suggesting having an `if` statement rather than if-else to check for proxyGroupAuthorizations like we do for proxyGroup ?

something like 
```
if(doesProxyUserConfigExist && doesProxyUserConfigExist(user)) {
        super.authorize(user, remoteAddress);
} 
else if (doesProxyGroupConfigExist) {
        checkProxyGroupAuthorization(user, remoteAddress, groups);
}
```

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
moresandeep: Yup! i'll fix this.

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
pzampino: I mean that the presence of proxyuser config should not preclude the evaluation of the group-level config. If there is user-level config that does not apply to the requesting user (super.authorize() fails), then the group-level config should be evaluated to see if there is matching authorization based on the requesting user's group membership.

pzampino: Probably means catching AuthorizationException from the super.authorize() call, and continuing if there is group-level config. Perhaps, something like
```
if(doesProxyUserConfigExist && doesProxyUserConfigExist(user)) {
     try {
        super.authorize(user, remoteAddress);
     } catch (AuthorizationException e) {
         if (!doesProxyGroupConfigExist) {
            throw e;
         }
     }
}

if (doesProxyGroupConfigExist) {
    checkProxyGroupAuthorization(user, remoteAddress, groups);
}
```

moresandeep: Ah, i see thanks for the explanation!

moresandeep:
@pzampino i addressed your comments, following are the changes in a nutshell:
1. Regex expressions are now constants
2.  For cases where proxy user impersonation fails proxy group proxy impersonation will be evaluated
3. Added unit tests for the item no. 2

I tested it locally and verified that it works. Let me know if there are more changes.

gateway-spi/src/main/java/org/apache/knox/gateway/i18n/GatewaySpiMessages.java
```diff
@@ -93,4 +96,22 @@ public interface GatewaySpiMessages {
 
   @Message(level=MessageLevel.DEBUG, text="Ignoring cookie path scope filter for default topology")
   void ignoringCookiePathScopeForDefaultTopology();
+
+  @Message(level=MessageLevel.DEBUG, text="Loaded proxy groups ACLs: {0}")
+  void loadedProxyGroupsAcls(String acls);
+
+  @Message(level = MessageLevel.ERROR, text = "Unauthorized connection for super-user: {0} from IP remoteAddress {1}")
```
pzampino: Suggestion: "User impersonation failed for user {0}. Connections from remote address {1} are not authorized."

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
```diff
@@ -0,0 +1,263 @@
...
+
+        /* Proxy user configuration takes precedence over proxy group configuration. */
+        if (doesProxyUserConfigExist) {
+            try{
+                /* check for proxy user authorization */
+                super.authorize(user, remoteAddress);
+            } catch (final AuthorizationException e) {
+                /*
+                 * Log and try group based impersonation.
+                 * Since this provider is for groups no need to check if
```
pzampino: Is this comment true? We only employ this implementation if there is group-level proxy config? That seems wrong to me.

gateway-spi/src/main/java/org/apache/knox/gateway/util/AuthFilterUtils.java
```diff
@@ -42,212 +44,247 @@
...
+                    conf.set(name, value);
+                });
+            }
         }
 
-        @Override
-        public Principal getUserPrincipal() {
-          return remoteRequestUgi::getUserName;
+        ImpersonationProvider impersonationProvider;
+        if(hasProxyGroupParams) {
```
pzampino: Why can't this be our default implementation that supports both user-level and the new group-level impersonation config? It would simplify a lot of things. So, rather than GroupBasedImpersonationProvider, it would be someyhing like KnoxImpersonationProvider (our implementation that also supports group-level config).

pzampino (commented):
The change you've made to address my previous comment appears to match what I had in mind. I just have a couple more questions.

gateway-spi/src/main/java/org/apache/knox/gateway/util/GroupBasedImpersonationProvider.java
moresandeep: The main reason was separation of functionality. I did not want the existing functionality to be affected by new code. So, unless users configure proxy group `GroupBasedImpersonationProvider` will not be engaged. This was my thinking.

PR #1045, opened 2025-05-14
KNOX-3150 - Fix a bug where jwks keys might not be cached properly

moresandeep:
Previous fix for KNOX-3150 might not have addresses the caching issue completely. This is a followup fix to that. 

To recap, https://github.com/apache/knox/pull/1044 added support for caching JWKS keys. Following configurable options were added for caching jwks keys
These parameters can be configured in `gateway-site.xml` (in line with other parameters around JWKS configuration e.g. `gateway.token.issuers.ignore.type.validation`)

1. `gateway.jwks.cache.ttl` (default value 20 mins)
2. `gateway.jwks.cache.refresh.interval` (default value 20 mins)

## How was this patch tested?
This patch was tested locally: 
1. Created a topology with a JWKS endpoint
2. ensured that the jwks endpoint was reachable
3. Verified that the token was validated
4. Made the jwks endpoint unreachable 
5. created a new token (we cache the authentication result for older tokens so had to be a new token)
6. verified that new token was validated even when JWKS endpoint was unreachable.
7. attached debugger to gateway.
8. Followed steps 4-6 and ensured that cache was hit.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
@@ -32,6 +32,7 @@ import java.security.interfaces.RSAPrivateKey;
 import java.security.interfaces.RSAPublicKey;
 import java.text.ParseException;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
```diff
@@ -232,13 +236,18 @@ public boolean verifyToken(JWT token, String jwksurl, String algorithm, Set<JOSE
...
+                  .cache(cacheTTL, cacheTimeOut)
+                  .refreshAheadCache(true)
+                  .retrying(true)
+                  .outageTolerant(outageTTL)
+                  .build();
+            cachedJwkSources.put(jwksurl, keySource);
+        }
+
+
+        JWSKeySelector<SecurityContext> keySelector = new JWSVerificationKeySelector<>(expectedJWSAlg, cachedJwkSources.get(jwksurl));
```
lmccay, line 250: Seems like you could do the get only once on the cachedJwkSources.

lmccay (commented):
It's a bit of a nit but I think you can just do the get once.

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/DefaultTokenAuthorityService.java
moresandeep: Sure, i'll update the PR

lmccay (approved):
+1

PR #1046, opened 2025-05-16
KNOX-3151: Refactored the test websocketclient awaitClose method, fix…

hanicz:
…ed related tests, fixed an issue caused by a racing condition that resulted in intermittent MessageFailureTest.testMessageTooBig test failures

- Refactored the test WebSocketClient awaitClose method to fail on not expected status codes and use the neglected countdown latch properly
- Fixed the tests that rely on the mentioned awaitClose method
- Fixed the intermittent `MessageFailureTest.testMessageTooBig` failures

I was able to reproduce the issue locally by copying the `AbstractWebSocketConnection` jetty class. I added a sleep to the close method between the `session.callApplicationOnClose(closeInfo);`and `close(closeInfo, new DisconnectCallback(this));` calls. Even a 1ms sleep is enough to consistently reproduce the issue. I believe the intermittent failures are due to a racing condition thats causing a failed close handshake resulting in 1006 (ABNORMALLY CLOSED) status code. In our `ProxyWebSocketAdapter` class the `onWebSocketError` method was checking for `!frontendSession.isOpen()` which should always be open at this point. By removing the `!` the issue is solved and the client properly receives a close event.

## How was this patch tested?

Unit tests

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
@@ -206,13 +206,13 @@ public class ProxyWebSocketAdapter extends WebSocketAdapter {
 
     LOG.onError(t.toString());
     if (t.toString().contains("exceeds maximum size")) {
-      if(frontendSession != null && !frontendSession.isOpen()) {
+      if(frontendSession != null && frontendSession.isOpen()) {
         frontendSession.close(StatusCode.MESSAGE_TOO_LARGE, t.getMessage());
       }
     }
 
     else {
-      if(frontendSession != null && !frontendSession.isOpen()) {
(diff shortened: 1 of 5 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/websockets/ProxyWebSocketAdapter.java
```diff
@@ -206,13 +206,13 @@ private void cleanupOnError(final Throwable t) {
 
     LOG.onError(t.toString());
     if (t.toString().contains("exceeds maximum size")) {
-      if(frontendSession != null && !frontendSession.isOpen()) {
```
moresandeep, line 209: That's a good catch, thanks!

moresandeep (approved):

smolnar82:
This is a very useful fix, @hanicz, your contributions are highly appreciated!

PR #1010, opened 2025-03-28
KNOX-3115: HA Dispatch for SSE

hanicz:
## What changes were proposed in this pull request?

[KNOX-3065](https://issues.apache.org/jira/projects/KNOX/issues/KNOX-3065) Introduced the proper support for SSE connections. However HA Dispatch is missing. This PR add a new SSEHaDispatch that is able to load balance and use sticky session cookies. Failover is not added yet. The SSEHaDispatch is based on ConfigurableHaDispatch however it can't extend it due to extending SSEDispatch instead.

Changes:
- New LBHaDispatch interface to hold LB related code to avoid duplication
- New SSEHaDispatch to handle LB, sticky session, user agents for SSE
- New unit tests for SSEHaDispatch
- Moved the SSEEvent creation into a static method inside SSEEvent 

## How was this patch tested?

Unit tests.
Local manual tests with a custom SSE service.

I was hosting a service locally on localhost:7435 and I tested this service with the new SSEHaDispatch and ConfigurableHaDispatch as well.

Ha provider:
```
<provider>
    <role>ha</role>
    <name>HaProvider</name>
    <enabled>true</enabled>
    <param>
        <name>SSERVICE</name>
        <value>enableStickySession=true;enabled=true;enableLoadBalancing=true;disableLoadBalancingForUserAgents=AgentY</value>
    </param>
    <param>
        <name>HUE</name>
        <value>
            maxFailoverAttempts=3;failoverSleep=1000;enableStickySession=true;enabled=true;enableLoadBalancing=true
        </value>
    </param>
</provider>
```

Services:
```
<service>
    <role>HUE</role>
    <url>http://localhost:7435</url>
    <url>http://local2host:7436</url>
</service>

<service>
    <role>SSERVICE</role>
    <url>http://localhost:7435</url>
    <url>http://local2host:7435</url>
</service>
```

Curl commands:
`curl -k -i -u admin:admin-password https://localhost:8443/gateway/my-topology/sservice/sse -H "User-Agent: AgentX"`
`curl -k -i -u admin:admin-password https://localhost:8443/gateway/my-topology/sservice/sse -H "User-Agent: AgentY"`
`curl -k -i -u admin:admin-password https://localhost:8443/gateway/my-topology/hue/sse -H "User-Agent: AgentX"`
`curl -k -i -u admin:admin-password https://localhost:8443/gateway/my-topology/hue/sse -H "User-Agent: AgentY"`

I tested load balancing, failover, user-agents and sticky session cookies from browser.

PR #947, opened 2024-10-22
KNOX-3065: SSE support for Knox

hanicz:
## What changes were proposed in this pull request?

SSE support for Knox.

Currently Knox is not working with SSE. In case of SSE the response header contains Content-Type = text/event-stream, in which case the server will send messages terminated by \n\n. Knox should send these messages to the client as they arrive. Currently Knox collects them all, and once the server closes the connection sends all of them concatenated to each other to the client.

Requirement:

Full support in Knox for Server Sent Events.
Definition of done:

Clients can connect to SSE endpoints in the backend via Knox.
Server Sent Events (https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events) are forwarded by Knox from the server to the client as soon as they are received.

## How was this patch tested?

Unit tests.
Locally tested with a custom SSE service.
Any SSE service that is the follows the specs is supported in Knox.
Setup:
Asynchronous behaviour and therefore SSE support is not enabled by default. Modify the below config to ‘true’ in _<KNOX-HOME>/conf/gateway-site.xml_ file.
```
<property>
   <name>gateway.servlet.async.supported</name>
   <value>true</value>
   <description>Enable/Disable async support.</description>
</property>
```

Change the existing topology by adding an SSE service to _<KNOX-HOME>/conf/{topology}.xml_
```
<service>
   <role>SSERVICE</role>
   <url>http://localhost:7435</url>
</service>
```

Add the SSE service to the _<KNOX_HOME>/data/services_ folder.

<KNOX-HOME>/data/services/{myservice}/{version}/rewrite.xml
```
<rules>
   <rule dir="IN" name="SSERVICE/sservice/inbound" pattern="*://*:*/**/sservice/*">
       <rewrite template="{$serviceUrl[SSERVICE]}/sse"/>
   </rule>
</rules>
```

_<KNOX-HOME>/data/services/{myservice}/{version}/service.xml_  (Make sure to add the SSEDispatch to the service)
```
<service role="SSERVICE" name="sservice" version="0.1">
   <routes>
       <route path="/sservice/**"></route>
   </routes>
   <dispatch classname="org.apache.knox.gateway.sse.SSEDispatch"/>
</service>
```
curl command to test the configured SSE service:
`curl -k -i -u admin:admin-password https://localhost:8443/gateway/sandbox/sservice/sse `

GET, POST, PUT, PATCH methods were tested.

PR #1049, opened 2025-05-22
KNOX-3146: Failover feature for SSE HA

hanicz:

This PR builds on [PR-947](https://github.com/apache/knox/pull/947) and [PR-1010](https://github.com/apache/knox/pull/1010). It completes the SSE feature by adding the failover capability to the HA dispatch. 

- Fixed a bug where service connection error wasn't sent to the client for non HA
- Renamed `LBHaDispatch` -> `CommonHaDispatch` since it holds failover logic as well now
- Moved failover logic from `ConfigurableHaDispatch` to `CommonHaDispatch` to avoid code duplication
- Added failover for `SSEHaDispatch`
- New tests for SSE HA to cover the new failover capability
- Moved `httpcore` dependency exclusions to the parent pom level

## How was this patch tested?

Unit tests
Manual tests

Local setup:

I had a custom SSE service running locally.

```
        <provider>
            <role>ha</role>
            <name>HaProvider</name>
            <enabled>true</enabled>
            <param>
                <name>SSERVICE</name>
                <value>
                    maxFailoverAttempts=3;failoverSleep=1000;enableStickySession=true;enabled=true;enableLoadBalancing=true;noFallback=true
                </value>
            </param>
            <param>
                <name>HUE</name>
                <value>
                    enableStickySession=true;enabled=true;enableLoadBalancing=true;maxFailoverAttempts=1;failoverSleep=1000
                </value>
            </param>
        </provider>

    <service>
        <role>HUE</role>
        <url>http://localhost3:7435</url>
        <url>http://localhost:7435</url>
        <url>http://localhost2:7435</url>
        <url>http://localhost1:7435</url>
        <url>http://localhost5:7435</url>
    </service>

    <service>
        <role>SSERVICE</role>
        <url>http://localhost3:7435</url>
        <url>http://localhost:7435</url>
        <url>http://localhost2:7435</url>
        <url>http://localhost1:7435</url>
        <url>http://localhost5:7435</url>
    </service>
```

`curl -k -i -u admin:admin-password https://localhost:8443/gateway/mytopology/hue/sse`
`curl -k -i -u admin:admin-password https://localhost:8443/gateway/mytopology/sservice/sse`

### DIFF
--- a/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/CommonHaDispatch.java
+++ b/gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/CommonHaDispatch.java
@@ -22,6 +22,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.http.client.methods.HttpRequestBase;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.client.utils.URIBuilder;
+import org.apache.knox.gateway.filter.AbstractGatewayFilter;
 import org.apache.knox.gateway.ha.dispatch.i18n.HaDispatchMessages;
 import org.apache.knox.gateway.ha.provider.HaProvider;
 import org.apache.knox.gateway.ha.provider.HaServiceConfig;
@@ -30,6 +31,7 @@ import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
@@ -38,14 +40,17 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
-public interface LBHaDispatch {
+public interface CommonHaDispatch {
 
     HaDispatchMessages LOG = MessagesFactory.get(HaDispatchMessages.class);
     Map<String, String> urlToHashLookup = new HashMap<>();
     Map<String, String> hashToUrlLookup = new HashMap<>();
+    String FAILOVER_COUNTER_ATTRIBUTE = "dispatch.ha.failover.counter";
+    List<String> nonIdempotentRequests = Arrays.asList("POST", "PATCH", "CONNECT");
 
     boolean isStickySessionEnabled();
 
@@ -71,7 +76,25 @@ public interface LBHaDispatch {
 
     void setActiveURL(String url);
 
-    default void initializeLBHaDispatch(HaServiceConfig serviceConfig) {
+    int getMaxFailoverAttempts();
+
+    void setMaxFailoverAttempts(int maxFailoverAttempts);
+
+    int getFailoverSleep();
+
+    void setFailoverSleep(int failoverSleep);
+
+    void setFailoverNonIdempotentRequestEnabled(boolean enabled);
+
+    boolean isFailoverNonIdempotentRequestEnabled();
+
+    void setNoFallbackEnabled(boolean enabled);
+
+    boolean isNoFallbackEnabled();
+
+    URI getDispatchUrl(HttpServletRequest request);
+
+    default void initializeCommonHaDispatch(HaServiceConfig serviceConfig) {
         setLoadBalancingEnabled(serviceConfig.isLoadBalancingEnabled());
         setStickySessionsEnabled(isLoadBalancingEnabled() && serviceConfig.isStickySessionEnabled());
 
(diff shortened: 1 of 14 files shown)


gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/CommonHaDispatch.java
```diff
@@ -38,14 +40,17 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
-public interface LBHaDispatch {
+public interface CommonHaDispatch {
```
lmccay, line 47: Not sure this was needed to be renamed. Failover is typically a LB feature as well. Nothing you need to change but it probably would have required fewer other changes to use the new name everywhere.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/CommonHaDispatch.java
```diff
@@ -240,4 +270,73 @@ default void shiftActiveURL(boolean userAgentDisabled, Optional<URI> backendURI)
             }
         }
     }
+
```
lmccay, line 241: It is super confusing to have so much code in an interface.
Maybe only so for old eyes like mine though.
Based on the default modifier being in this class previously, I assume that you are not introducing it here for the first time. I'll think about it a bit more further into the review where multiple inheritance starts to rear its head. :)

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/SSEHaDispatch.java
```diff
@@ -136,17 +149,100 @@ public void setActiveURL(String url) {
...
+    @Override
+    public void setNoFallbackEnabled(boolean enabled) {
+        this.noFallbackEnabled = enabled;
+    }
+
+    @Override
+    public boolean isNoFallbackEnabled() {
+        return noFallbackEnabled;
+    }
+
```
lmccay: All of the above methods have to be duplicated across all implementers of CommonHaDispatch? Why is that? If they were added to a common extension instead wouldn't that be better?

lmccay: Is this only because SSEHaDispatch doesn't extend the common one?

lmccay (commented):
My comments likely require education more than any changes.
Otherwise, it looks good to me. I'll wait for explanations though.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/SSEHaDispatch.java
hanicz: Yes, that is the reason. We could move them into `ConfigurableDispatch` or `DefaultDispatch`. However I don't think we should add HA code to non HA Dispatches.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/CommonHaDispatch.java
hanicz: Unfortunately the SSE async code is vastly different from the other dispatches that is the reason of the interface with default methods.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/SSEHaDispatch.java
pzampino: There is a ConfigurableHADispatch

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/CommonHaDispatch.java
pzampino: I think @hanicz added this LBHaDispatch relatively recently.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/SSEHaDispatch.java
hanicz: I can't extend that since SSEHaDispatch already extends SSEDispatch

smolnar82: We now have `HaConfiguration`  in place, I suppose this answers your doubts, right @lmccay , @pzampino ?

pzampino: Yes, this is a better solution.

gateway-provider-ha/src/main/java/org/apache/knox/gateway/ha/dispatch/CommonHaDispatch.java
```diff
@@ -225,19 +223,88 @@ default void shiftActiveURL(boolean userAgentDisabled, Optional<URI> backendURI)
...
+            }
+            return counter;
+        }
+    }
+
+    default HttpServletRequest prepareForFailover(HttpUriRequest outboundRequest, HttpServletRequest inboundRequest) {
+        //null out target url so that rewriters run again
+        inboundRequest.setAttribute(AbstractGatewayFilter.TARGET_REQUEST_URL_ATTRIBUTE_NAME, null);
+        // Make sure to remove the ha cookie from the request
+        inboundRequest = new StickySessionCookieRemovedRequest(getHaConfigurations().getStickySessionCookieName(), inboundRequest);
```
moresandeep, line 273: Could there be a case where there is an HA request but no cookie? i.e. no sticky session, will it blow up here?

hanicz: No, it just makes sure it is removed in case there is one. This logic was just moved into the interface, I didn't touch it.

moresandeep (approved):
Looks good! thanks for the explanation @hanicz !

PR #1052, opened 2025-06-03
KNOX-3155 - Isolate the CLIENTID and APIKEY param names from KNOXTOKEN

lmccay:

Issue: The extension APIs of KNOXTOKEN: APIKEY and CLIENTID APIs are unable to be colocated with the KNOXTOKEN API in the same topology due to param name conflicts. This results in unexpected results when the TTL or other params are set in one or more of the APIs.

Solution: I provided prefixes for all extension APIs that extend the PasscodeTokenResourceBase class via a getPrefix() method on each implementation. Those prefixes are used to disambiguate the params at initialization time and then are stripped from the params and reset to the expected names in KNOXTOKEN implementation.

This insures that the params do not collide and that the implementation has the expected values at runtime.

Note: there may be a more fundamental issue that has required this change that has been lurking for quite some time. The setting of all the params from the JAXRS APIs as init params within web.xml in a flat space for the GatewayServlet seems problematic. This will require additional investigation. In the meantime, this will work for our needs and will still be safe when we address that problem.

## How was this patch tested?

Existing unit tests were run and adjusted to accommodate the new names with the prefixes.
All other unit tests were also run.

Manually tested with APIKEY and CLIENTID colocated with KNOXTOKEN and tested API specific configs and defaults to insure they work properly together.

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/APIKeyResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/APIKeyResource.java
@@ -22,6 +22,7 @@ import org.apache.knox.gateway.services.security.token.TokenMetadataType;
 import org.apache.knox.gateway.util.JsonUtils;
 
 import javax.inject.Singleton;
+import javax.servlet.FilterConfig;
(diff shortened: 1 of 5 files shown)


pzampino (approved):
LGTM

PR #1053, opened 2025-06-04
KNOX-3156: Ehcache configs not on classpath, fixed persistence dir lo…

hanicz:
…cking issue

- New ability to add files as Ehcache config that are not on the classpath
- Fixed an issue when multiple topologies used the same configuration file and the persistence folder couldn't be locked due to it. In this case the exception is caught and the persistence folder's path is changed. This makes it possible for users to not provide configuration files and the default one used for every topology.
- New test case for multiple cache managers with the same configuration file.

## How was this patch tested?

Unit tests
Manual tests with ldap and pam

Enabled caching on two topologies and verified that the cache is working. Checked the gateway logs as well for related messages.

### DIFF
--- a/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxCacheManager.java
+++ b/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxCacheManager.java
@@ -39,6 +39,7 @@ import org.ehcache.xml.XmlConfiguration;
 import java.io.File;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.nio.channels.OverlappingFileLockException;
 import java.nio.file.Paths;
 import java.util.Optional;
 import java.util.UUID;
@@ -113,19 +114,33 @@ public class KnoxCacheManager implements org.apache.shiro.cache.CacheManager, In
       try {
         manager.init();
       } catch (StateTransitionException e) {
-        LOG.resolvePersistenceDirLockError(e.getMessage());
-        this.resolveLockConflict(xmlConfiguration);
-        if(manager.getStatus() != Status.UNINITIALIZED) {
+        if(containsOverlappingFileLockException(e)) {
+          LOG.resolvePersistenceDirLockError(e.getMessage());
+          this.resolveLockConflict(xmlConfiguration);
+          if(manager.getStatus() != Status.UNINITIALIZED) {
             manager.close();
+          }
+          manager = CacheManagerBuilder.newCacheManager(xmlConfiguration);
+          manager.init();
+        } else {
+          throw e;
         }
-        manager = CacheManagerBuilder.newCacheManager(xmlConfiguration);
-        manager.init();
       }
 
       cacheManagerImplicitlyCreated = true;
     }
 
     return manager;
+  }
+
+  private static boolean containsOverlappingFileLockException(Throwable throwable) {
+    while (throwable != null) {
+      if (throwable instanceof OverlappingFileLockException) {
+        return true;
+      }
+      throwable = throwable.getCause();
+    }
+    return false;
   }
 
   /**


gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/ShiroMessages.java
```diff
@@ -44,4 +44,7 @@ public interface ShiroMessages {
 
   @Message( level = MessageLevel.WARN, text = "There was an error closing the CacheManager, reason: {0}" )
   void errorClosingManagedCacheManager(@StackTrace(level=MessageLevel.WARN) Exception e);
+
+  @Message( level = MessageLevel.WARN, text = "Unable to create Ehcache manager, changing persistence dir, reason: {0}" )
```
pzampino: This message sounds like a failure. With the alternative persistence directory, the creation of the manager should succeed. Perhaps, something like "The default or configured ehcache persistence directory is unavailable; Choosing an alternative directory. reason: {0}" ?

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxCacheManager.java
```diff
@@ -99,16 +108,54 @@ private synchronized org.ehcache.Cache<Object, Object> createCache(String name)
 
   private org.ehcache.CacheManager ensureCacheManager() throws MalformedURLException {
     if (manager == null) {
-      manager = CacheManagerBuilder.newCacheManager(getConfiguration());
-      manager.init();
+      XmlConfiguration xmlConfiguration = getConfiguration();
+      manager = CacheManagerBuilder.newCacheManager(xmlConfiguration);
+      try {
+        manager.init();
+      } catch (StateTransitionException e) {
```
pzampino, line 116: This exception only results from a failure to acquire a lock on the persistence directory? If not, is there a way to determine that the exception was specifically for that reason?

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxCacheManager.java
```diff
@@ -99,16 +108,54 @@ private synchronized org.ehcache.Cache<Object, Object> createCache(String name)
...
+    Optional<ServiceCreationConfiguration<?>> serviceConfig = xmlConfiguration.getServiceCreationConfigurations().stream()
+            .filter(service -> service instanceof CacheManagerPersistenceConfiguration).findFirst();
+
+    if (serviceConfig.isPresent()) {
+      CacheManagerPersistenceConfiguration cachePersistenceConfig = (CacheManagerPersistenceConfiguration) serviceConfig.get();
+      String path = cachePersistenceConfig.getRootDirectory().getPath();
+      xmlConfiguration.getServiceCreationConfigurations().remove(cachePersistenceConfig);
+      String newFolder = DEFAULT_FOLDER_NAME + UUID.randomUUID().toString().substring(0, 4);
+      String newRootDirectory = Paths.get(path).getParent().resolve(newFolder).toAbsolutePath().toString();
+      xmlConfiguration.getServiceCreationConfigurations()
```
pzampino, line 163: Is there anything else possibly configured in a CacheManagerPersistenceConfiguration in addition to the location? Could we be omitting other user-defined config by replacing the original with this simplified version?

hanicz: No, the class only holds one attribute the rootDirectory. I create an instance of it the same way as the Ehcache code does.

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/ShiroMessages.java
hanicz: Yes that makes more sense. I changed the log message.

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxCacheManager.java
hanicz: Other issues could cause StateTransitionException as well. Since this is our only retry mechanic I don't see a point identifying the cause. We could only determine it from the error message however I don't think that's reliable since it could change with jdk versions for example.

hanicz: There should be an OverlappingFileLockException in the getCause() stack. I think I can check for that.

pzampino: Changing the log method name was good also. Thanks.

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxCacheManager.java
```diff
@@ -113,13 +114,17 @@ private org.ehcache.CacheManager ensureCacheManager() throws MalformedURLExcepti
...
+        if(containsOverlappingFileLockException(e)) {
+          LOG.resolvePersistenceDirLockError(e.getMessage());
+          this.resolveLockConflict(xmlConfiguration);
+          if(manager.getStatus() != Status.UNINITIALIZED) {
             manager.close();
+          }
+          manager = CacheManagerBuilder.newCacheManager(xmlConfiguration);
+          manager.init();
+        } else {
+          throw e;
```
pzampino, line 126: e is the result of getCause() at this point? Is that your intention?

hanicz: No, the 'e' here is the original StateTransitionException. The code only throws it if there was no locking exception in the getCause chain.

PR #1056, opened 2025-06-12
KNOX-3158: Enhancing the maintainability of knox-shell

Raghav-Mah3shwari:
Increasing the maintainability of the code by reducing code smells from gateway-shell

How was this patch tested?

The changes were made in local first then a maven build was triggered. The build was successful


smolnar82 (changes requested):
Please use Knox's `MesageFactory` feature and the relevant interface (in this case this would be `org.apache.knox.gateway.shell.KnoxShellMessages`) instead of the plain Logger approach.

PR #1057, opened 2025-06-17
KNOX-3154: Implement CM discovery support for the Apache Iceberg REST Catalog Service

bonampak:

Added and registered IcebergRestServiceModelGenerator for CM service discovery.

## How was this patch tested?

Added unit and integration tests and tested on a dev cluster.


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/hive/IcebergRestServiceModelGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  @Override
+  public ServiceModel generateService(ApiService service,
+                                      ApiServiceConfig serviceConfig,
+                                      ApiRole role,
+                                      ApiConfigList roleConfig, ApiServiceConfig coreSettingsConfig) throws ApiException {
+    String hostname = role.getHostRef().getHostname();
+    String scheme = "http";
+    String port = getHttpPort(serviceConfig);
```
pzampino, line 78: Does this service not have a separate HTTPS configuration?

pzampino (approved):

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/model/hive/IcebergRestServiceModelGenerator.java
bonampak: It only has http. Once they add https, we'll add a separate PR.

PR #1058, opened 2025-07-09
KNOX-3167 - Upgrade to Node v16 as it is the first version that supports ARM64 architecture

smolnar82:

Upgraded the following dependencies:
- NodeJS to `v16.x`: this was necessary because this is the first Node version that supports ARM64
- `frontend-maven-plugin` to `1.15`: without this upgrade the plugin tried to download the x64 artifact
- JNA to `5.9.0`: after Knox was built and redeployed, the Knox Home page was buggy due to the incompatible UnixCLibrary loading on ARM64 so that I'd to upgrade to a version where it's supported

I also had to
- remove the obsolete `angular/http` dependency
- remove `package-lock.json` files and exclude them from Git
- adjust node/types depdendency to satisfy both x64 and ARM64 architectures
- install `npm` with `--legacy-peer-deps` switch so that we don't need to get rid of old(ish) features such as `angular2-datatable` or `ACE` editor for now.

## How was this patch tested?

Confirmed Knox is built:
```
smolnar $ mvn -DskipTests -Dcheckstyle.skip=true -Dfindbugs.skip=true -Dpmd.skip=true -Drat.skip -Dspotbugs.skip=true -Dforbiddenapis.skip=true -Ppackage clean install
...
[INFO] BUILD SUCCESS
[INFO] Total time:  02:48 min
[INFO] Finished at: 2025-07-09T09:41:33+02:00
```

Did a smoke test on the Knox UIs and they seem to work properly:
image
image
image
I also confirmed that there were no errors in the logs.


gateway-admin-ui/package-lock.json
moresandeep, line 1: We most likely do not need to check in package-lock.json.

knox-homepage-ui/package-lock.json
moresandeep, line 1: We most likely do not need to check in package-lock.json.

knox-token-generation-ui/package-lock.json
moresandeep, line 1: We most likely do not need to check in package-lock.json.

knox-token-management-ui/package-lock.json
moresandeep, line 1: We most likely do not need to check in package-lock.json.

knox-webshell-ui/package-lock.json
moresandeep, line 1: We most likely do not need to check in package-lock.json.

moresandeep (changes requested):
Looks good, the only suggestion i have is that we don't need to check in package-lock.json file :)

gateway-admin-ui/package-lock.json
smolnar82: Ack.

knox-token-generation-ui/package-lock.json
smolnar82: Ack.

knox-webshell-ui/package-lock.json
smolnar82: Ack.

knox-token-management-ui/package-lock.json
smolnar82: Ack.

knox-homepage-ui/package-lock.json
smolnar82: Ack.

PR #1065, opened 2025-07-25
KNOX-3172: New intercepting socket and outputstream to handle FIPS bo…

hanicz:
…uncycastle exception, set max connections for PoolingHttpClientConnectionManager

The BC FIPS provider causes a SocketException with 'Broken Pipe' message on FIPS clusters. When there is a connection: close header in the response Knox tries to close the connection however there is an exception coming from BC. It tries to write to the already closed connection and we get the Broken Pipe issue and it results in HTTP 500 responses from Knox.

The solution catches and ignores this exception on the socket level. The intercepting socket would only load if FIPS arg is provided for Knox. This arg is defaults to com.safelogic.cryptocomply.fips.approved_only=true and can be changed in the gateway-site.xml.

## How was this patch tested?
Unit tests
Tested on FIPS cluster with fips arg and BC provider loaded by JDK.


smolnar82 (approved):
LGTM

gateway-spi/src/main/java/org/apache/knox/gateway/fips/BCInterceptingSocket.java
```diff
@@ -0,0 +1,275 @@
...
+
+    @Override
+    public void setPerformancePreferences(int connectionTime,
+                                          int latency,
+                                          int bandwidth) {
+        delegate.setPerformancePreferences(connectionTime, latency, bandwidth);
+    }
+
+    /**
+     * This method is only available in JDK9+ therefor reflection is used to call it.
```
smolnar82, line 243: nit: typo (`therefore` vs. `therefor`)

PR #1069, opened 2025-07-30
KNOX-3162 - Build support for JDK 17

moresandeep:
This PR adds builds support for JDK 17 and removes support of JDK 8.

## How was this patch tested?
This patch was locally tested
1. Ran a full build with unit tests on JDK 17
2. Installed Knox runtime on JDK 17 and tested basic functionality (SSO)


gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/ZookeeperTokenStateService.java
```diff
@@ -58,16 +62,77 @@ public ZookeeperTokenStateService(GatewayServices gatewayServices, AliasServiceF
   @Override
   public void init(GatewayConfig config, Map<String, String> options) throws ServiceLifecycleException {
     log.deprecatedServiceUsage(this.getClass().getCanonicalName());
-    final ZookeeperRemoteAliasService zookeeperAliasService = (ZookeeperRemoteAliasService) aliasServiceFactory.create(gatewayServices, ALIAS_SERVICE, config, options,
+
+    // Debug logging for troubleshooting NullPointerException
```
hanicz: I believe these DEBUG console calls were left here by accident.

moresandeep: Ahh, my bad! I will fix this thank you!

gateway-server/src/main/java/org/apache/knox/gateway/services/token/impl/ZookeeperTokenStateService.java
```diff
@@ -58,14 +62,46 @@ public ZookeeperTokenStateService(GatewayServices gatewayServices, AliasServiceF
...
+
+
     options.put(ZookeeperRemoteAliasService.OPTION_NAME_SHOULD_CREATE_TOKENS_SUB_NODE, "true");
     options.put(ZookeeperRemoteAliasService.OPTION_NAME_SHOULD_USE_LOCAL_ALIAS, "false");
-    zookeeperAliasService.registerRemoteTokenStateChangeListener(this);
-    zookeeperAliasService.init(config, options);
-    super.setAliasService(zookeeperAliasService);
-    super.init(config, options);
+
+    try {
```
hanicz: These try-catches don't seem to address anything.

moresandeep:
Moving to a different branch https://github.com/apache/knox/tree/feature/jdk17_upgrade

PR #1074, opened 2025-08-11
Moved client id validation into client secret parse, Modify getParameter in UrlEncodedFormRequest

hanicz:

- Modified the overridden getParameter in UrlEncodedFormRequest so it reads the client credentials from the payload instead of the query params.
- Moved the client id validation into the client credentials parse logic to avoid reading the request body during passcode flow.

## How was this patch tested?

Unit tests, tested the client credentials flow locally

```
curl -ivku admin:admin-password https://localhost:8443/gateway/test-client-id/clientid/api/v1/oauth/credentials

curl -ivk -X POST \
  "https://localhost:8443/gateway/tokenbox/knoxtoken/api/v2/token" \
  -H "Content-Type: application/x-www-form-urlencoded" -H "X-Client-Credentials-Flow: true" \
  -d "grant_type=client_credentials&client_id=197bac1e-7968-4a73-89e7-cb5f1ee8e590&client_secret=TVRrM1ltRmpNV1V0TnprMk9DMDBZVGN6TFRnNVpUY3RZMkkxWmpGbFpUaGxOVGt3OjpNelppTmpjNE1EWXRaREkzT0MwME1XVm1MV0prWm1VdE5XTTFZV1E0TTJObE5qSmw="

curl -k -X POST \
  "https://localhost:8443/gateway/tokenbox/knoxtoken/api/v2/token" \
  -H "Authorization: Basic UGFzc2NvZGU6VFcxWmVscFVTbWhOUkZWMFRrUkZNazE1TURCT2FrVTFURmRKTlZwRVNYUk5hbXMwVG0xSmQxcEhSVEJOVkVFeU9qcE5ha1Y1V21wbmVscHRWWFJaVjBWNVRWTXdNRnBVVm10TVYwVXhUWHBqZEU1RVJtcE9hbWhxVDBSTmVVMUhUWGs9"
```

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
@@ -262,7 +262,7 @@ public class JWTFederationFilter extends AbstractJWTFilter {
 
   private void validateClientID(HttpServletRequest request, String tokenValue) {
     final String clientID = request.getParameter(CLIENT_ID);
-    String tokenId = null;
+    String tokenId;
     try {
       final String[] base64DecodedTokenIdAndPasscode = decodeBase64(tokenValue).split("::");
       tokenId = decodeBase64(base64DecodedTokenIdAndPasscode[0]);
@@ -271,7 +271,7 @@ public class JWTFederationFilter extends AbstractJWTFilter {
     }
     // if there is no client_id then this is not a client credentials flow
     if (clientID != null && !tokenId.equals(clientID)) {
-     throw new SecurityException("Client ID mismatch");
+     throw new SecurityException(MISMATCHING_CLIENT_ID_AND_CLIENT_SECRET);
     }
   }
 


Raghav-Mah3shwari (approved):
LGTM

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -264,22 +260,19 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
+    String tokenId = null;
+    try {
+      final String[] base64DecodedTokenIdAndPasscode = decodeBase64(tokenValue).split("::");
+      tokenId = decodeBase64(base64DecodedTokenIdAndPasscode[0]);
+    } catch (Exception e) {
+      throw new SecurityException("Error while parsing the received client secret", e);
+    }
+    // if there is no client_id then this is not a client credentials flow
+    if (clientID != null && !tokenId.equals(clientID)) {
+     throw new SecurityException("Client ID mismatch");
```
pzampino: Could we not still use the MISMATCHING_CLIENT_ID_AND_CLIENT_SECRET message?

gateway-server/src/main/java/org/apache/knox/gateway/UrlEncodedFormRequest.java
```diff
@@ -77,6 +78,9 @@ private MultiMap<String> parseQueryString(String queryString) {
 
   @Override
   public String getParameter(String name) {
+    if(JWTFederationFilter.GRANT_TYPE.equals(name) || JWTFederationFilter.CLIENT_ID.equals(name) || JWTFederationFilter.CLIENT_SECRET.equals(name)) {
+      return super.getParameter(name);
+    }
     return queryParams.getValue(name, 0);
```
lmccay, line 84: I am assuming there is some reason that we were limiting this to just the query string but do you know why? If we checked whether queryParams contains the name first then use that otherwise fall back to the parent then we don't need to check that the name matches one of those three? I feel like if we aren't violating the original intent of this method for those three then why are we for any others?

hanicz: The below comment is on the `UrlEncodedFormRequest` class. I think it explains clearly the issue with calling the super's getParameter.

```
 * The problem is that various authentication filters (such as HadoopAuthFilter) check if there is a doAs parameter in request.
 * This will consume the input stream and the dispatch will forward an empty body to the service.
 *
 * To avoid this problem all "x-www-form-urlencoded" requests are wrapped into UrlEncodedFormRequest.
 *
 * This class ignores the request body when accessing the parameters (since KNOX as a proxy doesn't care about the payload either),
 * and it only cares about the query string.
```

pzampino: I had this question myself.

lmccay: Interesting. Really makes me wonder about rewrites which can require request body changes.
At any rate, I'll give my plus one for when other comments are resolved.

pzampino: For proxying cases, that makes sense, but when Knox itself is the terminus, and the API includes form data, we end up with this unusual situation. It seems like perhaps we should only be using UrlEncodedFormRequest when the content type is "x-www-form-urlencoded"  AND Knox is proxying. When Knox is providing the API itself, this restriction is not necessary and this extension need not be applied. The trick is discerning proxying requests from "Knox-native" API requests.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
hanicz: Modified as suggested

pzampino (approved):

PR #1075, opened 2025-08-12
KNOX-3181: PasscodeTokenResourceBase should extend TokenResourceV2

pzampino:

Changed the base class for PasscodeTokenResourceBase from TokenResource to TokenResourceV2.

## How was this patch tested?

Full build and verify, and manually validated the HTTP methods allowed for the various token lifecycle operations using the CLIENTID service (e.g., revoke requires DELETE rather than POST).


smolnar82 (approved):
LGTM

PR #1083, opened 2025-09-11
KNOX-3188 - Token renewal and revocation should be authorized based on the group membership of the authenticated subject.

smolnar82:

Implemented the proposed solution described in [KNOX-3188](https://issues.apache.org/jira/browse/KNOX-3188).

## How was this patch tested?

Updated existing unit tests, added new test cases to cover the new functionality and executed them successfully.

Additionally, I conducted manual testing using the OOTB LDAP server with different users:
```
analyst  group
member: uid=sam,ou=people,dc=hadoop,dc=apache,dc=org
member: uid=tom,ou=people,dc=hadoop,dc=apache,dc=org

scientist group
member: uid=sam,ou=people,dc=hadoop,dc=apache,dc=org

admin group
member: uid=admin,ou=people,dc=hadoop,dc=apache,dc=org
```

In order for the group-based authorization to work, I changed the `DEFAULT` identity assertion filter to `HadoopGroupProvider` in the `homepage` topology as well as edited the `gateway-site.xml` and added `sam` and `tom` to `gateway.knox.token.management.users.can.see.all.tokens`.

Other than this, I customized the `knox.token.renewer.whitelist` (RW, going forward) and `knox.token.renewer.group.whitelist` (GRW, going forward) service-level params.

Since token renewal is not exposed on the token management UI, only token revocation, I tested using that functionality (they both have been updated the same way).

Cases:
- Logged in user: admin, RW: admin, GRW: empty -> token revocation succeeded ✅
- Logged in user: admin, RW: empty, GRW: admin -> token revocation succeeded ✅
- Logged in user: sam, RW: empty, GRW: scientist -> token revocation succeeded ✅
- Logged in user: sam, RW: empty, GRW: analyst -> token revocation succeeded ✅
- Logged in user: sam, RW: empty, GRW: admin -> token revocation failed ✅
- Logged in user: tom, RW: empty, GRW: scientist -> token revocation failed ✅
- Logged in user: tom, RW: empty, GRW: analyst -> token revocation succeeded ✅

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -574,14 +574,8 @@ public class TokenResource {
       }
     } else {
       final String renewer = SubjectUtils.getCurrentEffectivePrincipalName();
-      final Set<GroupPrincipal> groups = SubjectUtils.getCurrentGroupPrincipals();
 
-      final boolean userAllowed = allowedRenewers.contains(renewer);
-      final boolean groupAllowed = groups.stream()
-              .map(GroupPrincipal::getName)
-              .anyMatch(allowedRenewerGroups::contains);
-
-      if (userAllowed || groupAllowed) {
+      if (tokenRenewalOrRevocationAuthorized(renewer)) {
         try {
           JWTToken jwt = new JWTToken(token);
           if (tokenStateService.isExpired(jwt)) {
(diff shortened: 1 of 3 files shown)


gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -562,8 +573,15 @@ public Response renew(String token) {
         errorCode = ErrorCode.INTERNAL_ERROR;
       }
     } else {
-      String renewer = SubjectUtils.getCurrentEffectivePrincipalName();
-      if (allowedRenewers.contains(renewer)) {
+      final String renewer = SubjectUtils.getCurrentEffectivePrincipalName();
+      final Set<GroupPrincipal> groups = SubjectUtils.getCurrentGroupPrincipals();
+
+      final boolean userAllowed = allowedRenewers.contains(renewer);
```
hanicz: Since it doesn't matter that it's `userAllowed` or `groupAllowed` it would be beneficial to move the authorisation into a separate method. It would solve the duplication and only one variable would hold the outcome.

smolnar82: Good idea, I'll submit a new PS soon.

smolnar82: Done.

PR #1090, opened 2025-09-25
KNOX-3195: Add support for Flink Dashboard and Flink SQL gateway

hanicz:

This change adds support for Flink Dashboard and Flink SQL gateway API. To maintain backward compatibility Flink as a service definition is kept around.

## How was this patch tested?

Tested on dev cluster with Flink


smolnar82 (approved):
LGTM

PR #1092, opened 2025-10-01
KNOX-3203: Remove jakarta activation-api duplication

hanicz:

- Removed `jakarta.activation-api` duplication.
- Specified the `com.sun.activation` version so it is easily distinguished from jakarta activation-api

## How was this patch tested?
Build, unit tests


moresandeep (approved):

pom.xml
```diff
@@ -1506,13 +1506,7 @@
             <dependency>
                 <groupId>com.sun.activation</groupId>
```
moresandeep, line 1507: com.sun.activation will go away with JDK17 changes.

PR #1197, opened 2026-04-10
KNOX-3297: Docker entrypoint script has issues that prevents this image being run as docker container and as a pod in k8s

sneethiraj:

KNOX-3297: Docker entrypoint script has issues that prevents this image being run as docker container and as a pod in k8s

1. Modified to use keytool from the PATH instead of hardcoding it.
2. Fixed a scripting error related to default init of MASTER_S... environment variable.
3. Added additional validation for file existance before additing it to truststore. Also, introduced a custom env variable TRUSTSTORE_IMPORTS to allow customers to add more number of certificates into truststore.

## How was this patch tested?

1. Validated the changes as standalone docker [run]
2. Validated by running the docker image as k8s pod

## Integration Tests
no new integration tests.

## UI changes
None.

### DIFF
--- a/gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
+++ b/gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
@@ -187,45 +187,40 @@ then
   importMultipleCerts "$CUSTOM_CERT"
 fi
 
-# Add Amazon Root CA 1
-keytool -importcert \
-  -keystore "${KEYSTORE_DIR}"/truststore.jks \
-  -alias amazon-ca-1 \
-  -file /home/knox/cacrts/AmazonRootCA1.cer \
-  -storepass "${ALIAS_PASSPHRASE}" \
-  -noprompt || true
+# This default was set to emulate the existing behaviour
+# Customer should be able to override this by specifying this via Docker environment settings
+if [[ ! -n ${TRUSTSTORE_IMPORTS} ]]
+then
+	TRUSTSTORE_IMPORTS="
+	 amazon-ca-1:/home/knox/cacrts/AmazonRootCA1.cer
+     amazon-ca-2:/home/knox/cacrts/AmazonRootCA2.cer
+     amazon-ca-3:/home/knox/cacrts/AmazonRootCA3.cer
+	 amazon-ca-4:/home/knox/cacrts/AmazonRootCA4.cer
+     letsencrypt-stg-root:/home/knox/cacrts/letsencrypt-stg-root-x1.pem"
+fi
 
-# Add Amazon Root CA 2
-keytool -importcert \
-  -keystore "${KEYSTORE_DIR}"/truststore.jks \
-  -alias amazon-ca-2 \
-  -file /home/knox/cacrts/AmazonRootCA2.cer \
-  -storepass "${ALIAS_PASSPHRASE}" \
-  -noprompt || true
-
-# Add Amazon Root CA 3
-keytool -importcert \
-  -keystore "${KEYSTORE_DIR}"/truststore.jks \
-  -alias amazon-ca-3 \
-  -file /home/knox/cacrts/AmazonRootCA3.cer \
-  -storepass "${ALIAS_PASSPHRASE}" \
-  -noprompt || true
-
-# Add Amazon Root CA 4
-keytool -importcert \
-  -keystore "${KEYSTORE_DIR}"/truststore.jks \
-  -alias amazon-ca-4 \
-  -file /home/knox/cacrts/AmazonRootCA4.cer \
-  -storepass "${ALIAS_PASSPHRASE}" \
-  -noprompt || true
-
-# Add letsencrypt staging root CA
-keytool -importcert \
-  -keystore "${KEYSTORE_DIR}"/truststore.jks \
-  -alias letsencrypt-stg-root \
-  -file /home/knox/cacrts/letsencrypt-stg-root-x1.pem \
-  -storepass "${ALIAS_PASSPHRASE}" \
-  -noprompt || true
+if [[ -n ${TRUSTSTORE_IMPORTS} ]]
+then
+    for certinfo in ${TRUSTSTORE_IMPORTS}
+    do
+        aliasId="`echo ${certinfo} | awk -F: '{ print $1 }'`"
+        certPath="`echo ${certinfo} | awk -F: '{ print $2 }'`"
+        if [[ -n "${aliasId}" ]] && [[ -n "${certPath}" ]] && [[ -f "${certPath}" ]]
+        then
+            echo "INFO: Importing certificate [${certPath}] into truststore"
+            keytool -importcert \
+                -keystore "${KEYSTORE_DIR}"/truststore.jks \
+                -alias "${aliasId}" \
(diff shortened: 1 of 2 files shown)


gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -86,11 +86,12 @@ fi
 
 if [[ -n ${MASTER_SECRET} ]]
 then
-  echo "Using provided knox master secret"
-  /home/knox/knox/bin/knoxcli.sh create-master --master "${MASTER_SECRET}"
+  echo "Using provided knox master secret [env:MASTER_SECRET}]"
```
smolnar82: nit: `}` is not needed.

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -187,45 +188,40 @@ then
...
+then
+	TRUSTSTORE_IMPORTS="
+	 amazon-ca-1:/home/knox/cacrts/AmazonRootCA1.cer
+     amazon-ca-2:/home/knox/cacrts/AmazonRootCA2.cer
+     amazon-ca-3:/home/knox/cacrts/AmazonRootCA3.cer
+	 amazon-ca-4:/home/knox/cacrts/AmazonRootCA4.cer
+     letsencrypt-stg-root:/home/knox/cacrts/letsencrypt-stg-root-x1.pem"
+fi
+
+if [[ -n ${TRUSTSTORE_IMPORTS} ]]
```
smolnar82: This is redundant: end-users either define their own `TRUSTSTORE_IMPORTS` or we set to default above -> the if is unnecesary.

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -187,45 +188,40 @@ then
...
+then
+    for certinfo in ${TRUSTSTORE_IMPORTS}
+    do
+        aliasId="`echo ${certinfo} | awk -F: '{ print $1 }'`"
```
smolnar82: Use $(...) notation instead of legacy backticks `...`.

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -187,45 +188,40 @@ then
...
+        certPath="`echo ${certinfo} | awk -F: '{ print $2 }'`"
```
smolnar82: Use $(...) notation instead of legacy backticks `...`.

smolnar82 (changes requested):
Thanks, @sneethiraj , for your contribution. I've only a couple small requests before we merge your PR :)
I'd also like to ask you to run [shellcheck](https://www.shellcheck.net/) before you submit you next changeset. As of today, it'd be failiing:

Image

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
sneethiraj: addressed in commit - d03092d8e

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
sneethiraj: addressed in commit - d03092d8e

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
sneethiraj: Changed to use $() notation. commit - d03092d8e

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
sneethiraj: Changed to use $() notation. commit - d03092d8e

sneethiraj:
Used shellcheck to address open issues on my changes:
image

smolnar82 (approved):
LGTM

smolnar82:
Thanks, @sneethiraj , for your contribution and the quick reaction to my review comments!

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -86,11 +86,12 @@ fi
 
 if [[ -n ${MASTER_SECRET} ]]
 then
-  echo "Using provided knox master secret"
-  /home/knox/knox/bin/knoxcli.sh create-master --master "${MASTER_SECRET}"
+  echo "Using provided knox master secret [env:MASTER_SECRET]"
```
moresandeep, line 89: We are echoing the secret, that can be an issue.

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -187,45 +188,35 @@ then
...
+then
+	TRUSTSTORE_IMPORTS="
+	 amazon-ca-1:/home/knox/cacrts/AmazonRootCA1.cer
+     amazon-ca-2:/home/knox/cacrts/AmazonRootCA2.cer
+     amazon-ca-3:/home/knox/cacrts/AmazonRootCA3.cer
+	 amazon-ca-4:/home/knox/cacrts/AmazonRootCA4.cer
+     letsencrypt-stg-root:/home/knox/cacrts/letsencrypt-stg-root-x1.pem"
+fi
+
+for certinfo in ${TRUSTSTORE_IMPORTS}
```
moresandeep, line 203: When we specify `TRUSTSTORE_IMPORTS` then the OOTB certs are not included? this might be a problem here is why: In docker we do not use system keystore which means we do not have green certs that are provisioned in the keystore as a result ALL communication to certs signed by public CAs will fail when using `TRUSTSTORE_IMPORTS` :(

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
moresandeep, line 33: Adding some docs here for `TRUSTSTORE_IMPORTS` would have been helpful.

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
smolnar82: Nope, the output will literally be:
`Using provided knox master secret [env:MASTER_SECRET]`

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
smolnar82: I might be wrong, but the new logic mimics what we had before: adding the Amazon and Lets
Encrypt certificates to `${KEYSTORE_DIR}"/truststore.jks`. Nothing more, nothing less.
The only difference is that you can now overide that with your own certs if you want. But the change - IMO - is fully backward compatible and preserves the previous behavior.

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
smolnar82: Yes, this is a good point, I think @sneethiraj agrees and will submit a new PR with the missing ENV variable docs.

PR #1207, opened 2026-04-21
KNOX-3302: Falling back to gateway-level credential store while looking up LDAP system password

smolnar82:
[KNOX-3302](https://issues.apache.org/jira/browse/KNOX-3302) - Falling back to gateway-level credential store while looking up LDAP system password

  This pull request improves the LDAP system password alias resolution in the Shiro realm (KnoxLdapContextFactory). Key changes include:
   * Fallback Alias Resolution: Updated `KnoxLdapContextFactory` to check the gateway-level credential store if a system password alias is not found in the cluster-level (topology) credential store.
   * Centralized Alias Logic: Added isAlias and extractAlias default methods to the `AliasService` interface to standardize how alias strings (e.g., `S{ALIAS=...`}) are identified and processed across the codebase.
   * Improved Testability: Refactored KnoxLdapContextFactory to allow for better unit testing by extracting the AliasService lookup into a protected method.
   * Code Cleanup: Replaced manual string parsing with the new AliasService methods and added validation for blank passwords using StringUtils.

 ## How was this patch tested?
The changes were verified by adding a new unit test class: `org.apache.knox.gateway.shirorealm.KnoxLdapContextFactoryTest`.

  The following scenarios were tested:
   1. Blank Password: Verified that null or empty strings do not trigger alias resolution.
   2. Literal Password: Verified that passwords not matching the alias pattern are used as-is.
   3. Cluster-level Alias: Verified successful resolution when the alias exists in the cluster-specific credential store.
   4. Gateway-level Fallback: Verified that the factory correctly falls back to the gateway credential store when the alias is missing from the cluster store.
   5. Missing Alias: Verified that the system password is set to an empty string and a warning is logged when an alias cannot be resolved in either store.

 ## Integration Tests
  No new integration tests were added as the logic was fully covered by the new unit tests in KnoxLdapContextFactoryTest. Existing Shiro-related integration tests should be monitored for regressions.

##  UI changes
  N/A

### DIFF
--- a/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxLdapContextFactory.java
+++ b/gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxLdapContextFactory.java
@@ -24,6 +24,7 @@ import javax.naming.Context;
 import javax.naming.NamingException;
 import javax.naming.ldap.LdapContext;
 
+import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.GatewayMessages;
 import org.apache.knox.gateway.GatewayServer;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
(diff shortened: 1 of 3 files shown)


lmccay (approved):
LGTM
Here is my +1 pending tests succssful completion.

PR #1212, opened 2026-04-23
KNOX-3306: Make server startup fail if a forbidden security algorithm…

hanicz:
… is configured for Knox in a FIPS environment

[KNOX-3306](https://issues.apache.org/jira/browse/KNOX-3306) - Make server startup fail if a forbidden security algorithm is configured for Knox in a FIPS environment

Server startup fails if ARC4, MD5 or SHA1 is used on FIPS environment.

## How was this patch tested?

Tested locally with allowed and forbidden algorithms while the system property was set.

## Integration Tests
N/A

## UI changes
N/A

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultCryptoService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultCryptoService.java
@@ -64,11 +64,7 @@ public class DefaultCryptoService implements CryptoService {
       throw new ServiceLifecycleException("Alias service is not set");
     }
     if (FipsUtils.isFipsEnabledWithBCProvider()) {
-      //invoking the following getters will throw IllegalArgumentException in case a forbidden algorithm is set
-      //so we can use them as a validation at service initialization time
-      config.getCredentialStoreAlgorithm();
-      config.getAlgorithm();
-      config.getPBEAlgorithm();
+      FipsUtils.validateAlgorithms(config);
     }
   }
 
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultCryptoService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultCryptoService.java
@@ -29,6 +29,7 @@ import java.util.Map;
 
 import org.apache.knox.gateway.GatewayMessages;
 import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.fips.FipsUtils;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.services.security.AliasService;
 import org.apache.knox.gateway.services.security.AliasServiceException;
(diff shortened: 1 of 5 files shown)


gateway-spi/src/main/java/org/apache/knox/gateway/fips/FipsUtils.java
```diff
@@ -17,11 +17,43 @@
...
+        if (!isFipsEnabledWithBCProvider() || algorithm == null || algorithm.isEmpty()) {
+            return true;
+        }
+
+        String upperCaseAlg = algorithm.toUpperCase(Locale.ROOT);
+        for (String forbidden : FORBIDDEN_ALGORITHMS) {
+            if (upperCaseAlg.contains(forbidden)) {
+                return false;
+            }
+        }
```
smolnar82: Why not a simple ` return !isFipsEnabledWithBCProvider() || algorithm == null || algorithm.isEmpty() || !FORBIDDEN_ALGORITHMS.contains(algorithm.toUpperCase(Locale.ROOT))`?

gateway-spi/src/main/java/org/apache/knox/gateway/fips/FipsUtils.java
```diff
@@ -17,11 +17,43 @@
...
+        return true;
+    }
+
+    public static void validateAlgorithm(String algorithm, String paramName) {
+        if (!isAlgorithmAllowed(algorithm)) {
+            throw new IllegalArgumentException("In a FIPS environment, you are not allowed to use " + algorithm + " as " + paramName);
+        }
+    }
```
smolnar82, line 52: Extracting the error message out to a constant template and re-use it in the tests?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultCryptoService.java
```diff
@@ -57,11 +58,18 @@ public void setAliasService(AliasService as) {
...
-  if (aliasService == null) {
+    if (aliasService == null) {
       throw new ServiceLifecycleException("Alias service is not set");
     }
+    if (FipsUtils.isFipsEnabledWithBCProvider()) {
+      //invoking the following getters will throw IllegalArgumentException in case a forbidden algorithm is set
+      //so we can use them as a validation at service initialization time
+      config.getCredentialStoreAlgorithm();
+      config.getAlgorithm();
+      config.getPBEAlgorithm();
```
smolnar82: We may want to add a new GatewayConfig method called `validateFipsAlgorithms` and have the `GatewayConfigImpl` self-contain what does it mean to be valid in FIPS envs?

smolnar82 (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/DefaultCryptoService.java
hanicz: Moved them into FipsUtils

smolnar82 (approved):

PR #1218, opened 2026-05-02
KNOX-3312 - Client Credentials Flow with HTTP Basic needs Unwrapped Servlet Request

lmccay:

[KNOX-3312](https://issues.apache.org/jira/browse/KNOX-3312) - A short description of the change

Current implementation can't get to the grant_type request param.

Unit tests mock out the requests and make it hard to tease this out as an issue.

When we know that there is an Authorization header and that it is Basic then we need to check whether there is the hardcoded username of token or passcode and if not, unwrap the request to check for a grant_type for OAuth client_credentials and handle it appropriately.

Current implementation tries to check that but the params are hidden by the wrappers.

## How was this patch tested?

All existing unit and integration tests were run and it was manually tested.


lmccay:
This has some unintended changes from another PR in it - closing and will create another.

lmccay:
actually don't need to close it. Will just fix it. :)

PR #1219, opened 2026-05-02
KNOX-3312 - Client Credentials Flow with HTTP Basic needs Unwrapped S…

lmccay:
[KNOX-3312](https://issues.apache.org/jira/browse/KNOX-3312) - Client Credentials Flow with HTTP Basic needs Unwrapped Servlet Request

Current implementation can't get to the grant_type request param.

Unit tests mock out the requests and make it hard to tease this out as an issue.

When we know that there is an Authorization header and that it is Basic then we need to check whether there is the hardcoded username of token or passcode and if not, unwrap the request to check for a grant_type for OAuth client_credentials and handle it appropriately.

Current implementation tries to check that but the params are hidden by the wrappers.

## How was this patch tested?

All existing unit and integration tests were run.
Manually tested the flow.

Also added a request wrapper to the unit tests of the mock http request.
This ensures that the access to the request params for OAuth flows does not regress.

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
@@ -361,12 +361,15 @@ public class JWTFederationFilter extends AbstractJWTFilter {
       String passcode = values[1].isEmpty() ? null : values[1];
       if (TOKEN.equalsIgnoreCase(username) || PASSCODE.equalsIgnoreCase(username)) {
           parsed = Pair.of(TOKEN.equalsIgnoreCase(username) ? TokenType.JWT : TokenType.Passcode, passcode);
-      } else if (request != null && CLIENT_CREDENTIALS.equals(request.getParameter(GRANT_TYPE))) {
-          // Allow client_credentials flow where client_id/client_secret are provided via HTTP Basic
-          if (passcode != null) {
-            validateClientID(username, passcode);
-            parsed = Pair.of(TokenType.Passcode, passcode);
-          }
+      } else if (request != null) {
+          HttpServletRequest unwrappedRequest = ServletRequestUtils.unwrapHttpServletRequest(request);
+          if (CLIENT_CREDENTIALS.equals(unwrappedRequest.getParameter(GRANT_TYPE))) {
+            // Allow client_credentials flow where client_id/client_secret are provided via HTTP Basic
+            if (passcode != null) {
+              validateClientID(username, passcode);
+              parsed = Pair.of(TokenType.Passcode, passcode);
+            }
+            }
       }
 
       return parsed;
--- a/gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/OAuthFlowsFederationFilterTest.java
+++ b/gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/OAuthFlowsFederationFilterTest.java
@@ -47,8 +47,55 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import javax.servlet.http.HttpServletRequestWrapper;
 
 public class OAuthFlowsFederationFilterTest extends TokenIDAsHTTPBasicCredsFederationFilterTest {
+
+    /**
+     * Test wrapper that hides access to parameters and request body to simulate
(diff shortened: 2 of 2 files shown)


pzampino (approved):
LGTM

moresandeep (approved):

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -361,12 +361,15 @@ private Pair<TokenType, String> parseFromHTTPBasicCredentials(final String heade
...
       if (TOKEN.equalsIgnoreCase(username) || PASSCODE.equalsIgnoreCase(username)) {
           parsed = Pair.of(TOKEN.equalsIgnoreCase(username) ? TokenType.JWT : TokenType.Passcode, passcode);
-      } else if (request != null && CLIENT_CREDENTIALS.equals(request.getParameter(GRANT_TYPE))) {
-          // Allow client_credentials flow where client_id/client_secret are provided via HTTP Basic
-          if (passcode != null) {
-            validateClientID(username, passcode);
-            parsed = Pair.of(TokenType.Passcode, passcode);
-          }
+      } else if (request != null) {
+          HttpServletRequest unwrappedRequest = ServletRequestUtils.unwrapHttpServletRequest(request);
```
smolnar82, line 365: What if we added a util method in ServletRequestUtils, like this:
```
    public static String getRequestParam(final ServletRequest request, final String paramName) {
        if (request == null) {
            return null;
        }
        String requestParamValue = request.getParameter(paramName);
        if (requestParamValue == null) {
            requestParamValue = ServletRequestUtils.unwrapHttpServletRequest(request).getParameter(paramName);
        }
        return requestParamValue;
    }
```
Then the above change would look like:
```
} else if (CLIENT_CREDENTIALS.equals(ServletRequestUtils.getRequestParam(GRANT_TYPE)))
```
I did the same in my KnoxIDF branch, which I realized it should not be scoped to my [KnoxIDF TokenResource.](https://github.com/smolnar82/knox/blob/knox_idf_smolnar/gateway-service-knoxidf/src/main/java/org/apache/knox/gateway/service/knoxidf/TokenResource.java#L387C1-L393C6)

lmccay: This touches on something that I've been discussing with @pzampino offline about. It is really not clear why we hide these params in the first place. I know that once you read them, the input stream has been opened and it can't be done again, so it breaks certain usecases. In cases where we fully know that we aren't passing it on to someone else to use, this is okay. These OAuth flows are an example of when its okay. So, I have 2 reactions to your question:

1. A convenience method to get the wrapped param would be fine but should make it clear that it is unwrapping with the name - getWrappedRequestParam. The method you have above would maybe return a query param or a hidden request body param. That would potentially break usecases without knowing it. Cracking open the request body should be an explicit action unless/until we remove the hiding of the params.
2. The fact that you've asked AND that you had to do the same elsewhere may be another smell of this implementation and maybe we shouldn't hide them at all.

Bottomline: if we keep them hidden then the interface should make it very clear what you are opting into. Additionally, we should check whether we really need to hide them. The unwrapping code will still work if we stop hiding them anyway.

Thoughts?

smolnar82: As indicated offline, the reason for the wrapping was added in #719 by @zeroflag a couple years ago.
I do agree with the proposal of having clear method names, and I still think that having a utility method is a good idea.

smolnar82: We may want to re-use the `UrlEncodedFormRequest.isUrlEncodedForm(servletRequest)` to determine if unwrapping is needed at all.

lmccay: I don't believe we want to unhide them since that will break the proxied interactions that need to interrogate the query string  for things like doAs. I also don't think we want to add a method that tries the params and then falls back. We could add a method like getUnwrappedRequestParam but the value there is questionable.

I think we should move forward the way it is for now and we can always go and refactor it after.

smolnar82: Yes, I agree.

smolnar82 (approved):

PR #1229, opened 2026-05-12
KNOX-3321 - KnoxToken Support for RFC 8693 Token Exchange act Claim

lmccay:
[KNOX-3321](https://issues.apache.org/jira/browse/KNOX-3321) - KnoxToken Support for RFC 8693 Token Exchange act Claim

To support use cases that need insight into access of a resource on behalf of user other than the token holder, we need to add the 'act' chain claim. The ability track a chain of interactions being done by services, pipelines or agents will allow for better audit detail and authorization decision making.

Based on the existence of the ImpersonatedPrincipal in the Java Subject, KnoxToken API will add the 'act' claim with a nested 'sub' to represent the entity acting on behalf of the primary 'sub' of the token.

This requires adding additional methods to our JWTTokeService for both adding the 'act' claim itself but also for extracting it from a parsed token.

## How was this patch tested?

All existing unit and integration tests were built and rain locally and new test cases were added and also run.


pzampino (approved):
LGTM

PR #1236, opened 2026-05-20
KNOX-3328: Implement recursive group resolution for LDAP proxy

smolnar82:
[KNOX-3328](https://issues.apache.org/jira/browse/KNOX-3328) - Support recursive group resolution in LDAP Proxy Service

## What changes were proposed in this pull request?
This PR introduces recursive group resolution to the LdapProxyBackend and integrates it into the Knox LDAP Service. These changes allow Knox to resolve transitive group memberships when proxying to an external LDAP server (e.g., Active Directory or OpenLDAP).

Key changes include:
   * **Recursive Logic in LdapProxyBackend**: Implemented a Breadth-First Search (BFS) algorithm to traverse group hierarchies. 
   * **Cycle Detection**: Uses a visited set of Distinguished Names (DNs) to track processed groups, preventing infinite loops in case of circular references (e.g., Group A → Group B → Group A).
   * **Configurable Depth Limiting**: Added gateway.ldap.recursive.group.resolution.max.depth (default: 3) to control the recursion depth and protect against excessive LDAP round-trips.
   * **Enriched User Entries**: `LdapProxyBackend.getUser` now populates the `memberOf` attribute with all recursively discovered groups, ensuring that the returned "proxy entry" contains a complete membership profile.
   * **HadoopGroupProvider Integration**: Updated `HadoopGroupProviderFilter` to leverage the KnoxLDAPService for group lookups when LDAP is enabled. This ensures that Hadoop-style group mapping can benefit from the recursive resolution logic.
   * **New KnoxCLI Test Command**: Added `ldap-user-groups-test` to the KnoxCLI to allow administrators to verify LDAP group resolution directly from the command line:
   ```knoxcli.sh ldap-user-groups-test --u <username>```
   * **Bug Fix**: Fixed an issue in `SimplePrincipalMapper` where group mappings were not populated if principal mappings were empty.

New Configuration Properties:
   * `gateway.ldap.recursive.group.resolution`: Boolean to enable/disable the feature (default: false).
   * `gateway.ldap.recursive.group.resolution.max.depth`: Integer to control recursion depth (default: 3).

## How was this patch tested?
The patch was verified using a combination of unit tests with embedded ApacheDS and integration tests.
Unit Tests:
   * Updated `LdapProxyBackendTest` with new test cases:
       * `testGetRecursiveUserGroupsDepth2`: Verifies 2-level nesting.
       * `testGetRecursiveUserGroupsDepth4`: Verifies deeper nesting and depth limiting.
[...]

PR #1227, opened 2026-05-11
KNOX-3318: document LDAP service

smolnar82:
[KNOX-3318](https://issues.apache.org/jira/browse/KNOX-3318) - Add Knox LDAP Server documentation to User Guide

 ## What changes were proposed in this pull request?

This pull request adds comprehensive documentation for the Knox LDAP Server feature (introduced in KNOX-3247 and improved in KNOX-3277) to the official Knox User Guide.

  Changes include:
   - Created knox-site/docs/service_ldap_server.md which provides:
       - An architectural overview of the embedded ApacheDS-based LDAP service.
       - Detailed configuration instructions for both file and proxy backends.
       - Specific guidance for Active Directory (AD) integration, including memberOf and sAMAccountName support.
       - Troubleshooting tips regarding logs and lock files.
   - Updated knox-site/mkdocs.yml to include the "LDAP Server" page in the Client/User Guide navigation.
   - Updated knox-site/docs/book_service-details.md to include the LDAP Server in the master list of supported services.

  ## How was this patch tested?

   - Manually verified the Markdown rendering and formatting of the new documentation.
   - Validated all internal and external (JIRA) links.
   - Cross-referenced all configuration property names (e.g., `gateway.ldap.backend.proxy.*`) with the implementation in `GatewayConfigImpl.java` and `LdapProxyBackend.java` to ensure accuracy.

   ## Integration Tests
  N/A - This is a documentation-only change. No functional code was modified.

   ## UI changes
  N/A

PR #1240, opened 2026-05-26
KNOX-3330: Refactor LDAP Proxy configuration to support multiple backends

handavid:
[KNOX-3330](https://issues.apache.org/jira/browse/KNOX-3330) - Refactor Knox LDAP Proxy configuration and implementation to allow multiple backends to be simultaneously configured

Gateway server configurations are updated to use 'gateway.ldap.interceptor.*' instead of 'gateway.ldap.backend.*' to allow specifying multiple types of interceptors as well as multiple backends to the LDAP proxy.

BackendFactory has been modified to use the java ServiceLoader to load a factory for a backend class instead of a backend instance directly. This allows multiple backends of the same class to be configured. InterceptorFactory has been implemented following the same pattern.

GroupLookupInterceptor is renamed to UserSearchInterceptor to more accurately describe what it does. Multiple UserSearchInterceptors can be configured with each forwarding the search to its backend and appending the results.

A DuplicateUserFilteringInterceptor has been implemented that will filter out search Entries with the same UID that are returned from different backends.

## How was this patch tested?

Unit tests were updated.
- KnoxLDAPServerManagerTest.java modified to configure interceptors instead of backends
- KnoxLDAPServerManagerTest.java modified to configure multiple backends simultaneously

Changes were manually tested against the test ldap server and an AD that I have access to.  
The following configuration was added to the gateway-site.xml

```
    <!-- LDAP Proxy Service Configuration -->
    <property>
        <name>gateway.ldap.enabled</name>
        <value>true</value>
        <description>Enable the embedded LDAP service for user and group lookups. Set to true to enable.</description>
    </property>
    <property>
        <name>gateway.ldap.port</name>
        <value>3890</value>
        <description>Port for the LDAP service to listen on. Default is 3890.</description>
    </property>
    <property>
        <name>gateway.ldap.base.dn</name>
        <value>dc=proxy,dc=com</value>
        <description>Base DN for LDAP entries in the proxy server. Default is dc=proxy,dc=com.</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.names</name>
        <value>localldap,testad,duplicatefilter</value>
        <description>Interceptor names for LDAP service.</description>
    </property>

    <!-- Local LDAP Server -->
    <property>
        <name>gateway.ldap.interceptor.localldap.interceptorType</name>
        <value>backend</value>
        <description>Type of interceptor. Currently supported: backend, duplicateuserfilter</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.localldap.backendType</name>
        <value>ldap</value>
        <description>Type of backend. Currently supported: file, ldap. Future: jdbc, knox.</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.localldap.url</name>
        <description>LDAP server URL for proxy backend</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.localldap.remoteBaseDn</name>
        <value>dc=hadoop,dc=apache,dc=org</value>
        <description>Base DN of the remote LDAP server</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.localldap.systemUsername</name>
        <value>uid=guest,ou=people,dc=hadoop,dc=apache,dc=org</value>
        <description>LDAP bind DN for proxy backend authentication</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.localldap.systemPassword</name>
        <value>guest-password</value>
        <description>LDAP bind password for proxy backend authentication</description>
    </property>
    
    <!-- Test AD -->
    <property>
        <name>gateway.ldap.interceptor.testad.interceptorType</name>
        <value>backend</value>
        <description>Type of interceptor. Currently supported: backend, duplicateuserfilter</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.backendType</name>
        <value>ldap</value>
        <description>Type of backend. Currently supported: file, ldap. Future: jdbc, knox.</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.url</name>
        <value>ldap://test-ad.example.com:389</value>
        <description>LDAP server URL for proxy backend</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.remoteBaseDn</name>
        <value>dc=test-ad,dc=example,dc=com</value>
        <description>Base DN of the remote LDAP server</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.systemUsername</name>
        <value>bind user to AD</value>
        <description>LDAP bind DN for proxy backend authentication</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.systemPassword</name>
        <value>password to AD</value>
        <description>LDAP bind password for proxy backend authentication</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.userIdentifierAttribute</name>
        <value>sAMAccountName</value>
        <description>Attribute used for identifying users</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.userSearchBase</name>
        <value>cn=users,dc=test-ad,dc=example,dc=com</value>
        <description>Search base for users</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.groupSearchBase</name>
        <value>ou=groups,dc=test-ad,dc=example,dc=com</value>
        <description>Search base for groups</description>
    </property>
    <property>
        <name>gateway.ldap.interceptor.testad.useMemberOf</name>
        <value>true</value>
        <description>Whether to use the memberOf attribute for efficiency when retrieving group memberships</description>
    </property>

    <!-- Duplicate Filter Interceptor -->
    <property>
        <name>gateway.ldap.interceptor.duplicatefilter.interceptorType</name>
        <value>duplicateuserfilter</value>
        <description>Type of interceptor. Currently supported: backend, duplicateuserfilter</description>
    </property>

    <!-- END LDAP Proxy Service Configuration -->
```

## Integration Tests
No integration test changes. PR can be updated after https://github.com/apache/knox/pull/1236 is merged

## UI changes
no UI changes

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
@@ -1754,8 +1754,8 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   }
 
   @Override
-  public String getLDAPBackendType() {
-    return get(LDAP_BACKEND_TYPE, "file");
+  public List<String> getLDAPInterceptorNames() {
+    return splitConfigValueToList(LDAP_INTERCEPTOR_NAMES);
   }
 
   @Override
@@ -1782,9 +1782,9 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   }
 
   @Override
-  public Map<String, String> getLDAPBackendConfig(String backendType) {
+  public Map<String, String> getLDAPInterceptorConfig(String interceptorName) {
     Map<String, String> config = new HashMap<>();
-    String prefix = "gateway.ldap.backend." + backendType + ".";
+    String prefix = "gateway.ldap.interceptor." + interceptorName + ".";
 
     for (String key : getPropertyNames()) {
       if (key != null && key.startsWith(prefix)) {
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
@@ -17,12 +17,22 @@
  */
 package org.apache.knox.gateway.services.ldap;
 
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.directory.api.ldap.model.cursor.Cursor;
+import org.apache.directory.api.ldap.model.entry.Attribute;
 import org.apache.directory.api.ldap.model.entry.DefaultEntry;
 import org.apache.directory.api.ldap.model.entry.Entry;
+import org.apache.directory.api.ldap.model.entry.Value;
+import org.apache.directory.api.ldap.model.exception.LdapException;
+import org.apache.directory.api.ldap.model.message.SearchRequest;
+import org.apache.directory.api.ldap.model.message.SearchRequestImpl;
+import org.apache.directory.api.ldap.model.message.SearchScope;
 import org.apache.directory.api.ldap.model.name.Dn;
 import org.apache.directory.api.ldap.model.schema.SchemaManager;
 import org.apache.directory.server.core.DefaultDirectoryService;
 import org.apache.directory.server.core.api.DirectoryService;
+import org.apache.directory.server.core.api.DnFactory;
 import org.apache.directory.server.core.api.InstanceLayout;
 import org.apache.directory.server.core.api.interceptor.Interceptor;
 import org.apache.directory.server.core.api.schema.SchemaPartition;
@@ -32,13 +42,16 @@ import org.apache.directory.server.ldap.LdapServer;
 import org.apache.directory.server.protocol.shared.transport.TcpTransport;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
-import org.apache.knox.gateway.services.ldap.backend.BackendFactory;
-import org.apache.knox.gateway.services.ldap.backend.LdapBackend;
+import org.apache.knox.gateway.services.ldap.interceptor.InterceptorFactory;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Manages the ApacheDS LDAP server instance with pluggable backends
@@ -46,13 +59,16 @@ import java.util.Map;
 public class KnoxLDAPServerManager {
     private static final LdapMessages LOG = MessagesFactory.get(LdapMessages.class);
 
-    private DirectoryService directoryService;
+    @VisibleForTesting
+    DirectoryService directoryService;
     private LdapServer ldapServer;
-    private LdapBackend backend;
+    private GatewayConfig gatewayConfig;
+    private List<Interceptor> interceptors;
     private File workDir;
     private int port;
     private String baseDn;
-    private String remoteBaseDn;
+    // Collection of DNs for the proxied backend LDAP servers
+    private Set<String> baseDns;
 
     /**
      * Initialize the LDAP server with the given configuration
@@ -60,6 +76,8 @@ public class KnoxLDAPServerManager {
      * @param config Gateway configuration
      */
     public void initialize(GatewayConfig config) throws Exception {
+        this.gatewayConfig = config;
+
         // Prepare work directory for LDAP data
         File gatewayDataDir = new File(config.getGatewayDataDir());
         this.workDir = new File(gatewayDataDir, "ldap-server");
@@ -67,27 +85,8 @@ public class KnoxLDAPServerManager {
         // Get configuration
         this.port = config.getLDAPPort();
         this.baseDn = config.getLDAPBaseDN();
-        String backendType = config.getLDAPBackendType();
 
-        // Get backend-specific configuration using prefixed properties
-        Map<String, String> backendConfig = config.getLDAPBackendConfig(backendType);
-
-        // Add common configuration
-        backendConfig.put("baseDn", baseDn);
-
-        // Add legacy dataFile property for backwards compatibility with file backend
-        if ("file".equalsIgnoreCase(backendType) && !backendConfig.containsKey("dataFile")) {
-            backendConfig.put("dataFile", config.getLDAPBackendDataFile());
-        }
-
-        backendConfig.put("recursiveGroupResolution", String.valueOf(config.isLDAPRecursiveGroupResolutionEnabled()));
-        backendConfig.put("recursiveGroupResolutionMaxDepth", String.valueOf(config.getLDAPRecursiveGroupResolutionMaxDepth()));
-
-        // For proxy backends, extract remoteBaseDn if present
-        this.remoteBaseDn = backendConfig.get("remoteBaseDn");
-
-        // Initialize backend
-        backend = BackendFactory.createBackend(backendType, backendConfig);
+        this.interceptors = createInterceptors(config);
 
         // Clean up previous run if it didn't shut down cleanly
         File lockFile = new File(workDir, "run/instance.lock");
@@ -97,6 +96,34 @@ public class KnoxLDAPServerManager {
         }
 
         workDir.mkdirs();
+    }
+
+    private List<Interceptor> createInterceptors(GatewayConfig config) throws Exception {
+        List<String> interceptorNames = config.getLDAPInterceptorNames();
+        List<Interceptor> interceptors = new ArrayList<>(interceptorNames.size());
+        for (String interceptorName : interceptorNames) {
+            // Get backend-specific configuration using prefixed properties
+            Map<String, String> interceptorConfig = config.getLDAPInterceptorConfig(interceptorName);
+
+            // Add common configuration
+            interceptorConfig.put("baseDn", baseDn);
+
+            // Add common LDAP Proxy configurations to backends
+            String interceptorType = interceptorConfig.get("interceptorType");
+            String backendType = interceptorConfig.get("backendType");
+            if ("backend".equalsIgnoreCase(interceptorType)) {
+                interceptorConfig.put("recursiveGroupResolution", String.valueOf(config.isLDAPRecursiveGroupResolutionEnabled()));
+                interceptorConfig.put("recursiveGroupResolutionMaxDepth", String.valueOf(config.getLDAPRecursiveGroupResolutionMaxDepth()));
+                if ("file".equalsIgnoreCase(backendType) &&
+                        !interceptorConfig.containsKey("dataFile")) {
+                    // Add legacy dataFile property for backwards compatibility with file backend
+                    interceptorConfig.put("dataFile", config.getLDAPBackendDataFile());
+                }
+            }
+
+            interceptors.add(InterceptorFactory.createInterceptor(interceptorName, interceptorConfig));
+        }
+        return interceptors;
     }
 
     /**
@@ -134,17 +161,11 @@ public class KnoxLDAPServerManager {
         partition.setPartitionPath(new File(workDir, "proxy").toURI());
         directoryService.addPartition(partition);
 
-        // Create partition for remote base DN if different from proxy base DN
-        // This allows backend entries with remote DNs to be returned in search results
-        if (remoteBaseDn != null && !remoteBaseDn.equals(baseDn)) {
-            JdbmPartition remotePartition = new JdbmPartition(schemaManager, directoryService.getDnFactory());
-            remotePartition.setId("remote");
-            remotePartition.setSuffixDn(new Dn(schemaManager, remoteBaseDn));
-            remotePartition.setPartitionPath(new File(workDir, "remote").toURI());
-            directoryService.addPartition(remotePartition);
-        }
+        baseDns = new HashSet<>();
+        baseDns.add(baseDn);
+        addRemotePartitions();
 
-        addGroupLookupInterceptor();
+        addInterceptors();
 
         // Allow anonymous access
         directoryService.setAllowAnonymousAccess(true);
@@ -152,8 +173,8 @@ public class KnoxLDAPServerManager {
         // Start the service
         directoryService.startup();
 
-        // Add base entries to the partition
-        createBaseEntries(schemaManager);
+        // Add base entries to the partitions
+        createBaseEntries(baseDns, schemaManager);
 
         // Create LDAP server on configured port
         ldapServer = new LdapServer();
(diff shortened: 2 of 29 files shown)


handavid:
@smolnar82 @lmccay

handavid:
rebased and fixed conflicts

gateway-server/data/security/keystores/__gateway-credentials.jceks
smolnar82, line 1: Please remove this file from the commit (and the other .jceks files too).

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -1748,8 +1749,16 @@ public String getLDAPBaseDN() {
...
 
   @Override
-  public String getLDAPBackendType() {
-    return get(LDAP_BACKEND_TYPE, "file");
+  public List<String> getLDAPInterceptorNames() {
+    String names = get(LDAP_INTERCEPTOR_NAMES, "");
+    String[] namesArray = names.split(",");
+    List<String> namesList = new ArrayList<>();
+    for (String name : namesArray) {
+      if (!Strings.isNullOrEmpty(name)) {
```
smolnar82: We use `org.apache.commons.lang3.StringUtils#isNotBlank` in the project for the same purpose.
This is already imported -> the new import will be eliminated too.
Even better, we already have a method to get comma separated config values as list: `org.apache.knox.gateway.config.impl.GatewayConfigImpl#splitConfigValueToList`

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/DuplicateUserFilteringInterceptor.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.directory.server.core.api.interceptor.BaseInterceptor;
+import org.apache.directory.server.core.api.interceptor.context.SearchOperationContext;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+public class DuplicateUserFilteringInterceptor extends BaseInterceptor {
```
smolnar82, line 38: Please add unit test coverage for this class.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/DuplicateUserFilteringInterceptor.java
```diff
@@ -0,0 +1,88 @@
...
+        // First execute the interceptor chain to get the results
+        EntryFilteringCursor originalResults;
+        originalResults = next(ctx);
+
+        List<Entry> originalEntries = new ArrayList<>();
+        try {
+            while (originalResults.next()) {
+                originalEntries.add(originalResults.get());
+            }
+            originalResults.close();
```
smolnar82: Please ensure that `originalResults.close()` is always invoked to avoid cursor leaks (using try-with-resources is a good option here).
E.g.
```
try (EntryFilteringCursor originalResults = next(ctx)) {
   ...
}
```

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/DuplicateUserFilteringInterceptorFactory.java
```diff
@@ -0,0 +1,35 @@
...
+package org.apache.knox.gateway.services.ldap.interceptor;
+
+import org.apache.directory.server.core.api.interceptor.Interceptor;
+
+import java.util.Map;
+
+public class DuplicateUserFilteringInterceptorFactory implements KnoxLdapInterceptorFactory {
+    @Override
+    public Interceptor create(String name, Map<String, String> config) {
+        // NOTE: no further configuration expected for this Interceptor
```
smolnar82: nit: I think this JavaDoc comment is redundant and can be obsolete quickly.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/DuplicateUserFilteringInterceptorFactory.java
```diff
@@ -0,0 +1,35 @@
...
+        return new DuplicateUserFilteringInterceptor(name);
+    }
+
+    @Override
+    public String getType() {
+        return "duplicateuserfilter";
```
smolnar82: nit: using a constant would be better

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/UserSearchInterceptor.java
```diff
@@ -98,47 +107,55 @@ public EntryFilteringCursor search(SearchOperationContext ctx) throws LdapExcept
...
-                        if (backendEntry != null) {
-                            // Return backend result directly without caching
-                            entries.add(backendEntry);
-                            LOG.ldapUserEntry(backendEntry.toString());
-                        } else {
-                            LOG.ldapUserNull(username);
+                        // if no results, perform single-user search
+                        if (entries.isEmpty()) {
+                            // Specific user lookup
+                            LOG.ldapUserLoaded(username);
```
smolnar82, line 126: This log line should be under the actual `getUser(username);` line

smolnar82 (changes requested):
I'll keep the review later today.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/UserSearchInterceptor.java
```diff
@@ -98,47 +107,55 @@ public EntryFilteringCursor search(SearchOperationContext ctx) throws LdapExcept
...
+            return new EntryFilteringCursorImpl(new ListCursor<>(entries), ctx, schemaManager);
         }
 
         return originalResults;
     }
 
+    private List<Entry> searchUsers(String userSearchString) throws Exception {
+        List<Entry> entries = new ArrayList<>();
+        entries.addAll(backend.searchUsers(userSearchString, schemaManager));
+        return entries;
```
smolnar82: Why not simply: `return backend.searchUsers(userSearchString, schemaManager)`?
In fact, if this is a simple inline backend invocation, why we have this private method at all? No any particular check, logic, etc...in here.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/UserSearchInterceptor.java
```diff
@@ -98,47 +107,55 @@ public EntryFilteringCursor search(SearchOperationContext ctx) throws LdapExcept
...
+    }
+
+    private Entry getUser(String username) throws Exception {
+        return backend.getUser(username, schemaManager);
```
smolnar82: The same as above: I don't think this private method is necessary.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
```diff
@@ -32,27 +34,34 @@
...
  */
 public class KnoxLDAPServerManager {
     private static final LdapMessages LOG = MessagesFactory.get(LdapMessages.class);
 
-    private DirectoryService directoryService;
+    @VisibleForTesting
+    DirectoryService directoryService;
     private LdapServer ldapServer;
-    private LdapBackend backend;
+    private List<Interceptor> interceptors = new ArrayList<>();
```
smolnar82: This is redundant, as you initialize interceptors anyway (i.e. not simply add entries into it).
See new line 80: `this.interceptors = createInterceptors(config)`

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/UserSearchInterceptorFactory.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class UserSearchInterceptorFactory implements KnoxLdapInterceptorFactory {
+    @Override
+    public Interceptor create(String name, Map<String, String> config) throws Exception {
+        return new UserSearchInterceptor(name, config);
+    }
+
+    @Override
+    public String getType() {
+        return "backend";
```
smolnar82: nit: Should be a constant (we could reuse it in KnoxLdapManager, see my comment about not using the `instanceof` operator below).

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
```diff
@@ -174,11 +183,30 @@ private void addGroupLookupInterceptor() {
...
 
-        final GroupLookupInterceptor interceptor = new GroupLookupInterceptor(directoryService, backend);
-        if (authIdx != -1) {
-            interceptors.add(authIdx, interceptor);
-        } else {
-            interceptors.add(interceptor);
+        // Add our configured interceptors
+        SchemaManager schemaManager = directoryService.getSchemaManager();
+        for (Interceptor interceptor : interceptors) {
+            if (interceptor instanceof UserSearchInterceptor) {
```
smolnar82: I personally am not a big fan of the `instanceof` operator, so this feel free to ignore this comment :) 

It might be an option to store the interceptors above in a `Map<String, Interceptor>`, where the key is the interceptor type. If that was the case, we could simply check if the Map.Entry.key.equals("backend") (this is where my comment about creating the `backend` interceptor type constant would be useful).

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
```diff
@@ -174,11 +183,30 @@ private void addGroupLookupInterceptor() {
...
+                if (!baseDns.contains(remoteBaseDn)) {
+                    //create partition
+                    String id = backend.getName().replaceAll("\\s+", "");
+                    JdbmPartition remotePartition = new JdbmPartition(schemaManager, directoryService.getDnFactory());
+                    remotePartition.setId(id);
+                    remotePartition.setSuffixDn(new Dn(schemaManager, remoteBaseDn));
+                    remotePartition.setPartitionPath(new File(workDir, id).toURI());
+                    directoryService.addPartition(remotePartition);
+                    baseDns.add(remoteBaseDn);
+                }
```
smolnar82, line 208: nit: a new private method (`addRemotePartition`, for instance) would increase the readability of this part of the code.

gateway-server/src/main/resources/conf/gateway-site.xml
```diff
@@ -68,61 +68,76 @@ limitations under the License.
...
-    <!-- Uncomment and configure based on backend type specified in gateway.ldap.backend.type -->
+    <!-- Uncomment and configure based on interceptor names specified in gateway.ldap.interceptor.names -->
 
-    <!-- File backend configuration (gateway.ldap.backend.type=file) -->
+    <!-- File backend configuration (gateway.ldap.interceptor.<interceptorName>.backendType=file) -->
     <!--
     <property>
-        <name>gateway.ldap.backend.file.dataFile</name>
+        <name>gateway.ldap.interceptor.ldapfile.backendType</name>
+        <value>ldap</value>
```
smolnar82: Should be `file`.

gateway-server/src/main/resources/conf/gateway-site.xml
```diff
@@ -68,61 +68,76 @@ limitations under the License.
...
         <value>${GATEWAY_DATA_HOME}/ldap-users.json</value>
         <description>Path to JSON file containing user and group data</description>
     </property>
     -->
 
-    <!-- LDAP proxy backend configuration (gateway.ldap.backend.type=ldap) -->
+    <!-- LDAP proxy backend configuration (gateway.ldap.interceptor.<interceptorName>.backendType=ldap) -->
     <!-- This backend proxies to an external LDAP server (e.g., demo LDAP) -->
     <!--
     Example 1: Using Knox demo LDAP server (default port 33389)
```
smolnar82, line 134: I'm not sure I like the idea of putting all these samples in the gateway-site.xml here.
We should rather create the new section in our user guide (see the `knox-site` module), and add all these sample configs there.

smolnar82 (changes requested):
Another batch of comments.
I'll review the tests tomorrow.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/DuplicateUserFilteringInterceptor.java
handavid: used try-with-resources. also verified close in new unit tests

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/UserSearchInterceptor.java
handavid: removed. this was a result of refactoring some intermediate changes.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
handavid: moved to get the remoteBaseDns from GatewayConfig directly since config is available to the KnoxLDAPServerManager

gateway-server/src/main/resources/conf/gateway-site.xml
handavid: I'm fixing the existing configs here for my config changes. should we remove the whole ldap proxy-related block?
I was waiting for https://github.com/apache/knox/pull/1227 to be merged before updating the documentation

handavid: these values need to be correct for the KnoxCLITest to work

handavid:
force push was due to rebase on master to work on resolution of conflicts.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
handavid: done. I pulled this out to get the values from the configs rather than the Interceptors.

handavid:
@smolnar82 
force pushed after rebase to pick up documentation changes.
I've updated the service_ldap_server.md to reflect the architectural and configuration changes.
I've also fixed the KnoxCLI to start the KnoxLDAPService. This is needed to start the directoryService. I tested this by running:
```
$ ./knoxcli.sh ldap-user-groups-test --u sam
Querying KnoxLDAPService for groups of user: sam
sam is a member of: scientist, analyst
```

gateway-server/src/main/resources/conf/gateway-site.xml
smolnar82: @handavid - I merged #1227, we can add the sample configs there.
IMO, we only need to keep the one above (`ldapproxy`) which is "active" (i.e. not commented out), the rest should be added in the documentation below.
We can do that in a follow-up JIRA though.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/DuplicateUserFilteringInterceptor.java
```diff
@@ -0,0 +1,87 @@
...
+    public EntryFilteringCursor search(SearchOperationContext ctx) throws LdapException {
+        // First execute the interceptor chain to get the results
+        List<Entry> filteredEntries = List.of();
+        try (EntryFilteringCursor originalResults = next(ctx)) {
+            List<Entry> originalEntries = new ArrayList<>();
+            try {
+                while (originalResults.next()) {
+                    originalEntries.add(originalResults.get());
+                }
+                originalResults.close();
```
smolnar82, line 54: Since `originalResults` is now opened using `try-with-resources`, we don't need this call.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/InterceptorFactory.java
```diff
@@ -0,0 +1,59 @@
...
+
+        // Use ServiceLoader to discover all available interceptors (built-in and external plugins)
+        // Indirect instantiation through a factory is used to allow configuration of multiple instances
+        // of the same class of interceptor. e.g., if multiple backends are configured
+        ServiceLoader<KnoxLdapInterceptorFactory> loader = ServiceLoader.load(KnoxLdapInterceptorFactory.class);
+        for (KnoxLdapInterceptorFactory interceptorFactory : loader) {
+            if (interceptorFactory.getType().equalsIgnoreCase(interceptorType)) {
+                LOG.ldapInterceptorCreating(interceptorType, "ServiceLoader");
+                Interceptor interceptor = interceptorFactory.create(interceptorName, config);
+                return interceptor;
```
smolnar82, line 51: nit: you can return directly

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
```diff
@@ -99,6 +98,34 @@ public void initialize(GatewayConfig config) throws Exception {
...
+                if ("file".equalsIgnoreCase(backendType) &&
+                        !interceptorConfig.containsKey("dataFile")) {
+                    // Add legacy dataFile property for backwards compatibility with file backend
+                    interceptorConfig.put("dataFile", config.getLDAPBackendDataFile());
+                }
+            }
+
+            interceptors.add(InterceptorFactory.createInterceptor(interceptorName, interceptorConfig));
+        }
+        return interceptors;
```
smolnar82, line 126: nit: since `interceptors` is a class member and this method is `private` (i.e. not used in tests), this might be `void` and set `interceptors directly. It really is just a nitpick comment tough.

smolnar82 (approved):
This looks good now. We can merge it and you'll submit a new PR to address the blow comment. The only ones that make me worry are the extra `close()` call on the cursor and the inactive sample configs listed in the `gateway-site.xml` we release.
The 2 other comments are just nitpicks.

smolnar82:
I also asked Gemini to focus on the test coverage. Here is the output. I removed some of the items as they are irrelevant for now (I know you'll be working on the "generalization" of the attributes to support not only user/group related searches):
```
While the PR introduces good unit tests for the new components, several critical areas remain under-tested or completely uncovered.

  1. Missing Test Scenarios
   * Partition ID Collisions: The logic interceptorName.replaceAll("\\s+", "") is prone to collisions. There are no tests ensuring that providing names like "My LDAP" and "MyLDAP" won't cause the server to fail during initialization.
   * Interceptor Ordering: The PR relies on the order of names in `gateway.ldap.interceptor.names`. There are no tests verifying that putting `duplicatefilter` before a backend interceptor correctly results in no deduplication of backend results (confirming "sink" behavior).

  2. Error Handling Resilience
   * Partial Failures: If multiple backends are configured and one is unreachable, the system's behavior is untested. Does the entire search fail, or does it gracefully continue to the next interceptor?
```

PR #1244, opened 2026-05-31
KNOX-3333 - Update Letsencrypt staging certs

moresandeep:

[KNOX-3333](https://issues.apache.org/jira/browse/KNOX-3333) - Update Letsencrypt staging certs

Letsencrypt staging certs were updated with few more, adding all. 

## How was this patch tested?

### DIFF
--- a/gateway-docker/src/main/resources/docker/Dockerfile
+++ b/gateway-docker/src/main/resources/docker/Dockerfile
@@ -56,7 +56,8 @@ ARG ENTRYPOINT
 COPY --chown=8000:0 ${ENTRYPOINT} /home/knox/knox/entrypoint.sh
 RUN chmod +x /home/knox/knox/entrypoint.sh
 
-# Add the Amazon Root CA and Let's Encrypt production + staging root certificates (best-effort)
+# Add the Amazon Root CA and Let's Encrypt production root certificates (best-effort).
+# Staging roots are downloaded at runtime when IMPORT_DEFAULT_STAGING_CERTS=true (default).
 RUN mkdir /home/knox/cacrts && \
     curl -sSLo /home/knox/cacrts/AmazonRootCA1.cer  https://www.amazontrust.com/repository/AmazonRootCA1.cer  || true && \
     curl -sSLo /home/knox/cacrts/AmazonRootCA2.cer  https://www.amazontrust.com/repository/AmazonRootCA2.cer  || true && \
@@ -64,13 +65,6 @@ RUN mkdir /home/knox/cacrts && \
     curl -sSLo /home/knox/cacrts/AmazonRootCA4.cer  https://www.amazontrust.com/repository/AmazonRootCA4.cer  || true && \
     curl -sSLo /home/knox/cacrts/isrgrootx1.pem     https://letsencrypt.org/certs/isrgrootx1.pem               || true && \
     curl -sSLo /home/knox/cacrts/isrg-root-x2.pem   https://letsencrypt.org/certs/isrg-root-x2.pem             || true && \
-    curl -sSLo /home/knox/cacrts/letsencrypt-stg-root-x1.pem               https://letsencrypt.org/certs/staging/letsencrypt-stg-root-x1.pem               || true && \
-    curl -sSLo /home/knox/cacrts/letsencrypt-stg-root-x2.pem               https://letsencrypt.org/certs/staging/letsencrypt-stg-root-x2.pem               || true && \
-    curl -sSLo /home/knox/cacrts/letsencrypt-stg-root-x2-signed-by-x1.pem  https://letsencrypt.org/certs/staging/letsencrypt-stg-root-x2-signed-by-x1.pem  || true && \
-    curl -sSLo /home/knox/cacrts/letsencrypt-stg-root-ye.pem               https://letsencrypt.org/certs/staging/gen-y/root-ye.pem                         || true && \
-    curl -sSLo /home/knox/cacrts/letsencrypt-stg-root-ye-by-x2.pem        https://letsencrypt.org/certs/staging/gen-y/root-ye-by-x2.pem                    || true && \
-    curl -sSLo /home/knox/cacrts/letsencrypt-stg-root-yr.pem               https://letsencrypt.org/certs/staging/gen-y/root-yr.pem                         || true && \
-    curl -sSLo /home/knox/cacrts/letsencrypt-stg-root-yr-by-x1.pem        https://letsencrypt.org/certs/staging/gen-y/root-yr-by-x1.pem                    || true && \
     chown -R 8000:0 /home/knox/cacrts
 
 WORKDIR /home/knox/knox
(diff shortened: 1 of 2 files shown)


smolnar82:
I don't think this is a serious security vulnerability as the added certificates are public CA roots from Let's Encrypt's official staging hierarchy, not arbitrary certificates.

However, I've a few questions:

- What use case requires trusting staging roots?
  - Is there a real customer scenario?
  - Is this only for automated testing?
- Can this be made optional?
  - For example via build argument or environment variable.
  - Then test users can enable it while production users keep a smaller trust set.
- Are all these roots necessary?
  - The patch adds multiple generations (X1, X2, YE, YR, cross-signed variants).
  - It may be worth confirming that all are actually needed.

Adding staging roots increases the set of trusted certificate authorities and allows Knox to trust certificates issued by Let's Encrypt's testing infrastructure. Can we clarify the use case and whether this trust should be enabled only for testing environments rather than all Docker deployments (see my question above about making them optional)?

moresandeep:
> I don't think this is a serious security vulnerability as the added certificates are public CA roots from Let's Encrypt's official staging hierarchy, not arbitrary certificates.
> 
> However, I've a few questions:
> 
> * What use case requires trusting staging roots?
>   
>   * Is there a real customer scenario?
>   * Is this only for automated testing?
> * Can this be made optional?
>   
>   * For example via build argument or environment variable.
>   * Then test users can enable it while production users keep a smaller trust set.
> * Are all these roots necessary?
>   
>   * The patch adds multiple generations (X1, X2, YE, YR, cross-signed variants).
>   * It may be worth confirming that all are actually needed.
> 
> Adding staging roots increases the set of trusted certificate authorities and allows Knox to trust certificates issued by Let's Encrypt's testing infrastructure. Can we clarify the use case and whether this trust should be enabled only for testing environments rather than all Docker deployments (see my question above about making them optional)?

Thanks @smolnar82 
1. Is there a real customer scenario? -> Yes, in enterprises there are staging environments which do not have prod certs. This change will help users test Knox in staging without a need to add staging certs into keystore (which they do not have to do in prod) mimicking prod env. 
2. Can this be made optional? -> This is interesting, the idea is to make sure the prod and staging are identical in all respects so making it optional means they will have to have a flag that they use to turn this ON in stage and OFF in prod so there will be a change in config although not a big one.
3. They are not test related changes these are runtime/config related changes. These changes do not affect tests.
4. Are all these roots necessary? -> Unfortunately yes. we do not know and cannot dictate what root CA users are going to use for staging.
5. It may be worth confirming that all are actually needed. -> Yes, this came up in an internal investigation, we were using a different gen and got switched to other.

moresandeep:
@smolnar82 I put the staging certs behind flag (by default `false`) I also added some truststore hardening.

moresandeep:
@sneethiraj when you have time can you do a quick review, thanks!

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -30,6 +30,9 @@
 # - DATABASE_CONNECTION_PASSWORD - (optional) gateway database password
 # - DATABASE_CONNECTION_TRUSTSTORE_PASSWORD - (optional) gateway database ssl truststore password
 # - CUSTOM_CERT - (optional) the location of a file containing the custom certs
+# - IMPORT_DEFAULT_STAGING_CERTS - (optional) when 'true' (default), download Let's Encrypt staging root
```
smolnar82: I'd call this ENV variable `IMPORT_LETS_ENCRYPT_STAGING_CERTS` (to be honest the `DEFAULT` word inside made me think what this is about).
Additionally: you indicate the default is `true`, but this isn't the case (see below, line 47 sets it to `false` if not configured).

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -76,6 +82,27 @@ importMultipleCerts() {
   return "$import_failed"
 }
 
+## Download Let's Encrypt staging root CAs (best-effort) when IMPORT_DEFAULT_STAGING_CERTS is true.
+downloadDefaultStagingCerts() {
```
smolnar82: `downloadLetEncryptStagingCerts` maybe?

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -247,7 +287,16 @@ do
...
-export KNOX_GATEWAY_DBG_OPTS="${KNOX_GATEWAY_DBG_OPTS} -Djavax.net.ssl.trustStore=${KEYSTORE_DIR}/truststore.jks -Djavax.net.ssl.trustStorePassword=${ALIAS_PASSPHRASE}"
+# To avoid leaking password into the process command line 
+# we pass the trust options through a 0600 Java argument file. 
+# Java launcher expands @file after exec, so only "@<path>" appears in the process args.
+TRUSTSTORE_JVM_OPTS_FILE="${KEYSTORE_DIR}/truststore-jvm.options"
+cat > "${TRUSTSTORE_JVM_OPTS_FILE}" <<EOF
+-Djavax.net.ssl.trustStore=${KEYSTORE_DIR}/truststore.jks
+-Djavax.net.ssl.trustStorePassword="${ALIAS_PASSPHRASE}"
+EOF
+chmod 600 "${TRUSTSTORE_JVM_OPTS_FILE}"
```
smolnar82, line 298: This is a nice addition, I like that a lot :)

smolnar82:
@moresandeep - Have you had the chance to read my comments above? If so, what do you think of them?

smolnar82:
@moresandeep - I'd like to hear you back from you on my comments above. Please let us know your thoughts by 31 July 2026. Thank you!

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
moresandeep: You are right, the name is misleading, also the defaults are flipped :( I'll fix this.

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
moresandeep: Yup

gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
moresandeep: :)

moresandeep:
@smolnar82 thanks for the review! and apologies for the delay. I modified the patch based on review comments, can you please take a look again. Thanks!

gateway-docker/src/main/resources/docker/Dockerfile
```diff
@@ -56,7 +56,8 @@ ARG ENTRYPOINT
 COPY --chown=8000:0 ${ENTRYPOINT} /home/knox/knox/entrypoint.sh
 RUN chmod +x /home/knox/knox/entrypoint.sh
 
-# Add the Amazon Root CA and Let's Encrypt production certificates (best-effort)
+# Add the Amazon Root CA and Let's Encrypt production root certificates (best-effort).
+# Staging roots are downloaded at runtime when IMPORT_DEFAULT_STAGING_CERTS=true (default).
```
smolnar82: You meant `IMPORT_LETS_ENCRYPT_STAGING_CERTS`, right?

smolnar82 (approved):
Juts a nitpick comment: in the new comment you failed to rename the ENV variable.
Otherwise it looks good.

gateway-docker/src/main/resources/docker/Dockerfile
moresandeep: Ah, yes! good catch, i will update it and commit.

PR #1251, opened 2026-06-08
KNOX-3330: Handle interceptor name collisions

handavid:
[KNOX-3330](https://issues.apache.org/jira/browse/KNOX-3330) - Ensure Partition IDs are unique when starting Knox LDAP Proxy.

Partition IDs are taken from the interceptor names with whitespace removed. This could lead to potential collisions in the Partition IDs. This change adds a numerical suffix in the case of collision to ensure that the Partition IDs are unique. A unit test was added to cover the case where Partition IDs would collide.

Unit tests were improved to check that interceptors are created in the expected order (reverse order based on the interceptorNames config).

gateway-site.xml examples were moved to doc section.

## How was this patch tested?

Unit tests were added to cover the Partition IDs change.

## Integration Tests
no changes

## UI changes
no changes


handavid:
@smolnar82

smolnar82 (approved):

PR #1257, opened 2026-06-10
KNOX-3334 - Introduce ActorChainPrincipal for RFC 8693 instead of ImpersonatedPrincipal

lmccay:

[KNOX-3334](https://issues.apache.org/jira/browse/KNOX-3334) Introduce ActorChainPrincipal for RFC 8693 instead of ImpersonatedPrincipal

[KNOX-3321](https://issues.apache.org/jira/browse/KNOX-3321) provided an initial implementation for 8693 and adding an 'act' claim to the returned JWT based on the presence of the ImpersonatedPrincipal and having delegated auth enabled on the KnoxToken service.

This falls short of what we need to support token exchanges that already include an 'act' claim in the subject token. To support this properly, we need the previous 'act' claim represented in the Java Subject with the full chain represented. We will then add the next actor subclaim to the chain from within the KnoxToken service, effectively continuing the chain as it flows through the actors for the given request.

To support this, we should introduce a new principal called ActorChainPrincipal which will have an extended interface to provide the list of 'act' claims within the presented token for building out the chain in the new token.

## How was this patch tested?

* Built and ran existing tests
* Added new tests
* Manually tested as follows:

curl -ivku admin:admin-password 'https://localhost:8443/gateway/knoxcreds/knoxtoken/api/v1/token?doAs=guest'

`{"access_token":"eyJqa3UiOiJodHRwczovL2xvY2FsaG9zdDo4NDQzL2dhdGV3YXkva25veGNyZWRzL2tub3h0b2tlbi9hcGkvdjEvandrcy5qc29uIiwia2lkIjoiS1M2enJFSUsxSFJTbUQ4Ri1FckhnOWc2bHlSQ1A1WUcxLUN0TFFaM2U2byIsInR5cCI6IkpXVCIsImFsZyI6IlJTMjU2In0.eyJzdWIiOiJndWVzdCIsImFjd...uIiwia2lkIjoiS1M2enJFSUsxSFJTbUQ4Ri1FckhnOWc2bHlSQ1A1WUcxLUN0TFFaM2U2byIsImlzcyI6IktOT1hTU08iLCJtYW5hZ2VkLnRva2VuIjoidHJ1ZSIsImtub3guaWQiOiI1YTk4MzM4ZC1kNjlkLTRlNTEtYWNiMS0yNzVhNzZkYzE4M2UifQ.VJOd0pJFwQpWFWv7Xo661pKFMl8md_1UmYf1kLdQIgeVfLhhwWINAKqpD-9Od6YwWIXlfr3SBGbPoHeMQYZp1fEOt4fx1gyFo08VGhDSJMI63FX93KpvDrdeECKwcKimIhnh9H9VEQLP56WOBqn3eYoc8aJFua4Ydh9dC0b0AbnVSrDqoS3hOJLwGsj602NvsIU5IoxmYz8s7rNO7CN6qynxmTp-w4g1Q3skmqU8zki9DvEMJXMRdMsflssFDWrybw0UvzMjjmKQxzKqcb9jsnM3lrsqit7-JEg6kPNA5M5IWIKgkIgt-P_iYAr6ouXt0BxxGfJplK_rF2qxkD98xg","token_id":"5a98338d-d69d-4e51-acb1-275a76dc183e","managed":"true","endpoint_public_cert":"MIIDZDCCAkygAwIBAgIISWFJksv5UD4wDQYJKoZIhvcNAQELBQAwXzESMBAGA1UEAwwJbG9jYWxob3N0MQ0wCwYDVQQLDARUZXN0MQ8wDQYDVQQKDAZIYWRvb3AxDTALBgNVBAcMBFRlc3QxDTALBgNVBAgMBFRlc3QxCzAJBgNVBAYTAlVTMB4XDTI2MDYxMDIxMTg1M1oXDTI3MDYxMDIxMTg1M1owXzESMBAGA1UEAwwJbG9jYWxob3N0MQ0wCwYDVQQLDARUZXN0MQ8wDQYDVQQKDAZIYWRvb3AxDTALBgNVBAcMBFRlc3QxDTALBgNVBAgMBFRlc3QxCzAJBgNVBAYTAlVTMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEApSDLZB9y0gV5W9HEmV49GAo4PH9JTmfISUCTGYZ5z6pwU0ierFz0Qm....OCNkLF4fj5iQAqyp+6lF7e1Wj50oaKCTsLkxV+3bn2PVdI7rYR9wS6t6V7MuoaLbgcL/RY8/ec3qUC6NvQIDAQABoyQwIjAgBgNVHREEGTAXggpHV1dZRkc2NU1Rgglsb2NhbGhvc3QwDQYJKoZIhvcNAQELBQADggEBAD8URUGKn4Em8AtVBrLxfwxyvPxWHXUONERbFKizIV2xkTcxgNnoKKUaHX2A8f0DW1OQ1KsiR2mp8ygeZcqgsCcEpPGnzYnfknD58nEDOXPpg2vy64ppIXzXPbR6PUBQo9zuIH5n+JlFhR5dP0WugqdNarKwtNCSjDRmjiLki4quuM8Wd4kSflHZmFwehqUH0Hc4mtdvrrCZUrz3RnK0N/oBW3Mbr5/0gj6CzC4YgKqCipRzCvJ9RKrewOumPLhg5m2Zb6NfDJXVFG4WhrYF+dRX0Dljngy2eLsTOdg4nlenRZYZWZL6/UpnpjSik/J3AFB7CPLStC280QP6m/89R5o=","token_type":"Bearer","expires_in":-1,"passcode":"TldFNU9ETXpPR1F0WkRZNVpDMDBaVFV4TFdGallqRXRNamMxWVRjMlpHTXhPRE5sOjpZVEF5TlRVeFlqSXRPR05rTVMwME56STFMVGhrTlRZdE1USmtNamxpT0dSbVpURmg="}%  `

curl -ivk -H 'Authorization: Bearer eyJqa3UiOiJodHRwczovL2xvY2FsaG9zdDo4NDQzL2dhdGV3YXkva25veGNyZWRzL2tub3h0b2tlbi9hcGkvdjEvandrcy5qc29uIiwia2lkIjoiS1M2enJFSUsxSFJTbUQ4Ri1....1ZSIsImtub3guaWQiOiI1YTk4MzM4ZC1kNjlkLTRlNTEtYWNiMS0yNzVhNzZkYzE4M2UifQ.VJOd0pJFwQpWFWv7Xo661pKFMl8md_1UmYf1kLdQIgeVfLhhwWINAKqpD-9Od6YwWIXlfr3SBGbPoHeMQYZp1fEOt4fx1gyFo08VGhDSJMI63FX93KpvDrdeECKwcKimIhnh9H9VEQLP56WOBqn3eYoc8aJFua4Ydh9dC0b0AbnVSrDqoS3hOJLwGsj602NvsIU5IoxmYz8s7rNO7CN6qynxmTp-w4g1Q3skmqU8zki9DvEMJXMRdMsflssFDWrybw0UvzMjjmKQxzKqcb9jsnM3lrsqit7-JEg6kPNA5M5IWIKgkIgt-P_iYAr6ouXt0BxxGfJplK_rF2qxkD98xg' 'https://localhost:8443/gateway/knoxidf/knoxtoken/v1/oauth/tokens?doAs=tom'

`{"access_token":"eyJqa3UiOiJodHRwczovL2xvY2FsaG9zdDo4NDQzL2dhdGV3YXkva25veGlkZi9rbm94dG9rZW4vdjEvb2F1dGgvandrcy5qc29uIiwia2lkIjoiS1M2enJFSUsxSFJTbUQ4Ri1FckhnOWc2bHlSQ1A1WUcxLUN0TFFaM2U2byIsInR5cCI6IkpXVCIsImFsZyI6IlJTMjU2In0.eyJzdWIiOiJ0b20iLCJhY3QiOn....ZTZvIiwiaXNzIjoiS05PWFNTTyIsIm1hbmFnZWQudG9rZW4iOiJ0cnVlIiwia25veC5pZCI6IjQwYTFmNmNmLWRmYzQtNDU3Ny05MmNjLWEwMTI5M2RkYWZiZCJ9.kUi8Mc7E7njrrh3B-UWgDjJVlWm23rRmd6p9en79Q-6Q_I5qw1LQf6bqUPqTwZ35JgZwfcSodFcSuitJwgOOoAbhx3qZR8mTVhoqNyoUtSqmtixkSvMQwXTBZtcWDKyjc208LZCs6Vv21T_oHXHuInSUZnTjghoM_gh1JrcIkxvLPaz8utL-aWc_1cJvNKNvQtz5uuU_FynLLhBCJQ-dJnKOlJ4-HXZwlnA9QhaEQDW-07n-R7ZhM_urqrrmYdYKNlyRztoCiD7M13pSwdjxf4AQura1vV7AyRwEEfI1gTPZ7L2xP52dVBU1R_sBz83NuXzSOumpGWJa8x4eWOIXsA","refresh_token":"d7622c25-4514-463d-a608-37aadbba0cce","issued_token_type":"urn:ietf:params:oauth:token-type:access_token","token_type":"Bearer","expires_in":-1}`

image

### DIFF
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
@@ -46,6 +46,7 @@ import org.apache.knox.gateway.filter.security.AbstractIdentityAssertionBase;
 import org.apache.knox.gateway.i18n.GatewaySpiResources;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.i18n.resources.ResourcesFactory;
+import org.apache.knox.gateway.security.ActorChainPrincipal;
 import org.apache.knox.gateway.security.GroupPrincipal;
 import org.apache.knox.gateway.security.ImpersonatedPrincipal;
 import org.apache.knox.gateway.security.PrimaryPrincipal;
@@ -146,6 +147,11 @@ public abstract class AbstractIdentityAssertionFilter extends
 
           final Set<TokenIdPrincipal> tokenIdPrincipals = SubjectUtils.getTokenIdPrincipals(currentSubject);
           subject.getPrincipals().addAll(tokenIdPrincipals);
+
+          // RFC 8693 Token Exchange: Preserve ActorChainPrincipal from the current subject
+          // This ensures the delegation chain is maintained through identity assertion
+          final Set<ActorChainPrincipal> actorChainPrincipals = currentSubject.getPrincipals(ActorChainPrincipal.class);
+          subject.getPrincipals().addAll(actorChainPrincipals);
 
           doAs(request, response, chain, subject);
         }
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
@@ -150,7 +150,7 @@ public abstract class AbstractIdentityAssertionFilter extends
 
           // RFC 8693 Token Exchange: Preserve ActorChainPrincipal from the current subject
           // This ensures the delegation chain is maintained through identity assertion
-          final Set<ActorChainPrincipal> actorChainPrincipals = currentSubject.getPrincipals(ActorChainPrincipal.class);
+          final Set<ActorChainPrincipal> actorChainPrincipals = SubjectUtils.getActorChainPrincipal(currentSubject, subject);
           subject.getPrincipals().addAll(actorChainPrincipals);
 
           doAs(request, response, chain, subject);
@@ -160,7 +160,7 @@ public abstract class AbstractIdentityAssertionFilter extends
         }
       }
 
-  private void doAs(final ServletRequest request, final ServletResponse response, final FilterChain chain, Subject subject)
+    private void doAs(final ServletRequest request, final ServletResponse response, final FilterChain chain, Subject subject)
       throws IOException, ServletException {
     try {
       Subject.doAs(
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -1108,36 +1108,40 @@ public class TokenResource {
       }
 
       // RFC 8693 Token Exchange: Build the actor chain if delegated auth is enabled
-      if (enableDelegatedAuth) {
-        // First check if there's an existing actor chain from a previous token exchange
-        Set<ActorChainPrincipal> actorChainPrincipals = subject.getPrincipals(ActorChainPrincipal.class);
-        List<Map<String, Object>> existingChain = null;
-        if (!actorChainPrincipals.isEmpty()) {
-          existingChain = actorChainPrincipals.iterator().next().getActorChain();
-          log.generalInfoMessage("Found existing actor chain with " + existingChain.size() + " actors");
-        }
-
-        // Check if impersonation is occurring to add a new actor to the chain
-        if (SubjectUtils.isImpersonating(subject)) {
-          String primaryPrincipalName = SubjectUtils.getPrimaryPrincipalName(subject);
-          String impersonatedPrincipalName = SubjectUtils.getImpersonatedPrincipalName(subject);
-          if (primaryPrincipalName != null && impersonatedPrincipalName != null && !primaryPrincipalName.equals(impersonatedPrincipalName)) {
-            // Build the new actor chain by adding the current actor (primary principal) to the existing chain
-            List<Map<String, Object>> newActorChain = TokenUtils.addActorToChain(existingChain, primaryPrincipalName);
-            jwtAttributesBuilder.setActorChain(newActorChain);
-            log.addingActorClaimToToken(primaryPrincipalName, impersonatedPrincipalName);
-          }
-        } else if (existingChain != null && !existingChain.isEmpty()) {
-          // No new impersonation, but preserve existing actor chain
-          jwtAttributesBuilder.setActorChain(existingChain);
-          log.generalInfoMessage("Preserving existing actor chain without adding new actor");
-        }
-      }
+      handleDelegatedAuthentication(subject, jwtAttributesBuilder);
     }
 
     jwtAttributes = jwtAttributesBuilder.build();
     token = ts.issueToken(jwtAttributes);
     return token;
+  }
+
+  private void handleDelegatedAuthentication(Subject subject, JWTokenAttributesBuilder jwtAttributesBuilder) {
+    if (enableDelegatedAuth) {
+      // First check if there's an existing actor chain from a previous token exchange
+      Set<ActorChainPrincipal> actorChainPrincipals = subject.getPrincipals(ActorChainPrincipal.class);
+      List<Map<String, Object>> existingChain = null;
+      if (!actorChainPrincipals.isEmpty()) {
+        existingChain = actorChainPrincipals.iterator().next().getActorChain();
+        log.generalInfoMessage("Found existing actor chain with " + existingChain.size() + " actors");
+      }
+
+      // Check if impersonation is occurring to add a new actor to the chain
+      if (SubjectUtils.isImpersonating(subject)) {
+        String primaryPrincipalName = SubjectUtils.getPrimaryPrincipalName(subject);
+        String impersonatedPrincipalName = SubjectUtils.getImpersonatedPrincipalName(subject);
+        if (primaryPrincipalName != null && impersonatedPrincipalName != null && !primaryPrincipalName.equals(impersonatedPrincipalName)) {
+          // Build the new actor chain by adding the current actor (primary principal) to the existing chain
+          List<Map<String, Object>> newActorChain = TokenUtils.addActorToChain(existingChain, primaryPrincipalName);
+          jwtAttributesBuilder.setActorChain(newActorChain);
+          log.addingActorClaimToToken(primaryPrincipalName, impersonatedPrincipalName);
+        }
+      } else if (existingChain != null && !existingChain.isEmpty()) {
+        // No new impersonation, but preserve existing actor chain
+        jwtAttributesBuilder.setActorChain(existingChain);
+        log.generalInfoMessage("Preserving existing actor chain without adding new actor");
+      }
+    }
   }
 
   private boolean shouldIncludeGroups() {
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -70,6 +70,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.context.ContextAttributes;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
+import org.apache.knox.gateway.security.ActorChainPrincipal;
 import org.apache.knox.gateway.security.GroupPrincipal;
 import org.apache.knox.gateway.security.SubjectUtils;
 import org.apache.knox.gateway.security.TokenIdPrincipal;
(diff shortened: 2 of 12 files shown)


gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/impl/JWTToken.java
```diff
@@ -100,14 +101,14 @@ public JWTToken(JWTokenAttributes jwtAttributes) {
...
-      // The act claim value is a JSON object containing a "sub" claim with the identity of the actor.
-      JWTClaimsSet actClaims = new JWTClaimsSet.Builder()
-          .subject(jwtAttributes.getActor())
-          .build();
-      builder.claim(ACT_CLAIM, actClaims.toJSONObject());
+    // RFC 8693 Token Exchange: The "act" (actor) claim provides a means within a JWT to express
+    // that delegation has occurred and identify the acting party to whom authority has been delegated.
+    // The actor chain is converted to the nested structure required by RFC 8693.
+    if (jwtAttributes.getActorChain() != null && !jwtAttributes.getActorChain().isEmpty()) {
+      Map<String, Object> nestedAct = org.apache.knox.gateway.services.security.token.TokenUtils.buildNestedActClaim(jwtAttributes.getActorChain());
```
smolnar82: nit: TokenUtils can be imported -> no need for the fully qualified class name to be added here.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/impl/JWTToken.java
```diff
@@ -100,14 +101,14 @@ public JWTToken(JWTokenAttributes jwtAttributes) {
...
-      // that delegation has occurred and identify the acting party to whom authority has been delegated.
-      // The act claim value is a JSON object containing a "sub" claim with the identity of the actor.
-      JWTClaimsSet actClaims = new JWTClaimsSet.Builder()
-          .subject(jwtAttributes.getActor())
-          .build();
-      builder.claim(ACT_CLAIM, actClaims.toJSONObject());
+    // RFC 8693 Token Exchange: The "act" (actor) claim provides a means within a JWT to express
+    // that delegation has occurred and identify the acting party to whom authority has been delegated.
+    // The actor chain is converted to the nested structure required by RFC 8693.
+    if (jwtAttributes.getActorChain() != null && !jwtAttributes.getActorChain().isEmpty()) {
```
smolnar82: This if is redundant -> TokenUtils.buildNestedActClaim checks for the same and returns null.

gateway-spi/src/main/java/org/apache/knox/gateway/security/ActorChainPrincipalImpl.java
```diff
@@ -0,0 +1,95 @@
...
+    Object sub = actorChain.get(actorChain.size() - 1).get("sub");
+    return sub != null ? sub.toString() : null;
+  }
+
+  @Override
+  public String getName() {
+    // Return the current actor's identity as the principal name
+    String currentActor = getCurrentActor();
+    return currentActor != null ? currentActor : "";
+  }
```
smolnar82, line 70: I did not find any use of this method, I think it can be simply removed.

gateway-spi/src/main/java/org/apache/knox/gateway/security/ActorChainPrincipal.java
```diff
@@ -0,0 +1,85 @@
...
+   */
+  List<Map<String, Object>> getActorChain();
+
+  /**
+   * Returns the subject (identity) of the most recent actor in the chain.
+   *
+   * <p>This is equivalent to calling {@code getActorChain().get(0).get("sub")}
+   * if the chain is not empty.</p>
+   *
+   * @return the subject of the most recent actor, or null if the chain is empty
```
smolnar82, line 71: Maybe return `Optional<String>` instead of `null`?

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -1106,14 +1107,30 @@ protected JWT getJWT(String userName, long expires, String jku) throws TokenServ
...
+          if (primaryPrincipalName != null && impersonatedPrincipalName != null && !primaryPrincipalName.equals(impersonatedPrincipalName)) {
+            // Build the new actor chain by adding the current actor (primary principal) to the existing chain
+            List<Map<String, Object>> newActorChain = TokenUtils.addActorToChain(existingChain, primaryPrincipalName);
+            jwtAttributesBuilder.setActorChain(newActorChain);
+            log.addingActorClaimToToken(primaryPrincipalName, impersonatedPrincipalName);
+          }
+        } else if (existingChain != null && !existingChain.isEmpty()) {
+          // No new impersonation, but preserve existing actor chain
+          jwtAttributesBuilder.setActorChain(existingChain);
+          log.generalInfoMessage("Preserving existing actor chain without adding new actor");
```
smolnar82: This logic deserves its own private method (e.g. `handleDelegatedAuth`), including the `if` statement -> here it would only be one method call (e.g. `handleDelegatedAuth(...)`.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
```diff
@@ -147,6 +148,11 @@ protected void continueChainAsPrincipal(HttpServletRequestWrapper request, Servl
           final Set<TokenIdPrincipal> tokenIdPrincipals = SubjectUtils.getTokenIdPrincipals(currentSubject);
           subject.getPrincipals().addAll(tokenIdPrincipals);
 
+          // RFC 8693 Token Exchange: Preserve ActorChainPrincipal from the current subject
+          // This ensures the delegation chain is maintained through identity assertion
+          final Set<ActorChainPrincipal> actorChainPrincipals = currentSubject.getPrincipals(ActorChainPrincipal.class);
```
smolnar82: We have named methods in `SubjectUtils` to get different principals from current subject (`getCurrentGroupPrincipals`, `getTokenIdPrincipals`, ...), I think we should add `getActorChainPrincipals` there too.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/TokenUtils.java
```diff
@@ -114,4 +118,143 @@ private static boolean useHMAC(char[] hmacSecret, String signingKeystoreName) {
...
+  @SuppressWarnings("unchecked")
+  public static List<Map<String, Object>> extractActorChain(JWT token) {
+    if (token == null) {
+      return Collections.emptyList();
+    }
+
+    Object actClaim = token.getClaimAsObject(JWTToken.ACT_CLAIM);
+    if (actClaim == null) {
+      return Collections.emptyList();
+    }
```
smolnar82: These 2 `if`s could be merged:
```
if (token == null || token.getClaimAsObject(JWTToken.ACT_CLAIM) == null) {
   return Collections.emptyList();
}
```
Maybe something like this?
```
public static List<Map<String, Object>> extractActorChain(JWT token) {
    final Object currentAct = token == null ? null : token.getClaimAsObject(JWTToken.ACT_CLAIM);
    final List<Map<String, Object>> actorChain = new ArrayList<>();

    while (currentAct instanceof Map<?, ?> actorMap) {
        actorChain.add(new LinkedHashMap<>((Map<String, Object>) actorMap));
        currentAct = actorMap.get(JWTToken.ACT_CLAIM);
    }

    return Collections.unmodifiableList(actorChain);
}
```

smolnar82 (changes requested):

gateway-spi/src/main/java/org/apache/knox/gateway/security/ActorChainPrincipalImpl.java
lmccay: I believe getName() is part of the Principal interface and cannot be removed.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/impl/JWTToken.java
lmccay: Got it.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/impl/JWTToken.java
lmccay: Got it.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
lmccay: Ahh, yes, I forgot to go back and do this but meant to.

gateway-spi/src/main/java/org/apache/knox/gateway/security/ActorChainPrincipal.java
lmccay: I don't think that this brings anything meaningful to the table. It really isn't expected that this would be an Optional return value. If there is an empty Map then a return for a given key will be a null. Adding the ceremony around checking for ifPresent type checks in this case is outside of what the contract would expect. I would even argue that the getActorChain method should return null but I went with the empty List to be somewhat safer. It really isn't optional to have the 'sub' as part of an actorChain that is non-empty. If you are looking for a specific key in an empty map or you are using a key that is invalid then it isn't really optional. If this were a map like the token metadata that we have elsewhere and certain metadata items are indeed optional, it makes sense for the contract to call that expectation out in the interface and nudge the ifPresent type checks.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/token/TokenUtils.java
lmccay: Got it.  I like the previous readability better but this is fine too.

gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
lmccay: Got it. One liner which may be why I didn't bother before but it is consistent.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
lmccay: Got it.

smolnar82:
LGTM. Sorry for the late approval.

PR #1258, opened 2026-06-11
KNOX-3340: Add Control for LDAPRolesLookupInterceptor

handavid:
[KNOX-3340](https://issues.apache.org/jira/browse/KNOX-3340) - Add Control to LDAPRolesLookupInterceptor

This commit adds a RolesLookupBypassControl for use with the LDAPRolesLookupInterceptor. The LDAPRolesLookupInterceptor will skip role mapping if this control is present and true in the request. This lets the client decide whether they will receive users' groups or roles.
  
## How was this patch tested?

Unit tests were added to cover the new code.

Manual testing was performed. The LDAP Proxy was configured with the RolesLookup interceptor and the following `ldapsearch` commands were run.
```
# add control by OID with value "true"
ldapsearch -v -x -H ldap://localhost:3890 -b 'ou=people,DC=proxy,DC=com' -e "1.3.6.1.4.1.18060.2.1379319520.35362.17433.40846.265936912329953=AQP/" '(uid=sam*)' '*'

# add control by OID with value "false"
ldapsearch -v -x -H ldap://localhost:3890 -b 'ou=people,DC=proxy,DC=com' -e "1.3.6.1.4.1.18060.2.1379319520.35362.17433.40846.265936912329953=AQMA" '(uid=sam*)' '*'

# don't add control
ldapsearch -v -x -H ldap://localhost:3890 -b 'ou=people,DC=proxy,DC=com' '(uid=sam*)' '*'
```

## Integration Tests
no integration tests added

## UI changes
no UI changes

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControl.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.services.ldap.control;
+
+import org.apache.directory.api.ldap.model.message.Control;
+
+public interface RolesLookupBypassControl extends Control {
+    // OID created from a UUID to ensure no collisions:
+    // Apache Root OID for core object classes 1.3.6.1.4.1.18060.2
+    // UUID "5236bee0-8a22-4419-9f8e-f1de43312ce1"
+    String OID = "1.3.6.1.4.1.18060.2.1379319520.35362.17433.40846.265936912329953";
+
+    boolean isBypassRolesLookup();
+    void setBypassRolesLookup(boolean bypassRolesLookup);
+}
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControl.java
@@ -20,11 +20,6 @@ package org.apache.knox.gateway.services.ldap.control;
 import org.apache.directory.api.ldap.model.message.Control;
 
 public interface RolesLookupBypassControl extends Control {
-    // OID created from a UUID to ensure no collisions:
-    // Apache Root OID for core object classes 1.3.6.1.4.1.18060.2
-    // UUID "5236bee0-8a22-4419-9f8e-f1de43312ce1"
-    String OID = "1.3.6.1.4.1.18060.2.1379319520.35362.17433.40846.265936912329953";
-
     boolean isBypassRolesLookup();
     void setBypassRolesLookup(boolean bypassRolesLookup);
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
@@ -60,7 +60,7 @@ public class RolesLookupBypassControlFactory extends AbstractControlFactory<Role
             RolesLookupBypassControl rolesLookupBypassControl = (RolesLookupBypassControl) control;
 
             buffer.put(BOOLEAN_TAG_BYTE);
-            buffer.put((byte) 3);
+            buffer.put((byte) 1); // Value is one byte long
             buffer.put((byte) (rolesLookupBypassControl.isBypassRolesLookup() ? 0xFF : 0x00));
         }
     }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
@@ -22,17 +22,18 @@ import org.apache.directory.api.asn1.util.Asn1Buffer;
 import org.apache.directory.api.ldap.codec.api.AbstractControlFactory;
 import org.apache.directory.api.ldap.codec.api.LdapApiService;
 import org.apache.directory.api.ldap.model.message.Control;
+import org.apache.knox.gateway.services.ldap.model.constants.SchemaConstants;
 
 public class RolesLookupBypassControlFactory extends AbstractControlFactory<RolesLookupBypassControl> {
     public static final byte BOOLEAN_TAG_BYTE = 0x01;
 
-    public RolesLookupBypassControlFactory(LdapApiService codec, String oid) {
-        super(codec, oid);
+    public RolesLookupBypassControlFactory(LdapApiService codec) {
+        super(codec, SchemaConstants.ROLES_LOOKUP_BYPASS_CONTROL_OID);
     }
 
     @Override
     public Control newControl() {
-        return new RolesLookupBypassControlDecorator(codec, new RolesLookupBypassControlImpl(oid), this);
+        return new RolesLookupBypassControlDecorator(codec, new RolesLookupBypassControlImpl(), this);
     }
 
     @Override
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.knox.gateway.services.ldap.control;
+
+import org.apache.directory.api.asn1.DecoderException;
+import org.apache.directory.api.asn1.util.Asn1Buffer;
+import org.apache.directory.api.ldap.codec.api.AbstractControlFactory;
+import org.apache.directory.api.ldap.codec.api.LdapApiService;
+import org.apache.directory.api.ldap.model.message.Control;
+
+public class RolesLookupBypassControlFactory extends AbstractControlFactory<RolesLookupBypassControl> {
+    public static final byte BOOLEAN_TAG_BYTE = 0x01;
+
+    public RolesLookupBypassControlFactory(LdapApiService codec) {
+        super(codec, RolesLookupBypassControl.OID);
+    }
+
+    @Override
+    public Control newControl() {
+        return new RolesLookupBypassControlDecorator(codec, new RolesLookupBypassControlImpl(), this);
+    }
+
+    @Override
+    public void decodeValue(Control control, byte[] controlBytes) throws DecoderException {
+        if (control instanceof RolesLookupBypassControl) {
+            RolesLookupBypassControl rolesLookupBypassControl = (RolesLookupBypassControl) control;
+            if (controlBytes == null || controlBytes.length < 3) {
+                throw new DecoderException("Invalid BER encoding for Boolean Control");
+            }
+
+            if (controlBytes[0] != BOOLEAN_TAG_BYTE) {
+                throw new DecoderException("Expected Boolean Tag (0x01), found: " + controlBytes[0]);
+            }
+
+            boolean value = (controlBytes[2] != 0x00);
+            rolesLookupBypassControl.setBypassRolesLookup(value);
+        } else {
+            throw new DecoderException("Cannot decode into "  + control.getClass().getSimpleName() + ". Control must be instance of RolesLookupBypassControl.");
+        }
+    }
+
+    @Override
+    public void encodeValue(Asn1Buffer buffer, Control control) {
+        if (control instanceof RolesLookupBypassControl) {
+            RolesLookupBypassControl rolesLookupBypassControl = (RolesLookupBypassControl) control;
+
+            buffer.put(BOOLEAN_TAG_BYTE);
+            buffer.put((byte) 3);
+            buffer.put((byte) (rolesLookupBypassControl.isBypassRolesLookup() ? 0xFF : 0x00));
+        }
+    }
+}
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
@@ -26,13 +26,13 @@ import org.apache.directory.api.ldap.model.message.Control;
 public class RolesLookupBypassControlFactory extends AbstractControlFactory<RolesLookupBypassControl> {
     public static final byte BOOLEAN_TAG_BYTE = 0x01;
 
-    public RolesLookupBypassControlFactory(LdapApiService codec) {
-        super(codec, RolesLookupBypassControl.OID);
+    public RolesLookupBypassControlFactory(LdapApiService codec, String oid) {
+        super(codec, oid);
     }
 
     @Override
     public Control newControl() {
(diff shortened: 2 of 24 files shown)


handavid:
@smolnar82

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControl.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+package org.apache.knox.gateway.services.ldap.control;
+
+import org.apache.directory.api.ldap.model.message.Control;
+
+public interface RolesLookupBypassControl extends Control {
+    // OID created from a UUID to ensure no collisions:
+    // Apache Root OID for core object classes 1.3.6.1.4.1.18060.2
+    // UUID "5236bee0-8a22-4419-9f8e-f1de43312ce1"
+    String OID = "1.3.6.1.4.1.18060.2.1379319520.35362.17433.40846.265936912329953";
```
handavid: @smolnar82 do you know how to get an official OID? This OID is generated from a UUID so we shouldn't have any collisions, but it would be nice to get an official OID before this control actually enters use.

smolnar82: That's an excellent question, @handavid !

You are correct that this UUID-based approach ensures no collisions, but we should aim for a cleaner, official arc under the Apache Private Enterprise Number (PEN). 

As a PMC member, I can help facilitate this. Here is the path forward:
   1. Apache Root: The ASF owns PEN 18060, so our official OID will start with 1.3.6.1.4.1.18060.
   2. Knox Sub-arc: I've checked the [IANA registry and confirmed Knox doesn't have an explicit entry](https://www.iana.org/assignments/enterprise-numbers/?q=Knox) because we fall under the Apache root. The PMC needs to request a dedicated sub-arc (e.g., 1.3.6.1.4.1.18060.X) for Knox by opening a JIRA with ASF Infrastructure (INFRA). - I think @lmccay can help us with that.
   3. Next Steps: For this PR, it's fine to keep the UUID-based OID as a placeholder so we don't block progress. Once we secure the official Knox sub-arc from INFRA, we can do a quick follow-up task to update this to a shorter, official OID (like 1.3.6.1.4.1.18060.X.1.1).

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    @Override
+    public void encodeValue(Asn1Buffer buffer, Control control) {
+        if (control instanceof RolesLookupBypassControl) {
+            RolesLookupBypassControl rolesLookupBypassControl = (RolesLookupBypassControl) control;
+
+            buffer.put(BOOLEAN_TAG_BYTE);
+            buffer.put((byte) 3);
```
smolnar82: The correct length is `1`.

smolnar82 (changes requested):
A BER-encoded Boolean consists of
   * Tag: 1 byte (0x01)
   * Length: 1 byte (The size of the value)
   * Value: 1 byte (0x00 or 0xFF)
   * Total: 3 bytes.
  
The error in the PR is specifically the content of the middle byte (the "Length" field of the TLV).

In BER, the "Length" byte should only represent the size of the Value portion. For a Boolean, the value is only 1 byte long. 
   * Current PR encoding: 0x01 (Tag) | 0x03 (Length) | Value (1 byte) -> This says "expect 3 more bytes" but only provides 1.
   * Correct BER encoding: 0x01 (Tag) | 0x01 (Length) | Value (1 byte).

Setting the length byte to 0x03 while only providing 1 byte of data might cause standard LDAP clients or strict decoders to fail.

Once this change is done, you may also have to change the documentation.

Great work on the implementation and the tests otherwise!

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControl.java
smolnar82: I filed [KNOX-3346](https://issues.apache.org/jira/browse/KNOX-3346) and assigned to @lmccay to create an official sub-arc for the Knox project under Apache.

handavid: thanks! In the meantime, I've moved the OID out of code and into configuration. That way we don't publish any unofficial OIDs. The configuration is optional, but the RolesLookupBypassControl will not be registered if it is not supplied. Anyone who wants to use the bypass control before we get an official oid will need to configure one of their own.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactory.java
handavid: fixed

handavid:
@smolnar82 I've updated the PR for your requested change and also to move the OID into configuration until we get an official OID.

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+        assertEquals("isBypassRolesLookup should match the value from the decorated Impl", rolesLookupBypassControl.isBypassRolesLookup(), rolesLookupBypassControlDecorator.isBypassRolesLookup());
+    }
+
+    @Test
+    public void setBypassRolesLookup() {
+        rolesLookupBypassControlDecorator.setBypassRolesLookup(true);
+        assertEquals("isBypassRolesLookup should match the value from the decorated Impl", rolesLookupBypassControl.isBypassRolesLookup(), rolesLookupBypassControlDecorator.isBypassRolesLookup());
+        rolesLookupBypassControlDecorator.setBypassRolesLookup(false);
+        assertEquals("isBypassRolesLookup should match the value from the decorated Impl", rolesLookupBypassControl.isBypassRolesLookup(), rolesLookupBypassControlDecorator.isBypassRolesLookup());
+    }
```
smolnar82: This is exactly the same test as the previous one. You might want to remove this one and rename the above to `testBypassLookup`?

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlDecoratorTest.java
```diff
@@ -0,0 +1,130 @@
...
+        rolesLookupBypassControl.setBypassRolesLookup(true);
+        byte[] expectedBytes = new byte[]{0x01, 0x01, (byte) 0xff};
+
+        ByteBuffer byteBuffer = ByteBuffer.allocate(3);
+        ByteBuffer encodedBuffer = rolesLookupBypassControlDecorator.encode(byteBuffer);
+        // transition from write mode to read mode
+        encodedBuffer.flip();
+        byte[] encodedBytes = new byte[encodedBuffer.remaining()];
+        encodedBuffer.get(encodedBytes);
+        assertArrayEquals(expectedBytes, encodedBytes);
```
smolnar82: Maybe creating a new private method (`testEncode(boolean encodeTrue)`) and call that one from here and from `encodeFalseValue`? The only difference is the content of `expectedBytes`, but we'd save some lines of duplication.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/LDAPRolesLookupInterceptorFactory.java
```diff
@@ -35,7 +37,16 @@ public Interceptor create(GatewayConfig gatewayConfig, String name, Map<String,
         if (ldapRolesLookupService == null || !ldapRolesLookupService.enabled()) {
             throw new ServiceLifecycleException("LDAP roles lookup service not found or disabled");
         }
-        return new LDAPRolesLookupInterceptor(ldapRolesLookupService);
+
+        // Configure the OID used for the RolesLookupBypassControl until an official OID is available
+        String rolesLookupBypassControlOid = gatewayConfig.getLdapRolesLookupBypassControlOid();
+        if (StringUtils.isNotBlank(rolesLookupBypassControlOid)) {
+            if (!Oid.isOid(rolesLookupBypassControlOid)) {
+                rolesLookupBypassControlOid = null;
```
smolnar82: I see you are throwing `throw new ConfigurationException("Roles Lookup Bypass Control OID is not valid");` below in `KnoxLDAPServerManager`. However, I think that exception belongs here, when the factory wants to create an instance of `LDAPRolesLookupInterceptor`.
The creation of this interceptor is also happening in `KnoxLDAPServerManager.init()` by calling `createInterceptors`, so it'd fail fast in case someone provided an invalid OID.
What do you think?

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactoryTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+        rolesLookupBypassControlFactory.encodeValue(asn1Buffer, control);
+
+        // expectedBytes in reverse because Asn1Buffer stores bytes in reverse order
+        byte[] expectedBytes = new byte[]{(byte) 0xff, 0x01, 0x01};
+        System.out.println(asn1Buffer.toString());
+        ByteBuffer encodedBuffer = asn1Buffer.getBytes();
+        byte[] encodedBytes = new byte[encodedBuffer.remaining()];
+        encodedBuffer.get(encodedBytes);
+        assertArrayEquals(expectedBytes, encodedBytes);
```
smolnar82: I've the same observation here as above wrt. encode/decode testing (one private method would eliminate the majority of duplicated lines here).

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/interceptor/LDAPRolesLookupInterceptorTest.java
```diff
@@ -68,14 +90,95 @@ public void testModifyEntryNoMemberOfNoRoles() throws Exception {
         assertNull(modifiedEntry.get("memberOf"));
     }
 
-    private LDAPRolesLookupInterceptor createInterceptor() {
+    @Test
+    public void testRolesLookupNoBypass() throws Exception {
```
smolnar82, line 92: With a little bit of refactor, we might save some duplicated lines here too (you may think this is "my thing" :) )
```
private TestContext createTestContext(boolean bypass, LDAPRolesLookupService rolesService) throws Exception {
    DirectoryService directoryService = new SimpleDirectoryService();
    directoryService.setShutdownHookEnabled(false);
    directoryService.setSchemaManager(SchemaManagerFactory.createSchemaManager());

    LDAPRolesLookupInterceptor interceptor =
            new LDAPRolesLookupInterceptor(rolesService, ROLES_LOOKUP_BYPASS_CONTROL_OID);
    interceptor.init(directoryService);
    directoryService.addLast(interceptor);

    ConfigurableEntriesTestInterceptor nextInterceptor =
            new ConfigurableEntriesTestInterceptor("NEXT");
    nextInterceptor.init(directoryService);
    directoryService.addLast(nextInterceptor);

    SearchOperationContext ctx =
            new SearchOperationContext(directoryService.getSession());
    ctx.setInterceptors(List.of(interceptor.getName(), "NEXT"));

    RolesLookupBypassControl control =
            new RolesLookupBypassControlImpl(ROLES_LOOKUP_BYPASS_CONTROL_OID);
    control.setBypassRolesLookup(bypass);
    ctx.addRequestControl(control);

    return new TestContext(interceptor, nextInterceptor, ctx);
}

private record TestContext(
        LDAPRolesLookupInterceptor interceptor,
        ConfigurableEntriesTestInterceptor nextInterceptor,
        SearchOperationContext ctx) {
}
```
Then tests are simpler:
```
@Test
public void testRolesLookupNoBypass() throws Exception {
    LDAPRolesLookupService mockRolesService = EasyMock.createMock(LDAPRolesLookupService.class);

    Collection<String> roles = List.of("roleA", "roleG");
    expect(mockRolesService.lookupRoles(anyString(), anyObject()))
            .andReturn(roles)
            .atLeastOnce();
    replay(mockRolesService);

    TestContext tc = createTestContext(false, mockRolesService);

    Entry userEntry = createUserEntry("alice", "cn=group1,ou=groups,dc=hadoop,dc=apache,dc=org");
    tc.nextInterceptor().setEntries(List.of(userEntry));

    EntryFilteringCursor entries = tc.interceptor().search(tc.ctx());

    assertTrue(entries.next());
    assertMemberOf(entries.get(),
            "cn=roleA,ou=groups,dc=hadoop,dc=apache,dc=org",
            "cn=roleG,ou=groups,dc=hadoop,dc=apache,dc=org");
    assertFalse(entries.next());
}

@Test
public void testRolesLookupWithBypass() throws Exception {
    TestContext tc = createTestContext(true, createMockRolesService());

    Entry userEntry = createUserEntry("alice", "cn=group1,ou=groups,dc=hadoop,dc=apache,dc=org");
    tc.nextInterceptor().setEntries(List.of(userEntry));

    EntryFilteringCursor entries = tc.interceptor().search(tc.ctx());

    assertTrue(entries.next());
    assertMemberOf(entries.get(), "cn=group1,ou=groups,dc=hadoop,dc=apache,dc=org");
    assertFalse(entries.next());
}
```

knox-site/docs/service_ldap_server.md
```diff
@@ -60,6 +60,31 @@ The duplicate user filter interceptor ensures that each `Entry` has a unique `ui
...
+
+For example, the control can be added to the `ldapsearch` cli using the `-e` option.
+```shell script
+ldapsearch -v -x -H ldap://localhost:3890 -b 'ou=people,DC=proxy,DC=com' -e "<oid>=<value>" '(uid=sam*)' '*'
+```
+The control is specified using it's OID. The OID used for this control is configurable until an official OID is generated. An OID generated from a UUID, e.g., `1.3.6.1.4.1.18060.2.1379319520.35362.17433.40846.265936912329953`, is guaranteed not to collide with existing OIDs.
+
+| Property | Default Value | Description |
+| :--- | :--- | :--- |
+| `gateway.ldap.roles.lookup.bypass.control.oid` | N/A | The OID to use for the bypass control. The control will not be registered if this value is not provided. The Knox LDAP Service will fail to initialize if the value provided is on an OID. |
```
smolnar82: typo: `if the value provided is on an OID.` -> `if the value provided is not an OID.`

smolnar82 (changes requested):

smolnar82:
Hi @handavid !

In addition to the above comments, I see that we have unit test failures now:
```
Error:  Tests run: 16, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.378 s <<< FAILURE! -- in org.apache.knox.gateway.services.ldap.KnoxLDAPServerManagerTest
Error:  org.apache.knox.gateway.services.ldap.KnoxLDAPServerManagerTest.testStartDontRegistersRolesLookupBypassControl -- Time elapsed: 0.166 s <<< FAILURE!
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertFalse(Assert.java:65)
	at org.junit.Assert.assertFalse(Assert.java:75)
	at org.apache.knox.gateway.services.ldap.KnoxLDAPServerManagerTest.testStartDontRegistersRolesLookupBypassControl(KnoxLDAPServerManagerTest.java:415)
```
The good news is that this test may go away, because we now have an official Knox sub-arc under ASF's OID base. See [KNOX-3346](https://issues.apache.org/jira/browse/KNOX-3346) for details.
Given the new KNOX OID base (`1.3.6.1.4.1.18060.18`), we can drop this temporary configuration and add our own as part of this PR. What do you think?

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/LDAPRolesLookupInterceptorFactory.java
handavid: I can get rid of this change since we have an OID.

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlDecoratorTest.java
handavid: minor difference between the tests.
One checks that the `decorator.isBypassRolesLookup` matches the decorated instance values by setting the values of the decorated instance.
The other checks that the decorator can set the values on the decorated instance.
I'll tweak the tests to make this more obvious.

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlDecoratorTest.java
handavid: done

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/LDAPRolesLookupInterceptorFactory.java
handavid: outdated comment because I removed the oid configuration

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlFactoryTest.java
handavid: done

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/interceptor/LDAPRolesLookupInterceptorTest.java
handavid: done

knox-site/docs/service_ldap_server.md
handavid: removed since OID will not be configurable

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/control/RolesLookupBypassControlDecoratorTest.java
smolnar82: Thanks, David!

smolnar82 (approved):

PR #1264, opened 2026-06-15
KNOX-3350: Populate groups in KnoxSSO cookie when configured

smolnar82:
[KNOX-3350](https://issues.apache.org/jira/browse/KNOX-3350) - Add group information into the generated JWT in WebSSOResource

This PR introduces the ability to include group information in the JWT tokens generated by `WebSSOResource` (KnoxSSO). 

 Key changes include:
  * Added a new configuration parameter `knoxsso.token.include.groups` to control whether group information should be included in the issued JWT.
  * Updated `getAuthenticationToken` in `WebSSOResource` to populate the `groups` claim in `JWTokenAttributes` when the feature is enabled.
  * Standardized group retrieval in `TokenResource` by utilizing `SubjectUtils.getCurrentGroupPrincipalNames()`.

## How was this patch tested?

The changes were verified by adding comprehensive unit tests in `WebSSOResourceTest`:
  * `testIncludeGroupsTrue`: Verifies that groups are correctly included in the JWT when `knoxsso.token.include.groups` is set to `true`.
  * `testIncludeGroupsFalse`: Verifies that groups are excluded when the parameter is set to `false`.
  * `testIncludeGroupsOmitted`: Verifies that the default behavior (when the parameter is missing) is to exclude groups.

## Integration Tests
I ran manual testing using my local Knox instance against the `knoxsso` topology. I logged in as `recursiveUser`, verified `hadoop-jwt` (extracted from DEV tools in Chrome) is generated, then checked its content on `jwt.io`:

1. Without `knoxsso.token.include.groups`:
image

2. `knoxsso.token.include.groups = false`:
image

3. `knoxsso.token.include.groups = true`:
image

## UI changes
N/A

### DIFF
--- a/gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
+++ b/gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
@@ -31,6 +31,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 
 import javax.annotation.PostConstruct;
 import javax.servlet.ServletContext;
@@ -51,6 +52,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.knox.gateway.audit.log4j.audit.Log4jAuditor;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
+import org.apache.knox.gateway.security.SubjectUtils;
 import org.apache.knox.gateway.services.GatewayServices;
 import org.apache.knox.gateway.services.ServiceType;
 import org.apache.knox.gateway.services.security.AliasService;
@@ -84,6 +86,7 @@ public class WebSSOResource {
   private static final String SSO_COOKIE_TOKEN_TYPE_PARAM = "knoxsso.token.type";
   private static final String SSO_COOKIE_TOKEN_AUDIENCES_PARAM = "knoxsso.token.audiences";
   private static final String SSO_COOKIE_TOKEN_SIG_ALG = "knoxsso.token.sigalg";
+  private static final String SSO_COOKIE_INCLUDE_GROUPS_PARAM = "knoxsso.token.include.groups";
   private static final String SSO_COOKIE_TOKEN_WHITELIST_PARAM = "knoxsso.redirect.whitelist.regex";
 
   private static final String SSO_SIGNINGKEY_KEYSTORE_NAME = "knoxsso.signingkey.keystore.name";
@@ -111,6 +114,7 @@ public class WebSSOResource {
   private List<String> targetAudiences = new ArrayList<>();
   private boolean enableSession;
   private String signatureAlgorithm;
+  private boolean includeGroups;
   private List<String> ssoExpectedparams = new ArrayList<>();
   private String clusterName;
   private String tokenIssuer;
@@ -224,6 +228,8 @@ public class WebSSOResource {
     }
     final String configuredTokenType = context.getInitParameter(SSO_COOKIE_TOKEN_TYPE_PARAM);
     tokenType = StringUtils.isBlank(configuredTokenType) ? JOSEObjectType.JWT.getType() : configuredTokenType;
+
+    includeGroups = Boolean.parseBoolean(context.getInitParameter(SSO_COOKIE_INCLUDE_GROUPS_PARAM));
   }
 
   @GET
@@ -314,7 +320,9 @@ public class WebSSOResource {
               .setSigningKeystorePassphrase(signingKeystorePassphrase)
               .setManaged(tokenStateService != null)
               .setType(tokenType)
+              .setGroups(groups())
               .build();
+
       JWT token = tokenAuthority.issueToken(jwtAttributes);
 
       // Coverity CID 1327959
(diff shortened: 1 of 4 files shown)


gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
```diff
@@ -224,6 +228,8 @@ private void handleCookieSetup() {
     }
     final String configuredTokenType = context.getInitParameter(SSO_COOKIE_TOKEN_TYPE_PARAM);
     tokenType = StringUtils.isBlank(configuredTokenType) ? JOSEObjectType.JWT.getType() : configuredTokenType;
+
+    includeGroups = Boolean.parseBoolean(context.getInitParameter(SSO_COOKIE_INCLUDE_GROUPS_PARAM));
```
moresandeep, line 232: We should have a default value `false`

moresandeep (changes requested):
@smolnar82 did you check https://issues.apache.org/jira/browse/KNOX-2679? 
AFAIK we already have support for groups but it is not hooked up to JWT. We have a param to turn off groups `pac4j.session.store.exclude.groups`

Looks like this PR includes groups from java Subject. For SSO groups are populated from SAML right? can you elaborate on how this will Subject groups are populated.

I think we should also support hooking up SAML groups here if `pac4j.session.store.exclude.groups=false` (currently they only show up in pac4j profile cookie)

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
smolnar82: It's a primitive `boolean` type field -> defaults to `false`.
The above line also defaults to false: if not `knoxsso.token.include.groups` declared -> the `context.getInitParameter()` will return `null` -> `Boolean.parseBoolean()` will return `FALSE` -> unboxed to `false`.
I added unit tests as well as tested this scenario manually (seePR description).

smolnar82:
@moresandeep - Thanks for your review; I replied back to the default value above.

> Subject. For SSO groups are populated from SAML right? can you elaborate on how this will Subject groups are populated.

Nope. SAML is one way to authenticate. But we do support other authN mechanisms, such as LDAP.
In case of LDAP, Knox needs to be configured with the `HadoopGroupProvider` for group lookup. If that's configured, Knox will place `GroupPrincipal` items in the current Subject during request processing flow. Ultimately, we arrive to the `KNOXSSO` service (`WebSSOResource` in the codebase), which is a terminating-service (i.e. non-proxying), but at this phase the Subject is already decorated and we can read what groups were resolved by Knox.

gateway-service-knoxsso/src/main/java/org/apache/knox/gateway/service/knoxsso/WebSSOResource.java
moresandeep: Cool, I explicitly declare the default value and add it to be in-line with the pattern elsewhere in the code and also helps me in reading in the future. I am okay with this, thanks!

moresandeep:
> @moresandeep - Thanks for your review; I replied back to the default value above.
> 
> > Subject. For SSO groups are populated from SAML right? can you elaborate on how this will Subject groups are populated.
> 
> Nope. SAML is one way to authenticate. But we do support other authN mechanisms, such as LDAP. In case of LDAP, Knox needs to be configured with the `HadoopGroupProvider` for group lookup. If that's configured, Knox will place `GroupPrincipal` items in the current Subject during request processing flow. Ultimately, we arrive to the `KNOXSSO` service (`WebSSOResource` in the codebase), which is a terminating-service (i.e. non-proxying), but at this phase the Subject is already decorated and we can read what groups were resolved by Knox.

Ahh, i see, that makes sense, don't you think we should also support SAML and not just LDAP? IMO we should support a way to fetch groups from SAML too if we are supporting fetching groups from other ways too. There have been internal requests for this.

lmccay (commented):
Please note that the reason we have an option to turn off groups in the cookie is that the cookies get too big and SSO will therefore not work. I don't actually think we should be passing all of the users' group memberships around on every request from the browser and allow the to get stale based on the TTL/expiry of the token and cookie. Also, we shouldn't be getting group memberships from different sources in a deployment. If we get this from the SSO SAML assertion and others are using LDAP Proxy Server then there could be divergence. Overall, I am more -1 on this as a change but if someone wants to use it and they make sure that all group lookup gets the same groups then we can add it but I think that it needs to be disabled by default.

smolnar82:
@lmccay made a point: cookie size can be an issue.

Modern browsers support cookies up to 4K:
image
Which opens the following question: shall we include groups only, and only if, when the cummulated cookie size doesn't exceed 4K?

lmccay:
> @lmccay made a point: cookie size can be an issue.
> 
> Modern browsers support cookies up to 4K: Which opens the following question: shall we include groups only, and only if, when the cummulated cookie size doesn't exceed 4K?

Well, it isn't really clear to me where those groups are even going to be used and what authorization check will see them.
Are we going to change JWTFederationFilter to extract them from the token and set them as GroupPrincipals?

Depending on the specific need, we may be able to handle this in another way.
We already have the ability to add a header for groups to a dispatched request, if this usecase in question here is for a proxied service that wants to get groups from Knox.

moresandeep:
If we decide to go with this PR, we need to add a filter for groups something like `preauth.group.filter.pattern` to prevent issues like [KNOX-2679](https://issues.apache.org/jira/browse/)

smolnar82:
> Depending on the specific need, we may be able to handle this in another way.
We already have the ability to add a header for groups to a dispatched request, if this usecase in question here is for a proxied service that wants to get groups from Knox.

That's a good idea. Let's figure that out...
I'll keep you posted.

smolnar82:
I'm closing this one for now based on the above points made by @lmccay and @moresandeep .

PR #1270, opened 2026-06-18
KNOX-3355 - Add TrustedOidcIssuerService schema and interface

hsheinblatt:
KNOX-3355 - Add TrustedOidcIssuerService schema and interface

Feature context:
Support OIDC JWKS discovery to validate tokens from external authorizers. Currently,
there is a fixed trusted JWKS list created as part of initial configuration that can be used
to validate tokens. Allow the dynamic registration of trusted issuers with a flag that they are
trusted for OIDC JWKS discovery. Tokens will be validated with discovered JWKS from
such trusted issuers.

An initial detailed use case is to validate kubernetes service account tokens from
external kubernetes clusters. When the kubernetes cluster is created or integrated,
the kubernetes issuer can be registered as trusted for JWKS discovery without knox
service disruption. Then, service account tokens from that cluster can be validated
using OIDC discovery for JWKS.

The implementation plan is bottom up: first the storage and interface for the trusted
issuers, then the implementation of registration through an admin API in knox idf, and
then the token validation integration. It will be part of knox_idf initially, but written to
be portable so that it can be moved into a different or separate service if more
OIDC discovery use cases emerge.

How this patch was tested:
New unit tests are added.

Follow on changes will implement the interface and then integrate in the implementation
into KnoxIDF flows, but now there is no integration or UI changes.


lmccay:
Hi @hsheinblatt - thank you for this initial contribution. Can you update the description to include some more context for this aspect of the larger feature? I believe that you are introducing the trusted issuer here as an extra qualifier to a given service account SVID/JWT to facilitate possible cross cluster interaction, is this correct?

Please add a description that articulates this need to disambiguate such credentials and for what use cases it is required.

lmccay (changes requested):
I'll leave the database changes for @smolnar82 or others to review in depth. Otherwise, it looks fine. I made a comment on the PR itself asking for more context in the description. Once that is done, you have a +1 from me but we will wait for someone more familiar with the schema layer and conventions to review there.

hsheinblatt:
I've updated the description with more detail on the planned feature, please review.

lmccay (approved):
+1 for the overall approach and detailed description.
I've reviewed the interface which looks fine and the tests.
Again, I'll defer db schema related changes to review by @smolnar82

gateway-server/src/main/resources/conf/gateway-site.xml
```diff
@@ -218,4 +218,53 @@ limitations under the License.
     </property>
     -->
 
+    <!-- KnoxIDF Trusted OIDC Issuer Service Configuration
```
smolnar82, line 221: I believe these configuration should be part of the feature doc (I even owe you proper documents in `knox-site`). It's ok to have them here for now, but we'll move them away to docs later.

gateway-server/src/main/resources/createKnoxIDFTrustedOidcIssuersTableDerby.sql
```diff
@@ -0,0 +1,22 @@
...
+--  License for the specific language governing permissions and limitations under
+--  the License.
+
+CREATE TABLE TRUSTED_OIDC_ISSUERS (
+    issuer_url    VARCHAR(2048) PRIMARY KEY,
+    dynamic_jwks  BOOLEAN DEFAULT false,
+    cluster_name  VARCHAR(256),
+    registered_at TIMESTAMP,
+    registered_by VARCHAR(2048)
+)
```
smolnar82, line 22: Missing `NOT NULL` constraints (see other DB type declarations).

gateway-spi/src/main/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/TrustedOidcIssuer.java
```diff
@@ -0,0 +1,63 @@
...
+    return registeredAt;
+  }
+
+  /**
+   * @return the identity that registered this issuer, or null if not recorded
+   */
+  public String getRegisteredBy() {
+    return registeredBy;
+  }
+}
```
smolnar82, line 63: Is there a case when we allow this to be `null`?

smolnar82 (changes requested):

gateway-server/src/main/resources/createKnoxIDFTrustedOidcIssuersTableOracle.sql
```diff
@@ -0,0 +1,23 @@
...
+--  Unless required by applicable law or agreed to in writing, software
+--  distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+--  WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+--  License for the specific language governing permissions and limitations under
+--  the License.
+
+CREATE TABLE TRUSTED_OIDC_ISSUERS (
+    issuer_url    VARCHAR2(2048) NOT NULL,
+    dynamic_jwks  NUMBER(1)      DEFAULT 0 NOT NULL,
+    cluster_name  VARCHAR2(256),
```
smolnar82, line 19: I'd move `cluster_name` to the end, this seems to be the only `NOT NULL` field (see my question about `registered_by`).

smolnar82 (changes requested):

gateway-server/src/main/resources/conf/gateway-site.xml
hsheinblatt: Filed https://issues.apache.org/jira/browse/KNOX-3384 to track this task.

gateway-server/src/main/resources/createKnoxIDFTrustedOidcIssuersTableDerby.sql
hsheinblatt: Added, but in a new PR.

gateway-server/src/main/resources/createKnoxIDFTrustedOidcIssuersTableOracle.sql
hsheinblatt: Done, but in a new PR.

gateway-spi/src/main/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/TrustedOidcIssuer.java
hsheinblatt: Yes. We try to pull the principal name from the request, but it may not be defined. It's possible that practically it should always be or you've configured security poorly, but in general I don't think we can guarantee the API all would fail if it's not.

hsheinblatt (commented):
Responded to comments here, but I'll put the actual changes in a new PR and close this one. The feature branch has been force pushed, so it's easier to create a new PR to merge.

smolnar82:
Closing this one in favor of #1311

PR #1272, opened 2026-06-19
KNOX-3356: Allow Cloudera Manager service discovery over cleartext HTTP

smolnar82:
[KNOX-3356](https://issues.apache.org/jira/browse/KNOX-3356) - Allow Cloudera Manager service discovery over cleartext HTTP

When Knox is configured with TLS but the target Cloudera Manager server is not, CM service discovery fails with:
```
com.cloudera.api.swagger.client.ApiException: java.net.UnknownServiceException: CLEARTEXT communication not enabled for client
```

The root cause is in `DiscoveryApiClient.configureSsl()`: it unconditionally replaced the OkHttp client's `connectionSpecs` with a single TLS-only spec (`ConnectionSpec.MODERN_TLS`), regardless of the discovery address scheme. OkHttp matches the request URL's scheme against the allowed connection specs, so an `http://` discovery address with no `CLEARTEXT` spec is rejected before any request is sent.

This PR makes TLS configuration conditional on the discovery address actually being HTTPS:
  - Added an `isSecure()` helper that checks whether the configured base path starts with `https:`.
  - `configureSsl()` now returns early (with a DEBUG log) for cleartext addresses, leaving OkHttp's default connection specs,  which include `CLEARTEXT`,  in place.
  - Added the `skippingSslConfigurationForCleartextAddress` discovery message.

  Behavior for HTTPS discovery addresses is unchanged.
## How was this patch tested?
Automated unit tests in `ClouderaManagerServiceDiscoveryTest`:

  - Corrected `testApiClientInterceptorsWhenKerberosIsDisabledAndPasswordIsNotSet` to use an
    HTTPS discovery address (it previously used an HTTP address while asserting a TLS-only spec,
    i.e. it asserted the buggy behavior); it still verifies the configured cipher/protocol are
    applied on the HTTPS path.
  - Added `testApiClientAllowsCleartextForHttpDiscoveryAddress`, which uses an `http://` address
    and asserts the client retains a `CLEARTEXT`-capable connection spec.

  Both tests pass: `Tests run: 2, Failures: 0, Errors: 0, Skipped: 0`.

## Integration Tests
N/A

## UI changes
N/A

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/ClouderaManagerServiceDiscoveryMessages.java
@@ -280,6 +280,9 @@ public interface ClouderaManagerServiceDiscoveryMessages {
   @Message(level = MessageLevel.ERROR, text = "Failed to configure truststore")
   void failedToConfigureTruststore();
 
+  @Message(level = MessageLevel.DEBUG, text = "Skipping SSL configuration for cleartext CM discovery address {0}")
+  void skippingSslConfigurationForCleartextAddress(String address);
+
   @Message(level = MessageLevel.DEBUG, text = "Looking up cluster services from service discovery repository...")
   void lookupClusterServicesFromRepository();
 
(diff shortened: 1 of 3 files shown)


pzampino:
Why are there CM deployments without TLS?

pzampino (approved):

smolnar82:
> Why are there CM deployments without TLS?

This is a really good question!

PR #1274, opened 2026-06-21
KNOX-3341: LDAP proxy backend handles general searches

handavid:

[KNOX-3341](https://issues.apache.org/jira/browse/KNOX-3341) - LDAP Proxy handles general search requests

This commit improves the behavior of the LdapProxyBackend to handle broader search requests such as retrieving all users, all groups, or filtering by attributes other than uid.

A "search" method was added to the LdapBackend interface to support broader search requests.

LdapProxyBackend implements this method by converting the search base, objectclass, and user identifier attribute from proxy values to values recognized by the remote LDAP backend. LdapProxyBackend.createProxyEntry was factored out into a new RemoteSchemaConverter class. This class also contains methods for converting the search filter and DNs. The conversions in the search filter are further supported by a new FilterMappingVisitor class which will traverse the ExprNode tree and replace values as needed.
Result entries are likewise converted from the remote values to the proxy values. As part of the mapping, the AD 'userAccountControl' attribute type is mapped into the 'nsAccountLock' attribute type. A new DisabledUserInterceptor is implemented to perform this conversion. This interceptor can also be configured to remove disabled entries from the results.

Group membership retrieval is modified to use 'getUserGroupEntries' as a common starting point then branch into separate codepaths for using 'memberOf' or not.

FileBackend simply maps the filter back into user search.

## How was this patch tested?

Unit tests were added.
The new features were manually exercised using commands such as:
```
ldapsearch -v -x -H ldap://localhost:3890 -b 'ou=people,DC=proxy,DC=com' '(&(objectclass=inetorgperson)(uid=test*))'
ldapsearch -v -x -H ldap://localhost:3890 -b 'ou=groups,DC=proxy,DC=com' '(objectclass=groupofnames)'
```

## Integration Tests
(Please add or update integration tests [`.github/workflows/tests`](.github/workflows/tests) for the feature you are adding. If no unit test is added, please explain why. Check out [`.github/workflows/tests/README.md`](./workflows/tests/README.md) for instructions)

## UI changes
no UI changes

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/FileBackend.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/FileBackend.java
@@ -20,6 +20,7 @@ package org.apache.knox.gateway.services.ldap.backend;
 import com.google.gson.Gson;
 import org.apache.directory.api.ldap.model.entry.DefaultEntry;
 import org.apache.directory.api.ldap.model.entry.Entry;
+import org.apache.directory.api.ldap.model.message.SearchScope;
 import org.apache.directory.api.ldap.model.name.Dn;
 import org.apache.directory.api.ldap.model.schema.SchemaManager;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
@@ -33,13 +34,20 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * File-based backend that reads user/group data from JSON
  */
 public class FileBackend implements LdapBackend {
     private static final LdapMessages LOG = MessagesFactory.get(LdapMessages.class);
+
+    private static final Pattern UID_PATTERN = Pattern.compile(".*\\(uid=([^)]+)\\).*");
+    private static final Pattern CN_PATTERN = Pattern.compile(".*\\(cn=([^)]+)\\).*");
+    private static final Pattern SAMAACCOUNTNAME_PATTERN = Pattern.compile(".*\\(sAMAccountName=([^)]+)\\).*");
 
     static final String TYPE = "file";
 
@@ -136,7 +144,7 @@ public class FileBackend implements LdapBackend {
     }
 
     @Override
-    public List<String> getUserGroups(String username) throws Exception {
+    public List<String> getUserGroups(String username, SchemaManager schemaManager) throws Exception {
         UserData userData = users.get(username);
         return userData != null && userData.groups != null ? userData.groups : Collections.emptyList();
     }
@@ -145,9 +153,13 @@ public class FileBackend implements LdapBackend {
     public List<Entry> searchUsers(String filter, SchemaManager schemaManager) throws Exception {
         List<Entry> results = new ArrayList<>();
 
+        String userFilter = extractUser(filter).toLowerCase(Locale.ROOT);
+
         // Simple filter matching - just check if username matches
         for (String username : users.keySet()) {
-            if (filter.contains("uid=" + username) || filter.contains("*")) {
+            String usernameLowerCase = username.toLowerCase(Locale.ROOT);
+            if (userFilter.equalsIgnoreCase(usernameLowerCase) ||
+                    (userFilter.contains("*") && userFilter.contains(usernameLowerCase))) {
                 Entry entry = getUser(username, schemaManager);
                 if (entry != null) {
                     results.add(entry);
@@ -169,5 +181,29 @@ public class FileBackend implements LdapBackend {
         }
 
         return false;
+    }
+
+    @Override
+    public List<Entry> search(String searchBase, SearchScope searchScope, String filter, SchemaManager schemaManager) throws Exception {
+        return searchUsers(filter, schemaManager);
+    }
+
+    private String extractUser(String filter) {
+        Matcher uidMatcher = UID_PATTERN.matcher(filter);
+        if (uidMatcher.matches()) {
+            return uidMatcher.group(1);
+        }
+
+        Matcher cnMatcher = CN_PATTERN.matcher(filter);
+        if (cnMatcher.matches()) {
+            return cnMatcher.group(1);
+        }
+
+        Matcher samaaccountnameMatcher = SAMAACCOUNTNAME_PATTERN.matcher(filter);
+        if (samaaccountnameMatcher.matches()) {
+            return samaaccountnameMatcher.group(1);
+        }
+
+        return null;
     }
 }
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
@@ -60,27 +60,31 @@ public class LdapProxyBackend implements LdapBackend {
     private String ldapUrl;
     private String bindDn;
     private String bindPassword;
-    private String userSearchBase;
-    private String groupSearchBase;
-    private String proxyBaseDn;  // Base DN for proxy entries (e.g., dc=proxy,dc=com)
-    private String remoteBaseDn;  // Base DN for remote server searches (e.g., dc=hadoop,dc=apache,dc=org)
     private int port;
     private String host;
 
+    private RemoteSchemaConverter remoteSchemaConverter;
+
+    // Proxy configuration
+    private String proxyBaseDn;  // Base DN for proxy entries (e.g., dc=proxy,dc=com)
+    private String proxyUserSearchBase;
+    private String proxyGroupSearchBase;
+
+    // Backend configuration
+    private String remoteBaseDn;  // Base DN for remote server searches (e.g., dc=hadoop,dc=apache,dc=org)
+    private String remoteUserSearchBase;
+    private String remoteGroupSearchBase;
+
     // Configurable attributes for AD/LDAP compatibility
-    private String userIdentifierAttribute = "uid"; // uid for LDAP, sAMAccountName for AD
+    private String remoteUserIdentifierAttribute = "uid"; // uid for LDAP, sAMAccountName for AD
     private String userSearchFilter = "({userIdAttr}={username})"; // Will be populated with userIdentifierAttribute
-    private String groupMemberAttribute = "memberUid"; // member for AD, memberUid for POSIX
+    private String remoteGroupMemberAttribute = "memberUid"; // member for AD, memberUid for POSIX
+    private String remoteUserObjectClass = "inetOrgPerson"; // user for AD, inetOrgPerson otherwise
+    private String remoteGroupObjectClass = "groupofnames"; // group for AD, groupofnames otherwise
     private boolean useMemberOf; // Use memberOf attribute for group lookup (efficient for AD)
     private boolean recursiveGroupResolution;
     private int recursiveGroupResolutionMaxDepth;
 
-    private List<String> proxyEntryAttributeTypes = List.of(
-            // "uid" will always be filled
-            "cn",
-            "dn",
-            "mail",
-            "description");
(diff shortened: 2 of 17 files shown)


handavid:
@smolnar82

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/FileBackend.java
```diff
@@ -145,9 +153,13 @@ public List<String> getUserGroups(String username) throws Exception {
     public List<Entry> searchUsers(String filter, SchemaManager schemaManager) throws Exception {
         List<Entry> results = new ArrayList<>();
 
+        String userFilter = extractUser(filter).toLowerCase(Locale.ROOT);
```
smolnar82, line 156: Possible NPE here: `extractUser` returns `null` when the filter has no `uid=/cn=/sAMAccountName=` term.
I'm aware that the file-based backend implementation is for tests only, but I'd like this to fix.

Analysis by Claude code:
```
extractUser returns null when the filter has no uid=/cn=/sAMAccountName= term (e.g. (objectclass=inetOrgPerson)). Since the new FileBackend.search() delegates every filter to searchUsers, and the new UserSearchInterceptor now forwards all searches to the backend, a general search against a file backend throws NPE. 
It's swallowed by UserSearchInterceptor's catch (Exception e) and logged as a search failure, so the file backend silently returns nothing for exactly the "general searches" this PR is about. Should guard the null (return empty or fall back to wildcard match).
```

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
```diff
@@ -302,10 +326,12 @@ public Entry getUser(String username, SchemaManager schemaManager) throws Except
             connection = getConnection();
             // Search for user using configurable attribute
             String filter = userSearchFilter.replace("{username}", username);
-            try (EntryCursor cursor = connection.search(userSearchBase, filter, SearchScope.SUBTREE, "*")) {
+            try (EntryCursor cursor = connection.search(remoteUserSearchBase, filter, SearchScope.SUBTREE, "*")) {
                 if (cursor.next()) {
                     Entry sourceEntry = cursor.get();
-                    return createProxyEntry(sourceEntry, username, connection, schemaManager, createResolvedParentsCache());
+                    addGroupMemberships(sourceEntry, connection, createEntryCache(), createResolvedParentsCache());
+                    return remoteSchemaConverter.convertRemoteEntryToProxyEntry(sourceEntry, schemaManager);
```
smolnar82, line 333: Attribute copying changed from an allowlist to copy-everything. 
Assessed by Claude:
```
  The old createProxyEntry copied a fixed set (cn, dn, mail, description, uid, objectClass, identifier, memberOf). The new RemoteSchemaConverter.convertRemoteEntryToProxyEntry copies all attributes from the remote entry. 

Two concerns:
  - Potential exposure of sensitive/operational attributes (e.g. userPassword) through the proxy that previously weren't surfaced. 
Please confirm the remote search ("*") can't return those, or re-introduce an allowlist/denylist. 
  - copyAttribute runs convertRemoteDnToProxyDn on every attribute value, not just DN-valued ones. 
A description/mail value that happens to contain the remote base DN substring would be rewritten. Low probability, but it's blind string rewriting.
```

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/RemoteSchemaConverter.java
```diff
@@ -0,0 +1,177 @@
...
+            }
+        }
+    }
+
+    /**
+     * Converts a remote dn string to a proxy dn string. This also works for search base strings.
+     * @param string the remote dn or search base
+     * @return the proxy search base
+     */
+    public String convertRemoteDnToProxyDn(String string) {
```
smolnar82, line 157: DN/filter conversion via String.replaceAll is regex-fragile.
Assessed by Claude:
```
convertRemoteDnToProxyDn/convertProxyDnToRemoteDn use the base DNs as regex patterns and the targets as replacement strings (where $/\ are special). 
Standard DNs are safe, but this will misbehave if a DN contains regex meta-characters.

Pattern.quote + Matcher.quoteReplacement (or proper DN parsing) would be more robust. Same fragility in convertProxyFilterToRemoteFilter's filter.replaceAll(":\\[.*?\\]", "")... cleanup, which is a bit of a hack (acknowledged in the comment).
```

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
```diff
@@ -540,27 +683,16 @@ private String extractGroupNameFromDn(String groupDn) {
         return null;
     }
 
-    private List<Entry> getUserGroupsInternal(LdapConnection connection, String userDn, String username) throws LdapException, CursorException, IOException {
+    // tODO reuse this method in recursive calls
```
smolnar82, line 686: nit: typo

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/DisabledUserInterceptor.java
```diff
@@ -0,0 +1,124 @@
...
+                // rethrow exception on incomplete iteration
+                throw new LdapException(e);
+            }
+            filteredEntries = filterDisabledUsers(originalEntries);
+        } catch (IOException e) {
+            // IOException would only occur after finishing iterating over results
+            // we can ignore this exception and return the filtered entries
+        } catch (Exception e) {
+            throw e;
+        }
```
smolnar82, line 76: This catch does nothing, we can remove it.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/RemoteSchemaConverter.java
```diff
@@ -0,0 +1,177 @@
...
+        if (!"uid".equals(remoteUserIdentifierAttribute)) {
+            Attribute idAttr = sourceEntry.get(remoteUserIdentifierAttribute);
+            if (idAttr != null) {
+                entry.add("uid", idAttr.getString());
+            }
+        }
+
+        // replace userObjectClass and groupObjectClass object classes
+        Attribute objectClassAttribute = sourceEntry.get("objectclass");
+        if (objectClassAttribute.contains(remoteGroupObjectClass)) {
```
smolnar82, line 98: Possible NPE if `sourceEntry.get("objectclass")` returns `null`.

gateway-server/src/test/resources/ldap-recursive-test.ldif
```diff
@@ -93,3 +105,77 @@ objectclass:top
...
+cn: memberOflevel1
+memberOf: cn=memberOflevel2,ou=recursiveMemberOfGroups,dc=hadoop,dc=apache,dc=org
+member: uid=memberOfUser,ou=recursiveMemberOfPeople,dc=hadoop,dc=apache,dc=org
+member: uid=memberOfUser2,ou=recursiveMemberOfPeople,dc=hadoop,dc=apache,dc=org
+
+# memberOf Level 2 Group (Member is Level 1)
+dn: cn=memberOflevel2,ou=recursiveMemberOfGroups,dc=hadoop,dc=apache,dc=org
+objectclass:top
+objectclass: groupofnames
+cn: memberOflevel2level2
```
smolnar82, line 146: Wrong `cn`: it should be `cn: memberOflevel2` (otherwise it'll be a mismatched RDN).

gateway-server/src/test/resources/ldap-recursive-test.ldif
```diff
@@ -93,3 +105,77 @@ objectclass:top
...
+memberOf: cn=memberOflevel3,ou=recursiveMemberOfGroups,dc=hadoop,dc=apache,dc=org
+memberOf: cn=memberOfCycleA,ou=recursiveMemberOfGroups,dc=hadoop,dc=apache,dc=org
+member: cn=memberOflevel1,ou=recursiveGroups,dc=hadoop,dc=apache,dc=org
```
smolnar82, line 149: points to `ou=recursiveGroups` instead of `ou=recursiveMemberOfGroups`.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
```diff
@@ -315,72 +341,182 @@ public Entry getUser(String username, SchemaManager schemaManager) throws Except
     }
 
     @Override
-    public List<String> getUserGroups(String username) throws Exception {
+    public List<String> getUserGroups(String username, SchemaManager schemaManager) throws Exception {
```
smolnar82, line 344: Claude's assessment:
```
`getUserGroups()` with `useMemberOf=true` + `recursiveGroupResolution=true` likely returns an incomplete CN list.

In the recursive memberOf path, populateParentCacheViaMemberOf fetches group entries via connection.lookup(groupDn, "memberOf") — requesting only memberOf, so the cached entries have no cn. Those entries flow back through getUserGroupsEntries

→ getCnsFromEntries, which does entry.get("cn") and drops anything without a cn. So the recursively-discovered (and even the directly-resolved) groups get dropped from the returned name list.

This path has no test — all the new recursive-memberOf tests go through search() and assert on the memberOf attribute (DN strings via addGroupMemberships), not on getUserGroups() CNs. Fix: request "cn", "memberOf" in the lookup, or derive the CN from the DN.
  
```

gateway-server/src/test/resources/ldap-recursive-test.ldif
```diff
@@ -93,3 +105,77 @@ objectclass:top
...
+
```
smolnar82, line 150: These 2 issues above don't fail the tests (the `memberOf` recursion reads the `memberOf` attribute, not `cn/member`), which is precisely why issue with `getUserGroups()` above slipped through untested. I'd fix the data and add a `getUserGroups()` assertion for the `useMemberOf+recursive` combo.

knox-site/docs/service_ldap_server.md
```diff
@@ -120,23 +128,27 @@ The proxy backend delegates lookups to a remote LDAP or Active Directory server.
...
 | `gateway.ldap.interceptor.<name>.groupSearchBase` | `ou=groups,{remoteBaseDn}` | Base DN for group searches on the remote server. |
 | `gateway.ldap.interceptor.<name>.userIdentifierAttribute` | `uid` | Attribute used for user lookup (e.g., `sAMAccountName` for AD). |
+| `gateway.ldap.interceptor.<name>.userObjectClass` | `inetOrgPerson` | Objectclass for identifying users in remote server (e.g., `person` for AD). |
+| `gateway.ldap.interceptor.<name>.groupObjectClass` | `groupOfNames` | Objectclass for identifying groups in remote server (e.g., `group` for AD). |
 | `gateway.ldap.interceptor.<name>.groupMemberAttribute` | `memberUid` | Attribute used for group membership (e.g., `member` for AD). |
 | `gateway.ldap.interceptor.<name>.useMemberOf` | `false` | If `true`, use the `memberOf` attribute for efficient group lookups. |
-| `gateway.interceptor.<name>.proxy.poolMaxActive` | `8` | Maximum number of active connections in the pool. |
+| `gateway.ldap.interceptor.<name>.proxy.poolMaxActive` | `8` | Maximum number of active connections in the pool. |
+| `gateway.ldap.interceptor.<name>.proxy.poolMaxActive` | `8` | Maximum number of active connections in the pool. |
```
smolnar82, line 143: Duplicated row -> has to be eliminated.

smolnar82 (approved):
Nice generalization of the proxy backend: the `RemoteSchemaConverter/FilterMappingVisitor` split is clean and the new `search()` test coverage is solid. 

My main concerns are a couple of correctness gaps (`FileBackend.searchUsers` NPEs on non-user filters, and `getUserGroups()` with `useMemberOf+recursive` drops group CNs) plus the shift from an attribute allowlist to copy-everything; details inline.

As discussed between us, I'll merge this PR and come up with another one that addresses those comments of mine.

PR #1279, opened 2026-06-25
KNOX-3360: create-k8s-alias command to save aliases from k8s secrets

hanicz:
[KNOX-3360](https://issues.apache.org/jira/browse/KNOX-3360) - New KnoxCLI command to save alias from k8s secret

- New KnoxCLI command `create-k8s-alias` that adds k8s secrets as aliases 
- It can take multiple secret names at the same time, optional `--ns` or `--namespace` flag to specify where the secrets live, it defaults to `knox`
- The secret has to have the below structure, topology is optional and it defaults to `__gateway`
```
  alias.name: ...
  topology: ...
  alias.value: ...
```

## How was this patch tested?

Unit tests, tested on local kind cluster

Secrets

```
apiVersion: v1
kind: Secret
metadata:
  name: stringdatasecret
  namespace: knox
type: Opaque
stringData:
  alias.name: my-alias2
  topology: sandbox
  alias.value: s3cr3t
---
apiVersion: v1
kind: Secret
metadata:
  name: datasecret
  namespace: knox
type: Opaque
data:
  alias.name: bXktYWxpYXM=
  alias.value: czNjcjN0
```

`./knoxcli.sh create-k8s-alias stringdatasecret datasecret`

## Integration Tests
N/A

## UI changes
N/A

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
@@ -1067,7 +1067,9 @@ public class KnoxCLI extends Configured implements Tool {
                                    + "Every Secret must contain 'alias.name' (the alias name)\n"
                                    + "and 'alias.value' (the secret value); 'topology' is optional\n"
                                    + "and defaults to the gateway-level credential store ('__gateway').\n"
-                                   + "Uses in-cluster Kubernetes config.";
+                                   + "All Secrets are fetched and validated before any alias is\n"
+                                   + "written, so a failure in the batch leaves the credential\n"
+                                   + "store untouched. Uses in-cluster Kubernetes config.";
 
    private static final String DEFAULT_NAMESPACE = "knox";
    private static final String ENTRY_NAME = "alias.name";
@@ -1085,6 +1087,7 @@ public class KnoxCLI extends Configured implements Tool {
    public void execute() throws Exception {
      AliasService as = getAliasService();
      String ns = (namespace == null || namespace.isEmpty()) ? DEFAULT_NAMESPACE : namespace;
+     List<ParsedAlias> parsed = new ArrayList<>(secretNames.size());
      try (KubernetesClient client = buildKubernetesClient()) {
        for (String secretName : secretNames) {
          Secret secret = client.secrets().inNamespace(ns).withName(secretName).get();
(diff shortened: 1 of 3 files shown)


gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -272,6 +282,16 @@ private int init(String[] args) throws IOException {
           printKnoxShellUsage();
           return -1;
         }
+      } else if (args[i].equals("create-k8s-alias")) {
```
moresandeep, line 285: Not a fan of hard coding but looks like this is the pattern we use so that's fine.

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -1029,6 +1058,77 @@ public String getUsage() {
...
+     String entry = optionalEntry(secret, entryKey);
+     if (entry == null || entry.isEmpty()) {
+       throw new IllegalStateException(
+           "Secret '" + secretName + "' is missing required entry '" + entryKey + "'.");
+     }
+     return entry;
+   }
+
+   private String optionalEntry(Secret secret, String entryKey) {
+     if (secret.getData() != null && secret.getData().containsKey(entryKey)) {
```
moresandeep, line 1126: This could possibly be a bug.
PR description shows a stringData example
```
type: Opaque
stringData:
  alias.name: my-alias2
  topology: sandbox
  alias.value: s3cr3t
```
but optionalEntry only reads `secret.getData():` Looking at Secrets docs and example [1], [2]
should the yaml file have `data`? also, looks like there is `stringData` option too so we should be handeling that too, something like 

```
if (secret.getStringData() != null && secret.getStringData().containsKey(entryKey)) {
  return secret.getStringData().get(entryKey);
}
// then fall back to getData() 
```


[1] https://github.com/fabric8io/kubernetes-client/blob/main/doc/CHEATSHEET.md#secret
[2] https://github.com/fabric8io/kubernetes-client/blob/main/kubernetes-itests/src/test/resources/secret-it.yml#L23

moresandeep (changes requested):
Few things:
1. I added my observation regarding `secret.getData()` on the review the other two below are minor nits.
2. in case of batch/bulk mode if `secret-a` succeeds and `secret-b` fails looks like secret-a’s alias will be persisted with no rollback, is this expected? 
3. I think we'll need a DOC jira for this :)

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
moresandeep: Please ignore this.

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
hanicz: When the secret gets created k8s api converts stringData into data and base64 encodes it.

hanicz:
> Few things:
> 
> 1. I added my observation regarding `secret.getData()` on the review the other two below are minor nits.
> 2. in case of batch/bulk mode if `secret-a` succeeds and `secret-b` fails looks like secret-a’s alias will be persisted with no rollback, is this expected?
> 3. I think we'll need a DOC jira for this :)
@moresandeep
1. As explained in my response the secret is converted into data and base64 encoded so I think the solution is correct. Tested with both stringData and data.
2. I added a secret validation.
3. I'll open a DOC jira.

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
moresandeep: Okay, as long as this is understood and verified I am good with it. thanks!

moresandeep (approved):
Looks good, thanks!

PR #1282, opened 2026-06-26
KNOX-3364: New loadAliasesFromK8sSecrets method in entrypoint.sh to l…

hanicz:
…oad aliases from k8s secrets. Install jq in the container as well.

[KNOX-3364](https://issues.apache.org/jira/browse/KNOX-3364) - Bootstrap Knox aliases from labeled Kubernetes Secrets on container startup

Helper function to load Knox aliases from labeled Kubernetes Secrets. Uses the pod's mounted ServiceAccount token to list Secrets in the pod's namespace that carry KNOX_ALIAS_LABEL and feeds the names to `knoxcli create-k8s-alias`, so aliases are restored on every pod restart. Skips silently when no ServiceAccount token is mounted (non-k8s runs); logs a warning and continues on API errors so Knox startup isn't blocked by transient cluster issues. Requires: jq in the image; the pod's ServiceAccount must have list on Secrets in its namespace.

- Installs jq on the container

## How was this patch tested?

Tested on a local kind cluster, tested by just running without k8s, tested with no secrets, tested with different labels

Secrets:

```
---
apiVersion: v1
kind: Secret
metadata:
  name: stringdatasecret
  namespace: knox
  labels:
    knox.apache.org/alias: "true"
type: Opaque
stringData:
  alias.name: my-alias2
  topology: sandbox
  alias.value: s3cr3t
---
apiVersion: v1
kind: Secret
metadata:
  name: datasecret
  namespace: knox
  labels:
    knox.apache.org/alias: "true"
type: Opaque
data:
  alias.name: bXktYWxpYXM=
  alias.value: czNjcjN0
```

`./bin/knoxcli.sh list-alias --cluster __gateway,sandbox`

### DIFF
--- a/gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
+++ b/gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
@@ -35,6 +35,8 @@
 #   TRUSTSTORE_IMPORTS="myRootCA:/mountedpath/enterprise_root_cert.pem myBizPartnerCA:/mountedpath/mybiz_partner_cert.pem"
 #   Description:
 #   This will import the specified certificates into the truststore JKS with the provided alias name(s) during the container startup process.
+# - KNOX_ALIAS_LABEL - (optional) label selector used to discover Knox alias
+#   Secrets in the pod's namespace, default 'knox.apache.org/alias=true'.
 
 
 set -e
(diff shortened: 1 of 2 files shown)


gateway-docker/src/main/resources/docker/gateway-entrypoint.sh
```diff
@@ -92,6 +94,49 @@ saveAlias() {
...
+  local namespace token resp_file http_code secret_names
+  namespace=$(/bin/cat "${sa_ns_file}")
+  token=$(/bin/cat "${sa_token_file}")
+  resp_file=$(mktemp)
+
+  http_code=$(curl -sS \
+    --cacert "${sa_ca_file}" \
+    -H "Authorization: Bearer ${token}" \
+    -o "${resp_file}" -w "%{http_code}" \
+    "https://kubernetes.default.svc/api/v1/namespaces/${namespace}/secrets?labelSelector=${label}") || http_code="000"
```
smolnar82, line 119: nit: there's no explicit timeout on the curl call. If the k8s API is slow or unreachable, this could stall container startup. Adding `--max-time 10` or similar would make the failure path faster.

smolnar82 (approved):
Just a minor observation on the `curl` command, it's not a blocker...

PR #1291, opened 2026-07-01
KNOX-3359 - Support Single-Purpose EKU Certificates

moresandeep:

[KNOX-3359](https://issues.apache.org/jira/browse/KNOX-3359) - Support Single-Purpose EKU Certificates

Public CAs are retiring dual purpose certificates that carry both serverAuth and clientAuth Extended Key Usages. This change allows Knox to use two separate keystores, one for its inbound TLS server identity (serverAuth only) and one for its outbound mTLS client identity (clientAuth only).

- `gateway.tls.single.eku.enabled` — new toggle that activates single-EKU mode. Off by default all existing behavior is unchanged when it is off.
- GatewayConfigImpl / GatewayConfig — new config accessors for the HTTP client keystore path, type, alias, and truststore, startup validation (validateSingleEkuConfig) that fails fast if any prerequisite is missing or misconfigured (wrong alias type, wrong EKU, missing truststore). Knox refuses to start rather than silently falling back to the server identity.
- DefaultKeystoreService.getKeystoreForHttpClient() / DefaultAliasService.getHttpClientKeyPassphrase() — load the dedicated client-identity keystore using the configured path, type, and alias.
- DefaultHttpClientFactory.createSSLContext() — when single-EKU is on and two-way SSL is active, loads the client keystore instead of the gateway server keystore.
- JettySSLService - enforce that the server keystore alias carries only serverAuth in single-EKU mode.
- `gateway.httpclient.twoWaySsl.enabled` — new global flag to activate outbound mTLS across all dispatches without requiring use-two-way-ssl="true" on every service definition. Automatically defaults to true when `gateway.tls.single.eku.enabled=true`, so enabling single-EKU mode is a single config change with sane defaults.

## Notes
Note that this feature is independent of mTLS. mTLS can be configured using the existing properties 

* gateway.client.auth.needed | default false | comments: INBOUND mTLS
* gateway.httpclient.twoWaySsl.enabled | default false | comments: OUTBOUND mTLS

### Unit Tests
- Unit tests added for GatewayConfigImpl, DefaultKeystoreService, DefaultAliasService, JettySSLService, and DefaultHttpClientFactory covering the single-EKU happy path, wrong-passphrase fail-closed, one-way SSL guard (no client cert presented), and the global flag OR-ing with the per-dispatch parameter.

### Manual Tests
Tested the following scenarios 

#### Happy path: clientAuth cert + basic auth → 200, backend sees knox-client
```
curl -sk --cert curl-client.crt --key curl-client.key -u guest:guest-password https://localhost:8443/gateway/sandbox/testmtls/hello
```
```
{"service":"test-https-backend","path":"/hello?user.name=guest","authenticated_client_cn":"knox-client","message":"mTLS OK -- backend saw your client certificate"}
```

####  No client cert → inbound TLS handshake rejected (client.auth.needed=true)

```
curl -iku guest:guest-password https://localhost:8443/gateway/sandbox/testmtls/hello-
```
```
curl: (56) LibreSSL SSL_read: LibreSSL/3.3.6: error:1404C412:SSL 
routines:ST_OK:sslv3 alert bad certificate, errno 0
```

####  serverAuth-ONLY cert presented as inbound client cert → rejected (EKU check)

```
curl -ik --cert knox-server.crt --key knox-server.key -u guest:guest-password https://localhost:8443/gateway/sandbox/testmtls/hello
```
```
curl: (55) Send failure: Broken pipe
```

####  Valid client cert but NO credentials → TLS ok, HTTP 401 from Knox'

```
curl -ik --cert curl-client.crt --key curl-client.key https://localhost:8443/gateway/sandbox/testmtls/hello
```
```
HTTP/1.1 401 Unauthorized
Date: Wed, 01 Jul 2026 18:56:23 GMT
WWW-Authenticate: BASIC realm="application"
Content-Length: 0
```

## Integration Tests
Added new integration tests

## UI changes
NA


gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
```diff
@@ -818,6 +818,45 @@ public String getHttpClientTruststorePasswordAlias() {
...
+  @Override
+  public boolean isSingleEkuEnabled() {
+    return getBoolean(TLS_SINGLE_EKU_ENABLED, DEFAULT_TLS_SINGLE_EKU_ENABLED);
+  }
+
+  @Override
+  public boolean isHttpClientTwoWaySslEnabled() {
+    String configured = get(HTTP_CLIENT_TWO_WAY_SSL_ENABLED);
+    if (configured != null) {
+      return Boolean.parseBoolean(configured);
```
pzampino: I believe parseBoolean returns false for null, so I think you could eliminate the null check unless this method intends to fall back to isSingleEkuEnabled() IFF two-way SSL is NOT explicitly configured. Is that the intent here? If so, this only applies to HTTP clients employed by Knox?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
```diff
@@ -132,6 +141,116 @@ private void logAndValidateCertificate(GatewayConfig config) throws ServiceLifec
...
+  //   - the outbound HTTP client truststore is configured, and
+  //   - both identities are single-purpose: the client identity carries clientAuth (not serverAuth)
+  //     and the server identity carries serverAuth (not clientAuth).
+  // Fail closed: never fall back to the server identity certificate for outbound calls, and never
+  // defer a cross-wired/dual-purpose certificate to a runtime handshake failure.
+  void validateSingleEkuConfig(GatewayConfig config) throws ServiceLifecycleException {
+    // 1. Outbound client-identity keystore must be configured and contain the configured key entry.
+    String clientKeystorePath = config.getHttpClientKeystorePath();
+    if (clientKeystorePath == null || clientKeystorePath.isEmpty()) {
+      throw new ServiceLifecycleException("Single-EKU mode is enabled (" + GatewayConfig.TLS_SINGLE_EKU_ENABLED
```
pzampino: Is it not valid to have single-purpose certs enabled WITHOUT two-way SSL, thus not requiring the client identity keystore?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
```diff
@@ -132,6 +141,116 @@ private void logAndValidateCertificate(GatewayConfig config) throws ServiceLifec
...
+            + "' (" + GatewayConfig.HTTP_CLIENT_KEY_ALIAS + "). Server will not start.");
+      }
+    } catch (KeystoreServiceException | KeyStoreException e) {
+      throw new ServiceLifecycleException("Single-EKU mode is enabled but the HTTP client keystore at "
+          + clientKeystorePath + " could not be loaded. Server will not start.", e);
+    }
+
+    // 2. Inbound client authentication must be enforced (server truststore + client-auth).
+    if (config.getTruststorePath() == null) {
+      throw new ServiceLifecycleException("Single-EKU mode is enabled but the server truststore ("
```
pzampino: Again, is it not permissible to have single-purpose certs enabled for Knox without assuming/requireing two-way SSL?

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
```diff
@@ -132,6 +141,116 @@ private void logAndValidateCertificate(GatewayConfig config) throws ServiceLifec
...
+          + GatewayConfig.GATEWAY_TRUSTSTORE_PATH + ") is not configured. Server will not start.");
+    }
+    if (!config.isClientAuthNeeded() && !config.isClientAuthWanted()) {
```
pzampino: Here too it seems two-way SSL is required just because single-purpose certs are enabled.

gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
moresandeep: Yeah, this method intends to fall back to `isSingleEkuEnabled()` this is because, for single EKU feature to work two-way-ssl is required as a result if two-way-ssl is not configured we default it to `isSingleEkuEnabled()`. 
So, if someone turned on `isSingleEkuEnabled()` then they not need to specifically turn on two-way-ssl as those go hand in hand.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
moresandeep: single-purpose certs need to have two-way SSL, this was pointed out by @smolnar82 in his offline review 

> This means Knox in "single-EKU mode" will start happily with a client keystore holding a serverAuth-only or dual-purpose cert — exactly the misconfiguration this feature exists to catch. The validation gives a false sense of assurance, and the tests reflect the gap (JettySSLServiceTest covers path-missing, truststore-missing, client-auth-disabled, and cert-not-a-key, but there is no EKU test).

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
moresandeep: Correct, two-way-SSL is required for this feature. But this change specifically ensures that we have truststore configured for clients who will talk to us over mTLS.

gateway-server/src/main/java/org/apache/knox/gateway/services/security/impl/JettySSLService.java
moresandeep: Correct.

pzampino: I don't think support for single-purpose certs necessitates mTLS; I think they are orthogonal concerns, but I may be wrong.

moresandeep: You might be right, let me check.

moresandeep: I changed it so now mTLS is independent of this feature! thanks for pointing this out!

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
```diff
@@ -173,25 +173,36 @@ private boolean doesRetryParamExist(final FilterConfig filterConfig) {
...
     KeyStore identityKeystore;
     char[] identityKeyPassphrase;
     KeyStore trustKeystore;
 
     KeystoreService ks = services.getService(ServiceType.KEYSTORE_SERVICE);
     try {
-      if (Boolean.parseBoolean(filterConfig.getInitParameter(PARAMETER_USE_TWO_WAY_SSL))) {
+      boolean singleEku = gatewayConfig != null && gatewayConfig.isSingleEkuEnabled();
+      boolean twoWaySsl = Boolean.parseBoolean(filterConfig.getInitParameter(PARAMETER_USE_TWO_WAY_SSL))
+          || (gatewayConfig != null && gatewayConfig.isHttpClientTwoWaySslEnabled());
```
pzampino, line 191: There is global config for mTLS for dispatching (i.e., gatewayConfig.isHttpClientTwoWaySslEnabled())?

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
```diff
@@ -201,6 +212,11 @@ SSLContext createSSLContext(GatewayServices services, FilterConfig filterConfig,
           trustKeystore = identityKeystore;
```
pzampino, line 212: Why is the truststore the same as the identity keystore? You're reusing the the client truststore as the Knox identity keystore in the mTLS scenario?

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/KeystoreService.java
```diff
@@ -34,6 +34,15 @@ public interface KeystoreService extends Service {
 
   void addSelfSignedCertForGateway(String alias, char[] passphrase, String hostname) throws KeystoreServiceException;
 
+  /**
+   * Adds a self-signed certificate to the Gateway keystore, stamping the given Extended Key Usage
+   * purpose OIDs (single-EKU identities). The default implementation ignores the OIDs.
+   */
+  default void addSelfSignedCertForGateway(String alias, char[] passphrase, String hostname, String... ekuOids)
```
pzampino, line 41: What is the purpose of this new method? Is it to support self-signed single-EKU certs for Knox?

gateway-util-common/src/main/java/org/apache/knox/gateway/util/X509CertificateUtil.java
```diff
@@ -73,7 +76,7 @@ public class X509CertificateUtil {
    * @param algorithm the signing algorithm, eg "SHA256withRSA"
    * @return self-signed X.509 certificate
    */
-  public static X509Certificate generateCertificate(String dn, KeyPair pair, int days, String algorithm) {
+  public static X509Certificate generateCertificate(String dn, KeyPair pair, int days, String algorithm, String... ekuOids) {
```
pzampino, line 79: We'll always generate single-EKU certs with this method now? Do we need to maintain the previous signature of this method for any reason and ADD this one instead?

moresandeep: Nope, this method does not always produce single EKU certs. In cases where ekuOids is empty certs will be generated as they have before.

The signature does not change! it still works with 4 args (assumes ekuOids is null) so existing calls still work. I did this to prevent yet another duplicate overloaded method. Let me know if you think overload is better here and I can switch it up.

gateway-spi/src/main/java/org/apache/knox/gateway/services/security/KeystoreService.java
moresandeep: Yes, it now supports adding self-signed single-EKU certs. It delegates to the original impl in case ekuOids is balnk/null so this is a non-breaking change.

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
moresandeep: THis does not look like a part of this PR, looks like it might have gotten shifted due to the changes. Looks like it is a fallback for when no outbound HTTP client truststore is configured then getTruststoreForHttpClient() returns null when gateway.httpclient.truststore.path isn't set. In that case we fall back to using its identity keystore as the trust material.

gateway-spi/src/main/java/org/apache/knox/gateway/dispatch/DefaultHttpClientFactory.java
moresandeep: Sorry, i did not understand your question. 
This is a global dispatch level mTLS flag.
Prior to this change two-way-ssl was dispatch level so we'd have to go to each service dispatch to enable mTLS. Now, we can just set it in gateay-site and it enables for all dispatches. Services can turn it off (override) if they want. But prior to this there was no global config for mTLS for dispatching.

pzampino: Ok, so you're introducing the new global config. Got it.

pzampino (approved):
LGTM

PR #1300, opened 2026-07-08
KNOX-3374: Fix inherited roles missing from auth headers when LDAP ro…

hanicz:
…les-lookup interceptor is active

[KNOX-3374](https://issues.apache.org/jira/browse/KNOX-3374) - Inherited roles missing from auth headers when LDAP roles-lookup interceptor is active

## What changes were proposed in this pull request?

- `KnoxLDAPServerManager.getUserGroups()` now attaches a `RolesLookupBypassControl` to its LDAP search, so it returns raw group names instead of interceptor-rewritten role names.
- Eliminates a double roles-lookup: `LDAPRolesLookupInterceptor` was rewriting memberOf to roles, and `AbstractAuthResource` then called the roles API again with those roles-as-groups — which returned only the user's direct roles, dropping inherited ones.
- External LDAP-proxy clients are unaffected; the interceptor still rewrites memberOf on their searches.
- New `KnoxLDAPServerManagerTest#testGetUserGroupsReturnsRawGroupsEvenWhenRolesInterceptorRewritesMemberOf` — installs the real roles interceptor over a test entries interceptor serving a user with memberOf group DNs; asserts getUserGroups returns raw group names, not the mocked roles.

## How was this patch tested?

Unit tests
E2Etests:

Roles:
```
[
  { "id": "admin", "type": "user",
    "roles": [ { "scope": "console", "name": "admin" } ] },
  { "id": "me-test-group-a", "type": "group",
    "roles": [ { "scope": "ws-1", "name": "viewer" } ] },
  { "id": "me-test-group-b", "type": "group",
    "roles": [ { "scope": "ws-2", "name": "user" } ] }
]
```

Without changes:
```
ldapsearch -x -H ldap://localhost:10389 \
  -D "uid=admin,ou=people,dc=hadoop,dc=apache,dc=org" -w admin-password \
  -b "dc=hadoop,dc=apache,dc=org" "(uid=admin)" memberOf

# admin, people, hadoop.apache.org
dn: uid=admin,ou=people,dc=hadoop,dc=apache,dc=org
memberOf: cn=ws-1:viewer,ou=groups,ou=groups,dc=hadoop,dc=apache,dc=org
memberOf: cn=ws-2:user,ou=groups,ou=groups,dc=hadoop,dc=apache,dc=org
```

```
curl -kvs -u admin:admin-password \
  https://localhost:8443/gateway/sandbox/auth/api/v1/extauthz/anything

< username: admin
< roles: console:admin
```

With the fix:
```
[...]

PR #1301, opened 2026-07-09
KNOX-3374: Revert LDAPServer Manager changes. Skip roles lookup if ro…

hanicz:
…lesLookupInterceptor is enabled

[KNOX-3374](https://issues.apache.org/jira/browse/KNOX-3374) - Fix inherited roles missing from auth headers when LDAP roles-lookup interceptor is active

Reworked the KNOX-3374 fix per review feedback.

- Revert previous [PR](https://github.com/apache/knox/pull/1300)
-  `AbstractAuthResource` checks for `rolesLookupInterceptor` enabled.
   1. **Interceptor configured**: the Subject's GroupPrincipal's are already role names.
   2. **Interceptor not configured**: Subject has raw groups, lookupRoles call is necessary.
- New unit tests to verify the behaviour 

## How was this patch tested?

Unit tests, manual e2e tests

Roles:

```
[
  { "id": "admin", "type": "user",
    "roles": [ { "scope": "console", "name": "admin" } ] },
  { "id": "me-test-group-a", "type": "group",
    "roles": [ { "scope": "ws-1", "name": "viewer" } ] },
  { "id": "me-test-group-b", "type": "group",
    "roles": [ { "scope": "ws-2", "name": "user" } ] }
]
```

Tested with rolesLookupInterceptor enabled and disabled

```
curl -kvs -u admin:admin-password \
  https://localhost:8443/gateway/sandbox/auth/api/v1/extauthz/anything

< username: admin
< roles: ws-1:viewer,console:admin,ws-2:user
```

## Integration Tests
N/A

## UI changes
N/A

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
@@ -45,7 +45,6 @@ import org.apache.directory.server.protocol.shared.transport.TcpTransport;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.services.ldap.control.RolesLookupBypassControlFactory;
-import org.apache.knox.gateway.services.ldap.control.RolesLookupBypassControlImpl;
 import org.apache.knox.gateway.services.ldap.interceptor.InterceptorFactory;
 import org.apache.knox.gateway.services.security.AliasService;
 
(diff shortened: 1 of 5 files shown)


hanicz:
@handavid Could you please review the changes?

handavid (commented):
@hanicz 
Can you explain to me your thoughts behind this PR? I don't understand why the `AbstractAuthResource` needs to do role lookup at all. Can't it just rely on the the KnoxLDAPServerManager to return the appropriate values (either groups or roles as configured)? Is there a use-case where the auth would need to return roles even if the LDAP service is configured to return groups? I find that a little confusing.
Also, I don't think it's really appropriate for auth to condition it's behavior on the configuration of the LDAP proxy. IMO, the LDAP proxy should be a black box to auth and auth should just take whatever the proxy returns.

PR #1305, opened 2026-07-14
KNOX-3377: KNOX-3378: KNOX-3379: Support secure LDAP (LDAPS) for the Knox embedded LDAP service

smolnar82:
[KNOX-3377](https://issues.apache.org/jira/browse/KNOX-XXXX) - Support secure LDAP (LDAPS) for the Knox embedded LDAP service

`KnoxLDAPService` previously operated over plaintext LDAP only, on both of its connection paths. This PR adds configurable, credential-store-integrated LDAPS support to both, and updates the docker-based integration tests to exercise the feature end-to-end. The three commits are self-contained:

1. Front-end (embedded server) LDAPS - the embedded ApacheDS server can now expose a TLS listener instead of accepting only plaintext.
    - New config in the existing gateway.ldap.* namespace (defaults off, behavior unchanged):
      - `gateway.ldap.ssl.enabled`
      - `gateway.ldap.ssl.keystore.path` (defaults to the gateway identity keystore)
      - `gateway.ldap.ssl.keystore.password.alias` (defaults to the gateway identity keystore password)
      - `gateway.ldap.ssl.enabled.cipher.suites` (optional)
    - The keystore password is resolved through `AliasService`, consistent with the existing bind-password handling. Invalid SSL config (missing keystore) fails fast at startup with a clear message.

2. Backend (proxy) LDAPS - `LdapProxyBackend` can now connect to the remote/backing LDAP server over LDAPS. Before, an `ldaps://` URL was parsed but SSL was never enabled on the connection, so secure backend connections silently did not work. TLS is applied to both the pooled search connections and the authentication bind.
Per-backend config (`gateway.ldap.interceptor.<name>.*`): `useSsl` (auto-enabled by an `ldaps://` URL), `trustAllCertificates`, `trustStore/trustStoreType/trustStorePassword`, `sslProtocol`, `enabledProtocols`, `enabledCipherSuites`, `connectionTimeout`.
    - Required dependency alignment for client-side LDAPS to work under the `MINA 2.2.8` that Knox ships (for CVEs): `Apache Directory client API 2.0.0 → 2.1.8` (2.0.0's SSL code calls MINA APIs removed in 2.2.x), and `ApacheDS server 2.0.0.AM26 → 2.0.0.AM27` (AM26's LdapsInitializer references `NoVerificationTrustManager`, removed in directory-api 2.1.8; AM27 is built for 2.1.x + MINA 2.2.x). MINA stays at 2.2.8. 
All directory-api artifacts are pinned to one version; `commons-collections4` is converged to 4.5.0; and the legacy `jdk15on` BouncyCastle that AM27 pulls is excluded in favor of Knox's `jdk18on` artifacts (same packages), keeping `dependency:analyze` and the shipped BouncyCastle consistent.

3. Integration test over LDAPS - the docker-based LDAP proxy search test now runs entirely over TLS: the client connects to the embedded Knox LDAP service over LDAPS, and Knox proxies to the demo LDAP backend over LDAPS. The demo LDAP (SimpleLdapDirectoryServer) gained optional system-property-driven SSL; ldap.sh/gateway.sh provision dev keystores and the knoxldap Shiro realm was moved to ldaps:// (its cert imported into the JVM cacerts so the JNDI realm trusts it).

## How was this patch tested?

- Unit tests (gateway-server):
  - Front-end: TLS handshake presenting the configured certificate (JSSE SSLSocket), plaintext-access-to-secure-port rejection, and missing-keystore fail-fast.
  - Backend: a new `LdapProxyBackendSslTest` starts an SSL-enabled embedded ApacheDS instance and drives the backend over TLS (user/group lookups and binds), and verifies an untrusted server certificate is rejected when validation is enabled.
  - Full LDAP package (129 tests) passes against the upgraded ApacheDS AM27 / directory-api 2.1.8, plus `gateway-demo-ldap` and `gateway-util-common`.
- Build: reactor-wide `mvn validate` (dependency convergence) and `dependency:analyze-only` pass after the version bumps and BouncyCastle exclusion.
- Manual: `mvn -Ppackage clean install` then the docker-compose stack.

## Integration Tests

Updated `.github/workflows/tests/test_knox_ldap_proxy_search.py` to connect over LDAPS (use_ssl=True, Tls(validate=CERT_NONE) for the self-signed dev cert), and the supporting docker fixtures (`gateway-site.xml`, `gateway.sh`, `ldap.sh`, `knoxldap.xml`) so both the client → Knox and Knox → demo-LDAP legs are TLS. 

`test_knox_auth_service_and_ldap.py` is unchanged but now implicitly exercises the secure backend (group resolution flows through it). 

Full stack result: 32/32 integration tests pass, including the `knoxldap/remoteauth/extauthz` auth suites that authenticate through the now-LDAPS embedded service.

## UI Changes
N/A

### DIFF
--- a/.github/workflows/build/conf/topologies/knoxldap.xml
+++ b/.github/workflows/build/conf/topologies/knoxldap.xml
@@ -35,7 +35,7 @@ limitations under the License.
 			</param>
 			<param>
 				<name>main.ldapRealm.contextFactory.url</name>
-				<value>ldap://localhost:33390</value>  <!-- Local Knox LDAP service which is configured to proxy to the demo LDAP as a backend -->
+				<value>ldaps://localhost:33390</value>  <!-- Local Knox LDAP service (LDAPS); configured to proxy to the demo LDAP as a backend -->
 			</param>
 			<param>
 				<name>main.ldapRealm.contextFactory.authenticationMechanism</name>
--- a/.github/workflows/build/gateway-site.xml
+++ b/.github/workflows/build/gateway-site.xml
@@ -153,6 +153,20 @@ limitations under the License.
         <name>gateway.ldap.interceptor.names</name>
         <value>demoldap</value>
     </property>
+    <!-- Expose the embedded Knox LDAP service over LDAPS using the dev keystore that
+         gateway.sh provisions; its password is resolved from the credential store. -->
+    <property>
+        <name>gateway.ldap.ssl.enabled</name>
+        <value>true</value>
+    </property>
+    <property>
+        <name>gateway.ldap.ssl.keystore.path</name>
+        <value>/knox-runtime/conf/ldaps-keystore.p12</value>
+    </property>
+    <property>
+        <name>gateway.ldap.ssl.keystore.password.alias</name>
+        <value>gateway_ldap_ssl_keystore_password</value>
+    </property>
 
     <!-- LDAP Backend specific configuration (proxying to demo ldap) -->
     <property>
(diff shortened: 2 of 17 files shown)


smolnar82:
Cc. @handavid

handavid (approved):
LGTM

hanicz (approved):

PR #1306, opened 2026-07-15
KNOX-3374: Skip redundant roles lookup in PreAuth/ExtAuthz when the identity-assertion layer already resolved roles

smolnar82:
[KNOX-3374](https://issues.apache.org/jira/browse/KNOX-3374) - Avoid duplicate LDAP roles lookup between the identity-assertion layer and the PreAuth/ExtAuthz auth services

During roles-lookup testing it turned out that Knox performed the remote roles lookup twice for a single request, and the second lookup could return the wrong roles:

- once on the identity-assertion layer, while `HadoopGroupProviderFilter` resolves groups against Knox's embedded LDAP server (the `LDAPRolesLookupInterceptor` does the lookup as part of the group search), and
- again in the `PreAuth` / `ExtAuthz` auth services (`AbstractAuthResource#lookupRoles`).

Because the interceptor already looked roles up on the authentication layer, the second call in the auth resource is redundant. This PR makes the auth-service-level roles lookup conditional: it now happens only if
- the roles-lookup service is enabled (existing condition), and
- roles were not already looked up on the authentication layer.

We deliberately keep the auth-service lookup for the case where the topology's Shiro config points to an external LDAP server (not Knox's embedded one) but roles are still wanted: a valid Apache scenario. The case where the interceptor is not configured but the roles-lookup service is present is not supported: those deployments must configure the `LDAPRolesLookupInterceptor`.

**How it works**

- A new request attribute constant `AbstractIdentityAssertionBase.ROLES_LOOKUP_EXECUTED` is introduced to flag, per request, that roles were already resolved on the authentication layer.
- `HadoopGroupProviderFilter` sets this attribute when it resolves groups via `KnoxLDAPService` and the embedded LDAP server has the roles-lookup interceptor active. To support this, `KnoxLDAPService/KnoxLDAPServerManager` now expose `hasRolesLookupInterceptor()`, computed once at startup from the configured interceptor chain.
- Because the interceptor already performs the lookup,` KnoxLDAPServerManager#getUserGroups` no longer attaches the `RolesLookupBypassControl` to the group search request (reverted changes from #1300).
- `AbstractAuthResource` reads the attribute (via a new abstract `getRequest()` implemented by PreAuthResource and ExtAuthzResource) and skips its own lookupRoles call - and emits a single roles header - when roles were already resolved upstream.
- The `mapGroupPrincipals(...)` SPI method gains a `ServletRequest` parameter so implementations can decorate the request with the flag. All identity-assertion filters (common, concat, hadoop-groups, no-doas, pseudo, regex, switchcase) are updated for the new signature.

## How was this patch tested?

Ran the affected module unit tests (`gateway-provider-identity-assertion-*, gateway-service-auth, gateway-server`).

Added / updated unit tests:
- `HadoopGroupProviderFilterTest#testLdapServiceIntegrationWithRolesLookupInterceptor` - verifies the `ROLES_LOOKUP_EXECUTED` attribute is set on the request when the interceptor is active.
- `PreAuthResourceTest#testRolesAreNotPopulatedTwice` - verifies the auth resource does not call lookupRoles again (strict mock) and emits the roles header only once when the flag is set.
- `ExtAuthzResourceTest / PreAuthResourceTest` updated to inject the `HttpServletRequest` and stub the `ROLES_LOOKUP_EXECUTED` attribute.
- All `mapGroupPrincipals(...)` call sites in existing filter tests updated for the new signature.
- Removed `KnoxLDAPServerManagerTest#testGetUserGroupsReturnsRawGroupsEvenWhenRolesInterceptorRewritesMemberOf`, which asserted the now-removed bypass-control behavior on the group search.

Manually tested against the sandbox topology, covering both roles-lookup paths:
- Embedded LDAP service + HadoopGroupProvider (use.ldap.service=true) — roles are looked up once on the identity-assertion layer via the interceptor, and the PreAuth service correctly skips the duplicate lookup.
- Demo LDAP + HadoopGroupProvider (use.ldap.service=false) — the identity-assertion layer does not run the interceptor, so the PreAuth service performs the roles lookup as before.
```
~ $ curl -iku admin:admin-password http://localhost:8443/gateway/sandbox/auth/api/v1/pre
HTTP/1.1 200 OK
Date: Wed, 15 Jul 2026 11:33:17 GMT
Set-Cookie: KNOXSESSIONID=node01clcgs9svska5dvd0c0dfdpzb4.node0; Path=/gateway/sandbox; Secure; HttpOnly
Set-Cookie: rememberMe=deleteMe; Path=/gateway/sandbox; Max-Age=0; Expires=Tue, 14-Jul-2026 11:33:17 GMT; SameSite=lax
x-knox-username: admin
x-knox-roles: platform:awc-admin-admin,ml-workspace-abc:viewer-admin
Content-Length: 0
```

## Integration Tests

N/A

## UI changes

N/A

### DIFF
--- a/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
+++ b/gateway-provider-identity-assertion-common/src/main/java/org/apache/knox/gateway/identityasserter/common/filter/AbstractIdentityAssertionFilter.java
@@ -54,8 +54,7 @@ import org.apache.knox.gateway.security.SubjectUtils;
 import org.apache.knox.gateway.security.TokenExchangePrincipal;
 import org.apache.knox.gateway.security.TokenIdPrincipal;
 
-public abstract class AbstractIdentityAssertionFilter extends
-  AbstractIdentityAssertionBase implements Filter {
+public abstract class AbstractIdentityAssertionFilter extends AbstractIdentityAssertionBase implements Filter {
 
   private IdentityAsserterMessages LOG = MessagesFactory.get(IdentityAsserterMessages.class);
 
@@ -70,15 +69,17 @@ public abstract class AbstractIdentityAssertionFilter extends
   }
 
   /**
-   * This method returns a Stringp[] of new group principal names to use
+   * This method returns a String[] of new group principal names to use
    * based on implementation specific mapping or lookup mechanisms.
    * Returning null means that whatever set of GroupPrincipals is in the
    * provided Subject is sufficient to use and no additional mapping is required.
+   *
    * @param mappedPrincipalName username for the authenticated identity - post mapUserPrincipal mapping.
-   * @param subject the existing Subject from the authentication event which may or may not contain GroupPrincipals.
+   * @param subject             the existing Subject from the authentication event which may or may not contain GroupPrincipals.
+   * @param request             the request which may or may not be decorated with extra attributes (e.g. if LDAP roles lookup happened already)
    * @return String[] of new principal names to use as GroupPrincipals or null.
    */
-  public abstract String[] mapGroupPrincipals(String mappedPrincipalName, Subject subject);
+  public abstract String[] mapGroupPrincipals(String mappedPrincipalName, Subject subject, ServletRequest request);
 
   /**
    * This method is used to map the username of the authenticated identity to some other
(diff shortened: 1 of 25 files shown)


smolnar82:
Cc. @handavid

handavid (approved):
lgtm

PR #1311, opened 2026-07-18
KNOX-3355 - Add TrustedOidcIssuerService schema and interface

hsheinblatt:
KNOX-3355 - Add TrustedOidcIssuerService schema and interface

Feature context:
Support OIDC JWKS discovery to validate tokens from external authorizers. Currently,
there is a fixed trusted JWKS list created as part of initial configuration that can be used
to validate tokens. Allow the dynamic registration of trusted issuers with a flag that they are
trusted for OIDC JWKS discovery. Tokens will be validated with discovered JWKS from
such trusted issuers.

An initial detailed use case is to validate kubernetes service account tokens from
external kubernetes clusters. When the kubernetes cluster is created or integrated,
the kubernetes issuer can be registered as trusted for JWKS discovery without knox
service disruption. Then, service account tokens from that cluster can be validated
using OIDC discovery for JWKS.

The implementation plan is bottom up: first the storage and interface for the trusted
issuers, then the implementation of registration through an admin API in knox idf, and
then the token validation integration. It will be part of knox_idf initially, but written to
be portable so that it can be moved into a different or separate service if more
OIDC discovery use cases emerge.

How this patch was tested:
New unit tests are added.

Follow on changes will implement the interface and then integrate in the implementation
into KnoxIDF flows, but now there is no integration or UI changes.

Note this was originally posted in https://github.com/apache/knox/pull/1270 . The knox_idf
branch was altered, so this PR is made on top of the updated branch.
Comments on that pull request are also addressed here:
* Filed https://issues.apache.org/jira/browse/KNOX-3384 to track moving
  gateway-site.xml content to doc
* Added NOT NULL statements to the Derby SQL file. Note that it's not necessary on the
  primary key column as that will default to enforce NOT NULL, but it's still included for
  clarity.
* Moved the cluster_name column to the end for all SQL files.
* For the requestedBy column: it can be null if we can't resolve the principal name of
  the API caller. Arguably one should not set up permissions so that is possible, but in
  general it could happen.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/database/AbstractDataSourceFactory.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/database/AbstractDataSourceFactory.java
@@ -50,6 +50,10 @@ public abstract class AbstractDataSourceFactory {
     public static final String DERBY_KNOXIDF_FED_IDENTITY_TABLE_CREATE_SQL_FILE_NAME = "createKnoxIDFFederatedIdentityTableDerby.sql";
     public static final String DERBY_KNOXIDF_FED_IDENTITY_ATTR_TABLE_CREATE_SQL_FILE_NAME = "createKnoxIDFFederatedIdentityAttributesTableDerby.sql";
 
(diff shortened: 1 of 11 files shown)


smolnar82 (approved):

lmccay (approved):
Lint test failures are unrelated to this change.
+1

PR #1313, opened 2026-07-21
KNOX-3387: Forward proxy-namespace searches to the LDAP backend when proxy and remote base DNs differ

smolnar82:
[KNOX-3387](https://issues.apache.org/jira/browse/KNOX-3387) - HadoopGroupProvider returns no groups when the embedded LDAP proxy base DN differs from the remote base DN

When the embedded Knox LDAP service (`gateway.ldap.enabled=true`) is configured with an LDAP proxy backend whose proxy base DN (`gateway.ldap.base.dn`) differs from the backend's `remoteBaseDn`, group lookups via the HadoopGroupProvider (`use.ldap.service=true`) always returned an empty group list, even when the user is a member of groups on the remote LDAP server.

**Root cause:** group lookups (`KnoxLDAPServerManager.getUserGroups()`) search under the proxy base DN and flow through `UserSearchInterceptor.search()`, which only forwards to the backend when `isUnderBackendBaseDn()` passes. That check compared the proxy-namespace search base against `LdapProxyBackend.getBaseDn()`, which returns the remote base DN, so `"dc=proxy,dc=com".endsWith("dc=example,dc=com")` was `false`, the backend was never queried, and no `memberOf` values were found. This contradicts the backend's own contract: `LdapProxyBackend.search()` expects a proxy-namespace base and translates it to the remote namespace internally via `RemoteSchemaConverter`.

**Changes:**
- Added `getProxyBaseDn()` to the `LdapBackend` interface. `FileBackend` returns its single base DN (proxy and remote are identical for non-proxying backends); LdapProxyBackend returns `proxyBaseDn` (existing already).
- Updated `UserSearchInterceptor.isUnderBackendBaseDn()` to forward a search when the base is under the proxy base DN or the remote base DN. Both namespaces are registered as partitions by `KnoxLDAPServerManager`, and `LdapProxyBackend.search()` handles either.

No new configuration is introduced: `getProxyBaseDn()` exposes the already-existing value derived from `gateway.ldap.base.dn`.

## How was this patch tested?

- Added `UserSearchInterceptorTest` (4 cases) exercising a backend where `proxyBaseDn != remoteBaseDn`:
  - proxy-namespace search is forwarded and the entry is returned exactly once (regression guard);
  - sub-tree search below the proxy base DN is forwarded;
  - remote-namespace search is still forwarded;
  - operational searches (e.g. ou=schema) are not forwarded.
- Verified the tests fail against the pre-fix code (backend never queried → empty results) and pass with the fix.
- Ran the full `org.apache.knox.gateway.services.ldap.**` suite (132 tests) — all green, checkstyle included.
- Manually verified end-to-end: with `gateway.ldap.base.dn = dc=proxy,dc=com` and `remoteBaseDn = dc=example,dc=com`, HadoopGroupProvider now resolves the user's groups from the remote server; previously it logged 
```
Found groups for principal admin : [].
```
now:
```
Found groups for principal admin : [engineering, operations]
```

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/FileBackend.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/FileBackend.java
@@ -91,6 +91,11 @@ public class FileBackend implements LdapBackend {
         return baseDn;
     }
 
+    @Override
+    public String getProxyBaseDn() {
+        return baseDn;
+    }
+
     private void loadData() throws Exception {
         Path path = Paths.get(dataFile);
 
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapBackend.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapBackend.java
@@ -45,9 +45,22 @@ public interface LdapBackend {
     String getType();
 
     /**
-     * Get the base dn of this backend
+     * Get the base dn of this backend.
+     * <p>
+     * For proxying backends this is the <em>remote</em> base DN (the namespace of the
+     * upstream LDAP server). For non-proxying backends it is the same as
+     * {@link #getProxyBaseDn()}.
      */
     String getBaseDn();
+
+    /**
+     * Get the proxy (client-facing) base DN this backend is mounted at in the embedded
+     * LDAP server, i.e. the namespace incoming searches are expressed in.
+     * <p>
+     * Proxying backends translate this namespace to {@link #getBaseDn()} internally. For
+     * non-proxying backends this is the same as {@link #getBaseDn()}.
+     */
+    String getProxyBaseDn();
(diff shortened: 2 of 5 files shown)


smolnar82:
Cc. @handavid

hanicz (approved):

handavid:
I have a change that I was about to post for this same problem. I opened https://issues.apache.org/jira/browse/KNOX-3385. I'll post that shortly.

I did have a question for @smolnar82 regarding the comments in code for
```
// System/operational searches (ou=schema, cn=config, root-DSE) must not be forwarded.
```
just using the proxy/remote base dn does not restrict these searches. My delegates the decision to the LdapBackend so the backend can use the user or group search base to restrict searches.

handavid:
@smolnar82 see https://github.com/apache/knox/pull/1314

smolnar82:
Closing this in favor of #1314 .

PR #1315, opened 2026-07-22
KNOX-3355 - Add OIDC trusted issuer discovery implementation

hsheinblatt:
KNOX-3355 - Add OIDCDiscoveryHelper, JdbcTrustedOidcIssuerService, and TrustedOidcIssuerServiceFactory

## How was this patch tested?

Unit tests are added.

There are no integration or UI tests added. Integration tests will be added once enough of feature is merged
that the full flow can be tested.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/JdbcTrustedOidcIssuerService.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/JdbcTrustedOidcIssuerService.java
@@ -1,0 +1,200 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.knox.gateway.services.knoxidf.trustedoidcissuer;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.knox.gateway.config.GatewayConfig;
+import org.apache.knox.gateway.database.DataSourceProvider;
+import org.apache.knox.gateway.i18n.messages.MessagesFactory;
+import org.apache.knox.gateway.services.ServiceLifecycleException;
+import org.apache.knox.gateway.services.security.AliasService;
+import org.apache.knox.gateway.util.knoxidf.KnoxIDFConstants;
+
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * JDBC-backed implementation of {@link TrustedOidcIssuerService}.
+ * <p>
+ * Maintains an in-memory registry snapshot as an {@link AtomicReference} to an immutable
+ * {@link Map}. Reads ({@link #isTrusted}, {@link #isDynamicJwks}, {@link #list}) are
+ * lock-free and always see a consistent snapshot. Writes ({@link #register},
+ * {@link #deregister}) are synchronized: the DB is committed first, then the snapshot is
+ * rebuilt from a fresh SELECT to guarantee the in-memory state cannot diverge from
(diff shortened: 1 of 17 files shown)


gateway-util-common/src/main/java/org/apache/knox/gateway/util/knoxidf/KnoxIDFConstants.java
```diff
@@ -57,4 +57,17 @@ public interface KnoxIDFConstants {
...
 
     String TOKEN_EXCHANGE_TOPOLOGY_NAME = "token.exchange.topology.name";
+
+    // TrustedOidcIssuerService gateway-level params (read from GatewayConfig / gateway-site.xml)
+    String TRUSTED_OIDC_ISSUER_DISCOVERY_CACHE_TTL_SECS =
+        "gateway.trustedoidcissuer.discovery.cache.ttl.secs";
+    String TRUSTED_OIDC_ISSUER_DISCOVERY_CONNECT_TIMEOUT_MS =
+        "gateway.trustedoidcissuer.discovery.connect.timeout.ms";
+    String TRUSTED_OIDC_ISSUER_DISCOVERY_READ_TIMEOUT_MS =
+        "gateway.trustedoidcissuer.discovery.read.timeout.ms";
```
smolnar82, line 67: They belong to `GatewayConfig.java`, just like any other gateway-level config names.

gateway-server/src/main/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/JdbcTrustedOidcIssuerService.java
```diff
@@ -0,0 +1,200 @@
...
+   * Rebuilds the registry snapshot from the current DB state.
+   * Called on init, after register, and after deregister.
+   * Synchronized on this to prevent concurrent rebuilds from interleaving with mutations.
+   */
+  private synchronized void reloadRegistrySnapshot() {
+    try {
+      final Map<String, TrustedOidcIssuer> fresh = database.selectAll().stream()
+          .collect(Collectors.toMap(TrustedOidcIssuer::getIssuerUrl, Function.identity()));
+      registrySnapshot.set(Collections.unmodifiableMap(fresh));
+    } catch (Exception e) {
```
smolnar82, line 196: The class javadoc states the snapshot is rebuilt after each write "to guarantee the in-memory state cannot diverge from persistent storage." But:
```
public synchronized void register(TrustedOidcIssuer issuer) {
    ...
    database.insert(issuer);        // committed
    reloadRegistrySnapshot();       // if the SELECT fails here, it's caught + logged, not rethrown
}
```
If the reload `SELECT` fails after a successful insert, `register()` returns normally (caller thinks it succeeded), the row is in the DB, but i`sTrusted(url) returns false` until the next successful mutation or a restart. 

That's a silent divergence - exactly what the javadoc promises can't happen. Consider propagating the failure (or at least having register/deregister surface it) rather than logging and returning success.

gateway-server/src/main/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/JdbcTrustedOidcIssuerService.java
```diff
@@ -0,0 +1,200 @@
...
+  @Override
+  public boolean isTrusted(String issuerUrl) {
+    return registrySnapshot.get().containsKey(issuerUrl);
+  }
+
+  @Override
+  public boolean isDynamicJwks(String issuerUrl) {
+    final TrustedOidcIssuer entry = registrySnapshot.get().get(issuerUrl);
+    return entry != null && entry.isDynamicJwks();
+  }
```
smolnar82, line 140: `fetchJwksUri` strips a trailing slash before building the discovery URL (`issuerUrl.replaceAll("/$", "")`), but the registry key is the raw string. So an issuer registered as `https://idp.example.com/` won't match an `iss` claim of `https://idp.example.com` (or vice-versa), and there's no normalization at `register()` either. 
OIDC does require exact `iss` matching, so this may be intentional - but if so, it's worth a comment, and if not, normalizing on both write and lookup would avoid a frustrating "why isn't my trusted issuer trusted" class of bug.

gateway-server/src/main/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/TrustedOidcIssuerDatabase.java
```diff
@@ -0,0 +1,101 @@
...
+    return result;
+  }
+
+  int count() throws SQLException {
+    try (Connection connection = dataSource.getConnection();
+         PreparedStatement ps = connection.prepareStatement(COUNT_SQL);
+         ResultSet rs = ps.executeQuery()) {
+      return rs.next() ? rs.getInt(1) : 0;
+    }
+  }
```
smolnar82, line 100: This method seems to be unused -> you may want to remove it.

pom.xml
```diff
@@ -259,6 +259,7 @@
         <mina.version>2.2.8</mina.version>
         <netty.version>4.1.127.Final</netty.version>
         <nimbus-jose-jwt.version>10.9.1</nimbus-jose-jwt.version>
+        <oauth2-oidc-sdk.version>11.37.2</oauth2-oidc-sdk.version>
```
smolnar82, line 262: Please move this new attribute below to honor the alphabetical order we try to keep. Thanks!

smolnar82 (changes requested):

smolnar82:
Based on our agreement (see [this e-mail thread](https://lists.apache.org/thread/2nlxcqqw06pw3sdmcvht6mrsdqdbnlz0)) I'm merging this change on the `knox_idf` branch. I filed [KNOX-3390](https://issues.apache.org/jira/browse/KNOX-3390) to address my comments above.

PR #1318, opened 2026-07-22
KNOX-3389: Embedded LDAP getUserGroups drops roles-lookup roles that have no backing group

smolnar82:
[KNOX-3389](https://issues.apache.org/jira/browse/KNOX-3389) - Embedded LDAP getUserGroups drops roles-lookup roles that have no backing group

`KnoxLDAPServerManager.getUserGroups` failed to return roles-lookup roles as groups when the user had no backing group.

When the roles lookup interceptor is enabled, resolved roles are written onto the user entry as `memberOf` values. If the user has no existing group, `LDAPRolesLookupInterceptor.addRoleAttribute` has no template group DN to build a
full DN from and stores each role as a bare RDN (e.g. `memberOf: cn=platform:admin-sam`) with no trailing DN components. `getUserGroups` extracted the group name as the substring between `cn=` and the first comma; a bare RDN has no comma, so the value was silently dropped and the roles never surfaced as groups.

This PR reworks the `memberOf` parsing in `getUserGroups`:
- Parses each `memberOf` value as a `Dn` and extracts the group name via the existing `LdapUtils.extractGroupName(Dn)` helper (leading RDN, `cn`-typed, `rdn.getValue()`) instead of manual string splitting. This is the same helper `LDAPRolesLookupInterceptor.fetchGroups` already uses, so the two `memberOf`-parsing paths are now consistent and resilient to RDN formatting (case, escaping, spacing).
- A full group DN (`cn=role,ou=groups,...`, i.e. `dn.size() > 1`) is always resolved. A bare RDN (`cn=role`) is only ever produced by the roles lookup interceptor, so it is accepted only when that interceptor is active (`hasRolesLookupInterceptor`); otherwise it is treated as unexpected data and skipped.
- Values that are not valid DNs are caught and skipped per-value, so a single malformed entry no longer aborts the whole group lookup.

## How was this patch tested?

Added two unit tests to `KnoxLDAPServerManagerTest` that start the embedded server, seed a user via the admin session, and call `getUserGroups` directly:
- `testGetUserGroupsResolvesBareRdnRolesWhenRolesLookupActive` - user with `memberOf: cn=platform:awc-admin-sam` and `cn=ml-workspace-abc:viewer-sam`, interceptor flag on: both roles resolve as groups (reproduces the reported scenario).
- `testGetUserGroupsIgnoresBareRdnWhenRolesLookupInactive` - a full-DN group plus a bare RDN, interceptor flag off: only the full-DN group resolves, proving the lenient parse is scoped to the roles-lookup case.

Manually tested as well (using the same setup as described in the JIRA). After my changes applied, I could see the resolved groups:
```
~ $ curl -iku sam:sam-password http://localhost:8443/gateway/sandbox/auth/api/v1/pre
HTTP/1.1 200 OK
Date: Wed, 22 Jul 2026 11:10:35 GMT
Set-Cookie: KNOXSESSIONID=node01qag1cyt0x3sat9xtybpnjtlb0.node0; Path=/gateway/sandbox; Secure; HttpOnly
Set-Cookie: rememberMe=deleteMe; Path=/gateway/sandbox; Max-Age=0; Expires=Tue, 21-Jul-2026 11:10:35 GMT; SameSite=lax
x-knox-username: sam
x-knox-roles: ml-workspace-abc:viewer-sam,platform:awc-admin-sam
Content-Length: 0
```

## Integration Tests
N/A

## UI changes
N/A

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
@@ -470,6 +470,13 @@ public class KnoxLDAPServerManager {
                             int commaIdx = groupDn.indexOf(',');
                             if (commaIdx > 0) {
                                 groups.add(groupDn.substring(3, commaIdx));
+                            } else if (hasRolesLookupInterceptor) {
+                                // When the roles lookup interceptor is active and no template
+                                // group DN was available, roles are stored as a bare RDN
+                                // (cn=role) with no trailing DN components (see
+                                // LDAPRolesLookupInterceptor#addRoleAttribute). Take everything
+                                // after "cn=".
+                                groups.add(groupDn.substring(3));
                             }
                         }
 
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
@@ -27,6 +27,7 @@ import org.apache.directory.api.ldap.model.entry.DefaultEntry;
 import org.apache.directory.api.ldap.model.entry.Entry;
 import org.apache.directory.api.ldap.model.entry.Value;
 import org.apache.directory.api.ldap.model.exception.LdapException;
+import org.apache.directory.api.ldap.model.exception.LdapInvalidDnException;
 import org.apache.directory.api.ldap.model.message.SearchRequest;
 import org.apache.directory.api.ldap.model.message.SearchRequestImpl;
 import org.apache.directory.api.ldap.model.message.SearchScope;
@@ -56,7 +57,6 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.IntStream;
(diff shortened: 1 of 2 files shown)


smolnar82:
Cc. @handavid

hanicz (approved):

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
```diff
@@ -470,6 +470,13 @@ public List<String> getUserGroups(String username) throws Exception {
                             int commaIdx = groupDn.indexOf(',');
                             if (commaIdx > 0) {
                                 groups.add(groupDn.substring(3, commaIdx));
+                            } else if (hasRolesLookupInterceptor) {
```
handavid: This looks ok, but I was wondering if we could convert the value to a `Dn` and pull the `Rdn` instead of string splitting. e.g., something like:
```
Dn groupDn = new Dn(value.getString());
String groupName = groupDn.getRdn().getValue();
```
I feel that would be more resilient regardless of what RDN is used

smolnar82: Thanks, @handavid , for your review. It's actually a really good idea so I submitted a new commit that uses an already existing group parser util (by `Dn`) instead of string operations.
I tested the updated code manually as well:
- backend LDAP with 0 groups + roles lookup enabled
- backend LDAP with 1+ groups + roles lookup enabled
- backend LDAP with 0 groups + roles lookup disabled
- backend LDAP with 1+ groups + roles lookup disabled

In all cases I got the desired output.

handavid (approved):

PR #1321, opened 2026-07-23
KNOX-3368 - Trusted OIDC Issuer admin API for Knox IDF

hsheinblatt:
KNOX-3368 - Trusted OIDC Issuer admin API for Knox IDF

Implement the admin API for the trusted OIDC issuer service
along with the contributor for it.

## How was this patch tested?
Unit tests were added

## Integration Tests
N/A will be added once the flow can be tested.


pom.xml
```diff
@@ -259,6 +259,7 @@
         <mina.version>2.2.8</mina.version>
         <netty.version>4.1.135.Final</netty.version>
         <nimbus-jose-jwt.version>10.9.1</nimbus-jose-jwt.version>
+        <oauth2-oidc-sdk.version>11.37.2</oauth2-oidc-sdk.version>
```
smolnar82, line 262: This must be left here because of the recent changes in #1320

gateway-service-knoxidf/src/main/java/org/apache/knox/gateway/service/knoxidf/deploy/KnoxIDFAdminServiceDeploymentContributor.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  protected String[] getPackages() {
+    return new String[] { "org.apache.knox.gateway.service.knoxidf" };
+  }
+
+  @Override
+  protected String[] getPatterns() {
+    return new String[] { "knoxidf/api/**?**" };
+  }
```
smolnar82, line 47: `KNOXIDF_ADMIN` and `KNOXIDF` share the same package AND URL pattern. Both contributors return:
- `getPackages()` → `{"org.apache.knox.gateway.service.knoxidf"}`
- `getPatterns()` → `{"knoxidf/api/**?**"}`

Because Jersey scans the whole package, the `TrustedOidcIssuersResource` admin endpoints are served by the `KNOXIDF` role too, not just `KNOXIDF_ADMIN`.
If the intent is for admin operations to sit behind stricter authorization (a separate role usually implies a separate authz/ACL binding in the topology), that intent is defeated: the same resource is reachable through the non-admin role. Conversely, deploying both roles in one topology with the identical knoxidf/api/**?** pattern invites a routing conflict. The test even notes this (`"Same prefix … so both roles can coexist … Jersey disambiguates by @Path"`), but path-based disambiguation doesn't give you role-based access control.

Is `KNOXIDF_ADMIN` meant to be independently protected, and if so, shouldn't its package/pattern be scoped to the admin resources only?

gateway-service-knoxidf/src/main/java/org/apache/knox/gateway/service/knoxidf/TrustedOidcIssuersResource.java
```diff
@@ -0,0 +1,227 @@
...
+  @Consumes(MediaType.APPLICATION_JSON)
+  public Response registerIssuer(String body) {
+    String issuerUrl = "INVALID_REQUEST";
+    final String operatorId = getOperatorId();
+    String outcome = ActionOutcome.FAILURE;
+
+    try {
+      final Map<String, Object> parsed;
+      try {
+        parsed = new ObjectMapper().readValue(body, new TypeReference<Map<String, Object>>() {});
```
smolnar82, line 95: `ObjectMapper` is expensive to construct and thread-safe to reuse: make it a `private static final`, or parse via `JsonUtils` for consistency with the rest of the class.

smolnar82 (changes requested):

smolnar82:
Taken care by #1327

PR #1322, opened 2026-07-23
KNOX-3390: Moved gateway-level config to GatewayConfig and implemented the missing methods

smolnar82:
[KNOX-3390](https://issues.apache.org/jira/browse/KNOX-3390) - Moved gateway-level config to GatewayConfig and implemented the missing methods

There is a missing items in #1320 which moves new gateway-level config from KnoxIDFConstants to where they belong -> `GatewayConfig` along with their own methods.
This PR adds these changes.

## How was this patch tested?

Ran unit tests (which I updated where it was necessary).

## Integration Tests
N/A

## UI changes
N/A

### DIFF
--- a/gateway-server/src/test/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/JdbcTrustedOidcIssuerServiceTest.java
+++ b/gateway-server/src/test/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/JdbcTrustedOidcIssuerServiceTest.java
@@ -17,7 +17,6 @@
 package org.apache.knox.gateway.services.knoxidf.trustedoidcissuer;
 
 import org.apache.knox.gateway.config.GatewayConfig;
-import org.apache.knox.gateway.config.impl.GatewayConfigImpl;
 import org.apache.knox.gateway.database.AbstractDataSourceFactory;
 import org.apache.knox.gateway.database.DatabaseType;
 import org.apache.knox.gateway.services.ServiceLifecycleException;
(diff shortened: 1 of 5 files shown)


smolnar82:
Cc. @hsheinblatt

hanicz (approved):

gateway-server/src/test/java/org/apache/knox/gateway/services/knoxidf/trustedoidcissuer/JdbcTrustedOidcIssuerServiceTest.java
```diff
@@ -234,13 +234,9 @@ public void testMaxTrustedIssuers() throws ServiceLifecycleException {
     limitedService.register(issuer("https://b.example.com", false));
     assertEquals("Second registration must succeed", 2, limitedService.list().size());
 
-    try {
```
hsheinblatt, line 237: Generally I prefer to use the try/catch with fail rather than an expected exception in any test that makes more than one call. In this case there are several calls to limitedService. With the general expected exception, you don't know where the exception came from, it could have happened well before you intend.

smolnar82: You are right, let me fix this.

smolnar82: Opened a follow-up PR for this: #1326

PR #1327, opened 2026-07-24
KNOX-3368 - Trusted OIDC Issuer admin API for Knox IDF

hsheinblatt:
KNOX-3368 - Trusted OIDC Issuer admin API for Knox IDF

Implement the admin API for the trusted OIDC issuer service
along with the contributor for it.

## How was this patch tested?

Unit tests were added

## Integration Tests
N/A will be added once the flow can be tested.

### DIFF
--- a/gateway-service-knoxidf/src/main/java/org/apache/knox/gateway/service/knoxidf/TrustedOidcIssuersResource.java
+++ b/gateway-service-knoxidf/src/main/java/org/apache/knox/gateway/service/knoxidf/TrustedOidcIssuersResource.java
@@ -54,13 +54,11 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import static org.apache.knox.gateway.util.knoxidf.KnoxIDFConstants.BASE_RESORCE_PATH;
-
 @Path(TrustedOidcIssuersResource.RESOURCE_PATH)
 @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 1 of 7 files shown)


smolnar82 (approved):

smolnar82:
There is one small change I'll open a follow-up PR for:
in line 122 you catch RuntimeException:
```
    } catch (RuntimeException e) {
      return errorResponse(Response.Status.INTERNAL_SERVER_ERROR, "storage_error",
          "Failed to register issuer");  <-- This would indicate either a JDBC issue or if max limit exceeded
```
I think it's better to separate those 2 and return a clean error message to the clients in case they face the limit.

Another finding (sort of related):

Untyped `Map<String,Object> + blind casts` causes client type errors surface as 500s. This is the main one. Lines 100/116:
```
final String rawUrl = (String) parsed.get("issuerUrl");
...
final String clusterName = (String) parsed.get("clusterName");
```
A parseable-but-wrong-typed body like `{"issuerUrl": 123}` or `{"clusterName": {}}` throws `ClassCastException` -> caught by `catch (RuntimeException)` at line 122 -> `500 storage_error "Failed to register issuer"`. That's a client mistake (should be `400 invalid_request`) misreported as a server storage failure. A typed request DTO deserialized by JAX-RS (`RegisterIssuerRequest`, for instance) would eliminate the casts, give automatic `400`s on type mismatch, and read more idiomatically. The current hand-rolled `Map` parse is the root of this.

PR #1328, opened 2026-07-24
KNOX-3396: Return 4xx for type-mismatched register bodies and issuer-limit-reached instead of 500

smolnar82:
[KNOX-3396](https://issues.apache.org/jira/browse/KNOX-3396) - Fix trusted-OIDC-issuer register API returning 500 for client-actionable errors

Follow-up to KNOX-3368 (#1327), hardening the `POST` register endpoint of the trusted-OIDC-issuer admin API so that client-actionable conditions return the correct `4xx` status instead of a misleading `500`:

1. **Typed request bean.** `registerIssuer` now deserializes the body into a new `RegisterIssuerRequest` DTO instead of a generic `Map<String, Object>` with blind casts.
2. **Differentiated exception handling.** The service throws `IllegalStateException` when the configured `MAX_TRUSTED_ISSUERS` limit is reached: an operator-facing capacity condition. This is now caught separately and mapped to `409 issuer_limit_reached`, so it is distinguishable from a genuine storage failure (still `500 storage_error`).

Parsing remains inside the `try` block, so the audit-on-every-path behavior and the malformed-JSON handling are unchanged. Validation ordering, the SSRF/HTTPS gate, the error-response contract, and the other endpoints `GET`/`DELETE`/`refresh-jwks`) are untouched.

## How was this patch tested?

Automated unit tests in `TrustedOidcIssuersResourceTest` (module `gateway-service-knoxidf`). Two cases were added:
- `testRegisterWrongTypeFieldReturnsBadRequest`: an array-valued `clusterName` now returns `400 invalid_request` (was `500` before the fix).
- `testRegisterIssuerLimitReached`: the service throwing `IllegalStateException` now returns `409 issuer_limit_reached`.

## Integration Tests
N/A

## UI changes
N/A

### DIFF
--- a/gateway-service-knoxidf/pom.xml
+++ b/gateway-service-knoxidf/pom.xml
@@ -77,10 +77,6 @@
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-core</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
         <dependency>
--- a/gateway-service-knoxidf/pom.xml
+++ b/gateway-service-knoxidf/pom.xml
@@ -73,6 +73,10 @@
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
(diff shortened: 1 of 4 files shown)


smolnar82:
Cc. @hsheinblatt

hanicz (approved):

hsheinblatt (approved):

PR #1331, opened 2026-07-29
KNOX-3386: LDAP Proxy pages backends

handavid:
[KNOX-3386](https://issues.apache.org/jira/browse/KNOX-3386) - LDAP Proxy pages backends

Adds a gateway.ldap.max.size.limit and max.time.limit configs to the LDAP Proxy. Thes configs are uses to set the LDAP server's maxSizeLimit and maxTimeLimit fields. The size limit indictes the maximum total number of search results for a request across all pages. The time limit is in milliseconds.

Implements paging to the backend LDAP servers. This is controlled by a new "pageSize" config on the LDAP Proxy Backends.

A bug was found during testing where the getUserGroups method will doubly retrieve the groups. This has been fixed.

## How was this patch tested?

Unit tests were added/updated in LdapProxyBackendTest and LdapProxyBackendSslTest

manual testing was performed against AD by setting the following values in the gateway-site.xml
```
    <property>
        <name>gateway.ldap.sizelimit</name>
        <value>5000</value>
        <description>Base DN for LDAP entries in the proxy server. Default is dc=proxy,dc=com.</description>
    </property>

    <property>
        <name>gateway.ldap.interceptor.localldap.pageSize</name>
        <value>3</value>
    </property>
    <property>
        <name>gateway.ldap.interceptor.ad.pageSize</name>
        <value>3</value> 
    </property>
```
Logging was increased to DEBUG level to see the paging-related log messages.

## Integration Tests
(Please add or update integration tests [`.github/workflows/tests`](.github/workflows/tests) for the feature you are adding. If no unit test is added, please explain why. Check out [`.github/workflows/tests/README.md`](./workflows/tests/README.md) for instructions)

The gateway-site was modified for the workflow tests to set a small page size. This ensure that the proxy needs to page to return the results in the tests.

## UI changes
none

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
@@ -74,8 +74,8 @@ public interface LdapMessages {
     void ldapInterceptorCreating(String interceptorName, String source);
 
     @Message(level = MessageLevel.INFO,
-            text = "Configuring LDAP interceptor {0}: {1} = {2})")
-    void ldapInterceptorConfiguring(String interceptorNamee, String configName, String configValue);
+            text = "Configuring LDAP interceptor {0}: {1} = {2}")
+    void ldapInterceptorConfiguring(String interceptorName, String configName, String configValue);
 
     @Message(level = MessageLevel.INFO,
             text = "Loading backend: {0} (via {1})")
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
@@ -101,6 +101,14 @@ public interface LdapMessages {
             text = "LDAP Search: {0} | {1}")
     void ldapSearch(String baseDn, String filter);
 
+    @Message(level = MessageLevel.DEBUG,
+            text = "LDAP Paged Search: {0} | {1}, page size {2}, page {3}")
+    void ldapPagedSearch(String baseDn, String filter, int pageSize, int pageNumber);
+
+    @Message(level = MessageLevel.DEBUG,
+            text = "LDAP Paged Search Completed: {0} | {1}")
+    void ldapPagedSearchCompleted(String baseDn, String filter);
+
     @Message(level = MessageLevel.ERROR,
             text = "LDAP Search failed: {0} | {1}, {2}")
     void ldapSearchFailed(String baseDn, String filter, @StackTrace(level = MessageLevel.DEBUG) Exception e);
@@ -133,9 +141,9 @@ public interface LdapMessages {
             text = "Backend user not found: {0}")
     void ldapUserNull(String username);
 
-    @Message(level = MessageLevel.ERROR,
+    @Message(level = MessageLevel.DEBUG,
             text = "Failed to copy attribute: {0}")
-    void ldapAttributeCopyError(@StackTrace(level = MessageLevel.DEBUG) Exception e);
+    void ldapAttributeCopyError(@StackTrace(level = MessageLevel.TRACE) Exception e);
 
     @Message(level = MessageLevel.DEBUG, text = "LDAP authentication succeeded for user: {0}")
     void ldapAuthSucceeded(String user);
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
@@ -74,6 +74,10 @@ public interface LdapMessages {
     void ldapInterceptorCreating(String interceptorName, String source);
 
     @Message(level = MessageLevel.INFO,
+            text = "Configuring LDAP interceptor {0}: {1} = {2})")
+    void ldapInterceptorConfiguring(String interceptorNamee, String configName, String configValue);
+
+    @Message(level = MessageLevel.INFO,
             text = "Loading backend: {0} (via {1})")
     void ldapBackendLoading(String backendName, String source);
 
@@ -104,6 +108,10 @@ public interface LdapMessages {
     @Message(level = MessageLevel.DEBUG,
             text = "LDAP Paged Search: {0} | {1}, page size {2}, page {3}")
     void ldapPagedSearch(String baseDn, String filter, int pageSize, int pageNumber);
+
+    @Message(level = MessageLevel.ERROR,
+            text = "LDAP Paged Search Exceeded Max Result Set Size: {0} | {1}")
+    void ldapPagedSearchExceededMaxResultSetSize(int resultSetSize, int maxResultSetSize);
 
     @Message(level = MessageLevel.DEBUG,
             text = "LDAP Paged Search Completed: {0} | {1}")
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
@@ -104,6 +104,7 @@ public class LdapProxyBackend implements LdapBackend {
     private boolean recursiveGroupResolution;
     private int recursiveGroupResolutionMaxDepth;
     private int pageSize;
+    private int maxResultSetSize;
 
     private final String proxyEntryGroupMembershipAttributeType = "memberOf";
 
@@ -194,6 +195,8 @@ public class LdapProxyBackend implements LdapBackend {
 
         // Configure search parameters
         pageSize = Integer.parseInt(config.getOrDefault("pageSize", "1000"));
+        maxResultSetSize = Integer.parseInt(config.getOrDefault("maxResultSetSize", "0")); // 0 means unlimited
+        LOG.ldapInterceptorConfiguring(name, "maxResultSetSize", Integer.toString(maxResultSetSize));
 
         // Configure secure transport (LDAPS) to the remote server. An ldaps:// URL enables it
         // by default; an explicit useSsl setting always wins.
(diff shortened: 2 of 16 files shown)


handavid:
@smolnar82

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
```diff
@@ -511,12 +521,10 @@ public List<Entry> searchUsers(String filter, SchemaManager schemaManager) throw
         try {
             connection = getConnection();
             String ldapFilter = "(" + remoteUserIdentifierAttribute + "=" + filter.trim() + ")";
-            try (EntryCursor cursor = connection.search(remoteUserSearchBase, ldapFilter, SearchScope.SUBTREE, "*")) {
-                while (cursor.next()) {
-                    Entry sourceEntry = cursor.get();
-                    addGroupMemberships(sourceEntry, connection, entryCache, resolvedParentsCache);
-                    results.add(remoteSchemaConverter.convertRemoteEntryToProxyEntry(sourceEntry, schemaManager));
-                }
+            List<Entry> searchResults = performPagedSearch(connection, remoteUserSearchBase, ldapFilter, SearchScope.SUBTREE, "*");
```
smolnar82, line 527: Previously `searchUsers` had no `try/catch` around its cursor, so an `LdapException` propagated to the caller. Now it routes through `performPagedSearch`, which catches and logs, returning partial/empty results. So a backend failure during user search now looks like "no users found" rather than an error. `search()` already behaved this way, so this may be intentional. But it's a behavior change for `searchUsers`. Please confirm silent degradation is desired there (auth/lookup failures being indistinguishable from empty results can be a real operational issue).

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackendTest.java
```diff
@@ -809,4 +874,28 @@ public Set<String> get(Object key) {
...
+        }
+
+        public List<SearchRequest> getRequests() {
+            return List.copyOf(requests);
+        }
+
+        @Override
+        public void handle(LdapSession session, SearchRequest message) throws Exception {
+            requests.add(message);
+            System.out.println(message.toString());
```
smolnar82: nit: this should be removed.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
```diff
@@ -133,9 +141,9 @@ public interface LdapMessages {
             text = "Backend user not found: {0}")
     void ldapUserNull(String username);
 
-    @Message(level = MessageLevel.ERROR,
+    @Message(level = MessageLevel.DEBUG,
             text = "Failed to copy attribute: {0}")
-    void ldapAttributeCopyError(@StackTrace(level = MessageLevel.DEBUG) Exception e);
+    void ldapAttributeCopyError(@StackTrace(level = MessageLevel.TRACE) Exception e);
```
smolnar82, line 154: Why are these changes needed here? Shouldn't this remain on `ERROR` level instead of `DEBUG`? We might hide real conversion problems.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
```diff
@@ -839,15 +841,67 @@ private List<Entry> getUserGroupsInternal(LdapConnection connection, Dn... dns)
...
+
+                    if (responseControl != null) {
+                        cookie = responseControl.getCookie();
+                    } else {
+                        cookie = null;
+                    }
+                }
+                pageNumber++;
+            } catch (LdapException e) {
+                LOG.ldapSearchFailed(baseDn, filter, e);
```
smolnar82: Here we log the failure but do not clear cookie or break the loop. As a result, it might end up in an infinite loop: `page 1` succeeds and returns a non-empty cookie. So the loop continues; `page 2`'s `connection.search(...)` (or cursor iteration) throws `LdapException` (server restart, timeout, connection drop mid-paging). The exception is caught, `cookie` still holds page 1's value, `pageNumber++` is skipped, and the loop condition is still `true`. It retries the identical failing page forever.

The old code (`search(...)`) caught `LdapException` around a single cursor and simply stopped with partial results. To preserve that, the catch should terminate the loop, e.g.:
```
} catch (LdapException e) {
    LOG.ldapSearchFailed(baseDn, filter, e);
    break; // or: cookie = null;
}
```
What do you think?

smolnar82 (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
handavid: good catch. I think that I'll have search propagate the exception also. I don't think it makes sense to return empty if there is an exception in all cases.

gateway-server/src/test/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackendTest.java
handavid: done

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
handavid: we get these messages because the entry returned from the backend holds attributes that are not in the proxy's schema. This isn't an "error", but rather an artifact of the design. At error level, the logs get flooded with 10s of lines per entry, which would make the log unusable.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
handavid: I removed the catch to let the calling methods handle or propagate the exception. This breaks the loop.

handavid:
@smolnar82 ready for re-review

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
smolnar82: Makes sense, thanks for the detailed explanation.

knox-site/docs/service_ldap_server.md
```diff
@@ -44,6 +44,8 @@ The service is configured in `gateway-site.xml`.
 | `gateway.ldap.roles.lookup.strategy` | N/A | The LDAP roles lookup strategy (`file` or `rest`). |
 | `gateway.ldap.roles.lookup.rest.api.endpoint` | N/A | The LDAP roles lookup REST API endpoint. |
 | `gateway.ldap.roles.lookup.file.path` | N/A | The LDAP roles lookup file path. |
+| `gateway.ldap.roles.max.size.limit` | 1000 | The maximum size limit of the result set returned by search requests. |
+| `gateway.ldap.roles.max.time.limit` | 60000 | The maximum time limit for search requests in milliseconds. |
```
smolnar82: Should be `gateway.ldap.max.size.limit` / `gateway.ldap.max.time.limit`.

knox-site/docs/service_ldap_server.md
```diff
@@ -198,6 +200,10 @@ The proxy backend delegates lookups to a remote LDAP or Active Directory server.
 | `gateway.ldap.interceptor.<name>.groupMemberAttribute` | `memberUid` | Attribute used for group membership (e.g., `member` for AD). |
 | `gateway.ldap.interceptor.<name>.useMemberOf` | `false` | If `true`, use the `memberOf` attribute for efficient group lookups. |
 | `gateway.ldap.interceptor.<name>.proxy.poolMaxActive` | `8` | Maximum number of active connections in the pool. |
+| `gateway.ldap.interceptor.<name>.pageSize` | `1000` | Page size for search requests. |
+| `gateway.ldap.interceptor.<name>.maxResultSetSize` | `0` | Maximum number of results to return from a search, regardless of paging. 0 means unlimited. |
+
+NOTE: If this value is undefined and the interceptor was created by the KnoxLDAPServerManager, the KnoxLDAPServerManager will set the `gateway.ldap.interceptor.<name>.maxResultSetSize` value to be 1 greater than the proxy's `gateway.ldap.roles.max.time.limit` configuration. This will ensure that the proxy returns a "Size limit exceeded" result if the backend has more results than the proxy's limit.
```
smolnar82: I think you meant `be 1 greater than the proxy's gateway.ldap.max.size.limit configuration`.
See `time` vs `size` and we don't need the `roles` prefix either (see my comment above).

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
```diff
@@ -73,6 +73,10 @@ public interface LdapMessages {
             text = "Creating LDAP interceptor: {0} (via {1})")
     void ldapInterceptorCreating(String interceptorName, String source);
 
+    @Message(level = MessageLevel.INFO,
+            text = "Configuring LDAP interceptor {0}: {1} = {2})")
```
smolnar82: See the additional `)` at the ned of the message. Please remove it.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
```diff
@@ -73,6 +73,10 @@ public interface LdapMessages {
...
+    void ldapInterceptorConfiguring(String interceptorNamee, String configName, String configValue);
```
smolnar82: nit: typo in the parameter name `interceptorNamee`

smolnar82 (changes requested):

knox-site/docs/service_ldap_server.md
handavid: fixed

knox-site/docs/service_ldap_server.md
handavid: fixed

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
handavid: fixed

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/LdapMessages.java
handavid: fixed

smolnar82 (approved):

PR #1333, opened 2026-07-29
KNOX-3400 - Bump up derby version from

moresandeep:


pzampino (approved):
LGTM

pom.xml
```diff
@@ -193,7 +193,7 @@
         <cryptacular.version>1.3.0</cryptacular.version>
         <curator.version>5.4.0</curator.version>
         <dependency-check-maven.version>6.0.3</dependency-check-maven.version>
-        <derby.db.version>10.14.2.0</derby.db.version> <!-- 10.15.1.3 requires Java 9 -->
+        <derby.db.version>10.17.1.0</derby.db.version>
```
smolnar82, line 196: org.apache.derby.jdbc.EmbeddedDataSource still exists in Apache Derby 10.17, but it is no longer in the derby artifact.
Starting with Derby 10.15 (to support the Java Platform Module System), the DataSource implementations were moved from derby.jar to derbytools.jar while keeping the same package name.

smolnar82 (changes requested):

pom.xml
moresandeep: The problem is Derby 10.17.1.0 is compiled for Java 19 (class-file version 63) and requires a Java 21 runtime. Knox targets Java 17. Java 17 compiler can't read Derby 10.17's class files, producing:
```
cannot access org.apache.derby.jdbc.EmbeddedDataSource
  bad class file: derbytools-10.17.1.0.jar(.../EmbeddedDataSource.class)
    class file has wrong version 63.0, should be 61.0
```
Derby's JDK requirements are: 10.15 --> Java 9, 10.16.x --> Java 17, 10.17.x ---> Java 21. So 10.17.1.0 is too new. 

The problem is that 10.16.x line has issues and so does 10.14.x

PR #1339, opened 2026-08-06
KNOX-3408 - Allow no actor token in JWTFederationFilter.handleTokenExchange

hsheinblatt:
KNOX-3408 - Allow no actor token in JWTFederationFilter.handleTokenExchange

RFC 8693 requires that the actor token be optional for token exchange.
Change the handleTokenExchange method to allow a missing actor token.
When actor_token is absent, the Subject has the subject from the subject token
as PrimaryPrincipal with no TokenExchangePrincipal. Note that headless
delegation using ImpersonatedPrincipal is currently not represented in this path.

## How was this patch tested?

A unit test for the new path is added. Regression tests for subject handling in
JWTFederationFilter.handleTokenExchange and TokenExchangePrincipal handling in AbstractIdentityAssertionFilter.continueChainAsPrincipal handling were also
added and run before the feature change to show no regression.

## Integration Tests
N/A


pzampino (approved):
LGTM

PR #1342, opened 2026-08-11
KNOX-3405 - Extend JWTFederationFilter for dynamic JWKS and iss attribute on token-exchange

hsheinblatt:
KNOX-3405 - Extend JWTFederationFilter for dynamic JWKS and iss attribute on token-exchange

JWTFederationFilter and the base AbstractJWTFilter are modified to use dynamically
discovered JWKS for validating token signatures for RFC 8693 token exchange.

AbstractJWTFilter.validateToken first checks if the issuer is an expected issuer.
That is, defined in the static topology constant list of allowed issuers. This is
the pre-existing path. If the token issuer is in the expected issuer list, then the
logic is unchanged, the token validate proceeds as it had.

If the token issuer is not part of the expected list, then it is checked against the
TrustedOIDCIssuer registry, if it's enabled. If it matches, then the JWKS dynamically
discovered for that issuer is used to validate the token signature. The rest of the token
validation is the same as for the existing case. That is, the other fields like expiry and nbf
remain validated in the path with the same logic.

JWTFederationFilter was also modified to store the validated token's iss claim as a request
attribute for downstream handlers.

Some RFC 8693 token exchange related constants were moved from TokenExchangeHandler
to JWTFederationFilter so that the tests outside the package, namely
JWTFederationFilterTokenExchangeTest, can read the values.

## How was this patch tested?

Unit tests were added for the JWKS dynamic discovery for trusted OIDC issuers paths.
These include success paths where the token signature is validated using the dynamically
discovered JWKS and a similar test with both actor and subject tokens where one is from
an issuer in the configured list and the other is discovered dynamically. Negative tests
for invalid tokens are added to show that signature validation must pass, and the same
checks conducted for the static issuer path is enforced in the dynamic issuer path as well,
such as expiry and nbf. Tests are added for paths where the token issuer is not in the
static list, but also not valid for dynamic discovery. These include an unregistered issuer--
one not registered as a trusted OIDC issuer--, when the TrustedOIDCIssuer
service is not configured, and when the request is not a token exchange request. Unit
tests were also added to show that token validation for an issuer that is statically defined
will not fail over to a dynamically registered issuer path, the existing logic is used when
the issuer is a statically defined issuer regardless of whether or not it's also a dynamically
trusted issuer, a sanity that the new dynamic path is not called in a successful
validation using the existing statically defined issuer path, and the logic for a non-token
exchange request does not use the new path.

A unit test was added for the iss request claim added.

## Integration Tests
No integration tests were added. They will be once the full flow is implemented.

## UI changes
N/A

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
@@ -469,65 +469,130 @@ public abstract class AbstractJWTFilter implements Filter {
     final String tokenId = TokenUtils.getTokenId(token);
     final String displayableTokenId = Tokens.getTokenIDDisplayText(tokenId);
     final String displayableToken = Tokens.getTokenDisplayText(token.toString());
-    // confirm that issuer matches the intended target
     if (expectedIssuers.contains(token.getIssuer())) {
-      // if there is no expiration data then the lifecycle is tied entirely to
-      // the cookie validity - otherwise ensure that the current time is before
-      // the designated expiration time
-      try {
-        if (tokenIsStillValid(token)) {
-          boolean audValid = validateAudiences(token);
-          if (audValid) {
-            Date nbf = token.getNotBeforeDate();
-            if (nbf == null || new Date().after(nbf)) {
-              final TokenMetadata tokenMetadata = tokenStateService == null ? null : tokenStateService.getTokenMetadata(tokenId);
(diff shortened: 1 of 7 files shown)


smolnar82:
> Some RFC 8693 token exchange related constants were moved from TokenExchangeHandler
to JWTFederationFilter so that the tests outside the package, namely
JWTFederationFilterTokenExchangeTest, can read the values.

I think this is the wrong approach: those constants should be in `TokenExchangeHandler` and the test should have been added in the same package (now everything is in the `...filter` package, but the test is outside. 

I filed a follow-up JIRA to resolve this: https://issues.apache.org/jira/browse/KNOX-3415

PR #1353, opened 2026-08-20
KNOX-3421 - Upgrade Shiro to 2.2.1

moresandeep:

[Knox 3421](https://issues.apache.org/jira/browse/KNOX-3421) - Upgrade Shiro to 2.2.1

- Update Shiro library (1.13.1) from EOLd to one that is supported (2.2.1)
- Code changes due to the upgrade.

## How was this patch tested?

- This patch was tested locally in a test env.
- Unit tests are added as part of this PR
- New docker based tests added to catch regression and test the upgrade.
- 
## Integration Tests
- New integration test added `test_knox_ldap_dn_injection.py`


gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
```diff
@@ -1866,9 +1865,7 @@ private char[] getSystemPassword(Topology t) throws NoSuchProviderException, Mis
     protected Subject getSubject(Ini config) throws BadSubjectException {
       try {
         ThreadContext.unbindSubject();
-        @SuppressWarnings("deprecation")
-        Factory factory = new IniSecurityManagerFactory(config);
-        org.apache.shiro.mgt.SecurityManager securityManager = (org.apache.shiro.mgt.SecurityManager) factory.getInstance();
+        org.apache.shiro.mgt.SecurityManager securityManager = new BasicIniEnvironment(config).getSecurityManager();
```
hanicz, line 1872: BasicIniEnvironment should be destroyed after use, this might be a leak.

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxLdapRealm.java
```diff
@@ -744,16 +752,15 @@ protected String getUserDn( final String principal ) throws IllegalArgumentExcep
...
+      final ByteSource credentialsSalt = new SecureRandomNumberGenerator().nextBytes();
+      final SimpleHash credentialsHash = new SimpleHash(HASHING_ALGORITHM, token.getCredentials(), credentialsSalt, HASHING_ITERATIONS);
+      return new SimpleAuthenticationInfo(token.getPrincipal(), credentialsHash.toHex(), credentialsSalt, getName());
     }
 
-  private static String expandTemplate(final String template, final Matcher input) {
-    return expandTemplate(template, input, false);
-  }
+  /** How a substituted template value must be escaped for its target context. */
+  private enum EscapeMode { NONE, FILTER, DN }
```
hanicz: nit: I think this should be moved to the beginning of the class. Its more readable since we use it before its introduction. Also NONE is never used.

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxLdapRealm.java
```diff
@@ -762,8 +769,10 @@ private static String expandTemplate( final String template, final Matcher input
       String lookupValue = input.group( lookupIndex );
       if (lookupValue == null) {
           lookupValue = "";
-      } else if (escapeForLdapFilter) {
+      } else if (escapeMode == EscapeMode.FILTER) {
           lookupValue = escapeLdapSearchFilterValue(lookupValue);
+      } else if (escapeMode == EscapeMode.DN) {
+          lookupValue = escapeDnValue(lookupValue);
```
hanicz, line 781: If someone provides {0} as username it results in an infinite loop which could lead to denial of service attacks.

hanicz (changes requested):
Did you test caching as well?

gateway-server/src/main/java/org/apache/knox/gateway/util/KnoxCLI.java
moresandeep: I'll get it fixed!

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxLdapRealm.java
moresandeep: You are right, that is clearer way.

gateway-provider-security-shiro/src/main/java/org/apache/knox/gateway/shirorealm/KnoxLdapRealm.java
moresandeep: Agreed, this is indeed an issue :( I'll followup with a patch!

moresandeep:
Thanks for the review @hanicz, i addressed the comments and also added docker tests for LDAP caching.

hanicz (approved):

PR #1354, opened 2026-08-21
KNOX-3423: Switch to JSON error messages in Knox token validation paths

hanicz:
[KNOX-3423](https://issues.apache.org/jira/browse/KNOX-3423) - Switch to JSON error messages in Knox token validation paths

RFC 8693 §2.2.2 / RFC 6749 §5.2 require OAuth token-exchange errors to be a JSON
body (`{"error", "error_description"}`) with `Content-Type: application/json`.
`JWTFederationFilter` previously emitted plain-text/HTML (`sendError`) for every
path, breaking standards-compliant clients.

- Errors raised while handling a **token-exchange request** now return the RFC JSON
  body with `Cache-Control: no-store` / `Pragma: no-cache`. HTTP status codes are
  unchanged (400 for param/nbf/audience, 401 for signature/expiry/parse).
- Scope is the **whole exchange path**, keyed off the existing
  `TOKEN_EXCHANGE_REQUEST_ATTR`: both the direct param/parse errors in
  `TokenExchangeHandler` **and** the shared JWT-validation errors emitted by
  `AbstractJWTFilter` (expired / bad-signature / wrong-audience / …). The latter is
  the gap a handler-only fix would miss.
- The normal **bearer** auth path and the **SSO-cookie** path are untouched — still
  plain text.

## How was this patch tested?

Unit tests, local tests

```
curl -skiv -X POST -H 'Content-Type: application/x-www-form-urlencoded' \
  --data 'grant_type=urn:ietf:params:oauth:grant-type:token-exchange' \
  'https://localhost:8443/gateway/knoxidf/knoxidf/api/v1/token'

HTTP/1.1 400 Bad Request
Content-Type: application/json;charset=utf-8
{"error":"invalid_request","error_description":"the subject_token parameter is required"}
```

```
curl -skiv -H 'Authorization: Bearer not-a-real-jwt' \
  'https://localhost:8443/gateway/tokenconsumer/auth/api/v1/pre'

Content-Type: text/html;charset=iso-8859-1
<body><h2>HTTP ERROR 401 Bad request: missing token passcode.</h2>
```

```
curl -skiv -X POST -H 'Content-Type: application/x-www-form-urlencoded' \
  --data 'grant_type=urn:ietf:params:oauth:grant-type:token-exchange&subject_token=abc&subject_token_type=urn:ietf:params:oauth:token-type:saml2' \
  'https://localhost:8443/gateway/tokenconsumer/auth/api/v1/pre'

HTTP/1.1 400 Bad Request
Content-Type: application/json;charset=utf-8
{"error":"unsupported_token_type","error_description":"unsupported subject_token_type urn:ietf:params:oauth:token-type:saml2"}
```

## Integration Tests
N/A

## UI changes
N/A

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
@@ -92,35 +92,35 @@ class TokenExchangeHandler {
     // RFC 8693 section 2.1: subject_token and subject_token_type are REQUIRED.
     if (subjectTokenValue == null || subjectTokenValue.isEmpty()) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "invalid_request: the subject_token parameter is required");
+          "invalid_request", "the subject_token parameter is required");
       return;
     }
     if (subjectTokenType == null || subjectTokenType.isEmpty()) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "invalid_request: the subject_token_type parameter is required");
+          "invalid_request", "the subject_token_type parameter is required");
       return;
     }
     // RFC 8693 section 2.1: actor_token_type is REQUIRED when actor_token is present and MUST NOT
     // be present otherwise.
     if (hasActorToken && !hasActorTokenType) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "invalid_request: actor_token_type is required when actor_token is present");
+          "invalid_request", "actor_token_type is required when actor_token is present");
       return;
     }
     if (!hasActorToken && hasActorTokenType) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "invalid_request: actor_token_type must not be present without actor_token");
+          "invalid_request", "actor_token_type must not be present without actor_token");
       return;
     }
     // Only JWT-family token types are supported.
     if (isNotSupportedTokenType(subjectTokenType)) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "unsupported_token_type: unsupported subject_token_type " + subjectTokenType);
+          "unsupported_token_type", "unsupported subject_token_type " + subjectTokenType);
       return;
     }
     if (hasActorToken && isNotSupportedTokenType(actorTokenType)) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "unsupported_token_type: unsupported actor_token_type " + actorTokenType);
+          "unsupported_token_type", "unsupported actor_token_type " + actorTokenType);
       return;
     }
 
@@ -148,7 +148,7 @@ class TokenExchangeHandler {
       filter.continueWithEstablishedSecurityContext(subject, request, response, chain);
     } catch (ParseException | UnknownTokenException e) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_UNAUTHORIZED,
-          "Failed to parse token in token exchange: " + e.getMessage());
+          "invalid_grant", "Failed to parse token in token exchange: " + e.getMessage());
     }
   }
 
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
@@ -115,12 +115,12 @@ class TokenExchangeHandler {
     // Only JWT-family token types are supported.
     if (isNotSupportedTokenType(subjectTokenType)) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "unsupported_token_type", "unsupported subject_token_type " + subjectTokenType);
+          "invalid_request", "unsupported subject_token_type " + subjectTokenType);
       return;
     }
     if (hasActorToken && isNotSupportedTokenType(actorTokenType)) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "unsupported_token_type", "unsupported actor_token_type " + actorTokenType);
+          "invalid_request", "unsupported actor_token_type " + actorTokenType);
       return;
     }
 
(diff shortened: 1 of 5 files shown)


gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandlerTest.java
```diff
@@ -113,7 +118,8 @@ public void testUnsupportedActorTokenType() throws Exception {
     filter.valid.put("subtok", jwt("alice", "KNOXSSO"));
     handler.handle(request("subtok", JWT_TYPE, "acttok", SAML2_TYPE), response, chain);
     assertEquals(HttpServletResponse.SC_BAD_REQUEST, filter.errorStatus);
-    assertTrue(filter.errorMessage.contains("unsupported_token_type"));
+    assertEquals("unsupported_token_type", filter.error);
```
hsheinblatt: Shouldn't this be invalid_request? unsupported_token_type isn't allowed as an error field, I believe: https://www.rfc-editor.org/info/rfc6749/#section-5.2

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
```diff
@@ -92,35 +92,35 @@ void handle(HttpServletRequest request, HttpServletResponse response, FilterChai
...
     if (isNotSupportedTokenType(subjectTokenType)) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "unsupported_token_type: unsupported subject_token_type " + subjectTokenType);
+          "unsupported_token_type", "unsupported subject_token_type " + subjectTokenType);
       return;
     }
     if (hasActorToken && isNotSupportedTokenType(actorTokenType)) {
       filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
-          "unsupported_token_type: unsupported actor_token_type " + actorTokenType);
+          "unsupported_token_type", "unsupported actor_token_type " + actorTokenType);
```
hsheinblatt: Isn't this supposed to be invalid_request not unsupported_token_type (https://www.rfc-editor.org/info/rfc6749/#section-5.2)? I realized this below, so same comment on the test for this.

gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandlerTest.java
```diff
@@ -88,23 +90,26 @@ public void testActorTokenTypeRequiredWhenActorPresent() throws Exception {
...
+    assertTrue(filter.errorDescription.contains("must not be present"));
     assertFalse(filter.continued);
   }
 
   @Test
   public void testUnsupportedSubjectTokenType() throws Exception {
     handler.handle(request("subtok", SAML2_TYPE, null, null), response, chain);
     assertEquals(HttpServletResponse.SC_BAD_REQUEST, filter.errorStatus);
-    assertTrue(filter.errorMessage.contains("unsupported_token_type"));
+    assertEquals("unsupported_token_type", filter.error);
```
hsheinblatt: Same here.

gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandlerTest.java
```diff
@@ -113,7 +118,8 @@ public void testUnsupportedActorTokenType() throws Exception {
     filter.valid.put("subtok", jwt("alice", "KNOXSSO"));
     handler.handle(request("subtok", JWT_TYPE, "acttok", SAML2_TYPE), response, chain);
     assertEquals(HttpServletResponse.SC_BAD_REQUEST, filter.errorStatus);
-    assertTrue(filter.errorMessage.contains("unsupported_token_type"));
+    assertEquals("unsupported_token_type", filter.error);
+    assertTrue(filter.errorDescription.contains("actor_token_type"));
     assertFalse(filter.continued);
   }
```
hsheinblatt, line 125: I think there are missing tests, at least for "Failed to parse token in token exchange", the error path through the `catch (ParseException | UnknownTokenException e)` block. I'd add that test as well.

gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/JWTFederationFilterTokenExchangeTest.java
```diff
@@ -324,19 +328,20 @@ public void testFutureNbfRejectedOnDynamicPath() throws Exception {
...
-    handler.doFilter(request, response, chain);
+    handler.doFilter(request, response.mock, chain);
 
     Assert.assertFalse(chain.doFilterCalled);
     if (capturedJwt.hasCaptured()) {
       Assert.assertEquals(EXTERNAL_ISSUER, capturedJwt.getValue().getIssuer());
     }
-    EasyMock.verify(mockAuth, issuerSvc, response);
+    Assert.assertEquals(HttpServletResponse.SC_BAD_REQUEST, response.status());
+    Assert.assertTrue(response.body(), response.body().contains("\"error\":\"invalid_request\""));
```
hsheinblatt, line 342: Shouldn't this be invalid_grant? The token is invalid:
`
         invalid_grant
               The provided authorization grant (e.g., authorization
               code, resource owner credentials) or refresh token is
               invalid, expired, revoked, does not match the redirection
               URI used in the authorization request, or was issued to
               another client.
`
nbf is kind of like a reverse expired, so I'd expect it to be that.

gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/JWTFederationFilterTokenExchangeTest.java
```diff
@@ -371,19 +376,20 @@ public void testAudienceMismatchRejectedOnDynamicPath() throws Exception {
...
```
hsheinblatt, line 390: Same here, shouldn't it be invalid_grant?

hsheinblatt (approved):
Thanks Tamás. LGTM. I think a couple of the error types might be incorrect, and there may be a missing test -- see comments -- but looks to make it spec compliant.

hanicz:
> Thanks Tamás. LGTM. I think a couple of the error types might be incorrect, and there may be a missing test -- see comments -- but looks to make it spec compliant.

I made some changes to address your comments.

`TokenExchangeHandler`: I switched `unsupported_token_type` to `invalid_request`. Catch block emits `invalid_request` as well now.
`TokenExchangeHandlerTest`: Added the missing test for the catch block.

As per [RFC 8693 §2.2.2](https://datatracker.ietf.org/doc/html/rfc8693#section-2.2.2): If the request itself is not valid or if either the subject_token or actor_token are invalid for any reason, or are unacceptable based on policy, the authorization server MUST construct an error response, as specified in [Section 5.2](https://www.rfc-editor.org/rfc/rfc6749#section-5.2) of [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)]. The value of the error parameter MUST be the invalid_request error code.

Due to the above `JWTFederationFilter` will send `invalid_request` for nbf/audience expiry/signature/parse. `invalid_grant` remains in KnoxIDF `authorization`/`refresh_token` grant paths

smolnar82 (approved):

PR #1355, opened 2026-08-24
KNOX-3406: LDAP Proxy backend improve error handling

handavid:
[KNOX-3406](https://issues.apache.org/jira/browse/KNOX-3406) - Fix LDAP Proxy error handling

The LdapProxyBackend.getUser code was modified to ensure that the search cursor is closed before attempting to augment the entry with group memberships to avoid having simultaneously open cursors.

The error handling in performPagedSearch was improved such that unsuccessful paging calls will result in exceptions for the caller to handle.

## How was this patch tested?

Workflow was run repeatedly to try reproducing the intermittent failures in the JIRA. Through this, I discovered the paging issue that I've fixed in this PR. Unit tests were added to cover the paged search changes.

## Integration Tests

no tests added

## UI changes

no UI changes

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
@@ -28,7 +28,10 @@ import org.apache.directory.api.ldap.model.entry.DefaultEntry;
 import org.apache.directory.api.ldap.model.entry.Entry;
 import org.apache.directory.api.ldap.model.entry.Value;
 import org.apache.directory.api.ldap.model.exception.LdapException;
+import org.apache.directory.api.ldap.model.exception.LdapOperationException;
+import org.apache.directory.api.ldap.model.message.LdapResult;
 import org.apache.directory.api.ldap.model.message.Response;
+import org.apache.directory.api.ldap.model.message.ResultCodeEnum;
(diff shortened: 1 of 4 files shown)


handavid:
@smolnar82 @hanicz

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
```diff
@@ -887,6 +892,11 @@ protected List<Entry> performPagedSearch(LdapConnection connection, String baseD
                 }
                 if (cursor.isDone()) {
                     SearchResultDone done = cursor.getSearchResultDone();
+
+                    LdapResult ldapResult = done.getLdapResult();
```
hanicz: `SearchResultDone done` can be null at this point due to errors. The `isDone()` docs don't mention its relation with `getSearchResultDone()` so this might be a genuine concern here.

hanicz (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/backend/LdapProxyBackend.java
handavid: added null-handling

handavid:
@hanicz I've updated PR per your comment. Thanks!

hanicz (approved):

PR #1357, opened 2026-08-26
KNOX-3422: Ldap Proxy used in-memory bind user

handavid:
[KNOX-3422](https://issues.apache.org/jira/browse/KNOX-3422) - Create LDAP Proxy bind user in-memory instead of in embedded LDAP

The bind user created from the gateway configuration is now created in-memory instead of being added to the embedded ldap server. This provides the benefit of being able to rotate the user just by changing the configuration. Previously, since the configured user is not tracked, there wasn't a way to automatically rotate the user.

## How was this patch tested?

Unit test were updated and run.

## Integration Tests

LDAP proxy search workflow tests were updated to use the configured bind user instead of a user in the remote LDAP.
A test was added to verify that anonymous bind does not work when the bind user is configured.

## UI changes

no ui changes

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
@@ -37,8 +37,11 @@ import org.apache.directory.server.core.DefaultDirectoryService;
 import org.apache.directory.server.core.api.DirectoryService;
 import org.apache.directory.server.core.api.DnFactory;
 import org.apache.directory.server.core.api.InstanceLayout;
+import org.apache.directory.server.core.api.InterceptorEnum;
 import org.apache.directory.server.core.api.interceptor.Interceptor;
 import org.apache.directory.server.core.api.schema.SchemaPartition;
+import org.apache.directory.server.core.authn.AuthenticationInterceptor;
+import org.apache.directory.server.core.authn.Authenticator;
 import org.apache.directory.server.core.partition.impl.btree.jdbm.JdbmPartition;
 import org.apache.directory.server.core.partition.ldif.LdifPartition;
 import org.apache.directory.server.ldap.LdapServer;
@@ -46,6 +49,7 @@ import org.apache.directory.server.protocol.shared.transport.TcpTransport;
 import org.apache.knox.gateway.config.GatewayConfig;
 import org.apache.knox.gateway.i18n.messages.MessagesFactory;
 import org.apache.knox.gateway.services.GatewayServices;
+import org.apache.knox.gateway.services.ldap.authn.InMemoryBindAuthenticator;
 import org.apache.knox.gateway.services.ldap.control.RolesLookupBypassControlFactory;
 import org.apache.knox.gateway.services.ldap.interceptor.InterceptorFactory;
 import org.apache.knox.gateway.services.ldap.interceptor.LDAPRolesLookupInterceptor;
@@ -56,6 +60,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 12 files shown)


handavid:
@smolnar82 @hanicz

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/authn/InMemoryBindAuthenticator.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public LdapPrincipal authenticate(BindOperationContext bindContext) throws LdapException {
+        Dn dn = bindContext.getDn();
+
+        // Check if authenticating as the configured in-memory user
+        if (bindDn.equals(dn)) {
+            byte[] passwordBytes = bindContext.getCredentials();
+            String password = new String(passwordBytes, StandardCharsets.UTF_8);
+            if (bindPassword.equals(password)) {
```
hanicz, line 53: I think if the password doesn't match this should throw a LdapAuthenticationException.

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
```diff
@@ -145,6 +151,19 @@ public void initialize(GatewayConfig config) throws Exception {
...
 
+    /**
+     * Validates that the bind dn, if configured, matches the base dn for the embedded
+     * LDAP service (e.g. {@code ou=system} or {@code ou=people,<baseDn>}).
+     */
+    private void validateBindUser() throws Exception {
+        if (!StringUtils.isBlank(bindUser)) {
+            Dn bindUserDn = new Dn(bindUser);
+            if (!(bindUserDn.isDescendantOf("ou=system") || bindUserDn.isDescendantOf("ou=people," + this.baseDn))) {
+                throw new IllegalArgumentException("Bind user must be a descendant of ou=system or ou=people" + this.baseDn);
```
hanicz: nit: missing `,` in message

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
```diff
@@ -435,22 +453,24 @@ private void createBaseEntriesForDn(SchemaManager schemaManager, String dn) thro
...
-            String rdnValue = bindDn.getRdn().getValue();
-            Entry bindEntry = new DefaultEntry(schemaManager);
-            bindEntry.setDn(bindDn);
-            bindEntry.add("objectClass", "top", "person", "organizationalPerson", "inetOrgPerson");
-            bindEntry.add("cn", rdnValue);
-            bindEntry.add("sn", rdnValue);
-            bindEntry.add("userPassword", bindPassword);
-            directoryService.getAdminSession().add(bindEntry);
+    private void configureInMemoryBindUser(String bindPassword) throws Exception {
+        String id = "inmemoryuser";
```
hanicz: nit: this is an unused variable

hanicz (changes requested):

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
handavid: removed

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/KnoxLDAPServerManager.java
handavid: added

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/authn/InMemoryBindAuthenticator.java
handavid: fixed

handavid (commented):
@hanicz I addressed you comments. thanks!

hanicz (approved):

PR #1358, opened 2026-08-26
KNOX-3425: LDAP Proxy disallows anonymous binds to remote backends

handavid:
[KNOX-3425](https://issues.apache.org/jira/browse/KNOX-3425) - LDAP Proxy disallows anonymous binds to remote backends

The LdapProxyBackend.authenticate method will now return false if either the given DN or password is empty to prevent unauthorized and anonymous binds.

The UserSearchInterceptor is also updated to not call the backend if the DN or password is empty.

## How was this patch tested?

Unit tests added.
Manual tests performed using ldapsearch to attempt anonymous bind that previously succeeded
```
ldapsearch -v -x -H ldap://localhost:3890 -b 'ou=people,DC=proxy,DC=com' '(uid=guest*)' '*'
```

## Integration Tests

integration tests were run locally

## UI changes

no UI changes


handavid:
@smolnar82 @hanicz

hanicz (approved):

PR #1372, opened 2026-09-02
KNOX-3431: UserSearchInterceptor throws exceptions on failed search

handavid:
[KNOX-3431](https://issues.apache.org/jira/browse/KNOX-3431) - UserSearchInterceptor throws exceptions on failed search

Improve the behavior such that the UserSearchInterceptor will throw exceptions when a search request fails rather than swallowing the exception. Swallowing exceptions make it difficult to tell if there is a problem communicating with the remote LDAP backend.

## How was this patch tested?

Added unit tests.

## Integration Tests

Ran existing workflow tests.

## UI changes

no UI changes

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/UserSearchInterceptor.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/UserSearchInterceptor.java
@@ -18,9 +18,12 @@
 package org.apache.knox.gateway.services.ldap.interceptor;
 
 import org.apache.directory.api.ldap.model.constants.AuthenticationLevel;
+import org.apache.directory.api.ldap.model.cursor.CursorException;
 import org.apache.directory.api.ldap.model.cursor.ListCursor;
 import org.apache.directory.api.ldap.model.entry.Entry;
 import org.apache.directory.api.ldap.model.exception.LdapException;
+import org.apache.directory.api.ldap.model.exception.LdapOperationException;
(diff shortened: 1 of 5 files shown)


handavid:
@smolnar82 @hanicz

gateway-server/src/main/java/org/apache/knox/gateway/services/ldap/interceptor/UserSearchInterceptor.java
```diff
@@ -71,6 +75,7 @@ public Entry lookup(LookupOperationContext ctx) throws LdapException {
                     entry = backend.getUser(username, schemaManager);
                 } catch (Exception e) {
                     LOG.ldapLookupFailed(ctx.getDn().toString(),e);
+                    throw new LdapOperationException(ResultCodeEnum.OTHER, "Lookup request to backend " + getName() + " failed.", e);
```
smolnar82, line 78: The only comment I have is if we can somehow get the original backend result code and use it here other than `OTHER`. I'm not sure about this, so raising it as a question and would not block the merge of this PR.

handavid: I picked OTHER because I didn't want to pass the backend code directly. I think that would be confusing. e.g., if the backend credentials were changed then passing the backend code would show "invalid credentials" to the caller. However, this is misleading since it's not something that the caller has control over since it's part of the proxy configuration. Similarly, it will be confusing to show the caller "unable to connect" if there's a network problem between the proxy and the remote backend.

PR #1373, opened 2026-09-02
KNOX-3424: Complete RFC 8707/8693 dynamic resource/audience handling in KNOXTOKEN and token exchange

smolnar82:
[KNOX-3424](https://issues.apache.org/jira/browse/KNOX-3424) - Complete RFC 8707/8693 dynamic `resource`/`audience` handling in KNOXTOKEN and token exchange

Follow-up to #1356, finishing the dynamic-audience work so Knox implements both RFC 8707 (Resource Indicators) and RFC 8693 (Token Exchange) end to end. Three self-contained commits:

1. **Replace the `audience` query param with a URL-validated `resource` param.**
The KNOXTOKEN service now reads the per-request target via `resource` instead of `audience`. Each value must be an absolute URI without a fragment (RFC 8707 §2 / RFC 3986 §4.3); a value that is not a valid absolute URI, or that carries a fragment, is rejected with `400 Bad Request` / `INVALID_RESOURCE`. Repeated and comma-separated values are supported; whitespace is trimmed. The pluggable audience validators (`static` / `whitelist` / `passthrough`) are unchanged in behavior.

2. **Wire the RFC 8693 `resource`/`audience` body parameters into token exchange.**
The JWTProvider's `TokenExchangeHandler` now parses the optional `resource`/`audience` body params from the exchange request and conveys them to the downstream KNOXTOKEN service via a new shared request attribute (`CommonTokenConstants.REQUESTED_AUDIENCES_REQUEST_ATTR`), so they land in the minted token's `aud` claim. `resource` values are URI-validated (malformed → `invalid_target`); `audience` values are logical service names taken verbatim. When present, the body values take precedence over the `resource` query parameter.

2. **JSON-escape the caller-supplied `resource` value in the token error body.** 
The token-issuance error echoed the caller-supplied value back in a hand-concatenated JSON body, so a value containing a `"` could break out of the JSON string. The `{"error": ..., "code": ...}` body is now rendered through `JsonUtils` (the same escaping the KNOX-3423/#1354 filter-layer error path uses), keeping the existing `{error, code}` / `ErrorCode` shape.

Documentation in `knox-site/docs/config_knox_token.md` is updated to describe the `resource` parameter, the absolute-URI requirement, and the RFC 8707 mapping into the `aud` claim.

## How was this patch tested?

Automated unit tests (all green, offline Maven build):

- `TokenExchangeHandlerTest` — 21 tests, including 8 new ones covering `resource`/`audience` body-param parsing, comma-splitting, invalid-URI and fragment rejection as `invalid_target`, empty-value rejection, and the no-param case leaving the request attribute unset.
- `TokenServiceResourceTest` — 101 tests, including new coverage for the `resource` query param, invalid-URI/fragment rejection, exchange-attribute precedence over the query param, and a regression test asserting a `resource` value containing double quotes yields a well-formed, parseable JSON error body with the value round-tripped intact.
- `JWTFederationFilterTokenExchangeTest` — 56 tests, re-run to confirm no regression on the broader exchange path.

Checkstyle passes with 0 violations across `gateway-spi`, `gateway-provider-security-jwt`, and `gateway-service-knoxtoken`.

## Integration Tests

No new integration test was added: the change is an extension of the existing token-exchange / dynamic-audience paths, which are covered comprehensively at the unit level, and the KnoxIDF federation broker flow (`test_knoxidf.py` / `test_knoxidf_federation.py`) already exercises the token-exchange handler this PR modifies. Because commit 2 touches the JWTProvider federation broker path, adding the **`test-federation`** label here before pushing, so the Keycloak-backed E2E federation suite runs against these changes.

### DIFF
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -35,6 +35,7 @@ import java.util.Enumeration;
 import java.util.Map;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Locale;
@@ -928,7 +929,7 @@ public class TokenResource {
     } catch (RequestedAudienceValidationException e) {
       log.rejectedAudienceRequest(e.getMessage());
       return new TokenResponseContext(null,
-          "{\n  \"error\": \"" + e.getMessage() + "\",\n  \"code\": " + e.getErrorCode().toInt() + "\n}\n",
+          errorResponseBody(e.getMessage(), e.getErrorCode()),
           Response.status(Response.Status.BAD_REQUEST));
     }
 
@@ -1211,6 +1212,19 @@ public class TokenResource {
           ErrorCode.INVALID_RESOURCE);
     }
     return value;
+  }
+
+  /**
+   * Renders a token-issuance error body as {@code {"error": ..., "code": ...}}. The {@code error}
+   * message may embed a caller-supplied value (e.g. an invalid {@code resource} indicator), so it is
+   * serialized through {@link JsonUtils} to escape it rather than being concatenated verbatim, which
+   * would allow the value to break out of the JSON string.
+   */
+  private static String errorResponseBody(String error, ErrorCode code) {
+    final Map<String, Object> body = new LinkedHashMap<>();
+    body.put("error", error);
+    body.put("code", code.toInt());
+    return JsonUtils.renderAsJsonString(body);
   }
 
   private JWT getJWT(UserContext userContext, long issueTime, long expires, String jku, List<String> audiences) throws TokenServiceException {
--- a/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
+++ b/gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
@@ -1169,15 +1169,14 @@ public class TokenResource {
   }
 
   private List<String> parseRequestedResources() throws RequestedAudienceValidationException {
-    // An RFC 8693 token-exchange request carries the requested resource/audience in the form body,
-    // which the JWTProvider's TokenExchangeHandler has already parsed and validated and stashed as a
-    // request attribute. When present, those body-supplied values take precedence over the resource
-    // query parameter.
+    // An upstream authentication/federation component may have already resolved and validated the
+    // requested audiences for this request and stashed them as a request attribute. When present,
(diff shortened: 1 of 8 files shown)


smolnar82:
Cc. @hsheinblatt

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -1163,26 +1168,65 @@ public ResponseMap(String accessToken, String tokenId, Map<String, Object> map,
...
+    // An RFC 8693 token-exchange request carries the requested resource/audience in the form body,
+    // which the JWTProvider's TokenExchangeHandler has already parsed and validated and stashed as a
+    // request attribute. When present, those body-supplied values take precedence over the resource
+    // query parameter.
+    @SuppressWarnings("unchecked")
+    final List<String> fromExchange =
+        (List<String>) request.getAttribute(CommonTokenConstants.REQUESTED_AUDIENCES_REQUEST_ATTR);
+    if (fromExchange != null) {
+      return fromExchange;
+    }
```
lmccay, line 1180: I'm not sure that I like how this is being done. We are leaking exchange specifics into the API service which should only be about issuing tokens. Perhaps this is happening already elsewhere but I want to be careful not to do this.

lmccay (changes requested):

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
smolnar82: The leak is really the _naming_, not the _mechanism_. I'll rename the attribute to a generic `knox.token.requested.audiences` and reword the docs/comment so TokenResource only knows "an upstream filter may have pre-resolved the requested audiences; use them over the query param" — no RFC 8693 / exchange references. The exchange specifics stay entirely in TokenExchangeHandler.
Pushing that shortly.

hanicz (approved):

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
lmccay: Okay. Let's move forward here but I think we need to do some careful testing. Having no way to ensure that the form body param is chosen over the query param without using the JWTProvider may be problematic.

lmccay (approved):

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
smolnar82: Filed [KNOX-3437](https://issues.apache.org/jira/browse/KNOX-3437) to not to forget about this.

PR #1374, opened 2026-09-03
KNOX-3426:  Embedded-Derby delegation policy service

smolnar82:
[KNOX-3426](https://issues.apache.org/jira/browse/KNOX-3426) - Embedded-Derby delegation policy service

- Add `DerbyDBDelegationPolicyService`, the self-provisioning embedded-Derby default for KnoxIDF delegation policies.
- Consolidate the duplicated DerbyDB boot/shutdown/alias plumbing from the token-state, federated-identity and trusted-OIDC-issuer services into a shared `EmbeddedDerbyDatabase` (+ `DerbyDatabaseCredentials`).
- Move multi-statement DDL parsing (`createTablesIfNotExist`, `parseCreateTableStatements`, license-header stripping) into `KnoxDatabase` and reuse it from `DelegationPolicyDatabase`; add a command-based `JDBCUtils.createTableFromSQL` overload.

## How was this patch tested?

- New `KnoxDatabaseTest` (in-memory HSQL) covers `createTablesIfNotExist`: statement parsing/ordering, table creation, idempotent re-runs, missing-table re-provisioning, and missing-script failure.
- Updated `DelegationPolicyServiceFactoryTest` / `ServiceFactoryTest`.
- Full gateway-server test suite + checkstyle pass; manually verified clean Derby boot/login/shutdown on a local gateway.

## Integration Tests
No new integration suite needed — the change is DB-provisioning plumbing covered by unit tests against HSQL/Derby.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/database/DerbyDatabaseCredentials.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/database/DerbyDatabaseCredentials.java
@@ -1,0 +1,49 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with this
+ * work for additional information regarding copyright ownership. The ASF
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.apache.knox.gateway.database;
+
+import static org.apache.knox.gateway.database.AbstractDataSourceFactory.DATABASE_PASSWORD_ALIAS_NAME;
+import static org.apache.knox.gateway.database.AbstractDataSourceFactory.DATABASE_USER_ALIAS_NAME;
+
+import org.apache.knox.gateway.services.security.AliasService;
+import org.apache.knox.gateway.services.security.MasterService;
+
+/**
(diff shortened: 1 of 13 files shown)


hanicz (changes requested):
I don't like this approach. Parsing an SQL script is too fragile. The current logic wouldn't work for postgres if we were to merge the KnoxToken tables (which we should to reduce the amount of sql files) into one script due to the advisory lock. The removeLicenceHeader logic would cut the lock. I think either we move  to separate SQL files for each table and remove the string manipulation (which I don't really like either because we have more tables now). Or we move the creation/exist check to the DB level. (only Derby presents challenges in that approach imo)

I also feel like the testing is lacking. Table generation should be verified for each DB with additional KnoxToken tests at least so there is no regression.

nit: licence vs license you have both in the changes, should be unified

smolnar82:
Thanks for the review discussion, @hanicz . Summarizing what we agreed so it's on the record:

This PR (now):
- Fix `removeLicenceHeader` to strip only `--` comment lines instead of cutting everything before the first `CREATE TABLE`. This removes SQL from nothing, so the Postgres `pg_advisory_lock(...)` statements survive. Addressing the fragility flagged in review while keeping the current parsing approach.
- Merge as-is once that's in.

Why not the fuller refactor yet: moving table creation to a pure SQL layer (DB-level `CREATE TABLE IF NOT EXISTS`, no Java-side splitting/guard) is blocked by two engines:
- Oracle:  no `IF NOT EXISTS`, but solvable purely in SQL by wrapping the DDL in a PL/SQL block in the Oracle files.
- Derby: no `IF NOT EXISTS` and can't execute a multi-statement script in one call; it's the one that forces the Java splitting. Derby is slated to be dropped.

Follow-up (separate JIRA, after [KNOX-3401](https://issues.apache.org/jira/browse/KNOX-3401) and once Derby is removed): switch to the pure SQL layer (Oracle handled via PL/SQL, everyone else via IF NOT EXISTS) and delete the Java-side parsing/splitting entirely. I'll file the JIRA to track this.

smolnar82:
@hanicz - I submitted a new commit to address the `removeLicenseHeader` fix.

smolnar82:
I filed [KNOX-3436](https://issues.apache.org/jira/browse/KNOX-3436) to switch to pure SQL after Derby is gone.

hanicz (approved):

PR #1376, opened 2026-09-03
KNOX-3432: Advertise RFC 8693 token-exchange grant type in KnoxIDF discovery metadata

smolnar82:
[KNOX-3432](https://issues.apache.org/jira/browse/KNOX-3432) - Advertise RFC 8693 token-exchange grant type in KnoxIDF discovery metadata

KnoxIDF supports OAuth 2.0 Token Exchange (RFC 8693), but its OIDC discovery document (`.well-known/openid-configuration`) only listed `authorization_code` and `refresh_token` in `grant_types_supported`. Per RFC 8414 §2 that field must list every grant type the server supports, so a spec-compliant reader would wrongly conclude token exchange is unavailable.

- Added `TOKEN_EXCHANGE_GRANT_TYPE = "urn:ietf:params:oauth:grant-type:token-exchange"` to `KnoxIDFConstants` (RFC 8693 §2.1, IANA-registered). Intentionally duplicated from `JWTFederationFilter.TOKEN_EXCHANGE` rather than shared; it's a fixed standard identifier, and duplicating avoids a module dependency on the JWT federation provider.
- `DiscoveryResource` now includes this value in `grant_types_supported`.

## How was this patch tested?

- Extended `DiscoveryResourceMetadataTest` to assert the discovery document advertises all three grant types, including the token-exchange URN.
- `mvn -pl gateway-service-knoxidf test -Dtest=DiscoveryResourceMetadataTest`: passes.

### DIFF
--- a/.github/workflows/tests/test_knoxidf.py
+++ b/.github/workflows/tests/test_knoxidf.py
@@ -69,6 +69,7 @@ class TestKnoxIDF(unittest.TestCase):
             [
                 "authorization_code",
                 "refresh_token",
+                "client_credentials",
                 "urn:ietf:params:oauth:grant-type:token-exchange",
             ],
         )
--- a/.github/workflows/tests/test_knoxidf.py
+++ b/.github/workflows/tests/test_knoxidf.py
@@ -66,7 +66,11 @@ class TestKnoxIDF(unittest.TestCase):
         self.assertEqual(config.get("response_types_supported"), ["code"])
         self.assertEqual(
             config.get("grant_types_supported"),
-            ["authorization_code", "refresh_token"],
+            [
+                "authorization_code",
+                "refresh_token",
+                "urn:ietf:params:oauth:grant-type:token-exchange",
+            ],
         )
         self.assertEqual(config.get("id_token_signing_alg_values_supported"), ["RS256"])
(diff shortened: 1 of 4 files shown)


moresandeep (approved):
Looks good.

hanicz (approved):

.github/workflows/tests/test_knoxidf.py
```diff
@@ -66,7 +66,11 @@ def test_discovery(self):
         self.assertEqual(config.get("response_types_supported"), ["code"])
         self.assertEqual(
             config.get("grant_types_supported"),
-            ["authorization_code", "refresh_token"],
+            [
+                "authorization_code",
+                "refresh_token",
+                "urn:ietf:params:oauth:grant-type:token-exchange",
+            ],
```
lmccay, line 74: Shouldn't we add client_credentials as well?

lmccay (changes requested):
Have only one question on this.
Otherwise, looks good!

.github/workflows/tests/test_knoxidf.py
smolnar82: Nice catch, Larry. Let me add it in a new commit soon.

PR #1377, opened 2026-09-03
KNOX-3434 - Allow missing typ when validating JWTs from TrustedOidcIssuers for token exchange.

hsheinblatt:
KNOX-3434 - Allow missing typ when validating JWTs from TrustedOidcIssuers for token exchange.

External trusted OIDC issuer tokens do not always have a typ header. In particular,
Kubernetes projected service account tokens do not. Change AbstractJWTFilter
so that RFC 8693 token exchange JWT validations for JWTs from TrustedOidcIssuers
allow a missing typ. This avoids having to dynamically update the list of issuers for which to
skip type verification separately from the trusted OIDC issuer registration, and restricts the
use case to only token exchange for externally trusted issuers.

## How was this patch tested?

A new unit test was added to show that when a JWT is validated for token exchange
and signature validation is to use a registered TrustedOidcIssuer, the typ verifier used
allows a missing typ. There were no typ validation unit tests, so two regression tests
were added to show that for statically defined JWKS validation and for non-token
exchange validation, the existing typ validation continues to be enforced.

## Integration Tests
No integration tests are added in this pull request. They will be added once the full delegation
enforcement flow is wired in a later task.

## UI changes
N/A

### DIFF
--- a/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
+++ b/gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/AbstractJWTFilter.java
@@ -584,7 +584,13 @@ public abstract class AbstractJWTFilter implements Filter {
       return true;
     }
     try {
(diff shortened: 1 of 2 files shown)


lmccay (approved):
LGTM
+1

smolnar82 (commented):
LGTM too.

PR #1379, opened 2026-09-06
KNOX-3438: Add delegation policy admin API with upsert support for RFC 8693 token exchange

hsheinblatt:
[KNOX-3438](https://issues.apache.org/jira/browse/KNOX-3438) - Add delegation policy admin API with upsert support for RFC 8693 token exchange

Adds a REST admin API for managing **RFC 8693 delegation policies** on the `KNOXIDF_ADMIN`
service under `/knoxidf/admin/v1/delegation-policies`.

The new `DelegationPolicyResource` (`gateway-service-knoxidf`) exposes:

| Method | Path | Purpose |
|--------|------|---------|
| `POST` | `/delegation-policies` | Register |
| `GET` | `/delegation-policies` | List, or get-by-id |
| `PUT` | `/delegation-policies/{registrationId}` | Full-replace update |
| `PUT` | `/delegation-policies` | Create-or-update (upsert) |
| `DELETE` | `/delegation-policies/{registrationId}` | Delete |

Three DTO classes accompany it — `DelegationPolicyRequest`, `DelegationPolicyResponse`, and
`DelegationPolicyListResponse` — along with a `DelegationPolicyDtoImmutabilityHelper` that
ensures defensive copies of `Set` and `Map` fields on both construction and getter access.

### Changes to common code and other services

**`gateway-util-common`**
- Adds `ResourceType.DELEGATION_POLICY`, used by audit records emitted from all three mutating
  endpoints (register, update, delete) in the same `Action.DELEGATION_LIFECYCLE` / try-finally
  pattern already used by `TrustedOidcIssuersResource`.

**`gateway-server`**
- Extracts the private `isUniqueConstraintViolation(SQLException)` method from
  `JdbcFederatedIdentityService` into a new `public static` on `JDBCUtils` (the existing
  JDBC-utility class in `org.apache.knox.gateway.database`). `JdbcFederatedIdentityService`
  now delegates to the shared copy; its own private method is deleted.
- The extracted helper classifies a `SQLException` as a unique-constraint violation portably
  across **Derby, MySQL, PostgreSQL, and Oracle** via SQLState `"23"`-prefix matching rather
  than vendor-specific syntax.

**`gateway-spi`**
- Adds two new exception types in the `delegation` package:
  - `DelegationPolicyAlreadyExistsException` — thrown by `register()` on a duplicate
    `actorAuthority` / `actorId` pair.
  - `DelegationPolicyNotFoundException` — thrown by `update()` and `delete()` when the supplied
    `registrationId` does not exist.
- The `DelegationPolicyService` interface gains a new `registerOrUpdate(DelegationPolicy)`
  method returning `RegisterOrUpdateResult` (a small value type carrying the persisted policy
  and a boolean indicating whether the call created or replaced), and `update()` now returns
  `DelegationPolicy` instead of `void`.
- Two new config keys and getters are added to `GatewayConfig` / `GatewayConfigImpl` for
  configurable token TTL bounds:
  - `gateway.delegation.service.min.token.ttl.sec` (default **60 s**)
  - `gateway.delegation.service.max.token.ttl.sec` (default **86400 s**)

### Service and database layer (`gateway-server`)

- `DelegationPolicyDatabase.updatePolicy()` and `deletePolicy()` now inspect the affected-row
  count returned by the core `UPDATE` and `DELETE` statements. A count of zero causes the
  transaction to be rolled back and the call to return `false`, which `JdbcDelegationPolicyService`
  converts into a `DelegationPolicyNotFoundException`. This replaces the previous silent no-op on
  a nonexistent `registrationId`.
- `JdbcDelegationPolicyService.register()` now wraps the database call in a specific `SQLException`
  catch block before the generic `Exception` catch. Inside it, `JDBCUtils.isUniqueConstraintViolation`
  is used to throw `DelegationPolicyAlreadyExistsException` on a genuine duplicate rather than a
  plain `RuntimeException`; non-duplicate `SQLException`s and all other exceptions still produce a
  plain `RuntimeException`.
- `JdbcDelegationPolicyService.registerOrUpdate()` composes the existing `findByActor()`,
  `register()`, and `update()` methods with exactly **one bounded retry per race branch**
  (concurrent delete between `findByActor` and `update`, or concurrent insert between
  `findByActor` and `register`). A second race on the same actor within the same call propagates
  the resulting `DelegationPolicyNotFoundException` or `DelegationPolicyAlreadyExistsException` to
  the caller unchanged. This design adds no new SQL and no new transaction boundary.

  > A database-level savepoint-based upsert was considered and rejected: savepoint semantics
  > differ across the databases supported in Knox in ways that are non-trivial to handle portably.

### Path ACLs and coexistence with `TrustedOidcIssuersResource`

`DelegationPolicyResource` is a sibling of `TrustedOidcIssuersResource` inside the same
`KNOXIDF_ADMIN` role. Both are served by `KnoxIDFAdminServiceDeploymentContributor`'s existing
Jersey package scan and `"knoxidf/admin/**"` URL pattern — **no deployment-contributor changes
are required**.

The two resources share the same base path prefix (`knoxidf/admin/v1/`) but have distinct final
segments (`trusted-oidc-issuers` vs. `delegation-policies`), which lets operators assign
independent `PathAclsAuthz` rules to each, for example:

```xml
<param>
  <name>KNOXIDF_ADMIN.rule_issuers.path.acl</name>
  <value>https://*:*/**/knoxidf/admin/v1/trusted-oidc-issuers**;awc-svc;issuer-admins;*</value>
</param>
<param>
  <name>KNOXIDF_ADMIN.rule_delegation.path.acl</name>
  <value>https://*:*/**/knoxidf/admin/v1/delegation-policies**;knox-operator;delegation-admins;*</value>
</param>
```
This means the two admin APIs can be administered by separate groups (or the same group) with no
change to PathAclsAuthorizationFilter or PathAclParser. PathAclsAuthz is what actually gates
access to both resources; the resource layer itself does not reject a null principal but does
audit the operator as "ANONYMOUS" in that case, matching TrustedOidcIssuersResource's precedent.

## How was this patch tested?

`DelegationPolicyResourceTest` (gateway-service-knoxidf, EasyMock) — covers every HTTP
status code enumerated in the error-handling contract for all five endpoint variants:
- POST happy paths (all-defaults, all-fields round-trip, unknown-field ignored,
duplicate-array-entry collapse via Set), boundary value tests for actorAuthority/actorId
(missing/null/empty/whitespace), status (omitted, empty, valid, invalid, wrong case),
tokenTtlSec (absent, at-min, at-max, below-min, above-max, negative, zero, wrong JSON type),
allowHeadlessExchange (omitted defaults to false, explicit true), canActForUsers/canActForGroups
combinations (both empty rejected, each individually sufficient, both non-empty), and
resourcePolicy shapes (omitted, empty object, empty-scope-array entry meaning all-scopes,
non-empty-scope-array, mixed). Error paths: 409 actor_exists, 500 storage_error, and
null-principal auditing ANONYMOUS.
- PUT (upsert) tests mirror the POST validation suite and additionally cover 201 (create
branch), 200 (update branch), 409/404 from the retry's second-race propagation, and 500.
- GET (list) covers no-filter, actorAuthority filter, empty-string filter treated as no
filter, empty result, and hasMore=true surfaced verbatim.
- GET (one) covers found, not-found (404), and storage failure (500).
- PUT /{id} covers full-replace semantics including fields reset to defaults on omission,
server-managed fields (registrationId, createdAt, createdBy, updatedAt) ignored when
present in the body, 404, and 500.
- DELETE covers 204, 404 (not the idempotent 204 that TrustedOidcIssuersResource
returns — a deliberate divergence since the storage layer now surfaces not-found explicitly),
and 500.
- Cross-cutting tests assert ISO-8601 serialization of Instant fields (guards against an
ObjectMapper without JavaTimeModule), and that @PostConstruct wires both the service and
the TTL bounds from GatewayConfig.

DelegationPolicyDtoImmutabilityTest (gateway-service-knoxidf) — confirms
DelegationPolicyRequest and DelegationPolicyResponse defensively copy Set and Map fields
on both construction and getter access, and that null inputs to
DelegationPolicyDtoImmutabilityHelper are handled cleanly.

JdbcDelegationPolicyServiceTest (gateway-server, real Derby in-memory):
- Duplicate-detection upgrade — register() twice on the same actorAuthority/actorId throws
the typed DelegationPolicyAlreadyExistsException with the original SQLException in its cause
chain — and a negative counterpart (a non-duplicate storage failure — VARCHAR overflow, SQLState
22001 — still throws plain RuntimeException, confirming isUniqueConstraintViolation does not
over-match).
- Not-found tests for update() and delete() on a nonexistent registrationId, and
identity-immutability tests rejecting an actorAuthority/actorId change on update(), verify
the database-layer not-found detection.
- Two real Derby-backed tests cover registerOrUpdate()'s create branch (new actor, isCreated()
true) and update branch (same actor called twice, same registrationId both times, isCreated()
false on the second call, mutable fields updated, createdAt unchanged, updatedAt advances).

## Integration Tests

Integration tests will follow once the rest of the delegation flow has been implemented.

## UI changes

N/A


gateway-service-knoxidf/src/main/java/org/apache/knox/gateway/service/knoxidf/DelegationPolicyResource.java
```diff
@@ -0,0 +1,368 @@
...
+  public void init() {
+    final GatewayServices services = (GatewayServices)
+        servletContext.getAttribute(GatewayServices.GATEWAY_SERVICES_ATTRIBUTE);
+    policyService = services.getService(ServiceType.DELEGATION_POLICY_SERVICE);
+
+    final GatewayConfig config = (GatewayConfig) servletContext.getAttribute(GatewayConfig.GATEWAY_CONFIG_ATTRIBUTE);
+    if (config != null) {
+      minTokenTtlSec = config.getDelegationServiceMinTokenTtlSec();
+      maxTokenTtlSec = config.getDelegationServiceMaxTokenTtlSec();
+    }
```
smolnar82, line 100: `min/maxTokenTtlSec` are read only from GatewayConfig here. The convention for per-service knobs like this is topology `<param>` via `context.getInitParameter(...)`, with a code-constant default fallback. For instance `TokenResource` (KNOXTOKEN) does for `token.ttl`.

I suggest:
- Drop the `DELEGATION_SERVICE_MIN/MAX_TOKEN_TTL_SEC` keys + getters from `GatewayConfig/GatewayConfigImpl`; keep the two default values as constants in the resource.
- In `init()`, read them using `getInitParameter`, fall back to the constant when absent.
- Fail fast when a supplied value is 
  - non-numeric, 
  - ≤ 0, 
  - or min > max

In this PR: if an operator sets `min.token.ttl.sec > max.token.ttl.sec`, every request carrying a TTL is rejected with a confusing message, with no startup warning.
With my proposal this goes away and shrinks the gateway-site surface.

gateway-server/src/main/java/org/apache/knox/gateway/services/knoxidf/delegation/DelegationPolicyDatabase.java
```diff
@@ -50,12 +50,16 @@ class DelegationPolicyDatabase extends KnoxDatabase {
...
+  // an update attempting to change them affects 0 rows -- indistinguishable from (and reported the
+  // same as) registrationId not existing at all.
   private static final String UPDATE_CORE_SQL =
       "UPDATE " + CORE_TABLE + " SET "
-          + "actor_authority = ?, actor_id = ?, name = ?, status = ?, token_ttl_sec = ?, "
-          + "description = ?, created_by = ?, created_at = ?, updated_at = ?, "
-          + "allow_headless_exchange = ? "
-          + "WHERE registration_id = ?";
+          + "name = ?, status = ?, token_ttl_sec = ?, description = ?, allow_headless_exchange = ?, "
+          + "updated_at = CURRENT_TIMESTAMP "
```
smolnar82, line 61: `INSERT` writes `created_at/updated_at` from the app clock (`Timestamp.from(Instant.now())`), but `UPDATE` sets `updated_at = CURRENT_TIMESTAMP` (DB clock). After an update, `createdAt` (app clock) and `updatedAt` (DB clock) come from different sources; if the DB host clock lags the app host, `updatedAt` could read earlier than `createdAt`. Same-host deployments make this negligible, but it's an avoidable inconsistency: either use the DB clock for both, or the app clock for both. (the `assertNotEquals` at `JdbcDelegationPolicyServiceTest.java:362` is safe from flakiness since two distinct clock sources will essentially never produce an identical instant; no change needed there.)

gateway-service-knoxidf/src/main/java/org/apache/knox/gateway/service/knoxidf/DelegationPolicyResource.java
```diff
@@ -0,0 +1,368 @@
...
+    } catch (RuntimeException e) {
+      return errorResponse(Response.Status.INTERNAL_SERVER_ERROR, "storage_error", "Failed to register or update delegation policy");
+    } finally {
+      auditor.audit(Action.DELEGATION_LIFECYCLE, auditId, ResourceType.DELEGATION_POLICY,
+          outcome, "event_type=" + eventType + " performed_by=" + auditLabel(operatorId));
+    }
+  }
+
+  @GET
+  public Response list(@QueryParam("actorAuthority") String actorAuthority) {
```
smolnar82, line 183: Only mutating endpoints emit audit records; `list/getOne` do not. Reading who-can-impersonate-whom is arguably sensitive. I'd ad auditing here too.

gateway-service-knoxidf/src/main/java/org/apache/knox/gateway/service/knoxidf/DelegationPolicyResource.java
```diff
@@ -0,0 +1,368 @@
...
+      return errorResponse(Response.Status.BAD_REQUEST, "invalid_request", "actorAuthority is required");
+    }
+    if (StringUtils.isBlank(req.getActorId())) {
+      return errorResponse(Response.Status.BAD_REQUEST, "invalid_request", "actorId is required");
+    }
+    final Integer tokenTtlSec = req.getTokenTtlSec();
+    if (tokenTtlSec != null && (tokenTtlSec < minTokenTtlSec || tokenTtlSec > maxTokenTtlSec)) {
+      return errorResponse(Response.Status.BAD_REQUEST, "invalid_request",
+          "tokenTtlSec must be between " + minTokenTtlSec + " and " + maxTokenTtlSec + " seconds (inclusive)");
+    }
```
smolnar82, line 286: `tokenTtlSec` is optional on register/update: omit it and the policy is stored with `token_ttl_sec = NULL`, so this fallback is the default path, not an edge case. The admin API bounds-checks an explicit `tokenTtlSec` against `[minTokenTtlSec, maxTokenTtlSec]`, but the fallback `configuredKnoxTokenTtlSec ` (in `JdbcDelegationPolicyService`) is never checked against those bounds, so every policy that doesn't pin its own TTL can yield an effective TTL outside the range the API enforces for explicit values.

smolnar82 (approved):
I'll submit a follow-up PR with the fixes for my review comments.

PR #567, opened 2022-04-26
KNOX-2738 On Fresh install JDBCTokenStateService initiation failed

zeroflag:
## What changes were proposed in this pull request?

Starting multiple knox instances can cause token state service initialization failures when they try to create the same database tables.

```
ERROR knox.gateway (TokenStateServiceFactory.java:createService(63)) - Error while initiatalizing org.apache.knox.gateway.services.token.impl.JDBCTokenStateService: Error while initiating JDBCTokenStateService: org.postgresql.util.PSQLException: ERROR: relation "knox_tokens" already exists
```

To fix this issue I added a `IF NOT EXISTS` clause to the create table statement. This is unfortunately not supported by Derby (which is only used for unit testing), so I replaced derby with hsql.

## How was this patch tested?

I tested manually with both mysql and postgres.

Postgres gateway-site config:

```xml

    <property>
        <name>gateway.service.tokenstate.impl</name>
        <value>org.apache.knox.gateway.services.token.impl.JDBCTokenStateService</value>
    </property>
     <property>
        <name>gateway.database.type</name>
        <value>postgresql</value>
    </property>
    <property>
        <name>gateway.database.connection.url</name>
        <value>jdbc:postgresql://localhost:5432/postgres?user=postgres</value>
    </property>
```

```bash
$ docker run --name posttest -d -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust postgres:alpine
```

```bash
   $ docker exec -it posttest psql -U postgres


   postgres=# select * from knox_tokens;
               token_id               |  issue_time   |  expiration   | max_lifetime  
--------------------------------------+---------------+---------------+---------------
 ad4953c9-38dd-4214-8ce1-e3b873a81f13 | 1650961012152 | 1650964612097 | 1651565812152
 d37f744c-ef5b-4086-a9b7-6c1c0640e596 | 1650961013088 | 1650964613065 | 1651565813088
 058b4e5c-642c-42fc-9e5a-9d23a56256d5 | 1650961026745 | 1650964626709 | 1651565826745
(3 rows)
```

MySql gateway-site config:

```xml
    <property>
        <name>gateway.service.tokenstate.impl</name>
        <value>org.apache.knox.gateway.services.token.impl.JDBCTokenStateService</value>
    </property>
     <property>
        <name>gateway.database.type</name>
        <value>mysql</value>
    </property>
     <property>
        <name>gateway.database.name</name>
        <value>mysql</value>
    </property>
     <property>
[...]

PR #1380, opened 2026-09-06
KNOX-3401: Replace Derby with H2

smolnar82:
[KNOX-3401](https://issues.apache.org/jira/browse/KNOX-3401) - Replace the embedded Apache Derby database with embedded H2

Apache Derby is being retired, and Knox no longer wants to bundle its driver. Knox has been using an embedded Derby database as its zero-configuration, on-disk store for server-managed token state (and, more recently, the KnoxIDF federation data). This PR replaces that embedded **Derby** backend with an embedded **H2** database.

See detailed analysis and form factors why I choose H2 [here](https://issues.apache.org/jira/browse/KNOX-3401?focusedCommentId=18111189&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-18111189).

From an operator's point of view the default experience is unchanged: the database is still provisioned automatically on first start, lives on the local file system under the gateway security directory, requires no external setup, and is intended for single-node (non-HA) deployments. Only the engine behind it changes.

### Highlights:

- **Embedded H2 replaces embedded Derby** as the default token-state backend and for the KnoxIDF federation stores. No configuration change is required for a fresh install.
- **The Apache Derby driver is removed from the entire distribution.** The `derbydb` database type and the Derby-backed service variants no longer exist. Externally hosted databases (PostgreSQL, MySQL, MariaDB, Oracle, HSQL) are unaffected.
- **Optional at-rest encryption for the embedded database.** It is off by default and opt-in; when enabled, the encryption passphrase is read from a credential-store alias (never the gateway master secret).
- **A supported upgrade path for existing token data.** Operators upgrading from a Derby-shipping release can move their existing tokens into the new H2 store with a new `knoxcli.sh migrate-derby-tokens` command. Because the Derby driver is no longer shipped, the operator temporarily drops a Derby driver jar into the `ext/` directory for the duration of the migration. The migration scope is tokens only: that was the only data ever persisted by Derby-shipping releases; and the command is safe to re-run.
- **UI:** the Token Management status banner that warns when tokens are stored in a local on-disk database (valid only for non-HA setups) now refers to H2.
- **KnoxShell:** the built-in Derby datasource convenience helper is removed. The KnoxShell JDBC datasource feature itself remains and works with any user-provided driver.
- **Documentation** updated across the token configuration guide, the KnoxCLI reference, and the KnoxIDF pages, including the new migration procedure and the encryption options.

## How was this patch tested?

- **Automated unit tests:** the Derby-specific tests were replaced with H2-backed equivalents, and a new test covers the Derby→H2 token migration logic (verbatim copy of tokens and metadata, preservation of encoded passcodes and absolute lifetime timestamps, and idempotent re-runs). The full `gateway-server` test suite passes and checkstyle reports no violations.
- **Manual testing** against the embedded H2 backend on a local deployment:
  - token generation via the Token Generation UI,
  - enabling and disabling tokens,
  - token revocation.
- The migration command is currently covered by its unit test; no manual tests were executed.

### Integration Tests

The existing end-to-end suite already exercises this change. The E2E tests spin up a fresh Knox Gateway that runs on the **default embedded backend - now H2** - and the token-state tests (`test_knoxtoken_jwt.py`) drive the full server-managed token lifecycle against it: minting, renewal, enable/disable, and revocation, including verifying that disable and revoke are enforced at federation. Because these run on the default backend, they now validate the H2-backed token store on every run, so no new integration test was required.

I'm also adding the `test-federation` label to this PR which ensures that recently added KnoxIDF tests are running as well (including federated tokens from KeyCloak).

## UI changes

The only UI change is the wording of the Token Management status banner shown when the token backend is the local embedded database: it now names H2 instead of Derby. A screenshot of the Token Management page showing that message can be attached here.
image


gateway-server/src/main/java/org/apache/knox/gateway/database/H2DataSourceFactory.java
```diff
@@ -0,0 +1,69 @@
...
+ * password and the user password). The file password is the operator-provisioned passphrase resolved
+ * from the credential-store alias named by {@link GatewayConfig#getDatabaseH2EncryptionPassphraseAlias()};
+ * initialization fails fast if that alias is unset (no silent unencrypted fallback).
+ */
+public class H2DataSourceFactory extends AbstractDataSourceFactory {
+
+    @Override
+    public DataSource createDataSource(GatewayConfig gatewayConfig, AliasService aliasService) throws AliasServiceException, SQLException {
+        final JdbcDataSource dataSource = new JdbcDataSource();
+        String url = "jdbc:h2:" + gatewayConfig.getDatabaseName();
```
hanicz, line 49: So we won't provide a way to use connectionUrl directly? Other DS implementation make it possible.

gateway-server/src/main/java/org/apache/knox/gateway/database/H2DataSourceFactory.java
```diff
@@ -0,0 +1,69 @@
...
+
+        if (gatewayConfig.isDatabaseH2EncryptionEnabled()) {
+            final String aliasName = gatewayConfig.getDatabaseH2EncryptionPassphraseAlias();
+            final String filePassword = getDatabaseAlias(aliasService, aliasName);
+            if (filePassword == null || filePassword.isEmpty()) {
+                throw new SQLException("H2 at-rest encryption is enabled (" + aliasName + ") but no passphrase is stored under credential-store alias '"
+                        + aliasName + "'. Provision the alias or disable gateway.database.h2.encryption.enabled.");
+            }
+            url += ";CIPHER=AES";
+            dataSource.setUser(getDatabaseUser(aliasService));
```
hanicz: nit: setUser is duplicated

gateway-server/src/main/java/org/apache/knox/gateway/database/H2DataSourceFactory.java
```diff
@@ -0,0 +1,69 @@
...
+        final String userPassword = getDatabasePassword(aliasService);
+
+        if (gatewayConfig.isDatabaseH2EncryptionEnabled()) {
+            final String aliasName = gatewayConfig.getDatabaseH2EncryptionPassphraseAlias();
+            final String filePassword = getDatabaseAlias(aliasService, aliasName);
+            if (filePassword == null || filePassword.isEmpty()) {
+                throw new SQLException("H2 at-rest encryption is enabled (" + aliasName + ") but no passphrase is stored under credential-store alias '"
+                        + aliasName + "'. Provision the alias or disable gateway.database.h2.encryption.enabled.");
+            }
+            url += ";CIPHER=AES";
```
hanicz: nit: Can you move this into a static variable?

hanicz (changes requested):
Can't we use HSQL instead of Derby going forward? What's the reasoning behind H2 vs HSQL? My only issue with H2 is that we pick up a new dependency where we have to manage CVE issues instead of defaulting to an already existing implementation.

If we decide to use H2 I think we should add a section in our docs that we DO NOT recommend using it in production.

smolnar82:
@hanicz - no we can't use HSQL; see my detailed analysis in the corresponding JIRA [here](https://issues.apache.org/jira/browse/KNOX-3401?focusedCommentId=18111189&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-18111189).

We might get rid of HSQL in the near future.

Update: I found that @zeroflag added HSQL support in #567 , because Derby did not support `IF NOT EXISTS`. Since the new H2 DB support this, we can remove HSQL DB entirely.

gateway-server/src/main/java/org/apache/knox/gateway/database/H2DataSourceFactory.java
smolnar82: Good point - the other factories honor `connectionUrl` because they front **external** databases where the operator owns the full JDBC URL. H2 here is deliberately the self-provisioning embedded backend, and I intentionally kept it URL-derived rather than URL-configurable, for two reasons:
1. The factory has to own the URL. For at-rest encryption it appends `;CIPHER=AES` and constructs the connection password as `<filePassword> <userPassword>` (H2 splits on the first space). A caller-supplied `connectionUrl` would bypass/conflict with that, so we'd be adding a config knob that silently breaks encryption.
2. The one thing operators legitimately need (the DB location) is already configurable via `gateway.database.name`.  The URL is `jdbc:h2:${name}`, a custom `connectionUrl` would only add H2-specific URL options, which isn't a use case we intend to support for the embedded store.

gateway-server/src/main/java/org/apache/knox/gateway/database/H2DataSourceFactory.java
smolnar82: Fixed.

gateway-server/src/main/java/org/apache/knox/gateway/database/H2DataSourceFactory.java
smolnar82: Fixed.

hanicz (approved):

smolnar82:
HSQL DB removal: https://issues.apache.org/jira/browse/KNOX-3440

PR #1381, opened 2026-09-07
KNOX-3438: Address review comments from delegation policy admin API

smolnar82:
[KNOX-3438](https://issues.apache.org/jira/browse/KNOX-3438) - Address review comments from delegation policy admin API

Follow-up #1379  (merged), addressing the review comments raised there in a separate PR.

- TTL bounds config: moved min/max token-TTL bounds from `GatewayConfig` to topology `<param>` init-params on `DelegationPolicyResource` (defaults as code constants), and fail fast in init() on a `non-numeric/non-positive` value or `min > max` - matching the `TokenResource/KNOXTOKEN `convention.
- Bounds-check the fallback default: `init()` now validates the service's gateway-wide default TTL (`DelegationPolicyService#getConfiguredTokenTtlSec`) against `[min, max]` and fails fast, so policies stored without an explicit `tokenTtlSec` can't yield an effective TTL outside the range the API enforces for explicit values.
- Consistent timestamp clock source: `DelegationPolicyDatabase` now stamps `updated_at` from the app clock (bound param) on update, matching how `INSERT` writes `created_at`/`updated_at`, instead of a DB-side `CURRENT_TIMESTAMP` that could read on a different clock/zone.
- Audit read endpoints: `list` and `getOne` now emit audit records under `Action.ACCESS` (reads of who-can-impersonate-whom are sensitive), alongside the existing mutating-endpoint audits.

## How was this patch tested?

- Existing + new unit tests: `DelegationPolicyResourceTest`, `JdbcDelegationPolicyServiceTest`, `DelegationPolicySchemaTest`, `DelegationPolicyServiceFactoryTest`; all passing.
- New tests: fail-fast on out-of-range configured default; read-endpoint audit assertions (`Action.ACCESS`, `policy_listed/policy_read`); TTL bounds read from init-params + fail-fast cases.
- `mvn checkstyle:check` (validate phase) clean on all touched modules.

## Integration Tests

No integration-test changes, these are behavior fixes to an existing feature, fully covered by module unit tests.

### DIFF
--- a/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
+++ b/gateway-server/src/main/java/org/apache/knox/gateway/config/impl/GatewayConfigImpl.java
@@ -1969,14 +1969,4 @@ public class GatewayConfigImpl extends Configuration implements GatewayConfig {
   public int getDelegationServiceListMaxPerAuthority() {
     return getInt(DELEGATION_SERVICE_LIST_MAX_PER_AUTHORITY, DELEGATION_SERVICE_LIST_MAX_PER_AUTHORITY_DEFAULT);
   }
-
-  @Override
-  public int getDelegationServiceMinTokenTtlSec() {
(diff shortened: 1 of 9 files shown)


smolnar82:
Cc. @hsheinblatt

hanicz (approved):

PR #1395, opened 2026-09-11
KNOX-2900: KNOX-2899 followup - reenable service based discovery filter

bonampak:
[KNOX-2900](https://issues.apache.org/jira/browse/KNOX-2900) - Re-enable service-based discovery filter

PollingConfigurationAnalyzer changes:
- Ignore excluded service types in CM config monitor event relevance
- Ignore excluded role types in scale events
- Prevent rediscovery of services that remained in an invalid config state (use the same workflow for service configuration data discovery as ClouderaManagerServiceDiscovery)
- Re-enable per-descriptor service filtering in CM discovery (cluster configuration cache data is merged after each topology discovery) 

## How was this patch tested?

Still needs to be tested.

### DIFF
--- a/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
+++ b/gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
@@ -59,7 +59,6 @@ import org.apache.knox.gateway.topology.simple.SimpleDescriptor;
 import org.apache.knox.gateway.topology.simple.SimpleDescriptorFactory;
 
 import java.io.File;
-import java.io.IOException;
 import java.math.BigDecimal;
 import java.security.KeyStore;
 import java.time.Instant;
@@ -427,6 +426,7 @@ public class PollingConfigurationAnalyzer implements Runnable {
     if (source != null && clusterName != null) {
       TopologyService ts = getTopologyService();
       if (ts != null) {
+        boolean referencesIncomplete = false;
         for (File f : ts.getDescriptors()) {
           try {
             SimpleDescriptor sd = SimpleDescriptorFactory.parse(f.toPath().toAbsolutePath().toString());
@@ -434,9 +434,18 @@ public class PollingConfigurationAnalyzer implements Runnable {
               remainingClusterRefs = true;
               break;
             }
-          } catch (IOException e) {
-            // Ignore these errors
+          } catch (Exception e) {
+            // A descriptor we cannot read/parse might reference this cluster. Concluding "no references" on
+            // incomplete information would tear down a still-referenced cluster's cache, so remember the gap and
+            // assume a reference remains below.
+            log.errorCheckingClusterReferences(f.getName(), source, clusterName, e);
+            referencesIncomplete = true;
           }
+        }
+        if (!remainingClusterRefs && referencesIncomplete) {
+          // Could not rule out a reference from an unreadable/unparseable descriptor; assume a reference remains and
+          // keep monitoring rather than evict the cache (mirroring the unavailable-TopologyService case below).
+          remainingClusterRefs = true;
         }
       } else {
         remainingClusterRefs = true; // If the TopologyService is unavailable, assume references remain
@@ -580,8 +589,15 @@ public class PollingConfigurationAnalyzer implements Runnable {
    * Determine the CM service types referenced by the deployed descriptors targeting the given discovery source and
    * cluster, by mapping each descriptor's declared Knox service names to CM service types via the registered service
    * model generators.
+   * <p>
+   * The set is built only from descriptors that could be read and parsed. An unreadable/unparseable descriptor is
+   * logged and skipped: a re-discovery can only ever refresh a descriptor that is parseable (an unparseable one cannot
+   * be regenerated until it is fixed, which itself triggers a fresh discovery), so filtering relevance on the services
+   * of the parseable descriptors never drops an actionable re-discovery. This intentionally differs from
+   * {@link #clusterReferencesExist(String, String)}, which must instead assume references remain so an unreadable
+   * descriptor never causes the monitored baseline to be discarded.
    *
-   * @return the referenced CM service types, or null if the TopologyService is unavailable (references undeterminable)
+   * @return the referenced CM service types (possibly empty), or null if the TopologyService is unavailable
    */
   private Set<String> getReferencedServiceTypes(final String source, final String clusterName) {
     final TopologyService ts = getTopologyService();
@@ -598,8 +614,11 @@ public class PollingConfigurationAnalyzer implements Runnable {
             referencedServices.add(service.getName());
           }
         }
-      } catch (IOException e) {
-        // Ignore these errors
+      } catch (Exception e) {
+        // A descriptor we cannot read/parse is skipped: it cannot be re-generated until it is fixed, so it cannot be
+        // the target of a useful re-discovery now. Log the offending file and keep filtering on the descriptors we
+        // could read.
+        log.errorDeterminingReferencedServiceTypes(f.getName(), source, clusterName, e);
       }
     }
     return serviceModelGeneratorsHolder.getServiceTypesForServices(referencedServices);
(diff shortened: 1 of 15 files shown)


gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
```diff
@@ -618,22 +751,33 @@ protected ServiceConfigurationModel getCurrentServiceConfiguration(final String
...
-        ApiConfigList configList = roleConfig.getConfig();
-
-        String roleName = roleConfig.getName();
-        String roleType = roleConfig.getRoleType();
-        ApiHostRef hostRef = roleConfig.getHostRef();
-        ApiRole role = new ApiRole().name(roleName).type(roleType).hostRef(hostRef);
-        roleConfigs.put(role, configList);
-      }
-      currentConfig = new ServiceConfigurationModel(svcConfig, roleConfigs);
+      final ApiService apiService = new ApiService().name(service).type(serviceType);
```
smolnar82, line 754: To reuse the discovery workflow, `getCurrentServiceConfiguration()` reconstructs the `ApiService` / `ApiRole` / config objects it feeds into `ServiceModelFactory.generateServiceModels()`. But this reconstruction doesn't populate the same fields that real discovery (`ClouderaManagerServiceDiscovery.discoverService`) hands to generators. Any `ServiceModelGenerator` that dereferences a field the reconstruction leaves null throws an NPE inside `generateService()`. This is not an edge case — it hits the majority of generators:

- **`role.getHostRef().getHostname()`** — used by **34 of 51 generators** (`OozieServiceModelGenerator`, `SolrServiceModelGenerator`, HBase, Impala, and most others). NPEs if the reconstructed role's `hostRef` (or its hostname) isn't populated exactly as discovery populates it.
- **`service.getClusterRef().getClusterName()`** — used by `YarnUIServiceModelGenerator` and `JobHistoryUIServiceModelGenerator`. NPEs because the synthetic `ApiService` sets only `name` / `type`, no `clusterRef`.
- Any other field a given generator's `handles()` / `generateService()` reads that the reconstruction omits.

**Failure path & blast radius:** a start/restart/scale event for almost any discoverable service (YARN, Oozie, Solr, HBase, …) → `hasConfigChanged` → `getCurrentServiceConfiguration` → `ServiceModelFactory.generateServiceModels` → the generator's `generateService()` → **NPE**. The NPE is not an `ApiException`, so the method's own try/catch doesn't catch it; it propagates to `monitorClusterConfigurationChanges`' outer `catch(Exception)`, aborting the **entire** polling cycle for **all** clusters. Because the triggering event is never marked processed, the same NPE recurs every polling interval — PCA is effectively dead for that gateway.

**Root cause / fix direction:** the reconstruction in `getCurrentServiceConfiguration()` diverges from how `ClouderaManagerServiceDiscovery` builds these objects. Rather than hand-rebuilding `ApiService` / `ApiRole` from `readServiceConfig`, PCA should obtain the model inputs through the **same** code path discovery uses (the shared component introduced here), so every field a generator may read is populated identically. A unit test that runs each registered generator against PCA-reconstructed inputs would have caught this and would guard against regressions as new generators land.

gateway-discovery-cm/src/main/java/org/apache/knox/gateway/topology/discovery/cm/monitor/PollingConfigurationAnalyzer.java
```diff
@@ -340,25 +344,31 @@ private boolean hasConfigChanged(String address, String clusterName, List<Releva
...
-          // configured (appropriately) at that time.
+        } else if (currentConfig != null) {
+          // No prior config, but the service now produces a model: new / became valid -> re-discover.
           log.serviceEnabled(re.getService());
           configHasChanged = true;
+        } else {
+          // Had a prior config but produces no model now: became invalid / was removed -> re-discover so the
+          // service is dropped from the affected topologies (and the scoped-replace merge clears its baseline).
+          log.serviceDisabled(re.getService());
+          configHasChanged = true;
```
smolnar82, line 371: This new branch treats `getCurrentServiceConfiguration() == null` as "the service is now invalid / disabled" and forces re-discovery. But that method **also** returns `null` on any `ApiException` - network blip, auth failure, transient 5xx:

```java
} catch (ApiException e) {
  log.clouderaManagerConfigurationAPIError(e);
}
return currentConfig; // still null on API error
```
So a `null` return is ambiguous: it can mean either "config genuinely produced no model" or "couldn't reach CM". hasConfigChanged's else branch (prior config present, currentConfig == null) can't tell them apart and unconditionally sets configHasChanged = true, logs serviceDisabled, and triggers a full re-discovery.

**Failure path:** CM API has a transient outage while computing the current config for a service that has a valid prior baseline → `getCurrentServiceConfiguration` catches the `ApiException` and returns `null` → `hasConfigChanged` forces re-discovery → repeats every polling cycle until CM recovers.

**Regression:** previously an API-error `null` was a harmless no-op (configHasChanged stayed false). This PR turns a transient CM error into a repeating full-cluster rediscovery storm — the same failure class KNOX-2900 set out to eliminate.

**Fix direction:** distinguish "no model produced" from "CM unreachable". Options: let the `ApiException` propagate (or rethrow) instead of collapsing it to `null`, or return a tri-state / Optional so the caller can skip the change decision on API errors and only treat a genuinely empty-but-successful result as "service invalid".

smolnar82 (changes requested):
I found 2 blockers, and the following efficiency issues:

#### 1. every descriptor is parsed twice per polling cycle

**File:** `PollingConfigurationAnalyzer.java` (`getReferencedServiceTypes()` / `clusterReferencesExist()`)

`clusterReferencesExist()` and `getReferencedServiceTypes()` each independently loop over `ts.getDescriptors()` and re-parse every descriptor file via `SimpleDescriptorFactory.parse()` (around lines 423 and 602). With `N` descriptors on disk, each polling cycle now parses every file **twice per cluster** — doubling filesystem reads + JSON parsing — and duplicates near-identical `IOException` handling.

**Fix direction:** merge the two into a single pass over `ts.getDescriptors()` that returns both "does any descriptor reference this `(address, cluster)`" and "the set of referenced service names", so each descriptor is parsed once.

#### 2. `CORE_SETTINGS` is re-fetched from CM once per relevant service type

**File:** `PollingConfigurationAnalyzer.java` (`getCoreSettingsConfig()`)

`getCoreSettingsConfig()` issues a fresh `readServices(clusterName, summary)` **plus** a `readServiceConfig()` round-trip every time `getCurrentServiceConfiguration()` runs — i.e. once per distinct relevant service type per polling cycle. Cluster discovery, by contrast, computes `CORE_SETTINGS` once for the whole cluster (`coreSettingsConfig(...)`).

**Failure path:** a polling cycle with events for three service types (e.g. `HDFS`, `HIVE_ON_TEZ`, `YARN`) makes three separate `readServices()` + `readServiceConfig()` calls just to re-derive the same `CORE_SETTINGS` value, adding avoidable CM API load on every interval.

**Fix direction:** compute `CORE_SETTINGS` once per polling pass and reuse it across all services handled in that pass (cache it for the duration of the cycle).

#### 3.: role-type exclusion config is cached for the analyzer's lifetime while service-type exclusion is read fresh

**File:** `PollingConfigurationAnalyzer.java` (`isExcludedRoleType()` / `isExcludedServiceType()`)

`isExcludedRoleType()` lazily builds and then **permanently caches** its `TypeNameFilter` (`roleTypeFilter`) from `gatewayConfig` on first use, whereas the adjacent `isExcludedServiceType()` re-reads `gatewayConfig.getClouderaManagerServiceDiscoveryExcludedServiceTypes()` fresh on **every** call.

**Failure path:** an operator updates `gateway.cloudera.manager.service.discovery.excluded.role.types` on a running gateway (without restarting the CM configuration monitor thread). `isExcludedServiceType` honors an equivalent excluded-service-types change immediately, but `isExcludedRoleType` keeps using the stale filter built at first invocation and silently ignores the update.

**Fix direction:** pick one policy for both. Either read both config values fresh per call, or resolve both once at construction — but don't split the behavior between the two sibling checks introduced in this PR.

PR #1400, opened 2026-09-15
KNOX-3458: enforce a configurable maximum actor-chain depth in token exchange

smolnar82:
[KNOX-3458](https://issues.apache.org/jira/browse/KNOX-3458) - Enforce a maximum actor-chain depth

The RFC 8693 `act` claim chain was walked and rebuilt with no depth bound anywhere:
`TokenUtils.extractActorChain()`, `addActorToChain()`, and `buildNestedActClaim()` all
iterate unbounded, and `ActorChainPrincipalImpl` does no validation. A deeply nested
`act` claim could therefore grow without limit across successive delegation exchanges.

This PR adds a configurable upper bound on the actor chain depth and enforces it at the
last point before a token is minted:

- **New topology provider param `delegation.max.actor.chain.depth`** (default `3`), read in `TokenResource.init()`. Invalid or non-positive values log a warning and fall back to the default.
- **Enforcement in `TokenResource.handleDelegatedAuthentication()`**, immediately after `TokenUtils.addActorToChain(...)` — the last chance before minting. If the resulting chain would exceed the configured maximum, the exchange is rejected with HTTP 400 and a new `ErrorCode.ACTOR_CHAIN_DEPTH_EXCEEDED(120)`, via the new `ActorChainDepthExceededException`. Only chain *growth* is bounded; an exchange that merely preserves an existing chain is not affected.
- **Audit**: `TokenExchangeHandler.auditMessage()` previously listed `act_chain_depth` as a deliberately-omitted field. It is now emitted, recording the *incoming* delegation depth (`0` for headless exchanges, which drop the incoming chain). Enforcement itself stays downstream in `TokenResource`; the audited value gives visibility into how deep chains are getting.
- Two new log messages (`actorChainDepthExceeded`, `rejectedTokenExchange`).

## How was this patch tested?

Automated unit tests (all passing):
- `TokenServiceResourceTest` — 4 new tests:
  - chain at exactly the max depth is allowed (resulting depth 3, default max 3 → 200);
  - chain that would exceed the max is rejected (resulting depth 11 → 400, body contains
    the "would exceed the configured maximum" message and code `120`);
  - a configured non-default max (`2`) is enforced;
  - an invalid config value falls back to the default of 3 rather than becoming unlimited.
- `TokenExchangeHandlerTest` — 1 new test asserting `act_chain_depth=2` is audited for a
  2-deep incoming chain, plus an assertion that `act_chain_depth=0` is audited when the
  subject token has no `act` claim.

Both affected modules (`gateway-service-knoxtoken`, `gateway-provider-security-jwt`)
compile cleanly and their targeted test classes pass.

## Integration Tests
N/A


smolnar82:
Cc. @hsheinblatt

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -155,6 +155,8 @@ public class TokenResource {
   private static final String KNOX_TOKEN_HARDCODED_CLAIM_MAPPINGS = TOKEN_PARAM_PREFIX + "hardcoded.claim.mappings";
   private static final String METADATA_QUERY_PARAM_PREFIX = "md_";
   private static final String TOKEN_ENABLE_DELEGATED_AUTH = TOKEN_PARAM_PREFIX + "enable.delegated.auth";
+  static final String DELEGATION_MAX_ACTOR_CHAIN_DEPTH = "delegation.max.actor.chain.depth";
+  static final int DELEGATION_MAX_ACTOR_CHAIN_DEPTH_DEFAULT = 10;
```
hsheinblatt: See comment on jira for reasoning, but 10 is risky, I suggest 5 (allows for most bespoke agentic flows), 3 (allows for standard agentic flows), or 1 (secure default, customers must set consciously to enable realistic agentic flows).

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -945,6 +972,11 @@ protected TokenResponseContext getTokenResponse(UserContext context) {
       } else {
         response = new TokenResponseContext(null, null, Response.serverError());
       }
+    } catch (ActorChainDepthExceededException e) {
+      log.rejectedTokenExchange(e.getMessage());
```
hsheinblatt, line 977: This doesn't seem like it is returned as an RFC 8693 compliant json response. It may be OK since this enforcement is a necessary best practice, not required by the spec, but I thought all failures for token exchange needed to conform.

hsheinblatt (commented):
Thanks Sandor, looks good to me. Questions about the default max depth and the error response format.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
smolnar82: Thanks, @hsheinblatt , for the detailed explanation. I'll update the default to 3.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
smolnar82: Fixed.

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
smolnar82: Fixed.

hanicz (approved):

PR #1401, opened 2026-09-15
KNOX-3459: honor delegation policy tokenTtlSec as the minted token's expiry basis

smolnar82:
[KNOX-3459](https://issues.apache.org/jira/browse/KNOX-3459) - Honor delegation policy tokenTtlSec

`PolicyDecision.getEffectiveTtlSec()` is already computed by `JdbcDelegationPolicyService.evaluate()` (from `DelegationPolicy.getTokenTtlSec()`, falling back to the configured default), but it was silently discarded — `TokenExchangeHandler` read only `getDenyReason()`, so the per-policy TTL had no effect on the minted token's lifetime.

This plumbs the effective TTL to KNOXTOKEN, mirroring the existing audience passthrough:

- **`CommonTokenConstants`** — adds `REQUESTED_TTL_REQUEST_ATTR` (`knox.token.requested.ttl`), an `Integer` seconds value an upstream component may set.
- **`TokenExchangeHandler`** — on the authorized delegation path (OBO / headless), sets the attribute from `policyDecision.getEffectiveTtlSec()`.
- **`TokenResource.getExpiry()`** — when the attribute is present and positive, uses it as the expiry basis directly, deliberately bypassing the topology `knox.token.ttl` upper bound and the client `lifespan` clamp.

The policy TTL is authoritative because `DelegationPolicy.tokenTtlSec` is trusted, operator-configured server-side state (a peer of the topology config, not untrusted client input like `lifespan`). Its main use case — longer-lived tokens for headless/batch delegations — is impossible if capped by the topology default. Plain (same-subject) exchanges are unaffected: the attribute is set only on the policy-governed delegation path.

## How was this patch tested?

- `TokenExchangeHandlerTest` (66 → 69): OBO delegation conveys the policy TTL, headless delegation conveys it, same-subject exchange does not.
- `TokenServiceResourceTest` (101 → 103): minted token honors the policy TTL while bypassing a lower topology `knox.token.ttl` cap, and bypasses a shorter `lifespan` clamp.
- Both modules compile clean; full suites green.


smolnar82:
Cc. @hsheinblatt

gateway-service-knoxtoken/src/main/java/org/apache/knox/gateway/service/knoxtoken/TokenResource.java
```diff
@@ -1346,6 +1346,16 @@ void addClientDataToMap(String[] tokenClientData,
   }
 
   protected long getExpiry() {
+    // An upstream component (the RFC 8693 token-exchange path) may have resolved an authoritative TTL
+    // from trusted, operator-configured delegation policy. When present it is the expiry basis
+    // directly: it deliberately bypasses the topology knox.token.ttl upper bound and the client-supplied
+    // lifespan clamp below, since the policy value is server-side state (a peer of the topology config),
+    // not untrusted client input, and its purpose is longer-lived tokens for headless/batch delegations.
+    final Object requestedTtlSec = request.getAttribute(CommonTokenConstants.REQUESTED_TTL_REQUEST_ATTR);
```
hsheinblatt, line 1357: Maybe worth referring to the authoring bounds: knox.delegation.max.token.ttl.sec defaults to 86400 sec (24 hours) and knox.delegation.min.token.ttl.sec defaults to 60 sec.

hsheinblatt (commented):
Thanks @smolnar82 , looks good to me. Just a comment that the operator allowed values are still bounded.

smolnar82:
> Thanks @smolnar82 , looks good to me. Just a comment that the operator allowed values are still bounded.

I made that bound clear in my comment in the code. Thanks for the review.

PR #1402, opened 2026-09-15
KNOX-3456: resolve canActFor.groups via KnoxLDAPService group lookup, failing with server_error when LDAP is unavailable

smolnar82:
[KNOX-3456](https://issues.apache.org/jira/browse/KNOX-3456) - LDAP group lookup for canActFor.groups

KNOX-3441 landed the token-exchange delegation policy with the `canActFor.groups` branch left as a stub (`UnsupportedOperationException` → HTTP 501). This PR implements the group-membership check:

- **Group lookup (AC1):** When the impersonated subject does not match `canActForUsers` and the policy's `canActForGroups` is non-empty, `JdbcDelegationPolicyService.evaluate()` resolves the subject's groups via `KnoxLDAPService.getUserGroups(subjectName)` and authorizes if any returned group is in the allowed set. Injected through `DelegationPolicyServiceFactory` (`ServiceType.LDAP_SERVICE`), mirroring `HadoopGroupProviderFilter`.
- **No mapping step (AC2):** Group names from the LDAP service (role CNs, including bare-RDN roles surfaced by the roles-lookup interceptor) are compared directly against the stored policy groups — a plain set-membership test, no transformation.
- **LDAP disabled/absent (AC3):** A group-based policy with LDAP absent or disabled is an operator misconfiguration, not a denial. It raises `DelegationGroupLookupUnavailableException`, which `TokenExchangeHandler` maps to HTTP 500 `server_error` with a description directing the operator to enable LDAP — rather than silently denying with `subject_not_allowed`.
- **Lookup failure (AC4):** If the LDAP group lookup call itself throws, the exception and stack trace are logged (at ERROR) and re-raised as `DelegationGroupLookupUnavailableException` (chaining the cause), again surfaced as `server_error`.

Ordering is unchanged otherwise: the group check runs last, after the cheaper user/resource/scope checks, and is skipped entirely when the subject already matches `canActForUsers`.

## How was this patch tested?

Automated unit tests, all green:
- `JdbcDelegationPolicyServiceTest` (58) — authorized-via-group, deny-when-in-no-group, user-check-short-circuits-group-lookup, and the three server_error paths (LDAP absent, LDAP disabled, lookup throws — the last asserting the originating exception is chained as the cause).
- `TokenExchangeHandlerTest` (67) — `canActFor.groups` denial → 400 `invalid_request`; group-lookup-unavailable → 500 `server_error` with an LDAP-oriented description.
- `DelegationPolicyServiceFactoryTest` (3) and `H2DBDelegationPolicyServiceTest` (2) confirm the LDAP service is wired in for both the H2 and JDBC paths.

## Integration Tests
N/A


gateway-server/src/test/java/org/apache/knox/gateway/services/knoxidf/delegation/JdbcDelegationPolicyServiceTest.java
```diff
@@ -859,14 +861,123 @@ public void testEvaluateDenyHeadlessNotAllowed() throws Exception {
...
+    EasyMock.expect(ldap.getUserGroups("alice")).andReturn(Collections.singletonList("users")).once();
+    EasyMock.replay(ldap);
+    service.setLdapService(ldap);
+
+    final PolicyDecision decision = service.evaluate(
+        new PolicyCheckRequest("oidc", "groups", "alice", Set.of("/api"), Collections.singleton("read"), false));
+
+    assertEquals("subject_not_allowed", decision.getDenyReason());
+    EasyMock.verify(ldap);
+  }
```
hsheinblatt, line 902: Arguably also boundary when the user is in no groups.

hsheinblatt (commented):
Thanks Sandor, looks good to me.

PR #1405, opened 2026-09-16
KNOX-3461: authorize same-subject requested audience against the subject token's aud via delegation.same.subject.requested.audience.enabled

smolnar82:
[KNOX-3461](https://issues.apache.org/jira/browse/KNOX-3461) - Authorize same-subject requested audience in RFC 8693 token exchange

Sub-task of KNOX-3441. When delegation is enabled, operators set the KNOXTOKEN audience validator to `passthrough` so the delegation policy authorizes requested audiences instead of a fixed list. But the delegation policy is only consulted for *delegation* exchanges, `same-subject` exchanges skip it entirely while still conveying the requested audience downstream. Under `passthrough` that let a same-subject caller request any audience with no authorization at all.

This adds a new topology provider flag `delegation.same.subject.requested.audience.enabled` (default off / fail-safe) that governs same-subject requested-audience handling in `TokenExchangeHandler`:

- **Delegation exchange** - unchanged; the requested audience is authorized by the delegation policy.
- **Same-subject, flag off (default)** - the requested audience is ignored (not conveyed), so `passthrough` cannot mint an arbitrarily-audienced token without authorization.
- **Same-subject, flag on** - the requested audience is honored but every value must be present in the subject token's own `aud` claim; otherwise the exchange is rejected with `invalid_target`.

Enforcement lives in `TokenExchangeHandler` because it is the only layer holding the raw subject JWT (the downstream KNOXTOKEN `TokenResource` sees only the resolved `Subject`).

Note: defaulting off changes behavior for existing `whitelist`/`static` deployments that today let a same-subject caller request an audience. They must opt in by setting the flag, which also brings the subject-token-`aud` enforcement. This matches the fail-safe convention of the other `delegation.*` flags.

## How was this patch tested?

Automated unit tests (all passing):
- `TokenExchangeHandlerTest` (75): added coverage for flag-off ignore, subset-authorized, audience-not-in-subject-token reject, subject-token-with-no-aud reject, and a delegation-path regression; updated existing same-subject conveyance tests for the new contract.
- `JWTFederationFilterDelegationConfigTest` (8): new flag defaults to false and is independently configurable.
- `JWTFederationFilterTokenExchangeTest` (60): regression.
- `HadoopAuthFilterTest` (12): added the new init-param stub.

`mvn -pl gateway-provider-security-jwt test -Dtest=TokenExchangeHandlerTest,JWTFederationFilterDelegationConfigTest,JWTFederationFilterTokenExchangeTest`

## Integration Tests

N/A (will come in a new PR soon)

## UI changes

N/A


smolnar82:
Cc. @hsheinblatt

smolnar82:
For future reviewers: none of these configs, behavior is documented. I filed [KNOX-3464](https://issues.apache.org/jira/browse/KNOX-3464) to add missing docs.

hanicz (approved):

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
```diff
@@ -131,6 +131,18 @@ public class JWTFederationFilter extends AbstractJWTFilter {
...
+  // same-subject token exchange may honor a requested audience/resource at
+  // all. While off (the fail-safe default), the requested audience is ignored
+  // for same-subject exchanges, so a passthrough audience validator cannot
+  // mint an arbitrarily-audienced token without authorization. While on, the
+  // requested audience is honored but must be authorized against the subject
+  // token's own aud claim; any requested value the subject token does not
+  // already carry is rejected. Unlike the delegation.* flags above, this one
+  // affects only same-subject exchanges (delegation exchanges are authorized
+  // by the delegation policy regardless of this flag).
+  public static final String DELEGATION_SAME_SUBJECT_REQUESTED_AUDIENCE_ENABLED = "delegation.same.subject.requested.audience.enabled";
```
hsheinblatt: I had been using 'delegation' to mean cross-subject token exchange specifically. For consistency, I wouldn't call this 'delegation'. I can see how it's mostly relevant if one is using delegation, but in principle one could enforce same subject token exchange audience validation with delegation exchange turned off. So, I suggest a different prefix for the name here, possibly "token-exchange.same.subject.requested.audience.enabled" or 'token.exchange' or 'rfc8693.token.exchange' since it only applies for the specific OIDC grant type.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
```diff
@@ -316,7 +316,29 @@ ActionOutcome.SUCCESS, auditMessage(policyDecision, actorIdentity, subjectToken,
       // present so that KNOXTOKEN falls back to its resource query parameter otherwise; when set,
       // the body value takes precedence over the query parameter.
       if (!requestedAudiences.isEmpty()) {
-        request.setAttribute(CommonTokenConstants.REQUESTED_AUDIENCES_REQUEST_ATTR, requestedAudiences);
+        final boolean sameSubjectExchange = !hasActorToken && !requestedSubjectDiffersFromSubject;
```
hsheinblatt: nit: I hadn't computed the boolean delegationExchange above for the big branch that handles delegation, because it was only used once, so I just inlined the AND expression.  It branches off for the delegation authorization logic. But if we're using it again here, we can compute it up there and share it. 

nit: Arguably this logic, or at least the enforcement logic could be moved into the above conditional, where we already have an if delegation block. We can add an else block to enforce the audiences are valid there, so all the enforcement is done before the subject setting. That is, if we later add more options for how to validate same-subject audience requests, we'd have a more complex validation step there. Grouping it with setting the attribute is a bit confusing.

gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandlerTest.java
```diff
@@ -571,6 +560,74 @@ public void testNoResourceOrAudienceLeavesRequestAttributeUnset() throws Excepti
...
+    // there is nothing to authorize against, so the request is rejected.
+    filter.delegationSameSubjectRequestedAudienceEnabled = true;
+    filter.valid.put("subtok", jwt("alice", "KNOXSSO"));
+    handler.handle(exchangeRequest("subtok", null, new String[] {"service-a"}), response, chain);
+
+    assertFalse(filter.continued);
+    assertEquals(HttpServletResponse.SC_BAD_REQUEST, filter.errorStatus);
+    assertEquals("invalid_target", filter.error);
+    assertFalse(requestedAudiencesAttr.hasCaptured());
+  }
```
hsheinblatt, line 645: I'd add another boundary test when the request has some audiences in the subject token claim and some not, to show it checks all.

hsheinblatt (commented):
Thanks Sandor, looks good to me. One boundary test suggested, but arguably not necessary.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
smolnar82: Good call, agreed, "delegation" is misleading since this applies to same-subject exchange and works independently of delegation.server.enabled. Renaming to `token.exchange.same.subject.requested.audience.enabled` (keeping the dot convention of the sibling flags). Will update the constant, field, getter, comment, and tests.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
smolnar82: > nit: I hadn't computed the boolean delegationExchange above for the big branch that handles delegation, because it was only used once, so I just inlined the AND expression. It branches off for the delegation authorization logic. But if we're using it again here, we can compute it up there and share it.

Makes sense. I'll hoist `delegationExchange = hasActorToken || requestedSubjectDiffersFromSubject` up where it's first used and reuse it here.

> nit: Arguably this logic, or at least the enforcement logic could be moved into the above conditional, where we already have an if delegation block. We can add an else block to enforce the audiences are valid there, so all the enforcement is done before the subject setting. That is, if we later add more options for how to validate same-subject audience requests, we'd have a more complex validation step there. Grouping it with setting the attribute is a bit confusing.

Agreed. I'll move the same-subject audience enforcement into an else on the delegation block so all authorization happens before the subject is set, leaving the downstream conveyance as a plain attribute-set.

gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandlerTest.java
smolnar82: Will add a boundary test where some requested audiences are in the subject token's `aud` and some aren't, asserting it's rejected, which confirms all values are checked, not just one.

smolnar82: Done.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
smolnar82: Fixed.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/JWTFederationFilter.java
smolnar82: Done.

smolnar82:
All code review comments are addressed, tests are green -> I'm merging this one.

PR #1408, opened 2026-09-17
KNOX-3471: KnoxIDF expires_in should return relative second instead o…

hanicz:
…f absolute epoch-ms

[KNOX-3471](https://issues.apache.org/jira/browse/KNOX-3471) - KnoxIDF expires_in should return relative second instead of absolute epoch-ms

- Overrides `buildResponseMap` in KnoxIDF's `TokenResource` to replace `expires_in` with the duration-based `getTokenLifetimeInSeconds()`. Native `KNOXTOKEN` is untouched. `expires_in` is removed from the response if there is no expiry.

## How was this patch tested?

Unit tests, local tests

`curl -vku admin:admin-password -X "POST" "https://localhost:8443/gateway/tokenissuer/knoxidf/api/v1/token"`

```
{
  "access_token": "eyJqa3UiOiJodHRwczovL2xv...",
  "token_id": "7c89743d-3e1b-4673-a3a4-afba906b69e9",
  "issued_token_type": "urn:ietf:params:oauth:token-type:jwt",
  "managed": "true",
  "endpoint_public_cert": "MIIDZDCCAkygAwIBAgIIVb...",
  "token_type": "Bearer",
  "expires_in": 3600,
  "passcode": "TjJNNE9UYzBNMlF..."
}
```


smolnar82 (changes requested):
The fix itself is right, but I don't think it's complete enough to merge yet. Three gaps:

- The `-1` (never-expires) path is broken and untested. `addExpiryIfNotNever` is a no-op when `getTokenLifetimeInSeconds()` returns -1, so on that path the inherited absolute epoch-ms expires_in from `super.buildResponseMap` survives; but we have no unit test coverage for that. I'm not even sure that `-1` is valid in OAuth 2.0 (see `expires_in` definition in [RFC 6749 / Section 5.1](https://www.rfc-editor.org/info/rfc6749/#section-5.1)

- No `expires_in` assertion in the Docker tests. `test_knoxidf.py` checks token_type/issued_token_type/etc. across all flows but never touches `expires_in`, the one field this PR changes. Since the unit test stubs the lifetime, the integration test is the only thing that verifies the real value end-to-end.

- Docs still show the old value. `knox-site/docs/knoxidf/endpoints.md:193` documents `"expires_in": 1699999999999 (epoch-ms)`, which this PR makes wrong.

smolnar82 (approved):

PR #1409, opened 2026-09-17
KNOX-3460: audit all RFC 8693 token-exchange paths pre-mint and add a mint-time TOKEN_EXCHANGE audit emitting jti/expiry/issued_subject

smolnar82:
[KNOX-3460](https://issues.apache.org/jira/browse/KNOX-3460) - Audit all RFC 8693 token-exchange paths and add a mint-time audit

Completes the RFC 8693 token-exchange audit trail:

- **Handler (pre-mint):** every exchange path now emits an `Action.TOKEN_EXCHANGE` record exactly once. The same-subject path (including the KNOX-3461 requested-audience authorization and its fail-safe default) previously emitted no `TOKEN_EXCHANGE` audit — it now audits SUCCESS/FAILURE like delegation. The delegation LDAP-unavailable path now emits an `UNAVAILABLE` record, so every delegation outcome (allow/deny/unavailable) is audited. Records carry an `audiences_honored` field so a requested-but-dropped audience (honoring disabled) is never mistaken for an honored one.
- **TokenResource (mint-time):** `getJWT()` emits a `TOKEN_EXCHANGE` SUCCESS audit for exchange-originated mints only, carrying the post-mint fields `jti`, `expiry`, and `issued_subject`. Ordinary (non-exchange) issuance stays unaudited. The pre-mint decision and the mint record share the request's audit correlation id.
- Relocated the exchange marker to `CommonTokenConstants.TOKEN_EXCHANGE_REQUEST_ATTR` (value unchanged) so the KNOXTOKEN service can distinguish exchange mints from ordinary issuance.
- Cleanup: extracted the token-exchange auditing into `TokenExchangeAuditing` and the actor-identity derivation into `ActorIdentity`; extracted request-parameter and delegation validation and the delegation branch into helper methods. Behavior-preserving.

## How was this patch tested?

Automated unit tests (all green):
- `TokenExchangeHandlerTest` — same-subject SUCCESS (audience honored / dropped), same-subject FAILURE on audience rejection, delegation UNAVAILABLE on group-lookup failure.
- `TokenServiceResourceTest` — mint audit fires with `jti`/`expiry`/`issued_subject` when the exchange marker is set; no mint audit for ordinary issuance.
- `JWTFederationFilterTokenExchangeTest` — confirms the constant relocation is transparent.
```
mvn -pl gateway-spi,gateway-provider-security-jwt,gateway-service-knoxtoken -am
  test -Dtest=TokenExchangeHandlerTest,JWTFederationFilterTokenExchangeTest,TokenServiceResourceTest
```
## Integration Tests
No new integration tests; the change is covered by unit tests at the handler and KNOXTOKEN-service layers. Existing RFC 8693 CI integration suites (KNOX-3465/3466) remain applicable and unaffected.


smolnar82:
cc. @hsheinblatt

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
hsheinblatt, line 140: I think every handleValidationError call needs an audit call first in this design, so missing here.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
```diff
@@ -348,6 +232,165 @@ ActionOutcome.SUCCESS, auditMessage(policyDecision, actorIdentity, subjectToken,
...
 
+  private boolean validateExchangeRequestParameters(HttpServletRequest request, HttpServletResponse response,
+                                                    HttpServletRequest bodyRequest, String subjectTokenValue, boolean hasActorToken) throws IOException {
+    final String subjectTokenType = bodyRequest.getParameter(SUBJECT_TOKEN_TYPE);
+    final String actorTokenType = bodyRequest.getParameter(ACTOR_TOKEN_TYPE);
+    final boolean hasActorTokenType = actorTokenType != null && !actorTokenType.isEmpty();
+
+    // RFC 8693 section 2.1: subject_token and subject_token_type are REQUIRED.
+    if (subjectTokenValue == null || subjectTokenValue.isEmpty()) {
+      filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
```
hsheinblatt, line 253: I think all these need audit messages with them. caller just returns, doesn't conditionally audit.

gateway-provider-security-jwt/src/main/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandler.java
```diff
@@ -348,6 +232,165 @@ ActionOutcome.SUCCESS, auditMessage(policyDecision, actorIdentity, subjectToken,
...
+  }
+
+  private boolean validateDelegationExchangeRequest(HttpServletRequest request, HttpServletResponse response,
+      boolean hasActorToken, boolean requestedSubjectDiffersFromSubject, String requestedSubjectValue,
+      Set<String> uniqueRequestedAudiences) throws IOException {
+    // Delegation exchanges are default denied unless DELEGATION_SERVER_ENABLED is set to true. When
+    // true, only authorized token exchanges will be permitted. Otherwise, any actor could impersonate
+    // any subject without authorization.
+    if (!filter.isDelegationServerEnabled()) {
+      filter.handleValidationError(request, response, HttpServletResponse.SC_BAD_REQUEST,
```
hsheinblatt, line 378: I think all these calls to handleValidationError need audit calls before them. The caller doesn't do it.

gateway-provider-security-jwt/src/test/java/org/apache/knox/gateway/provider/federation/jwt/filter/TokenExchangeHandlerTest.java
```diff
@@ -297,6 +320,77 @@ public void testSameSubjectExchangeNeverCallsPolicyEvaluation() throws Exception
...
+
+    assertFalse(filter.continued);
+    assertEquals(HttpServletResponse.SC_BAD_REQUEST, filter.errorStatus);
+    assertEquals("invalid_target", filter.error);
+    EasyMock.verify(auditor);
+    assertTrue(auditMessage.getValue().contains("event_type=token_exchange_denied"));
+    assertTrue(auditMessage.getValue().contains("deny_reason=requested_audience_not_authorized"));
+    assertTrue(auditMessage.getValue().contains("subject_token_sub=alice"));
+  }
+
```
hsheinblatt, line 424: If we add audits for all the other failure cases, then each needs a test too. Possibly we can simplify those to invalid requests with a simpler message, but it would be good to have the actor who requested an invalid request and for whom.

hsheinblatt (commented):
Thanks Sandor, looks good to me. Now we audit all the authorization decisions along with the exchange minting, and the refactor into helpers makes the handle method more readable.

I had thought we'd want to audit all failures. I think for SOC (number?) or FedRamp all request failures would need to be audited. We don't necessarily need to comply with any compliance specs, but it would limit use.  Perhaps that opens us up to flooding the audit logs on DOS attacks and we want to avoid it, though I'd think in an enterprise deployment you'd have to handle that already via other/external means.

smolnar82:
> Thanks Sandor, looks good to me. Now we audit all the authorization decisions along with the exchange minting, and the refactor into helpers makes the handle method more readable.
> 
> I had thought we'd want to audit all failures. I think for SOC (number?) or FedRamp all request failures would need to be audited. We don't necessarily need to comply with any compliance specs, but it would limit use. Perhaps that opens us up to flooding the audit logs on DOS attacks and we want to avoid it, though I'd think in an enterprise deployment you'd have to handle that already via other/external means.

You're right that all request failures should be audited; limiting compliance use isn't worth it. I've added auditing for every rejection, capturing who made the invalid request and (where known) for whom, with a reason code on each. The flooding concern is handled: the reason codes let downstream processing aggregate or rate-limit, on top of the external protections an enterprise deployment would already have. Tests added for each case.

smolnar82:
@hsheinblatt - I applied the changes you requested. If you think anything else should be changed, we'll do that in a separate JIRA. I'm merging this one now.

